PR #1, opened 2026-07-10
fix+docs: onboarding dry-run blockers, README rewrite, and Diátaxis docs tree

jozer-rami:
## What this is

We walked `evenfire-open-source` as an external developer would — using only the
public README and the repo's code — first through the Docker Compose quickstart,
then the unit tests, and stopped short of the full minikube stack. The compose
quickstart **works end to end** (LLM round-trip + a full MCP tool call through
the approval gate), but the docs strand a newcomer at every extension point past
"boot and chat." This PR fixes the 14 reproduced blockers.

## Findings fixed

| Finding | Sev | Fix |
|---|-----|-----|
| **F5** | S2 | `CLERUM_MCP_SERVERS` never reached the container (dropped at the compose layer). Add passthrough to `docker-compose.yml`. Verified: `docker compose config` now resolves it into the mcp-host env. |
| **F15** | S2 | The documented `CLERUM_MCP_SERVERS` JSON shape omitted the `status:{deployed,ready}` object, so a server was parsed but silently skipped (`manager.ts:43`). Fixed the shape in `mcp-host/README.md`, `.env.quickstart.example`, and the root README. |
| **F13** | S2 | The Testing section listed "Node 24+" then jumped to `npm test` with no `npm install`; every service (no root workspace) fails verbatim with `MODULE_NOT_FOUND`. Added the install step. |
| **F11** | S2 | README §3.6 documented approval routes/bodies that don't exist. Rewrote it with the real `mcp-host` runtime routes and the observed `waiting_approval` status. |
| **F16** | S3 | The runtime told users to `POST /approve` — a 404. Fixed the notification (and a doc-comment) to the real `POST /v1/runtime/approvals/approve`, **with a unit test** that fails against the old string. |
| **F4** | S2 | `cd workflow-sdk` → `cd packages/workflow-sdk` (Makefile already used the right path). |
| **F1/F2/F3/F6/F9** | S3 | Removed/redirected 15 dead doc links (root `CLAUDE.md`/`AGENTS.md`, `gcp.md`, two deploy runbooks, the absent `archive/`+`research/` trees, a stale demo-gif placeholder, and overlay comments citing unpublished paths). |
| **F7** | S3 | Rebranded doc titles/prose "Clerum" → "evenfire" (left the real `clerum.io` API group, `CLERUM_*` vars, `clerum/*` images, and `clerum-test` profile untouched). |
[...]

PR #2, opened 2026-07-10
fix(control-api): repair the imagePolicy base↔overlay boundary test

jozer-rami:
## Problem

`control-api` CI has been red on `main` (**1 failed / 2590 passed**) since
`e45ca03` ("Propagate base↔overlay genericization fix"). The failing test is
`test/imagePolicy.test.ts › assembled allowlist behavior › rejects the same
image against the genericized placeholder list (the bug)`:

```
AssertionError: expected { ok: true } to deeply equal { ok: false, reason: 'host_not_allowed' }
```

## Root cause

The `describe` block was added in `e45ca03` **already failing**, due to a
copy-paste error in the test fixtures — not a product bug:

- `REAL` and `PLACEHOLDER` were **byte-identical** strings, both containing the
  Artifact-Registry prefix `us-central1-docker.pkg.dev/${GCP_PROJECT}/clerum/`.
- `img` is an AR-host image, so `classifyPluginImage` matched it against **both**
  lists and correctly returned `{ ok: true }`.
- Test 1 asserts `(img, REAL) → ok:true` ✅ and Test 2 asserts `(img, SAME list)
  → ok:false` ❌ — a logical contradiction; both can't pass.

`classifyPluginImage` / `matchesAllowedImagePrefix` are correct — the matcher is
start-anchored (`image.startsWith(prefix)`). Only the fixture was wrong.

## Fix

Set `PLACEHOLDER` to the actual vendor-neutral base default that `deploy/base`
ships — `ghcr.io/evenfire-ai/,mongodb/,mcr.microsoft.com/,clerum/` — which is
named verbatim both in the `e45ca03` commit message and in the `@clerum/image-policy`
source comment. That list has **no** AR host, so a real AR image correctly fails
to match it and the boundary test asserts genuine fail-closed rejection.

## Verification

- `imagePolicy.test.ts`: **12/12 pass** (was 1 failing).
- Controlled local run proves net effect: unmodified suite = 33 failing tests,
  with this fix = 32 — exactly the one target test fixed, nothing else touched.
  (The other 32 local failures are environmental — they need CI's
  `workflow-runtime-core` / control-api build steps; they occur identically
  without this change and do not happen in CI.)
- One-line fixture change; no production code touched.

Unblocks `control-api` CI on `main` (and, once `main` picks this up, the same
check on #1).

### DIFF
--- a/control-api/test/imagePolicy.test.ts
+++ b/control-api/test/imagePolicy.test.ts
@@ -108,12 +108,11 @@ describe('assembled allowlist behavior (deploy/base vs gcp overlay boundary)', (
     'us-central1-docker.pkg.dev/${GCP_PROJECT}/clerum/,example.com/,mongodb/,mcr.microsoft.com/,clerum/'.split(
       ','
     )
-  // PLACEHOLDER mirrors a genericized base that leaked through unoverridden —
-  // the exact bug this boundary guards against.
-  const PLACEHOLDER =
-    'us-central1-docker.pkg.dev/${GCP_PROJECT}/clerum/,example.com/,mongodb/,mcr.microsoft.com/,clerum/'.split(
-      ','
-    )
+  // PLACEHOLDER mirrors the vendor-neutral base default that deploy/base ships
+  // (ghcr.io/evenfire-ai/,mongodb/,mcr.microsoft.com/,clerum/) — the exact list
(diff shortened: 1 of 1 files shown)


jozer-rami:
recheck

PR #82, opened 2026-07-15
docs: add "Desktop setup & updates" how-to

jozer-rami:
PR-2 of the docs improvement plan. A task-oriented how-to for **multi-environment setup and updates** — the flows the Desktop App surface doc explicitly defers ("auto-update not covered here").

## New: `docs/how-to/desktop-setup-and-updates.md`
- **Environment model** — save only the External REST API URL; the app discovers the RPC proxy + app name and caches it.
- **Three setup paths** — manual (`+ Environment`: name + REST API), Profile UI "Setup desktop app" (`evenfire://desktop-environment`), and invitation (`evenfire://desktop-setup`), with the member-registration gap noted.
- **Localhost** list behavior, and **updates** — `GET /desktop/release` + `minimumDesktopVersion` → update-required screen → release page (tag `desktop-app-<version>`), framed as **detection, not auto-install**.

## Verified against shipped code (this mattered)
Independently checked every claim against the code — the source design spec diverges in two places I did **not** propagate:
- The setup deep-link scheme is `evenfire://` (the app also registers `clerum://` for OAuth) — `main.ts:60-61`. Confirmed, not assumed.
- **The update check is *not* packaged-only.** The spec says "only in packaged builds," but the shipped code runs it after auth in dev too (no `isPackaged` gate — `useAppController.ts:585`, `appService.ts:126`). The doc reflects the code, not the spec.
- Confirmed: `/desktop/environment` (public) + `/desktop/release` (auth) shapes, `updateRequired = installed < minimumDesktopVersion`, manual add requires name + REST API only, Profile UI produces the deep link (`SettingsContent.tsx:332`).

Safety: no secrets, no tenant-specific hostnames, all links resolve.

## Wire-up
docs/README how-to table · llms.txt · learning path A · desktop-app + profile-ui surface docs (cross-links; auto-update/signing remain deferred there).

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -49,6 +49,7 @@ Use this index for long-form docs.
 | [Connect Telegram](how-to/connect-telegram.md)                 | Minikube env or CRD channel path    |
 | [Configure approvals](how-to/configure-approvals.md)           | Human-in-the-loop tool gates        |
 | [Add an MCP server](how-to/add-mcp-server.md)                  | Dev-mode JSON or CRD allowlist path |
+| [Desktop setup & updates](how-to/desktop-setup-and-updates.md) | Environments, invitation setup, updates |
 | [Minikube full stack](deploy/minikube.md)                      | Local Kubernetes platform           |
 | [Production notes](deploy/production.md)                       | Checklist and in-repo deploy assets |
 | [WorkflowRecipes operations](deploy/workflow-recipes-guide.md) | Recipe ops, RBAC, debugging         |
--- a/docs/get-started/learning-path.md
+++ b/docs/get-started/learning-path.md
@@ -7,7 +7,9 @@ Pick the path that matches what you want to do.
 1. [Quickstart](quickstart.md) — `make minikube-setup`, then message the seeded
    `chatllm` agent from the desktop app or the API
 2. Tour the [Desktop App](../surfaces/desktop-app.md) — screens, live activity,
-   in-chat approvals, and artifacts, beyond the one message you just sent
+   in-chat approvals, and artifacts, beyond the one message you just sent.
+   To point it at more than one instance or handle updates, see
+   [Desktop setup & updates](../how-to/desktop-setup-and-updates.md)
 3. Optional: connect [Telegram](../how-to/connect-telegram.md)
 4. Read [Why evenfire](../concepts/why-evenfire.md)
 
--- a/docs/how-to/desktop-setup-and-updates.md
+++ b/docs/how-to/desktop-setup-and-updates.md
@@ -1,0 +1,99 @@
+# How to: set up Desktop App environments and updates
+
+The Desktop App is **one packaged build** that any number of people point at any
+number of evenfire instances — your production tenant, a staging tenant, or a
+local `make`-forwarded cluster. This guide covers adding and switching
+**environments**, the three setup paths, and how the app handles **updates**.
+For packaging and shipping the build itself, see
+[Ship it to your users](../surfaces/desktop-app.md#ship-it-to-your-users).
+
+## When you need this
+
+- You run more than one instance (say, staging and production) and want to
+  switch between them in one app.
+- You are onboarding through an **invitation** rather than a shared login.
+- You distribute a **packaged** build and want users to know when a new release
+  requires them to update.
+
+## The environment model
+
+You save only the **External REST API URL** for an instance. The app then asks
+that API for the rest of the environment — the display name and the RPC proxy
+URL — and caches it locally. There is no second URL to type: the RPC proxy is
+**discovered**, not entered.
+
+## Setup paths
+
+Three ways to add an environment; all of them end with the same saved
+environment.
+
+### 1. Manual, from the sign-in screen
+
+Click the **+** beside **Environment**, give it a **name** and the **External
+REST API URL**, and save. Both fields are required. The app discovers the RPC
+proxy, and you can sign in.
+
+For a local minikube cluster, follow the
+[Quickstart](../get-started/quickstart.md) port-forwards and point the
+environment at the forwarded External REST API.
+
+### 2. From Profile UI, by deep link
+
+A member can hand the app its environment from the browser: in **Profile UI →
+Settings → Setup desktop app**, copy the External REST API and click **Open
+desktop app and setup**. That opens an `evenfire://desktop-environment` deep
+link the installed app handles — it saves or updates the environment, and if the
+External REST API was already saved, it updates the name and re-discovers the
+RPC proxy. Fleets can also be pre-seeded with a `CLERUM_DESKTOP_CONFIG_PATH`
+config file (see
+[Ship it to your users](../surfaces/desktop-app.md#ship-it-to-your-users)).
+
+### 3. Through an invitation
+
+When the app starts with **no selected environment**, it runs invitation
+detection. Completing the [Profile UI](../surfaces/profile-ui.md) invitation
+flow hands the app a setup token (an `evenfire://desktop-setup` deep link); the
+app then saves the tenant environment by name and discovers its RPC proxy.
+
+> **Gap:** invitation signup depends on `member-registration-service`, which is
+> not in this repo. See the
+> [member-registration gap](../surfaces/desktop-app.md#the-member-registration-service-gap)
+> for what still works without it.
+
+## Localhost
+
+**Localhost** always sits at the bottom of the environment list. In a packaged
+build it stays **unselected** until you pick it, so invitation detection can run
+on first launch; running through local dev commands selects it by default.
+
+## Updates
+
+After you authenticate, the app checks the **currently selected tenant** for its
+update policy, so different tenants can be on different versions at the same
+time. It asks that tenant's External REST API for its release info, including a
+`minimumDesktopVersion`. If the installed version is **below** that minimum, the
+app shows an **update-required** screen and opens the release page for the new
+build (tag `desktop-app-<version>`) in your browser. (A packaged distribution is
+where this matters most — a source checkout can just pull the latest.)
+
+This is update **detection**, not silent auto-install: the app tells the user an
+update is required and links the release. Code signing, notarization, and
+automatic download/install are not part of this repo — see
+[Ship it to your users](../surfaces/desktop-app.md#ship-it-to-your-users).
+
+## Troubleshooting
+
+- **Wrong instance after switching** — each environment caches its own RPC
+  proxy; re-open the **Environment** screen and confirm which instance is
+  selected.
+- **Stale RPC proxy** — re-running the Profile UI **Setup desktop app** for an
+  already-saved External REST API updates the name and clears the cached RPC
+  proxy so it is re-discovered.
+- **Signed in to the wrong tenant** — authentication is per environment; sign
+  out, reselect the environment, and sign in again.
+
+## Related
+
+- [Desktop App surface](../surfaces/desktop-app.md) — screens, packaging, env vars
+- [Profile UI](../surfaces/profile-ui.md) — the invited-member entry point
+- [Quickstart](../get-started/quickstart.md) — local minikube bring-up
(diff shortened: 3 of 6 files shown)


jozer-rami:
## Review — docs: Desktop setup & updates how-to

Reviewed at head `a7d00fb` against the shipped code that commit claims to follow (`desktop-app/src/main.ts`, `desktop-app/src/config.ts`, `desktop-app/src/appService.ts`, `desktop-app/ui/src/hooks/useAppController.ts`, `desktop-app/ui/src/hooks/domain/useAuthController.ts`, `desktop-app/ui/src/App.tsx`, `external-rest-api/src/routes/desktop.ts`, `profile-ui/app/settings/SettingsContent.tsx`). Scope is the diff: one new how-to plus five index/cross-link edits. No code changes. The PR was already merged; this is a post-merge accuracy review of the contract the body says was verified.

The shape of the page is right: task-oriented, three setup paths, the member-registration gap called out, updates framed as detection rather than auto-install, and the indexes (`docs/README.md`, `docs/llms.txt`, learning path A, desktop-app and profile-ui surface docs) all point at it. Several behavioral claims do not match the code they cite. Because the PR body treats those claims as verified, the mismatches are findings, not style notes (R4).

### Findings

#### R1-H1 — Release tag is documented as a shape that has never existed

`docs/how-to/desktop-setup-and-updates.md` (Updates) says the app opens the release page for tag `desktop-app-<version>`.

At this head, `external-rest-api/src/routes/desktop.ts` did build `desktop-app-${releaseManifest.desktopVersion}` and put it in `releaseUrl`. That is what the author verified. It was still wrong as user-facing documentation: real GitHub tags are `v0.1.0` … `v0.5.0`. `aa9ff214` (`fix(external-rest-api): link the update prompt at a release that exists`) later changed the route to `v${releaseManifest.desktopVersion}` with the comment that the `desktop-app-<version>` shape has never existed, so every update prompt linked to a 404. The how-to on current `main` still says `desktop-app-<version>` (`docs/how-to/desktop-setup-and-updates.md` around the Updates section). The doc and the code have drifted in opposite directions: the code was fixed, the sentence that taught the bad tag was not.

Also overstated in the same paragraph: the app does not open the browser by itself. `DesktopUpdateRequiredDialog` (`desktop-app/ui/src/App.tsx`) renders a modal with a **Download update** button; `handleOpenDesktopRelease` calls `shell.openExternal` only after that click. "Opens the release page … in your browser" reads as automatic.

Suggested wording: the selected tenant's `GET /api/v1/desktop/release` returns `releaseUrl` (tag `v<desktopVersion>`). If the installed version is below `minimumDesktopVersion`, a blocking **Desktop update required** dialog appears; **Download update** opens that URL.

#### R1-H2 — Re-running "Setup desktop app" does not clear a cached RPC proxy

Troubleshooting says: re-running Profile UI **Setup desktop app** for an already-saved External REST API "updates the name and clears the cached RPC proxy so it is re-discovered." Path 2 says the same thing ("if the External REST API was already saved, it updates the name and re-discovers the RPC proxy").

What the code does:

1. The deep link only carries `externalRestApiBaseUrl` and `tenantName` (`SettingsContent.tsx` `openDesktopAppSetup`).
2. The renderer stores a pending config with `rpcProxyBaseUrl: ''` and waits for an explicit confirm (`useAuthController.ts` `onDesktopEnvironmentSetup`). It does not save on receipt.
3. Confirm calls `saveDesktopRuntimeConfig`. On an existing profile (matched by External REST API URL), `config.ts` writes `rpcProxyBaseUrl: validated.rpcProxyBaseUrl || ''` — the empty string **overwrites** the previously cached proxy. It does not merge.
4. `resolveRuntimeConfigIfNeeded` (`appService.ts`) returns immediately when `config.rpcProxyBaseUrl?.trim()` is non-empty. It runs from `saveRuntimeConfig` only as `.catch(() => undefined)`, and from `passwordLogin`. After step 3 the proxy is empty, so the next login can rediscover it — but only if discovery succeeds. If `GET /desktop/environment` fails, the catch swallows it and the profile is left with an empty proxy. The troubleshooting step does not mention the confirm dialog, and it describes a clear-then-rediscover that is actually an overwrite-with-empty plus a best-effort refill.

#### R1-M1 — Manual setup is not "name + REST API, then the app discovers the proxy, and you can sign in"

Path 1 says both fields are required, the app discovers the RPC proxy, and you can sign in.

`handleSaveRuntimeConfig` does require a non-empty name and External REST API URL. The manual form can also submit `runtimeConfigSetupRpcProxyBaseUrl`. When that field is non-empty it is persisted, and `resolveRuntimeConfigIfNeeded` then **skips** discovery. "There is no second URL to type" / "the RPC proxy is discovered, not entered" is only true if the operator leaves the RPC field blank. The how-to never says that field exists.

Discovery is also not part of Save for the manual path in a way the user can see fail: `saveRuntimeConfig` calls `resolveRuntimeConfigIfNeeded().catch(() => undefined)`. A bad External REST API URL still toasts "Environment saved". The failure shows up later, at login (`passwordLogin` awaits `resolveRuntimeConfigIfNeeded` and will throw).

#### R1-M2 — Invitation detection is not "app starts with no selected environment"

Path 3 says: when the app starts with no selected environment, it runs invitation detection; completing the Profile UI invitation flow hands the app an `evenfire://desktop-setup` token; the app then saves the tenant by name and discovers the RPC proxy.

The actual gate is `runtimeConfigMissing` (`AuthPage`), which is "no runtime config is configured," not "an environment exists but none is selected." The user types an invitation email and the app calls `startDesktopSetup`, which looks up `member-registration-service` and opens Profile UI `/desktop-setup` in the browser. Only the return deep link (`evenfire://desktop-setup?email&authorizationToken`) calls `completeDesktopSetup`, which saves `externalRestApiBaseUrl` + `appName` with `rpcProxyBaseUrl: ''` and then best-effort discovery. The how-to collapses a three-step, email-gated flow into "detection on launch." The member-registration callout itself is accurate and the anchor `#the-member-registration-service-gap` resolves.

#### R1-M3 — Localhost is not "always at the bottom, unselected in packaged builds"

`buildRuntimeConfigOptions` appends Localhost after stored profiles, so it is last **only when it is included**. `getDesktopRuntimeConfigState` does pass `includeLocalhostOption: true` today, so the list order claim holds for the current caller — worth stating as current behavior, not as an invariant, because the helper can omit it.

"In a packaged build it stays unselected until you pick it" does not match `hydrateDesktopRuntimeConfig` / the module init in `config.ts`:

- A persisted `activeProfileId` of `__localhost__` selects Localhost in a packaged build.
- `envRuntimeConfigured` is correctly packaged-off (`!app?.isPackaged`).
- `shouldPreferLocalhostByDefault()` is the dev-only path (`!app?.isPackaged` && `ELECTRON_RENDERER_URL` && localhost env config). That part of the sentence ("running through local dev commands selects it by default") is right.
- Packaged first launch with no stored profile sets `activeRuntimeOptionId = null` and `desktopRuntimeConfigured = false`. Localhost is in the list and not auto-selected. That is narrower than "stays unselected until you pick it," which is false after the user has selected it once (the choice is persisted).

#### R1-L1 — "After you authenticate" and "different tenants can be on different versions" overclaim the check

`loadSession` (`useAppController.ts`) calls `refreshDesktopReleaseStatus()` once, fire-and-forget, when a session is already authenticated. It compares `compareSemverLike(currentVersion, policy.minimumDesktopVersion) < 0` against the **currently selected** environment's `GET /api/v1/desktop/release` (`requireAuth`). There is no `isPackaged` gate. The PR note that the design spec's "packaged-only" claim was not copied is correct, and the how-to's softer "a packaged distribution is where this matters most" matches that.

What is missing:

- The check does not re-run on environment switch. `handleSelectRuntimeConfig` does not call `refreshDesktopReleaseStatus`. A stale `updateRequired` from the previous tenant stays until the next `loadSession`. "Different tenants can be on different versions at the same time" implies the policy tracks the selection; it tracks the last session load.
- Failures are swallowed (`refreshDesktopReleaseStatus` returns `null` on throw). A tenant with a broken release endpoint silently skips the gate.
- Below-minimum blocks the signed-in shell with a dialog. At-or-above minimum is silent. The dialog copy says the workspace requires "the latest version," while the predicate is "below `minimumDesktopVersion`," which may be older than `desktopVersion` (the "Latest" line in More info). The how-to's "below that minimum" is the accurate one; don't let a later edit drift toward "not on latest."

#### R1-L2 — Deep link is not applied until the user confirms, and the Profile UI path is easy to miss

Path 2's click path matches `SettingsContent.tsx`: **Setup desktop app** opens a modal; copy puts the External REST API URL on the clipboard; **Open desktop app and setup** navigates to `evenfire://desktop-environment?...`. Good.

The app does not "save or update" on handling the link. `onDesktopEnvironmentSetup` only sets `pendingDesktopEnvironmentSetup`. `App.tsx` then shows a confirm dialog ("Only continue if you trust this External REST API host") before `handleConfirmDesktopEnvironmentSetup`. A how-to that skips the confirm step will send operators looking for a save that never happened.

`CLERUM_DESKTOP_CONFIG_PATH` is mentioned as a fleet pre-seed. In `saveDesktopRuntimeConfig`, a non-empty `CLERUM_DESKTOP_CONFIG_PATH` replaces the entire stored profile list with a single `custom-file` entry. That is a different contract from "also pre-seeded," and this page sends the reader to the surface doc instead of saying so. Acceptable only if the surface doc actually says the override is exclusive — worth one clause here so the how-to doesn't imply it merges with saved environments.

#### R1-L3 — Index rows are slightly inconsistent, not wrong

- `docs/README.md` how-to table row is aligned with the surrounding rows.
- `docs/llms.txt` uses the same relative path style as its siblings.
- Learning path A, step 2, now has two sentences but still no terminal period on the new sentence, and the step is a different task from "tour the Desktop App." A separate numbered step would match the path's "do this, then this" shape better than a subordinate clause.
- `docs/surfaces/desktop-app.md` still says code signing, notarization, and auto-update are not covered here, then links the how-to for how a packaged app surfaces a required update. That split is correct. The new sentence says "a packaged app," which is the audience that cares, not a claim that the check is packaged-only. Fine.
- `docs/surfaces/profile-ui.md` "Next" link is accurate. The profile-ui page body still doesn't mention **Setup desktop app** at all; the how-to is the only place that procedure lives. Not a defect in this PR, but the cross-link is doing more work than a "see also" usually does.

### What checked out

- Scheme is `evenfire://` for `desktop-environment` and `desktop-setup`; `clerum://` is registered alongside it (`main.ts`). The body is right not to document `clerum://` as the setup scheme.
- `GET /desktop/environment` is unauthenticated and returns `appName`, `externalRestApiBaseUrl`, `rpcProxyBaseUrl`. `GET /desktop/release` is `requireAuth` and returns `minimumDesktopVersion`. The how-to's "External REST API" shorthand hides the `/api/v1` prefix, which is fine for a how-to if it stays user-facing — the body is the place that says `GET /desktop/release`.
- Manual save rejects an empty name or an empty External REST API URL.
- Update predicate is installed `< minimumDesktopVersion`, not "not equal to latest."
- No `isPackaged` gate around `getDesktopReleaseStatus` / `refreshDesktopReleaseStatus`.
- Anchors `#ship-it-to-your-users` and `#the-member-registration-service-gap` exist on `docs/surfaces/desktop-app.md`.
- No secrets, no tenant hostnames, no new runtime behavior.

### Verdict

Do not treat this page as the verified contract the PR body describes. The environment model (save the External REST API URL; RPC proxy is discovered when it is empty) is the right idea and is worth keeping. Before relying on the page, correct the release tag to `v<version>` (the code on `main` already did — the doc did not), state that the update dialog opens the URL only when the user clicks **Download update**, and rewrite the three setup paths so they include the confirm step, the optional RPC field, the invitation email step, and the fact that an existing profile's RPC proxy is overwritten with empty rather than cleared and refreshed.

This PR is already merged, so the fix belongs in a follow-up docs commit, not a re-open.

---

*Review was done by Evenfire Reviewer.*

PR #136, opened 2026-07-22
refactor: make LLM secrets editor additive per provider

zach88:
## What
Converts the LLM secrets editor (`/secrets/llm` update modal and `/secrets/new` create flow) from rendering all ~21 provider credential groups unconditionally to an additive model: only providers with a stored key, a non-empty draft value, or manually added via a new "Add provider" picker are shown.

## Why
After #788 the agent configuration flow became additive and clean, but the secrets editor still rendered every provider group, producing a very long, noisy form even when the operator uses a single provider. This aligns both flows on one UX pattern.

## How
- `LlmCredentialFields` derives visible providers from three sources: `existingKeys`, non-empty draft values (also self-heals the create-flow step remount), and a session-local `manuallyAdded` set. The "Add provider" picker lists only non-visible providers and hides when exhausted.
- New pure helper `providerForDataKey()` in `lib/llm.ts`, prefix-aware, so extra slots (`claude-api-key-fb1`, `aws-access-key-id-fb1`, …) resolve their provider; secrets holding only an extra slot still surface their section.
- Existing extra slots are seeded into component state, making stored non-canonical keys visible/editable (previously invisible — pre-existing gap).
- Providers added in-session with no committed value can be removed from view; removal cleans the draft via `removeExtraSlot` (no orphan keys) and the gate also checks committed extra-slot state to cover custom-named slots.
- Payload shape (`POST` / `PUT merge:true`), slot-aware validation, and both call-site contracts are unchanged. Shared types extracted to `LlmCredentialFields/types.ts`.

## Testing
- `npm test` (Vitest): 111 files / 1056 tests passing, including the rewritten `LlmCredentialFields` suite (20 tests: additive default, prefix derivation, removal gate, step remount, extra-slot seeding).
- `next build` (with type-check): green.
- Reviewed by code-review + security-review agents; the one Medium finding (removal gate missing custom-named committed extra slots) was fixed and re-verified.

## Risks / Follow-ups
- Low risk: presentation-only change; backend contract untouched.
- Follow-up candidates: custom-named extra slots don't survive a create-flow step remount in the UI (value still submits — pre-existing); "renaming" a stored extra slot leaves the old key in the Secret (`merge:true` never deletes).


jozer-rami:
**Quick review pass (AI-assisted).** LGTM with nits. The main risk for an "additive per provider" editor is clobbering other providers' stored secrets, and that path was traced end-to-end and is clean: both payload call sites (`control-ui/components/SecretsTable.tsx:184-210`, `control-ui/app/secrets/new/page.tsx:217-243`) filter empty drafts before building `stringData`, hidden providers never fire `onChange`, the UI never sends `removeKeys`, and the server-side `merge:true` overlay (`control-api/src/routes/admin/secrets.ts:164-172`) preserves keys not present in the payload. No clobber path found.

Also checked and clean:
- `isProviderRemovable()` gating (`LlmCredentialFields/index.tsx:241-251`): a provider with a stored key can never appear in `manuallyAdded`, so the remove button cannot show next to real data. Matches the fix described for the self-reported finding.
- Secret handling: no logging or echo of values, `seedExistingExtraSlots()` seeds names only with empty values, consistent with the write-only contract, and extra-slot state is fully dropped on removal.

One follow-up worth a ticket (author already flagged it as pre-existing): a custom-named extra slot that `providerForDataKey` cannot attribute loses its UI row on a create-flow step remount while its value silently survives in the parent draft and still gets submitted. Not a regression from this PR, but the silent-survival part is the surprising half.

josue-tech-94 (dismissed):
**Stored extra credentials are not actually removed or renamed**
Stored extra slots are initialized with existingKey set and committedKey: null. Therefore, removeExtraSlot() only removes the React row and never communicates the stored key’s deletion.
SecretsTable.saveSecret() subsequently sends merge: true without removeKeys, so control-api preserves the original Kubernetes Secret key. Renaming has the same problem: the new key is written, but the original existingKey remains.
I reproduced this with a focused component test: clicking remove for claude-api-key-fb1 resulted in zero onChange calls.
Please track removed/renamed stored keys and include them in the update request’s removeKeys. For rename, retire the original key only after the replacement name and value are valid. Please also add regression tests covering removal and rename of seeded stored extras.

zach88:
@claude take a look at this, consider the existing comments in this PR

zach88:
@jozer-rami thanks for the review pass — re: the follow-up you flagged (a custom-named extra slot losing its UI row on a create-flow step remount while its value silently survives in the draft and still gets submitted): agreed it's real, but it's out of the scope of this PR, so we opened #203 to track it. The issue carries the traced mechanics and acceptance criteria for both the attributable and the unattributable key cases.

josue-tech-94 (approved):

PR #147, opened 2026-07-23
fix(hcc,rpc-proxy): per-host reconciliation priority, principal-scoped wake, and delete convergence

alfredolopez80:
Port of an upstream change: HCC per-host reconciliation priority, principal-scoped wake coordination, and delete convergence.

## What this is

A port of 16 commits that fix three validated defects in the stateless-host lifecycle. The two repositories share no git history and their `dev` branches have genuinely diverged, so this is **not a mechanical cherry-pick**: 6 of the 16 commits conflicted and were resolved by adapting each change's intent onto this tree's code shape. Every resolution is described in its own commit message.

## The three defects being fixed

1. **Global head-of-line blocking in HCC.** Every per-Host watch event, retry, and fleet pass shared one process-wide promise tail, so a slow Host, an API timeout, or a full-fleet pass delayed unrelated Hosts. Replaced by three lanes: deduplicated LIST→WATCH authority recovery, urgent per-Host admission through the existing per-Host serializer, and a bounded fleet worker pool (`HCC_HOST_FULL_RECONCILE_CONCURRENCY`, integer 1..8, default 2, fail-loud validator). Same-Host ordering, watch-generation and uid fences, and cleanup-authority guards are preserved; `reconcileDelete` now routes through the per-Host chain.
2. **Wake-authorization displacement in rpc-proxy.** The coordinator keyed only by `hostRef` and selected the latest authorization by expiry, so a non-wake or different-principal token could displace the only wake-capable one — and the resulting 403 collapsed every waiter on that host. Now keyed by principal `{typ, sub, accessScope, teamId ?? '', hostRef}` with a non-displaceable wake-capable slot, readiness probed via the unauthenticated runtime health endpoint, one-shot waiter settlement, and a graceful shutdown path.
3. **Desktop wake scopes.** `host:wake:write` is requested on finite operations only — and deliberately **not** on status/health/activity reads, SSE streams, McpServer catalog/invocation, Desktop view, or Sandbox UI, so polling and observability cannot defeat scale-to-zero.

Plus: delete-intent compensation across watch retirement with a uid identity fence and all-kinds orphan discovery; and a workspace-layout migration fix (below).

## Read this before reviewing: what did NOT transfer

This port is **not pre-validated**. The upstream change was validated against upstream's tree with a full cluster-backed campaign; that evidence does **not** carry to this repository. Three areas are materially different code here and need their own runtime validation:

**1. Session routes in `rpc-proxy/src/routes/rpc.ts` — new behaviour, not ported behaviour.**
`GET /sessions` and `GET /sessions/:agent/:chatId/messages` are still the plain-passthrough shape in this tree and had **no wake-and-hold at all**. The upstream commit's delta assumed a route shape that does not exist here, so the wake-and-hold wiring on those two routes was **written against this tree's shape**, not ported from running code. It fires on network-down and on a `503 host_draining` body via `sessionDrainingFence`. This is the single most review-worthy hunk in the PR.

**2. `host-context-controller/src/statelessDeployment.ts` — carries an out-of-scope upstream commit, and it can now block pod start.**
This tree still carries the original unguarded script: `mkdir -p` followed by an unconditional `mv "$root/workspace/$db" "$root/state/$db"`, which silently overwrites `state/state.db` on a stateful↔stateless round-trip. Upstream's newest-wins fix is *layered on top of* a different, earlier commit — **`6be3531ba` "fix: close stateless session continuity review gaps"**, which is **not part of the ported PR** but is an unavoidable prerequisite: it introduces `buildWorkspaceLayoutInitScript`, and the newest-wins delta is meaningless without it.

That prerequisite converts an unconditional `mv` into a **validate-then-move** script with new **fail-closed `exit 1`** paths: symlinked root, symlinked legacy source, non-directory managed path, duplicate non-SQLite destination, and split SQLite artifacts. **Layouts that previously migrated silently will now block stateless pod start.** That is the intended behaviour (the alternative is silent data loss), but it is a real operational change on every stateless pod start in this repository and deserves review on the actual runtime image — particularly the `date -u` and backup `mv` paths. The accompanying 331-line test executes the generated script against temp directories and asserts real filesystem state (12 cases including symlink escape, split SQLite, WAL/SHM lockstep, backup uniqueness) rather than asserting on the string.

**3. Stateless channel / cache-authority semantics — only half the upstream model exists here.**
Upstream defends an invariant: an *unknown* CommunicationChannel cache state must never flip a Host's template; only a **confirmed** association hard-rejects. In this tree, `!isCommunicationChannelCacheSynced()` is itself a hard rejection — both `effectiveLifecycleFromCache` and `communicationChannelPolicyRejection` flip an unsynced-cache Host to stateful-active. That is the invariant violation upstream guards against.

**It was deliberately left in place, and a third option was checked and does not exist here.** The conservative gate — keep the template stable but forbid suspension — needs a suspension-blocked signal. This tree's `EffectiveHostLifecycle` is `{stateless, state}` with no `suspension` field, and the tracker gates only on `getEffectiveLifecycle`, so returning `{stateless: true, state: 'active'}` would **not** block a suspend: an idle stateless *active* Host is precisely the suspend candidate. Building the gate means importing the suspension-block model wholesale.

Removal was rejected because taking the flip out without upstream's suspension-block machinery (`hostLifecycleDecision`, the `effective.suspension` model, the `StatelessSuspensionBlocked` condition, `status.lifecycle.effectiveMode` — none of which exist here) would leave an unsynced-cache stateless Host *suspend-eligible on unproven evidence*, which is strictly worse than the current conservative behaviour. Porting `hostLifecycleDecision` is a separate change.

**Operational consequence, stated explicitly.** This tree couples the storage layout to the *effective* lifecycle (upstream decouples it via `usesStatelessStorageLayout`, keyed on `spec.lifecycle.stateless`). Combined with the cache-unsync flip above, that means: **on an HCC restart or watch re-list window, a stateless Host can roll to the non-stateless template** — no initContainer, no dual subPath, no `CLERUM_SESSION_DB_DIR` — so mcp-host writes `state.db` at the PVC root. When the cache syncs it rolls back, and the init script hits exactly the root-vs-`state/` collision this PR's newest-wins resolution handles. **Non-destructive by design** (nothing is deleted; the displaced file becomes `state/state.db.pre-<ts>.bak`), but operators will observe restart churn and an accumulating set of `.bak` files until the lifecycle-decision model is ported. This is pre-existing behaviour in this tree — the port does not introduce it — but the port is what makes it routinely reachable. What did land is the confirmed-association half: the rejection message now names the channel count **and** the disassociation recovery action, which is the operator-visible string the UI renders verbatim.

## Two ported gates assert on a field this repository does not have

`status.lifecycle.effectiveMode` is not part of this repo's `HostLifecycleStatus` (`host-context-controller/src/types.ts`) and is never written by this tree's HCC. Two ported gates read it:

**`scripts/e2e/e2e-stateless-suspend-wake.sh` — Test 2d-pre — FIXED in this PR.** This gate is not opt-in: it ran unconditionally and its `exit 1` **aborted the entire script**, so every check after it — including 2d-pre's own recovery assertion and the whole Desktop Playwright phase — never executed. That is a truncated suite, not one red test, and a permanently-red gate carries no information: after the first run it is indistinguishable from a broken one. The `effectiveMode` conjunct is now skipped while the field is absent and **auto-arms if it is ever published**. Nothing was weakened — every other conjunct (the `StatelessEnableRejected` condition, `replicas: 1`, the stateful pod template) already proves the hard rejection, and the reason-text check that follows is unchanged.

**`control-ui/e2e/host-stateless-channel-rejection.spec.ts` — knowingly red, left red.** This one *is* opt-in (`E2E_STATELESS_CHANNEL_REJECTION=1`) and therefore already quarantined, so the assertion stands rather than being weakened. Its docstring, however, claimed the spec was red "until the Addendum-6 executor fix lands" — that fix is *in this port*, so the claim would have become false on merge and sent the next reader chasing a phantom. **The docstring is corrected in this PR** to name the real blocker. The hard rejection itself does work here; it is observable through the condition, the pod template and the banner text.

## Deliberately dropped

Each of these tests or threads a subject that does not exist in this tree; none is referenced by anything that landed: `parseCommunicationChannelCacheRecoveryRetryMs`; the session telemetry emissions (`emitSessionTerminal`, `sessionOutcome`, `sessionExceptionOutcome`, `fetchSessionResponse`) and their `session read telemetry contract` test block; `sessionWakeTerminalOutcome`; upstream's `blockedHost` / `confirmedIncompatibilityHost` fixtures; and the `ReconcileMutationDiagnostic` parameter threading (the commit's lane accounting was applied on top of this tree's one-argument signatures instead).

## Validation performed here

Unit and typecheck only, measured against a baseline on `dev` with the new test files removed, so every delta is attributable:

| Service | Baseline | After port | Typecheck |
|---|---:|---:|---|
| host-context-controller | 933 | **993** (51 files) | clean |
| rpc-proxy | 375 | **423** (31 files) | clean |
| desktop-app | 1040 | **1057** (124 files) | clean |
| control-ui | 1043 | **1045** (111 files) | clean |
| control-api | 3371 | **3373** (11 skipped, 0 failed) | — |

Zero failures. `npm ci` clean on all five services, no lockfile mismatch. The three ported bash e2e scripts pass `bash -n`. `HCC_HOST_FULL_RECONCILE_CONCURRENCY: "2"` landed in `deploy/base/control-plane/host-context-controller.yaml` and the new metrics are registered.

**Unit-green here is not evidence that the cluster-level e2e gates would pass**, and two of them provably will not (above). No cluster-backed validation was run against this repository.

Two pre-existing, unrelated conditions observed while validating: `control-api` needs `packages/workflow-runtime-core` built locally (`npm run build` there) or 42 files fail to resolve the workspace package; and `host-context-controller`'s `npm run lint` fails on an ESLint flat-config migration error.

## Deploy surface

One change: the `HCC_HOST_FULL_RECONCILE_CONCURRENCY` environment variable on the HCC Deployment, inherited by all overlays. **No CRD, RBAC, NetworkPolicy, auth-endpoint, grant-model, or login-path change.**

## Where reviewers should spend their attention

1. The session routes' newly-written wake-and-hold wiring (`rpc.ts`).
2. The workspace-layout init script replacement, on the real runtime image.
3. Whether to follow up by porting `hostLifecycleDecision` — and until then, that the two `effectiveMode` gates are expected red.
4. The `ActiveCommunicationChannels` rejection message string changed; anything matching the old text downstream will break.
5. `hostReconciler.ts` lane metrics are wired without upstream's diagnostic threading — check any dashboard or alert expectations.


jozer-rami:
**Quick review pass (AI-assisted).** Strong work overall: the lane split, principal-scoped wake keying, and the adversarial test suites all check out. Findings for your evaluation:

**Medium: missing TOCTOU recheck on the recovery-triggered delete path.** The orphan-cleanup path (`collectHostCleanupFailures`) re-checks the live cache *inside* `serializeByHost` right before deleting (the "F2 / #827 TOCTOU" guard, with tests). The new `enqueueRecoveredHostDeletes` / `dispatchRecoveredHostDelete` path in `host-context-controller/src/k8sClient.ts` has no equivalent: `reconcileDelete()` runs `deleteHostRuntimeResources` unconditionally once admitted to the per-Host serializer. Since the dispatch is fire-and-forget and races any watch-driven ADDED for the same name, a Host deleted then quickly recreated with the same name during the LIST→WATCH recovery gap could have its freshly created Deployment/PVC/Secrets wiped. Narrow window, high impact, and not covered by a test (the existing delete-serialization test only pins the opposite ordering). Suggest adding the same in-serializer recheck, or an explicit comment accepting the risk.

**Low: stale comment.** The "Wake fast-path" comment in `reconcileCore()` (`host-context-controller/src/hostReconciler.ts`) still says the Host watch callback is serial across the whole fleet, which is exactly the model this PR removes. Worth updating so it doesn't mislead future readers about the new concurrency model.

**Low: naming nit.** `rpc-proxy/src/services/wakeAndHold.ts:81` `trackedHostCount()` now counts principal+Host coordination entries, not Hosts. Test-only today, but misleading if it ever becomes a metric.

**Reminder (author already flagged):** the `ActiveCommunicationChannels` rejection message string changed; please double-check any downstream string matching (control-ui banner, alerts) against the new text before merge.

alfredolopez80:
@jozer-rami — thank you for the thorough pass. All four findings are addressed in this PR; details and evidence below.

**Medium — TOCTOU on the recovery-triggered delete path** — **Fixed** (`15bdfb79`). You were right that `dispatchRecoveredHostDelete` → `reconcileDelete` had no equivalent of the F2/#827 in-serializer recheck that `collectHostCleanupFailures` carries, and that it was the only unfenced `reconcileDelete` caller. `reconcileDelete` now takes an optional `skipIf` fence evaluated **inside** `serializeByHost`, immediately before anything destructive; the recovery path passes `skipIf: () => this.hosts.has(name)`, so a same-name Host recreated during the admission window is not wiped. On skip it emits `hostDeleteCleanupTotal{outcome:'superseded'}` — the same label the watch path already uses. A red-first test was added covering exactly the gap the existing test missed (recreation *during* the admission window, not visible at snapshot time). Runtime-verified: driving the recovery-delete path deliberately made `clerum_hcc_host_delete_cleanup_total` emit `queued/confirmed/completed 0→1` with the bundle cleaned and 0 HCC restarts. The underlying defect pre-dates this PR (it is byte-identical upstream) and is recorded as #152.

**Low — stale "serial across the whole fleet" comment** — **Fixed** (`e177cb8f`), reworded to per-Host serialization.

**Low — `trackedHostCount` misnomer** — **Fixed** (`9a9de828`). Verification found it is more than a nit: `entries.size` also backs the capacity limit at the `maxTrackedHosts` guard, so the cap is per-(principal, Host). Renamed both surfaces: `trackedCoordinationCount` / `maxTrackedCoordinations` / `MAX_TRACKED_WAKE_COORDINATIONS`. Confirmed no env var, ConfigMap or manifest references the old name.

**Reminder — `ActiveCommunicationChannels` message string** — Verified clean: the change is a pure suffix append (old text is a strict prefix of new), production control-ui branches on `condition.type` not the message text, and every string-matching consumer already carries the new text. No downstream breakage.

**Runtime validation** was run against a live single-node minikube profile (rebuilt HCC + rpc-proxy images from this head; image chain of custody verified). The recovery-delete fence, the storm/lane-accounting gate (22/22), the rpc-proxy rename inertness (cold `/sessions` wake, generation advance), and the `/sessions` transcript route (`.turns` payload proven both warm and on a cold wake, generation 146→147) are all RUNTIME-PROVEN. One item is named rather than claimed: fix 1 case (b), the `superseded` recreation race, is UNIT-PROVEN + CODE-DECISIVE but RUNTIME-UNPROVEN — its sub-second in-HCC window was not hand-landable on a single-node profile after 5 attempts, and forcing it would require a test-only hook in production code, which we don't add. The fix logic itself is correct and covered; only the live observation of that specific race is missing.

Branch is synced to current `dev` (CI green on the synced head) and **ready for review**.

zach88:
Heads-up from a review pass, not a blocker — just flagging one thing to sanity-check.

The new `gatherCleanupCandidates` orphan sweep in `hostReconciler.ts` now issues `listNamespaced*` calls on **ServiceAccount / Role / RoleBinding / PersistentVolumeClaim** (plus Deployment/Service) in the host namespace. Those are new `list` verbs on the HCC hot path, so if the SA's Role didn't already grant `list` on all of them, cleanup would just 403. It's fail-closed (the error goes to `listFailures` and cleanup is deferred, never a fail-open delete), so worst case is a silently-deferred sweep rather than anything unsafe — but a permanently-deferred cleanup is easy to miss.

I checked, and we're actually fine: the `host-context-controller-host-runtime` Role in `deploy/base/mcp-host/rbac.yaml` already grants `list` on `serviceaccounts`, `roles`, `rolebindings`, `persistentvolumeclaims`, `deployments`, and `services` in that namespace, bound to the `host-context-controller` SA. So no 403, no RBAC change needed.

Leaving this here mostly as a paper trail: if the sweep ever grows to a kind that Role *doesn't* cover, it'll fail this same quiet way. No action needed for this PR.

zach88 (approved):

alfredolopez80:
@zach88 — thanks, and confirmed: no RBAC change is needed. I ran an adversarial pass specifically trying to find a namespace gap, and it holds — but the paper trail is worth making precise, because the sweep is a bit wider than the host namespace.

**The real sweep spans three namespaces, not one.** `gatherCleanupCandidates` (`hostReconciler.ts:4001-4127`) issues 11 `list` calls:
- `mcp-host`: Deployment, Service, PVC, ServiceAccount, **Secret**, Role, RoleBinding, **NetworkPolicy**
- `channels`: Deployment, Service, Secret, NetworkPolicy
- `rpc-proxy`: NetworkPolicy

So the all-clear rests on **three** Roles, all bound to the `host-context-controller` SA, not just the host-namespace one:
- `deploy/base/mcp-host/rbac.yaml` → `host-context-controller-host-runtime`
- `deploy/base/channels/rbac.yaml` → `host-context-controller-channels-runtime`
- `deploy/base/rpc-proxy/rbac.yaml` → `host-context-controller-rpc-proxy-runtime`

Each grants `list` on the kinds it lists in that namespace (roles/rolebindings under the correct `rbac.authorization.k8s.io` apiGroup), so every call maps to a grant — no 403.

**One correction to the "new list verbs" framing:** the sweep *code* is new, but it introduces **no new RBAC verb**. All three RBAC files are unchanged since the merge-base, and HCC already held `list` on every one of these kinds for its normal reconcile/informer/provisioning paths — the new code reuses pre-existing grants. So there's nothing a `base_sha` bump needs to carry for RBAC.

**Fail-closed confirmed rigorously:** each list is individually try/caught into `listFailures`, and deletion is never driven by absence-in-list — a candidate only becomes a delete after owner-label derivation, cache absence, a **fresh authoritative 404**, `destructiveCleanupAllowed`, and an in-serializer TOCTOU re-check. A failed list in any namespace can only under-clean (deferred), never fail-open.

Your forward-looking point stands and is the useful takeaway: if the sweep ever grows to a kind/namespace not covered by one of these three Roles, it'll fail this same quiet, deferred way — so a new listed kind should come with its matching grant in the same-namespace Role.

PR #158, opened 2026-07-23
feat(gfs): per-agent access and atomic recursive copy

### LINKED ISSUE #156: Per-agent GFS access + atomic recursive copy
## Summary
Add per-agent GFS (Global File System) access so a user's autonomous agents can read/write GFS files directly, plus an atomic recursive copy operation.

## Motivation
Today only users and teams can be granted access to GFS resources. Agents (hosts) need direct, audited access to act on files on the user's behalf. This introduces managed per-agent grants with a strict read/write capability, hot revocation, and an atomic copy tool.

## Scope
- Managed `1st:<ns>/<name>` host subjects with a read/write grant cap
- Audited create/revoke + hot (steady-state) revocation
- Atomic recursive copy with a permission-epoch TOCTOU guard
- Reader/writer database role split
- Agent-side `clerum__gfs_*` tools
- control-api migrations 0071-0074

## Notes
Port of internal, already-validated work (unit + typecheck + runtime E2E green). The PR replicates the exact commits onto the OSS base (the #791 reconciliation branch).

alfredolopez80:
Port of internal work (issue-797 in the private repo) — per-agent GFS access + atomic recursive copy.

Closes #156

## What
Replicates the validated per-agent GFS backend onto the OSS base (stacked on the reconciliation branch, PR #147):
- Managed `1st:<ns>/<name>` host subjects, read/write grant cap, audited create/revoke, hot revocation
- Atomic recursive copy with a permission-epoch TOCTOU guard
- Reader/writer database role split
- Agent-side `clerum__gfs_*` tools
- control-api migrations 0071-0074
- External grants/shares rate limiter

## How this was ported
The exact changes were replicated (net diff, grouped into 3 thematic commits). Internal-only CI/deploy infra was intentionally omitted for the OSS repo.

## Validation
Per-service `npm test` + `npx tsc --noEmit` green: control-api, gfs-controller, host-context-controller, mcp-host, external-rest-api.

## Base
Stacked on `claude/mcp-reconciliation-wakeup-50fdd7` (#147). Retarget to `dev` when #147 merges.


zach88:
Did a full pass with our code-review + security-review agents over the PR's own diff (isolated against `dev` — merge-base is dev's tip now that #147 landed, so this is just the 797 work, not #147's).

**TL;DR: looks great, no blockers.** The load-bearing stuff (atomic recursive copy, permission-epoch TOCTOU guard, blob staging, migrations 0071-0074, reader/writer role split, grant hardening) is solid and fail-closed all the way through. Security verdict came back SAFE — no new exploitable authz/traversal/injection/secret issues.

Build + tests I ran green:
- `gfs-controller`: `tsc --noEmit` OK, vitest → 468 passed / 3 skipped
- `control-api`: `tsc` OK, touched suites (grants, seed/tx, withTransaction, migrations 0071-0074, externalGfs, legacy-standalone, mcpInvocable, accessReconciliation) → 148 passed
- `mcp-host`: touched suites (gfs tools, gfsClient, nativeToolRegistry scopes) → 34 passed

A few non-blocking notes for your call:

**Should-fix — observable contract change** · `control-api/src/routes/external/teams.ts:129-159` & `users.ts:217-248`
Dropped the K8s-reconciliation filtering of `agentNames`. On K8s success it now returns only resolved Hosts; on K8s failure it returns all DB-authorized names. Reads as a deliberate security fix (don't enumerate hidden Hosts) and tests were updated — just flagging it's a behavior change for any client consuming `agentNames`, so land it knowingly.

**Low — `getGfsToolScopes` too strict** · `mcp-host/src/internalTools/gfsClient.ts:30-52`
If the token carries *any* scope outside `{gfs.read, gfs.write}` it returns `null` and hides *all* GFS tools. Safe today since the policy only mints those two, but adding a new scope later would silently nuke every GFS tool. Suggest `scopes.filter(isKnown)` instead of rejecting the whole set.

**Low — rename vs copy TOCTOU asymmetry** · `gfs-controller/src/db/renamePublication.ts`
Rename doesn't re-check the permission-epoch inside the txn (copy does). Fine because it's a single short transaction with no async blob-staging gap (few-ms window), but worth a comment documenting the asymmetry in case rename grows.

**Low — possible fd leak on the legacy copy path** · `gfs-controller/src/db/copyBlobStaging.ts:61` + `copyPublication.ts:296-312`
For a `legacy_flat` source, the `source.reopen()` stream may not get destroyed if `writeImmutable` throws before consuming it (the `generation` path is covered by the outer catch's `generationStream?.destroy()`). Depends on whether `writeImmutable` destroys its input on error — worth a quick check on the reopened-stream cleanup in the legacy error path.

**Nits (no action needed):**
- `copy.ts:244-252` — the write-only-folder name-enumeration oracle (copy doesn't require `read` on the dest parent) is noted as intentional/risk-accepted per #797. Just leaving it visible.
- `GET /gfs/grants/legacy-standalone` has no rate limiter, but it's behind `requireAuthForControlUI` + `LIMIT 1000`, so only exposes inventory to authed admins. Fine.

Nice work — the TOCTOU handling (epoch captured pre-authorize, revalidated under `pg_advisory_xact_lock` + `FOR UPDATE` before insert) and the least-privilege reader role are both clean.

alfredolopez80:
Thanks @zach88 — did an adversarial pass over each item (default: refuted unless the code proves a real defect). **One confirmed and fixed in `2d1e78382`; the rest are intentional and I'm landing them knowingly.**

### #4 (fd leak, legacy copy path) — confirmed, fixed ✅
Real. `stageCopiedBlob` reopens the legacy source stream itself, then hands it to `writeImmutable`, whose `pipeline()` destroys its input only *once reached*. But several guards throw **before** `pipeline` runs — the `checkDeadline()` between `reopen()` and the write, and `writeImmutable`'s own parent/symlink/`O_EXCL` and deadline checks. On those paths the reopened fd had no owner in any `catch`: the generation path is covered by `copyPublication`'s `generationStream?.destroy()`, but the legacy stream is reopened *inside* `stageCopiedBlob`, so it needed an owner there. Fix tracks the reopened legacy stream and destroys it in the catch (idempotent once piped). Added two regression tests — legacy stream destroyed on a pre-`pipeline` throw, and a caller-owned generation stream is **not** touched by the stage path. `gfs-controller`: `tsc --noEmit` clean, 470 passed / 3 skipped.

### #3 (rename vs copy epoch re-check) — intentional, now documented
Agreed it's asymmetric by design. `publishCopy` re-checks the permission epoch because it stages blob bytes *outside* its commit txn (an async gap where a revocation can land); rename has no such gap — it authorizes, then runs a single short structure txn under advisory lock + `FOR UPDATE`. Added a comment documenting the asymmetry plus the "if rename grows an async pre-commit phase, add the revalidate" note.

### #1 (agentNames K8s filter) — intentional, landing knowingly
No cross-tenant exposure: the boundary is `getUserAgents(userId)` / `getTeamAgents(teamId)` behind `requireExternalUserParamMatch` / `requireExternalTeamParamMatch`, so the K8s-failure path only ever returns names already authorized **for that caller**. K8s-up hides DB-authorized-but-unreconciled Hosts (don't enumerate hidden Hosts); K8s-down preserves the caller's own authorized names without enriched DTOs (spec §7.2 graceful degradation). The behavior change for `agentNames` consumers is real but security-positive, and it's already called out in the route comments.

### #2 (getGfsToolScopes fail-closed) — keeping fail-closed
It's explicitly **not** an authz boundary (gfsc re-verifies the same bearer token). Given that, a token carrying a scope we don't recognize is anomalous, and hiding all tools until this code is updated is the defensive default we prefer over partially trusting it. `scopes.filter(isKnown)` is reasonable for forward-compat — I'd just rather add a new scope here deliberately than silently accept an unrecognized one. Happy to switch if you feel strongly.

### Nits
Write-only-folder name-enumeration oracle stays (intentional / risk-accepted per #797); `legacy-standalone` GET stays as-is (behind `requireAuthForControlUI` + `LIMIT 1000`).

CI is green on the branch after the fix (19/19 checks).

Thanks again for the thorough pass — the epoch/TOCTOU and reader-role review was exactly the right place to push.

zach88 (dismissed):

alfredolopez80:
## Status summary — green, pending one re-approval

**All required checks are green** (Build / test / typecheck across all services + Repo hygiene), the branch is up-to-date with `dev`, and there are no unresolved review threads.

### The red "CodeQL" check is a non-blocking false positive
Verified `isRequired=false` — it is **not** a required status check and does not gate the merge. CodeQL's `js/missing-rate-limiting` does not model the repo's custom `rateLimitMiddleware`, so it flags every route that uses it. This is the **same false positive already dismissed** on `grants.ts` (#777–780) and `shares.ts` (#773–776).

The `GET /gfs/grants/legacy-standalone` route **does** enforce the PG-backed token-bucket limiter (bucket `gfs_grants_legacy_report`, 30/min per admin, **fail-open** — no operational impact; it never blocks a legitimate request, only caps abuse). Alert #795 was dismissed with that rationale; #793 and #794 are fixed.

### Only remaining gate: one fresh approval
The earlier approval was auto-dismissed when the PR base was retargeted during a `dev`→`main`→`dev` branch promotion. `reviewDecision` is now `REVIEW_REQUIRED`. Nothing on the code or review surface changed — the dismissal was purely a side effect of the base change. One re-approval unblocks the merge.

zach88:
Re-checked the latest delta (the CodeQL fix `f2dfeb0` + the two real-PG test commits). **The CodeQL fix is real, not just alert-silencing — verified end to end.**

Poked at the spots where a "resolve CodeQL alert" commit usually goes cosmetic, and all four came back clean:

- **Middleware order** — `requireAuthForControlUI` sets `req.adminAuth` before chaining, and the limiter runs after it, so `adminAuth.sub` is always present when it keys. Same ordering as the sibling `grants.ts`.
- **`getBucketKey` → `null` fail-open** — that branch is unreachable here: auth already requires a truthy `sub` (`findAdminById(claims.sub)`), so the key is never null and the limiter always enforces. No silent bypass.
- **`bucketType` registration** — the token bucket keys off the string `bucketKey`, not `bucketType` (which only feeds the metric label), so no enum/table needs updating and the limiter isn't silently "unrecognized".
- **Disjoint budgets** — `gfsgrants-legacy:${sub}` vs `gfsgrants:${sub}` are separate key namespaces, so the read-only report doesn't eat the grant-mutation budget. Matches what the commit claims.

The two test commits aren't tautological either: the concurrency one runs two writers via `Promise.allSettled` against real PG and asserts exactly-one-winner + version bumps once + served bytes == winner's (never a mix); the `reconcileExpiredBlobs` one checks real on-disk bytes + manifest rows.

Verified locally: `tsc --noEmit` clean on control-api and gfs-controller; touched vitest suites green (22 tests).

Minor, non-blocking:
- **Low** — no route-level test asserting a `429` on the new limiter (`legacyStandaloneGrants.ts`). The `rateLimiterService` machinery is tested and the wiring mirrors the existing sibling, but a regression that dropped the middleware would slip through. A supertest firing >30 req expecting `429` + `Retry-After` would close it.
- **Nit** — in `gfsBlobReconciliation.realPostgres.integration.test.ts`, `metricsStub()`'s `failures` getter isn't read by any assertion; the "proving the failure counter is never touched" comment oversells it. Cosmetic, test-only.
- Heads-up: the `*.realPostgres.integration.test.ts` suites `describe.skip` without `CONTROL_API_REAL_PG_ADMIN_URL`, so they only add coverage when CI exports it (which the `efed1fe` change now does).

Net: delta's clean, nothing to fix before merge. 👍

zach88 (approved):

alfredolopez80:
@claude — closing review request. CI/CD is green on `998188b15`; I'd like an independent pass on the two CodeQL points below before merge. Please challenge the false-positive calls rather than confirm them.

### CI/CD status (head `998188b15`)
- **CI: success** · **CodeQL workflow: success** · all required checks green · `reviewDecision=APPROVED` · `mergeStateStatus=UNSTABLE` (mergeable; the only red is a non-required check).
- The red **"CodeQL"** check is **not a required status check** (verified `isRequired=false` via GraphQL). It reports "1 new alert in code changed by this PR".

### Point 1 — alert #795, `js/missing-rate-limiting` (dismissed by me)
`GET /gfs/grants/legacy-standalone` **does** enforce the repo's PG-backed token bucket (`gfs_grants_legacy_report`, 30/min, keyed `gfsgrants-legacy:<adminAuth.sub>`). CodeQL only models known limiter libraries, not this custom middleware — the same false positive was already dismissed on `grants.ts` (#777–780) and `shares.ts` (#773–776).

Please verify: middleware order (auth publishes `adminAuth` before the limiter keys), that the bucket namespace is disjoint from the grant-mutation budget, and that the new route test is a real regression guard — `control-api/test/gfsLegacyStandaloneGrantsRateLimit.test.ts` asserts the allowed path still serves the report, the exhausted bucket returns 429 with `Retry-After`/`X-RateLimit-*` **and never reaches Postgres**, and auth precedes the limiter. I verified it is non-tautological: removing the middleware fails the suite (`checkAndIncrement` called 0 times; `expected 500 to be 429`).

### Point 2 — alert #620, `js/remote-property-injection` (NOT dismissed — second opinion wanted)
`control-api/src/gfs/seedResources.ts:104` → `byPath[path] = parentResourceId`.

This alert is **pre-existing**: it is open on `refs/heads/main` and `refs/heads/dev` (created 2026-07-20) and only surfaces on this PR because the PR touches that file. It is **not introduced by this change**.

My read is that it is a false positive for two independent reasons — please confirm or refute both:
1. `normalizeSegments` rejects any path that does not start with `/`, so every key written is `/<something>`. `__proto__` / `constructor` / `prototype` are unreachable as property names.
2. Even if they were reachable, the assigned value is a **string** (`parentResourceId`), and assigning a primitive to `__proto__` is a silent no-op — prototype pollution requires assigning an object.

If either argument is wrong, this is a real high-severity finding and should be fixed (in its own PR, since it predates this one).

### Point 3 — review nits closed in `998188b15`
Both items from @zach88's re-review are addressed: the missing route-level limiter test (above), and the reconciler suite's `metricsStub` whose `failures` getter no assertion read — the test now holds the recorder and asserts `recorder.failures === 0`, making the comment's claim true.

Local validation: `tsc --noEmit` clean on control-api; touched suites green (6 tests, reconciler run against a fresh `postgres:16-alpine`).

alfredolopez80:
### Correction on `913ee7811`, and how alert #620 was closed

**Correcting the record first:** the commit message on `913ee7811` claims the null-prototype change "clears CodeQL js/remote-property-injection". **That turned out to be wrong** — CodeQL re-analysed the change and still reports #620 (now at `seedResources.ts:109`, the same sink after the added comment lines). The `js/remote-property-injection` query flags *writing an attacker-influenced property name at all*, not prototype pollution specifically, so removing the prototype is not a sanitizer it recognises. Treat that sentence in the commit body as inaccurate; the rest of the message stands.

**The change itself is kept**, on its own merits rather than for the scanner: a lookup keyed by caller-supplied strings has no reason to inherit from `Object.prototype`, and dropping it makes prototype pollution structurally impossible instead of resting on invariants a future refactor could break. Verified the observable contract is unchanged (`JSON.stringify` — the route returns this via `res.json` — plus `Object.keys` and spread all behave identically); seed suites green (31 tests), `tsc --noEmit` clean.

**#620 is now dismissed as a false positive**, following the precedent already set for this query family in this repo (`grants.ts` #777–780, `shares.ts` #773–776). Four independent reasons it is not exploitable:

1. `normalizeSegments` rejects any path that does not start with `/`, so every key written is `/<something>` — the bare `__proto__` / `constructor` / `prototype` names are unreachable.
2. The assigned value is a **string** (`parentResourceId`); assigning a primitive to `__proto__` is a spec-level no-op (Annex B.2.2.1.2) that creates no own property — verified empirically, not just by reading the spec.
3. The only caller, `POST /gfs/seed`, sits behind `requireInternalControlJwt` — an internal-service surface, not an external or anonymous one.
4. The lookup now has a null prototype (above).

The finding is also **pre-existing**: open on `refs/heads/main` and `refs/heads/dev` since 2026-07-20, surfacing here only because this PR touches the file. Worth noting the opposite of a regression — on `dev`, `normalizeSegments` was `path.split('/').filter(...)` with no leading-slash requirement, so this PR is what forecloses the bare-key shape going forward.

I did not deviate from the "no production code merely to satisfy a tool" rule by iterating further: CodeQL cannot be run locally here, so any additional change would have been unverifiable guesswork against the scanner.

PR #159, opened 2026-07-23
feat(gfs): desktop per-agent GFS delegation UI

### LINKED ISSUE #157: Desktop app: per-agent GFS delegation UI (grant / list / revoke)
## Summary
Let a user who holds `manage_acl` on a GFS folder/file grant `read`/`write` to their own agents from the desktop app, see the resource's current grants, and revoke them.

## Motivation
The backend for per-agent GFS access exists, but users cannot yet delegate to their agents from the desktop app. This closes that gap so autonomous agents can access GFS files directly.

## Scope
- Manage-modal agents section: multi-select the user's own agents, grant read/write, inherit toggle for directories
- Grants list ('who has access') with per-row revoke
- Server-side eligibility hardening (host target must be in the caller's agent directory)
- External grants/shares rate limiting (fail-closed)
- Desktop IPC: host-subject grammar, listGrants/revoke, agents:listMine

## Notes
Stacked on the per-agent GFS backend. Port of internal, already-validated work.

alfredolopez80:
Port of internal work (issue-797 follow-up in the private repo) — desktop per-agent GFS delegation UI. Stacked on the per-agent GFS backend PR.

Closes #157

## What
- Manage-modal agents section: select the user's own agents, grant read/write, inherit toggle for directories
- Grants list ('who has access') with per-row revoke
- Server-side eligibility guard (host target must be in the caller's agent directory), external rate limiting (fail-closed)
- Desktop IPC: host-subject grammar, listGrants/revoke, agents:listMine

## Validation
`npm test` + `npx tsc --noEmit` green: control-api, external-rest-api, desktop-app (1108 tests).

## Base
Stacked on `codex/issue-797-per-agent-gfs-access`. Retarget when the base merges.


zach88:
Reviewed with code-reviewer + security-reviewer. No blockers, authz is solid and server-side enforced. A few things worth a look:

**Should-fix**
- Client sends one PUT per subject (`runGfsSubjectBatch` in `FilesPage.tsx`) instead of using the bulk `subjects[]` the server already supports (up to 100). Granting to >30 agents in one action blows past the 30/min limiter and the extras 429 (partial success). Degrades gracefully + same pattern as the admin plane, so not a regression — but sending one bulk PUT would make it 1 request against the quota and atomic. Worth doing.

**Nits**
- `parseInvalidIndexes` / `parseRetryAfterSeconds` in `gfsGrantErrors.ts` are effectively dead in prod: the message that reaches the renderer over IPC only carries `error` + `message`, so `invalidIndexes` and the "try again in Ns" hint never actually render outside tests. Fallback is fine, just cosmetic. If we want them, `describeGfsGrantError` needs the body/object, not the string.
- GET `/external/gfs/grants` shares the same 30/min bucket as writes, so repeatedly opening "Who has access" can self-throttle grants. Consistent with the admin plane, just noting it.
- `SelectableOption` change (`aria-pressed ?? selected`) is a nice a11y fix but touches every consumer of the shared component, not just GFS. Low risk, verified explicit overrides still win.

Tests green on the branch: control-api 49/49, desktop-app 73/73, ui 66/66.

alfredolopez80:
Thanks @zach88 — adversarial pass over each item. **Both actionable items are fixed (`70739aa0d`); the two nits are intentional/verified.**

### Bulk PUT (should-fix) — done
You're right: `runGfsSubjectBatch` sent one PUT per subject, so >30 in one action blew past the 30/min bucket with partial success. Switched the whole chain to one atomic bulk `subjects[]` PUT (control-api already accepts 1..100):
- uriHandler `GfsGrantInput`/`GfsShareInput` carry `subjects[]`; `grant()`/`createShare()` send `body.subjects`.
- appService parses each key up front (one malformed key throws before any round-trip — never a partial write); ipc `sanitizeSubjectKeys` fails loud on an empty/non-array payload; preload/typings/controller/FilesPage follow.
- FilesPage's three loops become single atomic calls. `runGfsSubjectBatch`/`GfsSubjectBatchError` deleted (dead after the change).

One request against the quota, atomic all-or-nothing.

### Dead parsers (nit) — fixed by making them live
Confirmed your read: `invalidIndexes`/`retryAfterSeconds` are separate response-body fields, and only `Error.message` survives the Electron IPC boundary (`.bodyText` is dropped), so `parseInvalidIndexes`/`parseRetryAfterSeconds` never fired in prod and their tests used a wire format prod never produces. Rather than delete them, I made the feature real: uriHandler now reads `ApiError.bodyText` and re-throws a `GfsUriError` embedding the fields in the exact shape `gfsGrantErrors` parses, preserving the original message (and thus the server error code). Verified end to end — control-api emits them top-level (`sendGfsGrantError` / `rateLimitMiddleware`), external-rest-api forwards the body for 400/403/429 (`PROPAGATED`), httpClient keeps it on `.bodyText`. Tests now feed the real prod message, and `delegation.tsx` routes through `describeGfsGrantError` so the user/team panel surfaces them too. So "try again in 45s" and "subjects 2, 5" now actually render.

### Nits — no change (verified)
- GET grants shares the writes rate bucket: left as-is, consistent with the admin plane; opening "Who has access" 30×/min to self-throttle grants isn't a realistic path.
- `SelectableOption aria-pressed ?? selected`: backward-compatible — explicit overrides win, and only AppHeader consumes it pre-existing.

desktop-app: tsc (src) clean, 151 grant-suite tests green (full suite 1112). Branch CI passing so far (17/19, 2 still running).

zach88 (approved):

PR #151, opened 2026-07-23
fix(control-api): restore member-registration revoke migration

alfredolopez80:
## Summary
- restore migration 0070 required by the public database bootstrap verifier
- revoke obsolete DELETE access for the control-api runtime role while preserving its credential-revocation UPDATE capability
- align database readiness and runtime-access migration coverage with the required schema version

## Validation
- focused migration/readiness tests: 5 passed
- full control-api test suite and production build: passed
- clean dedicated minikube profile: migration job succeeded; all services report OK; no unhealthy pods

Fixes #150

PR #168, opened 2026-07-24
fix(control-api): add 0070 migration revoking runtime DELETE on membe…

andypavia:
…r_registration_credentials

Migration 0069 grants control_api_runtime full DML (incl. DELETE) on member_registration_credentials, but the runtime access contract classifies the table as `upsert` (SELECT/INSERT/UPDATE only). Without this follow-up, the minikube migration gate fails verifying that the 0070 delete-revoke migration was recorded, blocking `make minikube-setup`.

Idempotent: REVOKE of an already-absent privilege is a no-op.

## Testing

- [ ] `npm test` passes for changed services
- [ ] `npx tsc --noEmit` clean

## Checklist

- [ ] This is not a new third-party MCP server / WorkflowRecipe (those go to the registry)

### DIFF
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.365",
+  "version": "0.1.366",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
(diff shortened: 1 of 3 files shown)


alfredolopez80:
this Pr replicated the PR #151 and it was implemented en `dev` y `master`

PR #201, opened 2026-07-28
fix: make desktop-app nudges track agent active time and fix chat selection across routes

zach88:
## What
Three fixes to the desktop app around time-based nudges and chat selection, plus test hardening:
- Time-nudge tiers now measure the agent's active time instead of wall-clock time.
- Nudges no longer escalate while the agent is suspended waiting for a tool approval.
- Opening a session from another route no longer flashes the new-chat view before landing on the selected chat.

## Why
Nudge escalation was unfair to agents that spent most of the elapsed time paused on approvals: wall-clock timing pushed them into higher tiers while they were not actually working. The new-chat flash made navigation from notifications and other routes feel broken, briefly showing a wrong state before the real session loaded.

## How
- `AgentTaskTrackerContext` now records pause segments (suspended / pending approval) and derives `active_ms` = elapsed minus paused time; tier computation and terminal tier stamping run on active time.
- `NudgeArea` renders nothing while the task is suspended or has a pending approval, so no escalation is visible during waits.
- `useAppController` guards its imperative fast path on being on the chat route; from any other route the selection is stored as pending and applied on arrival, avoiding the new-chat flash.
- Includes a sync merge of `dev` (textual conflicts only in the desktop-app version files).
- Test coverage added for the chat-route guard through its real entry points (`handleSelectChatAgent` and the notification open path), mutation-verified against the guard, with assertions hardened for CI contention; the ad-hoc clerum test harness was consolidated into a single shared fixture.

## Testing
- Full desktop-app vitest suite: 129 files / 1134 tests, all green.
- New guard tests verified by mutation: removing the route condition from either call site fails the matching test.
- Flakiness check: 20/20 consecutive runs green under CPU contention.
- `tsc` main process clean; renderer typecheck shows only the 4 pre-existing errors in unrelated test files.

## Risks / Follow-ups
- Low risk: behavior changes are scoped to nudge timing and chat-selection navigation.
- Follow-up candidates (minor, reported in review): trim unused fixture options and spies, type the release-status option with the real `DesktopReleaseStatus` type.


zach88:
@claude take a look at this

josue-tech-94 (approved):
The desktop behavior looks correct:
- Nudge tiers now exclude completed approval waits.
- Replayed suspension events do not reset the pause timestamp.
- Suspended tasks do not show misleading “still working” nudges.
- Chat and notification navigation use pending selection when entering Chat from another route.
I ran the focused tests locally: 86/86 passed. CI and CodeQL are also green.

No functional blockers found.

PR #206, opened 2026-07-29
feat(control-api): auto-claim the registry's one-time token instead of discarding it

jozer-rami:
## What this fixes

Registry PR #59 adds open self-hosted registration: with `CLERUM_REGISTRY_OPEN_REGISTRATION` on, `POST /api/v1/deployments/register` auto-approves inside the transaction and returns `201` with a one-time `claim_token`.

The self-hosted control-api could not consume that response. In `control-api/src/routes/admin/registryConnect.ts` on `dev`:

- `:187` parsed the response as `{ deployment_id, key_id }` — the `claim_token` was dropped on the floor.
- `:179` used `if (!regRes.ok)`, so `201` passed as success and nothing distinguished auto-approved from pending.
- `:199` always replied `{ state: 'pending' }`.
- `:72` the status poller parsed `{ status?: unknown }` and never read `claimed`.

The token only ever exists in that response body. Discarding it meant every open registration stranded permanently: approved at the registry, holding a valid keypair, with no credentials and no way to obtain them.

A second defect rode along: PR #59's new rejections (`409 org_name_taken`, `429 registration_capacity`, `429 RATE_LIMITED`, `400 invalid_contact_email`) all collapsed into an opaque `502 registry_integration_error`. A self-hoster who picked a taken org name was told "registry integration error".

## What changed

**control-api**

- `redeemClaim()` is now the single claim implementation, shared by the manual paste route and the new automatic paths. It is exception-total: `markConnected` encrypts and writes to Postgres *after* the registry has already burned the one-time secret, so a throw there must be a value the caller can act on.
- The register handler branches on status. On `201` it persists the connection row **first**, then redeems the token. On `202` it is byte-identical to before.
- New `POST /admin/registry/connect/recover` finishes a connection whose inline claim failed: it reads `/status` first, returns terminal errors without spending a rotate, then rotates a fresh token and redeems it.
- `GET /admin/registry/connect` stays read-only.
- Registry rejections map to distinct, actionable codes via allowlists with stated defaults.
- Every registry `fetch` in the file is now bounded with `AbortSignal.timeout`.

**control-ui**

- New `connecting` wire state with a **Finish connecting** button. Under auto-approval no operator ever sees a claim token, so the old "paste the token your operator gave you" view is the wrong thing to show.
- `load()`'s catch-all default is replaced by an explicit `disconnected` branch plus a `never` exhaustiveness guard.
- `handleRequest` branches the view **and** the toast on the response.
- Distinct copy for each new registry rejection, plus client-side email validation so a typo does not spend one of the registry's ~5 daily per-IP registration attempts.

**docs** — `docs/how-to/connect-to-registry.md` documents both approval paths and drops the claim that the panel polls (it does not; it loads once on mount).

## Safety properties

Four properties are load-bearing. Each was verified against the final integrated state, not per-commit, and each has a test that fails under mutation:

1. **The operator flow cannot be broken.** A `202 pending` registration must never cause a `POST /:id/claim-token` rotate — rotating would silently invalidate the one-time token an operator delivered out of band to a human. There is exactly one rotate call site, behind a `status !== 'approved'` gate, and exactly one writer of local `'approved'`, derived from `regRes.status === 201` alone.
2. **`GET` is read-only.** It is called by `RegistryCatalog.tsx` on every Marketplace mount and is reachable by cross-site top-level navigation (`SameSite=Lax`, no CSRF check in this repo). Recovery is therefore an explicit `POST`, not a side effect of `GET`.
3. **Persist before claim.** The keypair is the only artifact that can recover an approved deployment, because rotate is proof-of-possession gated. Claiming first and dying before the write would burn the token *and* lose the key.
4. **Credentials cannot cross-wire.** `markConnected` had no `WHERE` clause; it now scopes to the deployment that actually claimed and reports a 0-row match instead of silently succeeding.

`claim_token` and `client_secret` are never logged, returned, or persisted outside the encrypted column. The four names are also added to `TOKEN_LEAK_REDACT_PATHS` and `REQUIRED_REDACT_PATHS`, the latter enforced by an existing test so redactions can grow but never shrink.

## Testing

- control-api: **3519 passed**, 51 skipped
- control-ui: **1112 passed**
- `node scripts/prettier/run-all.mjs --check` clean on all 14 files this branch touches (pre-existing `tests/e2e/` drift on `dev` is untouched)

## Release ordering — read before flipping the flag

**This branch is safe to merge on its own. It is behaviour-neutral while the registry flag is off.**

`CLERUM_REGISTRY_OPEN_REGISTRATION` must **not** be enabled on the shared registry until a capability gate lands there. Without it, the flag is all-or-nothing: every control-api older than this branch receives a `201`, discards the token, and strands permanently. That gate makes auto-approval conditional on the client declaring `deployment_info.auto_claim === true`, which this branch now sends.

The gate is written and its suite is green, but it is **not pushed** — PR #59's worktree is held by another active session, so it lives on a separate local branch pending coordination.

## Known follow-ups (deliberately not in this PR)

- **Register timeout strand window.** If the registry commits a registration and burns the claim token but answers after the 10s timeout, nothing is persisted locally and re-registering that org name returns `409 org_name_taken` permanently, while the operator sees a bare 500. This pre-dates the change for any register-fetch failure; the timeout widens it. Wants a distinct error code and panel copy.
- **`authEnabled` on the recovery response** has no negative-case test, so hardcoding `true` there would survive the suite. Blast radius is a missing "enable registry auth" banner for a self-hoster running `CLERUM_REGISTRY_AUTH_ENABLED=false`, and it self-heals on the next page load.
- **`connection_superseded` on the manual claim path** renders terminal copy telling the user to disconnect, but it renders in the `approved` view, which has no Disconnect control. Rare race; a dead end rather than a data-loss path.

### DIFF
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.367",
+  "version": "0.1.368",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.367",
+      "version": "0.1.368",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.373",
+  "version": "0.1.374",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.373",
+      "version": "0.1.374",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.368",
+  "version": "0.1.369",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.368",
+      "version": "0.1.369",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.374",
+  "version": "0.1.375",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.374",
+      "version": "0.1.375",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.375",
+  "version": "0.1.376",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.375",
+      "version": "0.1.376",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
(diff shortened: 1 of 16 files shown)


alfredolopez80:
Cross-PR validation note for the self-hosted registration rollout:

This PR is correctly paired with keyper-labs/evenfire-registry#60. This client
declares `deployment_info.auto_claim: true`, consumes the Registry `201` +
`claim_token` response, and persists the connection before redemption. The
Registry gate then ensures older self-hosted clients remain on the existing
operator-approved `202 pending` flow.

The required rollout order is:

1. Deploy Registry #60.
2. Merge and deploy this PR.
3. Enable `CLERUM_REGISTRY_OPEN_REGISTRATION` only after both are live.

The separate `403` observed while editing egress metadata through the
Evenfire-hosted Control UI is outside this PR's self-hosted connection scope.
Managed deployments use provisioned Registry credentials and do not enter the
self-hosted register/claim flow implemented here.

Please track it as a hosted-environment readiness check: verify the live
Control API uses its intended curator Registry identity and can update
`mcpServer.egressSummary` for a controlled `@evenfire/mcp-*` entry. That is a
configuration and operational verification, not a reason to broaden
self-hosted clients into curators.

PR #208, opened 2026-07-29
chore: add column sorting to LLM models table

zach88:
**What**
Adds ascending/descending sorting to every column of the LLM models table in control-ui, except the actions column. Sorting runs client-side on the filtered rows, so active filters are preserved and only the visible subset is reordered.

**Why**
The models list could only be read in server order, which makes it slow to find or compare models once the catalog grows. Operators can now reorder by provider, model, context window, enabled state, staleness or source.

**How**
- `TableHeaderRow` gains opt-in sortable columns (`sortable` on the column plus `sortKey`/`sortDir`/`onSortToggle` row props) and exposes `aria-sort`. Without the new props the rendered markup is unchanged, so existing consumers are unaffected.
- `LlmModelTable` holds the sort state with a two-state toggle (asc first) and typed comparators: provider sorts by display label, blank values go last in both directions, context window compares numerically with non-finite values treated as blank, and ties fall back to a stable provider/model comparison.
- Accessibility: the header's accessible name stays the visible label, direction is conveyed by `aria-sort`, the next-click hint lives in the button `title`, and the sort arrow is hidden from assistive tech.
- A new `.cu-link--sort` style keeps sortable headers at the table's 0.7rem uppercase typography, matching plain headers and the loading skeleton.

**Testing**
- Full control-ui suite: `npm test`, 112 files / 1102 tests passed.
- 14 new `LlmModelTable` cases: toggle behavior, filter preservation, blank-last ordering, provider-by-label ordering, numeric and NaN context window, tie-breaking, and server-order default.

### DIFF
--- a/control-ui/app/globals.css
+++ b/control-ui/app/globals.css
@@ -1524,10 +1524,18 @@ textarea.cu-input {
   font-size: 0.8125rem;
 }
 
-/* Sortable table header label: reads as header text, behaves as a button. */
+/* Sortable table header label: reads as header text, behaves as a button.
+   Inherits the <th> typography (0.7rem, uppercase) through `.cu-link`'s
+   `font: inherit` — deliberately NOT combined with `.cu-link--sm`, which would
+   bump it to 0.8125rem and break size parity with plain headers and with the
+   loading skeleton. */
 .cu-link--sort {
   color: var(--cu-text-muted);
   white-space: nowrap;
+}
+
+.cu-link--sort:hover {
+  color: var(--cu-text);
 }
 
 .cu-link--danger {
--- a/control-ui/app/globals.css
+++ b/control-ui/app/globals.css
@@ -1524,6 +1524,12 @@ textarea.cu-input {
   font-size: 0.8125rem;
 }
 
+/* Sortable table header label: reads as header text, behaves as a button. */
+.cu-link--sort {
+  color: var(--cu-text-muted);
+  white-space: nowrap;
+}
+
 .cu-link--danger {
   color: var(--cu-danger);
 }
--- a/control-ui/components/LlmModelTable.tsx
+++ b/control-ui/components/LlmModelTable.tsx
@@ -16,7 +16,23 @@ import { TablePanelHeader } from './TablePanelHeader'
 import { IconPencil, IconRefresh, IconX } from './icons'
 import { SelectInput } from './ui'
 
-const MODEL_COLUMNS: TableHeaderColumn[] = [
+type ModelSortKey =
+  | 'provider'
+  | 'model'
+  | 'vendor'
+  | 'displayName'
+  | 'contextWindow'
+  | 'enabled'
+  | 'source'
+  | 'stale'
+
+// A column may only opt into sorting if its key is one `compareByKey` knows how
+// to order — marking `sortable: true` on any other column is a compile error,
+// so MODEL_COLUMNS and the comparator cannot drift apart.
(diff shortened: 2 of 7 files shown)


zach88:
@claude take a look at this

zach88:
Thanks for the notes. Went through all three, applied one and kept the other two as-is. Reasoning below so the decisions are on the record.

**1. `aria-sort="none"` on sortable but unsorted columns: keeping the current behavior.**
Omitting the attribute is not the same thing as leaving the state undefined. `none` is the implicit value of `aria-sort` in the ARIA spec, so a sortable `<th>` with no attribute already exposes exactly the state that `aria-sort="none"` would set. The gap you are describing is about specific AT and browser combinations diverging from the spec there, which is real but narrow.

The tradeoff is what tips it: the current markup guarantees at most one `<th>` in the table carries `aria-sort`, and the tests lean on that invariant (`th[aria-sort]` selectors in `LlmModelTable.test.tsx:150`, `:158`, `:162`, `:212`). Making it explicit means every sortable header gains the attribute, so those four assertions have to start matching on the value instead of the presence. That is trading a verified invariant for a speculative gain, so I would rather leave it. Happy to revisit if we ever get a real screen reader report against this table.

**2. Redundant self-comparison in the tie-break: confirmed, leaving it.**
You are right that it is dead work. When `sortKey === 'provider'` and the primary comparison returns `0`, the tie-break call is guaranteed to return `0` too, since it is the same `localeCompare` and only the sign differs (`sign * 0` is still `0`).

Skipping it needs a `sortKey !== 'provider'` guard inside the comparator, which is a special case in the one place I most want to stay obvious, and it buys nothing measurable: the allowlist is tens to low hundreds of rows, and the extra call only fires on ties. Leaving the straightforward version, and your comment here is a good marker for the next reader.

**3. `sortTitle` next-direction contract: applied.**
This one was a genuine gap. `TableHeaderRow` has around thirty consumers, so a second sortable one is plausible, and the tooltip assumption was documented nowhere. Added the contract next to `onSortToggle` in `types.ts`: implementations invert `sortDir` when the key is already active and reset to `asc` when switching columns, because the tooltip announces the next click as ascending for every inactive column.

Verified locally, since the earlier pass could not run anything: type-check clean, full control-ui suite green at 112 files and 1127 tests.

zach88:
@claude re run please

jozer-rami (approved):

josue-tech-94 (approved):
No blocking issues found. PR `evenfire-ai/evenfire#208` looks safe to approve from my pass.

What I checked:

- Final PR diff against `dev`, not just individual commits.
- `LlmModelTable` sort behavior, blank handling, provider-label sorting, tie-breaks, and filter interaction.
- Shared `TableHeaderRow` backwards compatibility across existing consumers.
- Accessibility behavior: button name stays the visible header label, active sort uses `aria-sort`, arrow is hidden from AT.
- CSS consistency with `control-ui` token/style rules.
- Package/version noise: only `control-ui` auto-version bump, consistent with this repo’s commit behavior.

Validation I ran in an isolated PR worktree:

```bash
cd control-ui
npm test -- --run components/__tests__/LlmModelTable.test.tsx
npx tsc --noEmit -p tsconfig.json
npm test
npm run build
```

Results:

- Focused `LlmModelTable` tests: `23 passed`
- Full `control-ui` tests: `112 files / 1127 tests passed`
- TypeScript: clean
- Production build: passed

PR #211, opened 2026-07-29
ci: auto-delete merged head branches, except dev

jozer-rami:
## Summary

Deletes a PR's head branch when the PR is merged — except for long-lived branches, `dev` above all.

## Why not the repo setting

GitHub's `delete_branch_on_merge` ("Automatically delete head branches") has no exception list. The dev → main promotion PR uses **`dev` itself as the head branch**, so that setting deleted `dev` during the v0.5.0 promotion; it has been off ever since, which is why nothing gets cleaned up today.

Branch protection does not save `dev` either: the `dev protection` ruleset has had a `deletion` rule since 2026-07-23, and `dev` was still auto-deleted on 2026-07-24 — repo admins sit in `bypass_actors` with `bypass_mode: always`. Removing that bypass would also start enforcing the 1-approval and required-status-check rules against admins on `dev`, so it is not a free fix.

So the repo setting stays **off** and the cleanup happens here, where the exception list is explicit.

## Behaviour

| Merge | Head branch | Result |
| --- | --- | --- |
| feature → `dev` | `feat/x` | deleted |
| feature → `main` | `chore/y-main` | deleted (once this reaches `main`) |
| `dev` → `main` (promotion) | `dev` | **kept** |
| fork PR | contributor's branch | untouched |

Keep-list: `dev`, `main`, `cla-signatures`, `gh-pages`. Already-deleted branches are a no-op, so clicking "Delete branch" at merge time still works.

## Notes

- `pull_request_target` (not `pull_request`) so the workflow file is read from the **base** branch — this applies to the ~50 PRs already open, rather than only to branches cut after it lands. Nothing from the PR is checked out or executed, so the write-scoped token is never exposed to contributor code.
- Default workflow token is read-only repo-wide, hence the explicit `permissions: contents: write`.
- Takes effect for `main`-based PRs only after it flows to `main` via the next promotion.

## Test plan

- `actionlint` clean (includes shellcheck on the `run` block).
- `gh api repos/.../git/ref/heads/<ref>` existence probe verified against real refs, including a slashed branch name (`docs/readme-redesign` → found) and a missing one (→ not found).
- End-to-end verification is the merge of this PR itself: `ci/auto-delete-merged-branches` should disappear on merge.

### DIFF
--- a/.github/workflows/delete-merged-branch.yml
+++ b/.github/workflows/delete-merged-branch.yml
@@ -1,0 +1,55 @@
+name: Delete merged branch
+
+# Deletes a PR's head branch once the PR is merged, EXCEPT for long-lived
+# branches. The exception that matters is `dev`: the dev -> main promotion PR
+# uses `dev` itself as the head branch, so GitHub's repo-level "Automatically
+# delete head branches" setting deleted `dev` on the v0.5.0 promotion. That
+# setting is all-or-nothing, so it stays OFF and this workflow does the cleanup
+# with an exception list instead.
+#
+# `pull_request_target` rather than `pull_request` so the workflow file is read
+# from the base branch — cleanup then applies to PRs opened before this landed,
+# instead of only to branches cut afterwards. Nothing from the PR is checked
+# out or run here, so the write-scoped token never touches contributor code.
+
+on:
+  pull_request_target:
+    types: [closed]
+
+permissions:
+  contents: write
+
+jobs:
+  delete:
+    name: Delete head branch
+    runs-on: ubuntu-latest
+    # Fork branches are not ours to delete.
+    if: >
+      github.event.pull_request.merged == true &&
+      github.event.pull_request.head.repo.full_name == github.repository
+    steps:
+      - name: Delete head branch unless it is long-lived
+        env:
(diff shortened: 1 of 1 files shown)


jozer-rami:
Superseded by the native configuration — no workflow needed.

GitHub's docs do state that `Branch protection rules and repository rules can also prevent branches being automatically deleted`. The reason `dev` was deleted anyway on the v0.5.0 promotion is the `dev protection` ruleset's bypass list (repo admin, `bypass_mode: always`) — the merging admin's bypass carries the auto-delete through. This is a known trap, reported in community discussion #152926 with the exact same dev→main setup.

Rulesets aggregate and bypass is scoped per-ruleset, so the fix is a second ruleset with **no bypass actors** carrying only the `deletion` rule, layered over the existing one:

- New ruleset `no-delete: dev and default branch` (id 19985235) — targets `refs/heads/dev` + `~DEFAULT_BRANCH`, rule `deletion`, `bypass_actors: []`.
- Repo setting `delete_branch_on_merge` set back to `true`.

Verified with a throwaway ref covered by the new ruleset: deletable as admin before the ruleset existed, `422 Repository rule violations found — Cannot delete this branch` after.

The existing `dev protection` / `main protection` rulesets keep their admin bypass, so self-merge still works.

PR #212, opened 2026-07-30
feat(control-api): derive registry auth from the connection row in self-hosted mode

jozer-rami:
## Why

`CLERUM_REGISTRY_AUTH_ENABLED` was a manual env var an operator had to set **after** connecting a self-hosted deployment to the registry. Setting it too early was a boot-order deadlock:

```
auth on + no connection row → assertRegistryConnectionReady throws (main.ts:57)
  → control-api never listens → no Control UI → cannot run the connect flow
  → cannot create the row → crashloop
```

So the flag could not safely default to `true`, and an operator who set it early bricked the install. Since [#206](https://github.com/evenfire-ai/evenfire/pull/206) made connecting a single click with no operator involved, this manual flag was the last remaining friction in self-hosted onboarding.

Verified on a real cluster before starting: a freshly auto-claimed minikube deployment returned `409 registry_auth_disabled` from the API-keys panel until the var was patched to `true` and control-api restarted.

## What changed

**In self-hosted mode the env var is now inert.** Registry auth activates the moment the deployment holds machine credentials, which is when the connect flow completes. No env var, no restart.

- New `isRegistryAuthActive()` in `registryConnectionDb.ts` — managed returns the env flag verbatim, self-hosted returns `resolveMachineCreds() !== null`. It checks *credentials*, not row presence, so a `pending` row and an `approved`-but-unclaimed row both correctly report inactive.
- `registryBootGuard.ts` no longer throws. It logs whether a registry identity is present, so the "row disappeared after connecting" signal survives without the deadlock.
- The connection-row cache gains a 15s TTL, bounding staleness in both directions across replicas. Previously an access-control input could stay stale until pod restart.
- Six response fields, two admin route guards, and `registryClient.mintToken` now derive from the accessor.
- **Managed mode is untouched** — same env vars, same boot validation, same failure modes.

## A real hole this closes

The URL allowlist previously ran only when `registryAuthEnabled`, which ships `false`. Every default self-hosted deployment therefore **POSTed its registration and claimed machine credentials against a completely unvalidated registry URL**. The allowlist now applies to any explicitly self-hosted deployment with a URL configured, regardless of auth.

## Rate limiters on two connect routes

CodeQL flagged three `js/missing-rate-limiting` alerts on `registryConnect.ts`. One is a false positive — `POST …/recover` has had a limiter since #206; CodeQL cannot see this repo's PG-backed middleware. The other two were real pre-existing gaps, surfaced only because this branch touched those handlers. Both are now closed:

- **`GET /admin/registry/connect` → 30/min per admin.** Mounted by `RegistryCatalog` on every Marketplace visit, not just the connect panel, and reachable by top-level cross-site navigation (session cookie is `SameSite=Lax`, no CSRF or origin check). Each call makes an outbound registry `/status` fetch, so this bounds that traffic under a CSRF-driven loop while staying generous for normal navigation.
- **`POST …/request` → 3/min per admin.** Each call consumes one of the shared registry's ~5/day per-IP registration attempts and can create a deployment row there. 3/min allows a legitimate retry after a typo'd org name without letting a looped admin burn the daily budget in one minute.

Both mirror the per-admin token-bucket shape already used by `registry.ts` and the recover route. Tests assert the bucket key and limit (proving the middleware sits *after* `requireAuthForControlUI`) and that a 429 blocks before any DB or registry work.

## Managed-mode safety

This is the branch's hard constraint, and two design revisions broke it in review before landing — `parseRegistryConnectionMode()` defaults to `'managed'` when `REGISTRY_CONNECTION_MODE` is unset, which makes the obvious gates wrong in one direction or the other.

The final review built a 27-cell truth table over `REGISTRY_CONNECTION_MODE` × `CLERUM_REGISTRY_AUTH_ENABLED` × `CLERUM_REGISTRY_URL` and found **zero regressions**. One cell changes outcome — explicitly self-hosted + auth off + unlisted URL now refuses to boot, which is the intended hole closure. Three cells change only the error message. Every managed cell is identical.

`control-api/test/startupGuards.registryConsumer.test.ts` is **byte-identical to `dev`** and green — it is the regression proof and was deliberately never edited. The actual mutation-killer for the gate is a new test: mode unset + *unlisted* URL + auth off still boots. Dropping the self-hosted disjunct fails exactly that one test and nothing else.

## Testing

- control-api: **3553 passed**, 51 skipped
- control-ui: **1113 passed**
- `tsc --noEmit` clean in both; prettier clean on every touched file

## Behavioural change operators need to know

An explicitly self-hosted deployment whose `CLERUM_REGISTRY_URL` is not allowlisted now **refuses to start**. `https://registry.evenfire.ai` and `http://registry-api.registry.svc.cluster.local:8085` are built in; anything else needs `CLERUM_REGISTRY_URL_ALLOWLIST`. This is documented in the connect guide's Prerequisites, where the person configuring it will actually see it.

No existing deployment is affected: there are currently no self-hosted deployments, the minikube overlay uses an allowlisted URL, and both `gcp-prod` and `gcp-dev` are `managed` with the allowlisted shared registry.

## Known follow-ups (deliberately not in this PR)

- **`DELETE /admin/registry/connect` is blocked when no registry URL is configured.** The guard lives in the shared `requireSelfHostedAdmin`, but `DELETE` never builds a registry URL. An operator who removes `CLERUM_REGISTRY_URL` from a connected deployment cannot clear the stale row via the API. Recoverable by restoring the var and restarting; the fix is moving the guard into the four routes that need it.
- **Diagnostic ordering.** An auth-on deployment with no `REGISTRY_CONNECTION_MODE` now reports a URL error before the missing-mode error. Fails closed either way, but it is a two-step diagnosis of a one-step misconfiguration.
- **Disconnect still does not revoke registry-side.** `deleteConnection()` drops the local row only, and `registryClient` holds its own Bearer-token cache that is never cleared, so a disconnected deployment keeps working against the registry until token expiry. Pre-existing, but this change makes disconnect a routine action.
- A few test-hardening nits that do not guard live behaviour.

### DIFF
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.387",
+  "version": "0.1.388",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.387",
+      "version": "0.1.388",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.386",
+  "version": "0.1.387",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.386",
+      "version": "0.1.387",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.376",
+  "version": "0.1.377",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.376",
+      "version": "0.1.377",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.383",
+  "version": "0.1.384",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.383",
+      "version": "0.1.384",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.385",
+  "version": "0.1.386",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.385",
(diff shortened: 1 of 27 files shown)


jozer-rami:
@claude deep review

jozer-rami:
@claude reveiew fixes

jozer-rami:
@claude review latest commit fixes

PR #218, opened 2026-07-30
docs: add a self-hosted registry quickstart

jozer-rami:
## What

Adds `docs/get-started/registry-quickstart.md` — a single front-to-back path for a self-hoster who wants to *use* the registry, ending in a working install and a published entry.

The two existing how-tos each cover one leg (`connect-to-registry.md` stops at connected; `publish-plugin-to-registry.md` starts from an `efrk_` key), so there was no guide that took someone from a fresh cluster to installed-and-publishing. This fills that gap and links out to both rather than duplicating them.

## Sections

1. Confirm self-hosted mode — minikube already ships it; table of settings for a real cluster, including the allowlist that makes control-api refuse to boot
2. Connect — `/marketplace/connect`, auto-approve vs operator-approve, with a warning that **Start over** is destructive
3. Install a connector — required Context, all-or-none credentials, egress review, `kubectl` verification
4. Install a plugin — Package / Security / Install, verified in `sandbox-recipes`
5. Mint an org API key — auth derives from the connection, no flag and no restart
6. Publish — UI fields and the CI `curl`, plus the `scope_required` and `metadata.name` traps
7. Troubleshooting — 10 rows, including the `403` on grants that is expected for connect-onboarded tenants

## Also

- Fixes the stale `/registry/connect` → `/marketplace/connect` in `connect-to-registry.md`. The rename landed a while ago and the old path only survives via a 301.
- Indexes the new guide in `docs/README.md` and `docs/llms.txt`.

## Verification

Every UI label, route, namespace, and error code was read out of current `dev` rather than carried over from the older docs — button text from `RegistryConnectPanel.tsx`, routes from `app/constants/routes.ts`, the required `contextRef` from the install saga, `sandbox-recipes` from `constants/workflowRecipes.ts`. Prettier clean; all relative links resolve.

**Not** exercised against a live registry — the flow is documented from code and config, not from a recorded end-to-end connect.

Docs-only; no source or manifest changes.


jozer-rami:
## Update — operator-approval path removed

`CLERUM_REGISTRY_OPEN_REGISTRATION` is now on at the shared registry, so a self-hoster always takes the auto-claim path. Both connect guides described the `pending → operator approves → paste a one-time claim token` branch as a live alternative, which told people to wait on a human who is never involved.

Second commit rewrites both to a single flow, `disconnected → connecting → connected`:

- **`connect-to-registry.md`** — the two-state-machine framing, the claim-token step, and the **Refresh status** poll are gone. The **Start over** warning is pulled out of the numbered steps into its own callout, since it is the one irreversible action in the panel.
- **`registry-quickstart.md`** — step 2 collapses to the auto path; troubleshooting swaps the operator rows for what the auto path actually returns.

Failure handling now reflects the real error surface: `org_name_taken` / `org_blocklisted` refuse registration before anything is saved, **Finish connecting** recovers a redeem that did not land, and `already_claimed` / `deployment_suspended` are the two terminal states.

### Scope note

The operator-approved path still exists in `control-api` **by design** — PR #206 treats "the operator flow cannot be broken" as a hard invariant so a flag-off registry keeps working, and `registryConnect.ts:29-31` still documents `approved` as the human-paste state. This change is documentation-only: it stops describing a branch self-hosters on the shared registry cannot reach. No code was touched, and a flag-off deployment would need that branch documented again.

jozer-rami:
## Update — image push to your org namespace

Publishers keep getting the OCI coordinate wrong, and nothing in the docs covered it. The Publisher UI shows the correct coordinate, but anyone working from the how-to had to infer it from the entry name — and the entry name is precisely what misleads, since entries are scoped `@org/name` while the image path must be `org/name`.

Third commit adds a push section to `publish-plugin-to-registry.md` (new step 2, later steps renumbered) and a compact `## 6. Push an image` to the quickstart, both organized around the actual failure modes:

| Mistake | What happens |
|---|---|
| `registry.evenfire.ai/@acme/db-tools` | `docker` rejects it as `invalid reference format` — `@` is the digest delimiter |
| `registry.evenfire.ai/db-tools` | Outside the namespace the key grants |
| entry `@acme/db-tools` + image `acme/dbtools` | `422` at publish **and** install; unchecked it denies the cross-org pull and surfaces as `ImagePullBackOff` in someone else's cluster |
| OCI-only manifest | Registry advertises `http.compat: ["docker2s2"]`; re-tag to Docker manifest format |

Also captured: the login username is the literal `_` (`dockerCredential.ts:6`), the credential comes from **Publisher → Docker credentials** as a downloadable `dockerconfigjson` for CI, and the tag is unconstrained because only the repo path is matched against the entry name.

Sourced from `dockerCredential.ts`, `DockerCredentialModal.tsx`, and `registryImageRefIdentity.ts` rather than restated from the UI copy.

PR #173, opened 2026-07-24
Feat/desktop loading times improvements

josue-tech-94:
## Summary

Adds the backend pagination and persistence foundation required for bounded desktop session loading.

The `desktop-app` portion is isolated in draft PR #220 for independent review:
https://github.com/evenfire-ai/evenfire/pull/220

## Current scope

- Stable, scoped cursor pagination for session catalogs.
- Bounded before/after transcript windows.
- Strict and consistent route/query validation across `mcp-host` and `rpc-proxy`.
- Correct session-key parsing for unusual subjects, agents, and chat IDs.
- Materialized SQLite session summaries with convergence and migration coverage.
- Query/index alignment and proxy forwarding/error-behavior coverage.

The unrelated `profile-ui` lockfile changes were reverted.

## Validation

- `mcp-host`: 3,047 tests passed, with one todo; typecheck, build, and lint passed.
- `rpc-proxy`: 460 tests passed; build passed.
- Session ownership is enforced for catalog, transcript, context, and model reads,
  including overlapping colon-bearing user subjects.
- Dual-store catalog pagination prevents duplicate keys across cursor pages.
- Session activity remains monotonic across migration, replacement, compaction,
  and cold load.

PR #220, opened 2026-07-30
feat: improve desktop chat loading and recovery

josue-tech-94:
## Summary

Separates the `desktop-app` loading-time work from #173 so the desktop changes can be
reviewed independently. This PR intentionally targets
`feat/desktop-loading-times-improvements`; after it is reviewed and merged, #173 will
contain the complete service and desktop result without mixing both review surfaces in
one diff.

## What changed

- Adds durable paged chat storage with bounded reads, atomic snapshot replacement,
  recovery, quarantine, cleanup, and downgrade compatibility.
- Replaces settled legacy turnless cache windows without discarding completed server
  turns. An in-flight optimistic echo is kept until the server confirms its turn (at most
  a transient on-screen duplicate, never lost server history); on the next idle reconcile
  it collapses only when its content matches the authoritative server row, and the echo's
  side metadata (attachments, tool-steps) is preserved onto that surviving row. Durable
  local system, error, and failed-send messages are always retained.
- Quarantines torn or structurally invalid known-schema indexes while preserving paged
  transcript directories, and fails closed without rewriting unsupported future index versions.
- Reads known v3 indexes and normalizes them during the next ordinary catalog update through the
  crash-durable temporary-file, file-fsync, rename, and parent-directory-fsync protocol.
- Validates and rederives page ordering, counts, server-sync state, and turn bounds before
  append or pruning so corrupt metadata cannot discard local-only messages.
- Adds bounded session/message pagination and lazy older-history loading while preserving
  local, optimistic, failed-send, system, error, attachment, and active-task messages.
- Uses monotonic chat-switch and sidebar request identities, including fresh A→B→A
  reconciliation, and retries failed initial catalog loading for the same session.
- Persists authoritative content and metadata changes even when deterministic message IDs
  remain unchanged.
- Makes file-backed token migration atomic and crash-durable before deleting legacy
  sources, including Windows-compatible directory-sync handling.
- Runtime-validates session pagination responses and rejects invalid IPC pagination
  values before service calls.
- Surfaces dropped malformed session-catalog items instead of losing them silently: the
  sessions-list parser keeps its tolerant partial-drop policy (one corrupt session must not
  empty the sidebar) but logs each dropped item with its index and reports an additive
  `droppedItemCount` on the result so the renderer can observe the loss.
- Refuses symlinked quarantine roots during retention cleanup.
- Defers non-critical startup, catalog, activity, and prewarm work, and uses stable CSS
  transcript chunking to reduce initial rendering work.
- Restores the unrelated AuthPage runtime-profile behavior so this PR remains focused on
  desktop loading and chat reliability.

## Why

PR #173 combined backend pagination work with a large desktop persistence and rendering
change. Splitting the desktop portion makes the risk boundary and review ownership
explicit while preserving the intended merge order.

## User impact

Long chats render and hydrate incrementally, startup performs less blocking work, and
local transcripts remain recoverable across interrupted writes, corrupt metadata,
partial server windows, network failures, identity changes, and application restarts.
Stale async work cannot replace a newer sidebar or chat selection. Completed server turns
are never dropped by an in-flight optimistic bubble, and a corrupt session no longer
disappears from the sidebar without a trace.

## Validation

- Full desktop-app suite: 1,411/1,412 passing across 135 files. The single failure is a
  known-flaky renderer anti-coupling test (`useHostPrewarmController.chatIntegration`) that
  passes in isolation — a test-ordering artifact unrelated to this diff.
- Merge/reconciliation review suite `chatMessageMerge.test.ts`: 35/35 (hand-written cases +
  `fast-check` property tests, including idle-echo collapse, metadata preservation, the
  `task_id`-no-leak guard, and the R1-B1 no-loss invariant).
- `chatStore.test.ts`: 78/78. `rpcProxyClient.test.ts`: 57/57.
- `npm run build:main` (`tsc -p tsconfig.json`): clean.
- `npm run build:renderer` (existing bundle-size warning only).
- UI typecheck has only four known pre-existing fixture errors in
  `useGfsBrowserController.test.tsx` and `SettingsPage.test.tsx`.
- Prettier verification for all changed files.
- `git diff --check`


josue-tech-94:
@claude Please perform a deep, focused review of this draft PR at its current head.

This PR contains the `desktop-app` changes extracted from PR #173:
https://github.com/evenfire-ai/evenfire/pull/173

It intentionally targets `feat/desktop-loading-times-improvements`, allowing the desktop
work to be reviewed independently before being merged back into PR #173's branch. Review
only this PR's diff against its target branch; do not re-review the backend service
changes already present in the base.

Please independently verify the earlier desktop findings and examine their underlying
failure modes. Focus especially on:

- Paged chat durability, atomic replacement, recovery, quarantine, cleanup, and downgrade
  compatibility.
- Preservation of local-only, failed-send, optimistic, system, error, attachment, and
  active-task messages during partial-window reconciliation.
- Reconciliation idempotency, renderer/main merge behavior, ordering, paging gaps, and
  bounded delta recovery.
- Chat-switch, reset, logout, identity-scope, and stale async-response races.
- IPC pagination validation, filesystem path safety, permissions, and transient I/O
  failures.
- Startup deferral, cache invalidation, long-chat rendering, and actual loading-time
  behavior.
- Similar root-cause issues and regressions not explicitly covered by previous comments.

Report only current, reproducible actionable findings by severity with file and line
references. Distinguish genuine defects from false positives or documented tradeoffs,
and identify missing regression coverage.

josue-tech-94:
@claude Please perform a deep, focused follow-up review of the latest PR head.

Review the PR diff and the directly affected code paths. Verify findings against
the actual implementation; do not report speculative style issues.

Focus especially on:

1. Crash durability and data-loss safety
   - Atomic JSON/index protocol: write, file fsync, rename, and parent-directory fsync.
   - Normal snapshot promotion, recovery promotion, rollback, and quarantine moves.
   - Error handling and cleanup at every interruption point.
   - Windows behavior when directory fsync is unsupported.
   - Confirm that missing, torn, or unknown index files cannot trigger deletion of
     recoverable paged transcripts.
   - Class-check sibling rename/write paths for the same durability defect.

2. Async agent/chat races
   - Confirm stale `getIndex` results and retry failures cannot overwrite or clear
     another agent's sidebar.
   - Confirm `stillRelevant` is forwarded end-to-end and checked before and between
     delta-pagination requests.
   - Look for equivalent post-await state mutations in directly related hooks.

3. Token safety
   - Confirm every legacy-token migration persists successfully before deleting its
     source across keychain, encrypted-file, and plaintext backends.
   - Confirm logout and rejected-session cleanup remove only the current environment
     and explicitly supported aliases without leaking across environments.

4. IPC and metadata boundaries
   - Confirm the three changed IPC limits reject non-positive, fractional, infinite,
     NaN, and unsafe integer values while preserving valid defaults.
   - Confirm invalid paged metadata counts enter a safe recovery path without data
     loss or arithmetic corruption.
   - Check for contract or backward-compatibility regressions caused by these checks.

5. Regression tests
   - Verify each new test exercises the real failure mode and would fail before the
     fix.
   - Identify missing interruption points, race orderings, or backend cases that
     remain materially untested.

For every finding, provide:

- Severity: Blocker, Critical, High, Medium, or Low.
- Exact `file:line`.
- The defect in one sentence.
- A concrete failure scenario.
- Whether it was reproduced or established by code inspection.
- The smallest safe remediation.

Deduplicate related findings and clearly separate merge blockers from follow-ups.
Also include a “Verified clean” section for the focus areas that passed review.
If no merge-blocking issue remains, state that explicitly.

josue-tech-94:
@claude Please perform a deep, focused follow-up review of the latest PR head.

Review the PR diff and the directly affected code paths. Verify findings against
the actual implementation; do not report speculative style issues.

Focus especially on:

1. Crash durability and data-loss safety
   - Atomic JSON/index protocol: write, file fsync, rename, and parent-directory fsync.
   - Normal snapshot promotion, recovery promotion, rollback, and quarantine moves.
   - Error handling and cleanup at every interruption point.
   - Windows behavior when directory fsync is unsupported.
   - Confirm that missing, torn, or unknown index files cannot trigger deletion of
     recoverable paged transcripts.
   - Class-check sibling rename/write paths for the same durability defect.

2. Async agent/chat races
   - Confirm stale `getIndex` results and retry failures cannot overwrite or clear
     another agent's sidebar.
   - Confirm `stillRelevant` is forwarded end-to-end and checked before and between
     delta-pagination requests.
   - Look for equivalent post-await state mutations in directly related hooks.

3. Token safety
   - Confirm every legacy-token migration persists successfully before deleting its
     source across keychain, encrypted-file, and plaintext backends.
   - Confirm logout and rejected-session cleanup remove only the current environment
     and explicitly supported aliases without leaking across environments.

4. IPC and metadata boundaries
   - Confirm the three changed IPC limits reject non-positive, fractional, infinite,
     NaN, and unsafe integer values while preserving valid defaults.
   - Confirm invalid paged metadata counts enter a safe recovery path without data
     loss or arithmetic corruption.
   - Check for contract or backward-compatibility regressions caused by these checks.

5. Regression tests
   - Verify each new test exercises the real failure mode and would fail before the
     fix.
   - Identify missing interruption points, race orderings, or backend cases that
     remain materially untested.

For every finding, provide:

- Severity: Blocker, Critical, High, Medium, or Low.
- Exact `file:line`.
- The defect in one sentence.
- A concrete failure scenario.
- Whether it was reproduced or established by code inspection.
- The smallest safe remediation.

Deduplicate related findings and clearly separate merge blockers from follow-ups.
Also include a “Verified clean” section for the focus areas that passed review.
If no merge-blocking issue remains, state that explicitly.

josue-tech-94:
@claude Review the latest PR #220 head only. Ignore previous Claude bot comments
and prior failed Claude runs. Use only the PR body, latest diff against
origin/feat/desktop-loading-times-improvements, and human comments as context.

Focus on crash durability, async races, token safety, IPC/meta boundaries, and
regression tests. Report merge blockers separately from follow-ups.

zach88:
Took a deep pass through this one at `ab3dd91`, across durability, merge/reconcile, IPC + tokens, controller races, rendering, security, and contract consistency.

Short version: this is careful work and a lot of it holds up (the fsync protocol, `pathSafety.ts`, the token migration, the delta loop bounds), but I'm blocking on 4 issues. Three of them lose or corrupt data *durably*, meaning it gets written to disk and survives a restart. The damage is concentrated in `chatMessageMerge.ts` and `chatStore.ts`.

The thing I'd flag hardest: **4 of the 10 round 1 fixes broke something else**. Details in the regressions section, since that's the most actionable part of this.

## Regressions from the round 1 fixes

| Finding | Introduced by | Was closing |
|---|---|---|
| `R2-B3` | `b3a5cce` | R1-H1 |
| `R2-H1` | `65e402e` | R1-M5 |
| `R2-H2` | `74bd85b` | R1-H3 |
| `R2-H4` | `ab3dd91` | R1-H2 |
| `R2-L1` | `74bd85b` | R1-H3 |

That's 5 of this round's 33 findings, and 1 of the 4 blockers. Each of those fixes was individually well formed (one finding per commit, with `Class-check:` and `Repro-test:`). The failure mode wasn't sloppiness, it was that the collateral of each fix didn't get audited before it landed.

## Blockers

**1. [R2-B1] Turnless local paired with a suppressed server message drops both**

`desktop-app/src/chatMessageMerge.ts:219`

Loop 1 (line 186) filters `!suppressedServerMessages.has(serverMessage)`. Loop 2 (line 219) doesn't. So a turnless local gets marked as replaced against a server message that then gets filtered out at line 240, and `insertUnmatchedTurnlessIncoming` won't re-add it because its id is already in `removedIds`. Both vanish.

Repro, with the server fixture derived from the real producer (`turnsToChatMessages`):

```
local:   turn-2-user, echo-answer-A (turnless, task-A done), optimistic-B x2 (task-B live)
server:  turnsToChatMessages([{ number: 2, user_input: 'question A', response: 'answer A' }])
activeTaskIds: { task-B }

result:  ["turn-2-user", "optimistic-user-B", "optimistic-assistant-B"]
```

The answer to turn 2 is gone entirely, neither the server's `turn-2-assistant` nor the local echo. `replaceMessages` writes `merged` wholesale, so this lands on disk. It's also the reason `merge(merge(a,b), b) !== merge(a,b)`: a message can survive the first reconcile and disappear on the second.

The affected sequence is "ask A, get an answer, ask B", which is the most common thing the app does.

Fix is one line, adding the guard loop 1 already has:

```ts
const authoritativeCandidates = authoritative.filter(
  serverMessage =>
    !suppressedServerMessages.has(serverMessage) &&   // <- missing
    serverMessage.role === message.role &&
    serverTurnAllowedForTurnlessLocal(serverMessage, message, index)
)
```

I verified that patch: it closes both repros and the full suite stays at 1331/1331 with no collateral.

**2. [R2-B2] A local cache without `serverTurnNumber` gets duplicated wholesale, and the duplicate is persisted**

`desktop-app/src/chatMessageMerge.ts:228`, `desktop-app/ui/src/hooks/domain/useAgentChatController.ts:1451-1471`

`hydrateActiveChatFromServer` went from *replacing* with the server window to *merging* with it. The new guard only bails when the server didn't grow. When it did grow, the merge can't resolve turnless locals if more than one server turn is a positional candidate (`if (candidateTurns.size !== 1) continue`), so the locals survive *in addition to* the server's.

Repro:

```
local:  4 messages with no serverTurnNumber (q1,a1,q2,a2)
server: turnsToChatMessages(turns 1,2,3) = 6 messages

base c6dee60:  ["q1","a1","q2","a2","q3","a3"]                       (6)  ok
head ab3dd91:  ["q1","a1","q2","a2","q1","a1","q2","a2","q3","a3"]   (10) duplicated
```

The scope here is the part I'd want you to look at first: the pre-PR adapter (`ui/src/hooks/sessionAdapter.ts` at `c6dee60`) has **zero** references to `serverTurnNumber`, and the new `src/serverTurnAdapter.ts` writes it. So every local cache currently on disk lacks the field. First reconcile after upgrading, any chat that grew server side duplicates and persists that way. It's stable in the wrong state, so it doesn't self heal.

Minimal fix: when `!localHasServerTurns && localMessages.length > 0`, keep using the replace path. A cache with no turn numbers has no comparable identity against the authoritative window. Keeping the merge means resolving turnless locals first, which is new decision logic and probably wants a design pass rather than a patch.

**3. [R2-B3] Index fail-closed has no repair path, so a torn `index.json` bricks writes for that agent permanently**

`desktop-app/src/chatStore.ts:1416-1433` (regression from `b3a5cce`)

The store used to self heal here: `getIndex` returned `emptyIndex()` and the catalog got rebuilt from the server. Now it throws, and there's no repair path anywhere. `maybeWipeLegacyCache` deliberately preserves unreadable catalogs, and `index.json` has no quarantine and no rebuild. `meta.json` does have both (`recoverPagedChatUnlocked` -> `.corrupt/` -> `rebuildPagedChatFromSurvivingFilesUnlocked`), so the index just got the "refuse" half without the "recover" half. I grepped for quarantine/rebuild against the index specifically and there's nothing.

Repro:

```
index.json = '{"version":2,"chats":[{"id":"c1"'     # torn write

                                     ed7ff03    ab3dd91
loadMessages('agent-1','c1')         ["m1"]     ["m1"]
appendMessages('agent-1','c1',[m2])  ok         throws
createChat('agent-1','c-new')        ok         throws
```

`appendMessages` is the send path, so the user can't send another message in any chat for that agent, ever, with no UI to fix it. Same for `renameChat`, `deleteChat`, `setLastActiveChatId`, `replaceMessages` (11 `saveIndex` sites).

To be clear, the intent of R1-H1 is right and the code comment explains it well. Treating corrupt bytes as an empty writable index would orphan the preserved chats. It just needs the other half: on `parseChatIndex` failure, rename to `index.json.corrupt-<uuid>` once and return `emptyIndex()`. The catalog rebuilds from the server and the paged transcripts (the thing R1-H1 was protecting) stay untouched on disk. That closes `R2-M6` too.

**4. [R2-B4] `hydrateActiveChatFromServer` can't see `stillRelevant`, so an A-B-A superseded reconcile repaints and persists a stale window**

`desktop-app/ui/src/hooks/domain/useAgentChatController.ts:1434-1494`

`supersede()` protects the gate checkpoints, and `attachLiveTask`/`settleIdle` both take `stillRelevant` and recheck it. But `hydrateActiveChatFromServer`, which is where *all* the state mutation and persistence happens, only uses `isActive()`. That compares `selectedAgent` + `activeChatId` and doesn't include `activeChatSwitchRequestRef` or the reconcile generation, so after A-B-A `isActive()` is true again.

Repro:

1. `switchToChat(A)`, reconcile #1 gets `turn(1,'question','stale answer')`, hydrate parks on its `loadMessages`
2. `switchToChat(B)` then `switchToChat(A)` (second A, legitimately fresh), fresh reconcile paints `['question','fresh answer']`
3. run #1's deferred `loadMessages` resolves, `isActive()` is true, `setChatMessages` + `chatStore.replaceMessages` both fire
4. result: `['question','stale answer']`, telemetry says `reconcile_replaced` rather than `stale_drop`

The old answer overwrites the new one and gets persisted. `switchToChat.unified.test.tsx:122` covers the case where the old run is parked in the *fetch*, not parked inside hydrate, which is where the side effects live.

Fix: thread the gate's `stillRelevant` into `hydrateActiveChatFromServer` (both callers already have it) and recheck before `setChatMessages`, `setHasOlderMessages`, `chatStore.replaceMessages` and `loadedLocalMessageCountRef`.

## High

**5. [R2-H1] The wire parser requires turn >= 1 but the producer emits 0** (regression from `65e402e`)

`desktop-app/src/rpcProxyClient.ts:186,196,244`

`wireSafeInteger(..., 1)` on `oldestTurnNumber`, `latestTurnNumber` and `turns[].number`. The real producer floors at 0, in three places:

- `mcp-host/src/core/conversation/persistence/reconstruct.ts:149` -> `const turnNum = m.turn_number ?? 0`
- `mcp-host/src/db/statements.ts:110` -> `COALESCE(MAX(turn_number), 0)`
- `mcp-host/src/db/worker/protocol.ts:67` -> `turn_number: number | null`

The rpc-proxy passthrough test on this PR's base literally uses `?limit=5&afterTurn=0`.

Repro: a response with `oldestTurnNumber: 0` / `turns[0].number: 0` throws `Invalid session messages response.oldestTurnNumber`. `appService.loadSessionMessages` only catches the 401 branch, so it propagates to the renderer and the chat never hydrates. Before this PR the raw `response.json() as ...` passed it through.

Worth calling out how this got through: the only fixture in `rpcProxyClient.test.ts` uses `number: 1`, and there's no test anywhere with `number: 0`. The green test pins the contract as imagined rather than the one that's deployed, so validation added to protect against malformed responses now rejects legitimate ones. Fix is dropping the minimum to 0 in all three and adding the `number: 0` case.

**6. [R2-H2] `switchToChat` supersedes unconditionally, killing the loud `stream_lost` reconcile** (regression from `74bd85b`)

`desktop-app/ui/src/hooks/domain/useAgentChatController.ts:795`

`supersede(key)` fires on the first line even when the target is the already active chat. That reconcile is the only owner of the Resend / `getTaskResult` recovery UX.

Repro, same test on both heads:

```
ed7ff03: LOUD OUTCOME = fell_through_to_resend -> applyFallback() -> Resend banner
ab3dd91: LOUD OUTCOME = noop -> dropActivity() + setIdle(), no banner, no durable result lookup
```

Reachable in production: `switchToChat` on the already active chat fires from auto-select (route change, agent/team change) and from clicking the same chat in the sidebar. The loud reconcile easily runs over a second because of the network retries, which is exactly the stream-loss window. The lost message disappears with no recovery affordance.

Note for whoever picks this up: "detach stale same-chat single-flight work" was a *suggestion* from the previous round, and applying it literally is what caused this. Worth auditing the collateral before re-applying any variant.

Fix: scope it, e.g. `if (activeChatVisibilityRef.current.activeChatId !== chatId)`, or skip supersede when the in-flight reconcile came from a loud caller.

**7. [R2-H3] `loadChatList`'s failure path isn't generation guarded, so a stale failure blanks a freshly loaded sidebar**

`desktop-app/ui/src/hooks/domain/useChatListController.ts:381`

This is the same asymmetry R1-M1 was closing, one branch over:

- success (line 178): `selectedAgentRef.current !== agentRef || requestGenerationRef.current !== requestGeneration`
- failure (line 381): `if (selectedAgentRef.current === agentRef) setChatList([])`, agent only

Repro: `getIndex` rejects (the store-rebind transient the code comment itself describes), waits 300ms, a second load lands `['c1']` in that window, then the stale retry fails again and calls `setChatList([])`. Sidebar goes empty with the chats already loaded.

Suggestion: fix this together with **[R2-M1]** below in a single class-level commit ("guard every post-await setter in useChatListController") rather than two instance-level ones.

**8. [R2-H4] Page contiguity invariant was added to the reader but not the writer** (regression from `ab3dd91`)

`desktop-app/src/chatStore.ts:531-542` vs `:898-916`

`adoptDurablePageWritesUnlocked` adopts orphans without requiring contiguity, under a comment that's now false ("it is not referenced by metadata, so it cannot break reads"). So the store writes metadata its own reader rejects.

Repro:

```
                            ed7ff03                   ab3dd91
loadMessages (1st open)     6 messages                6 messages
loadMessages (2nd open)     6 messages, meta intact   4 messages, meta rewritten,
                                                      prunedBeforeCount 8 -> 10
```

Every affected chat pays a full transcript rewrite and loses local window. Fix: `break` on the first skipped orphan so the written list satisfies the invariant, or relax the reader to "strictly increasing".

**9. [R2-H5] `tool_steps` is the one wire field that skips validation, and it ends in a blank window that survives restart**

`desktop-app/src/rpcProxyClient.ts:275`

```ts
...(entry.tool_steps !== undefined ? { tool_steps: entry.tool_steps as never } : {}),
```

Every sibling field gets validated. Chain, verified end to end:

1. `parseSessionMessagesResult` returns it verbatim
2. `serverTurnAdapter.ts:55` does `turn.tool_steps?.length`, so `"PWNED".length === 5` and it lands on the message
3. persisted unvalidated (`ipc.ts` casts `messages as ChatMessage[]`)
4. `ChatThread.tsx:485-487` passes the guard, then `toolSteps.map(...)` throws

There's no error boundary anywhere in `ui/src` (grepped for `componentDidCatch|getDerivedStateFromError|ErrorBoundary`, nothing), so the throw takes down the whole tree and you get a blank window. The poison is on disk and the chat gets reselected via `lastActiveChatId`, so it survives restart and isn't recoverable without wiping `~/.clerum/chats`.

`as never` is also assignable to anything, so it turns off the compiler at exactly the boundary this PR is hardening. Fix: a `parseToolSteps` mirroring `parseTokens`, plus an `Array.isArray` guard in `renderHydratedToolSteps` for defense in depth.

**10. [R2-H6] A failed local read gets persisted as zeroed aggregates and never retried**

`desktop-app/ui/src/hooks/domain/useActivityController.ts:122-125`

```ts
const messages = await chatStore.loadMessages(...).catch(() => [])   // error == empty
await chatStore.backfillCounters(selectedAgent, chatId, messages)     // persists the empty
```

`updateChatIndexFromMessages` with `preserveExistingTotals` does `Math.max(chat.errorCount ?? 0, 0)`, so `errorCount` goes from `undefined` to `0`, and the retry gate is `errorCount === undefined`. That chat is excluded forever. Repro'd against a real `ChatStore` in a temp dir: expected `{errorCount:1, toolCallCount:1}`, got `{0, 0}`.

Same defect without any read error: `ACTIVITY_SUMMARY_MESSAGE_SCAN_LIMIT = 1000` freezes the count at the most recent window for longer chats.

**11. [R2-H7] A dev run with localhost env vars discards the persisted environment profile**

`desktop-app/src/config.ts:503-511`

`shouldPreferLocalhostByDefault()` lost the `EVENFIRE_RENDERER_URL` condition, so `!app.isPackaged` plus the two env vars is now enough.

Repro, same inputs on both heads (`isPackaged:false`, saved `activeProfileId:'prod'`):

| head | `externalRestApiBaseUrl` | `activeOptionId` |
|---|---|---|
| `c6dee60` | `https://api.example.com` | `prod` |
| `ab3dd91` | `http://127.0.0.1:8091` | `__localhost__` |

`config.runtimeConfig.test.ts:118` pins this as golden for the packaged+flag case, but the implementation applies it to all `!app.isPackaged`. Knock-on: `getActiveEnvKey()` changes, so different token slot and different chat-store namespace, and the dev loses their session and chats for the remote profile on every launch.

## Medium

**12. [R2-M1] `handleCreateChat` writes agent A's chat into agent B's list.** `useChatListController.ts:538`, no guard at all after the `createChat` await. Repro'd. Pre-existing rather than new, but it's the exact class R1-M1 declared closed. Pairs with `R2-H3`.

**13. [R2-M2] Turn anchoring collapses both slots of a turn and reorders interleaved locals.** `chatMessageMerge.ts:260-282`, `exactTurnAnchor` always returns the first index for that turn. Repro: `[turn-1-user, local-warning, turn-1-assistant]` comes back as `[turn-1-user, turn-1-assistant, local-warning]`. Also means `merge(a,a) !== a`.

**14. [R2-M3] Suppressing a server slot for a live optimistic bubble drops the authoritative message from the persisted snapshot.** `chatMessageMerge.ts:202-210`. Found by fuzzing (4k seeds), minimized: the first reconcile discards an authoritative turn, the second one accepts it.

**15. [R2-M6] Index `version: 3` reads fine but blocks every write.** `chatStore.ts:158` accepts v2 and v3, `saveIndex:1436` throws on anything but v2, and the `types.ts:810` comment claims a normalization that doesn't exist. Repro'd. Closes with `R2-B3`.

**16. [R2-M7] Optional wire fields treat `null` as invalid.** `rpcProxyClient.ts:152,38,56`. Doesn't break today since the host omits rather than nulls, but the sibling endpoint (`activityHub.ts:111`, `main.ts:1989`) already emits `nextCursor: null`, so that convention exists in the same service. If `/sessions` aligns with it, `useChatListController.ts:190` swallows the throw and the sidebar silently stops seeing remote sessions.

**17. [R2-M8] The parser fails closed on the whole response for one bad item or an unknown `state`.** `rpcProxyClient.ts:96-119`. A new lifecycle state (an additive, backward compatible change by contract) leaves the desktop with no session catalog at all rather than one degraded session.

**18. [R2-M9] Pagination errors are classified by regex over the serialized message.** `useChatListController.ts:79-82`. `ApiError.status` doesn't survive the IPC hop, so a 429 or 408 matches `/\b4\d\d\b/`, "Load more" disappears, and older sessions are unreachable until the list reloads.

**19. [R2-M10] `messageCount` mixes visible-only semantics with the paged count.** `chatStore.ts:940,991` push `PagedChatMeta.messageCount` (includes `system`) into a field `updateChatIndexFromMessages` maintains as user+assistant only, and `updateIndexedMessageCount` only ever increases. Repro'd.

**20. [R2-M11] Local counters are monotonic and never reconcile.** `chatStore.ts:1676` (`+=`, no dedupe) and `:1569` (`Math.max`). A failed send plus a retry with a new id counts 2, the authoritative merge consolidates to 1, `max(2,1)` stays 2 forever.

**21. [R2-M12] Auto-select assumes the server returns sessions newest-first without checking.** `useChatListController.ts:239` takes `serverSessions[0]`, and neither `parseSessionsListResult` nor the renderer sorts this list (unlike the cross-agent loader, which does `.sort(byUpdatedDesc)`).

**22. [R2-M13] Positional `#N` chunk suffix remounts chunks when a local group sits between server turns.** `ChatThread.tsx:65-95`. Verified with mounted DOM: append and prepend both preserve node identity (that part is a real improvement), interleaved local doesn't. Remount closes expanded `<details>`, recreates `SecureHtmlPreview` iframes, and drops the recorded size so `contain-intrinsic-size` falls back to the 720px estimate.

**23. [R2-M14] "Load older" prepend doesn't compensate `scrollHeight`.** `styles.css:6066-6071` plus `useAgentChatController.ts:949-1020`. No explicit scroll compensation anywhere and no `overflow-anchor` in the stylesheet. The worst case (yank to bottom) doesn't happen since `isNearBottom()` is false when the user is scrolled up, but reading position isn't preserved and chunks resolving from the 720px estimate shift content as you scroll up.

**24. [R2-M15] Counter backfill ships up to 20x1000 messages over IPC to the process that already has them on disk.** `useActivityController.ts:119-140`. `ipc.ts:1189` just forwards to `updateChatIndexFromMessages`. Messages carry `attachments` with `dataBase64`, so this can move megabytes, and it sits in the deferred startup path of a PR whose goal is reducing startup work.

**25. [R2-M16] The REST-only legacy env key is shared across profiles with the same REST origin.** `appService.ts:561`, `chatStoreBinding.ts:110-146`. Two profiles with the same REST origin and different RPC origins compute the same legacy key, so whichever opens first inherits the other's token and renames its chat tree in. The diff's own doc comment explains why REST origin alone isn't a sufficient boundary.

**26. [R2-M4] `legacyEnvKeys` is wired into 1 of 6 `bindChatStoreForUser` call sites.** `appService.ts:591` passes it, but `:406`, `:414`, `:656` (`completePasswordLogin`), `:727` (`googleLogin`) and `:1510` don't. `maybeMigrateEnvScopedCache` only runs when it's present, so a user who upgrades and does a fresh login rather than a token restore hits `:656` first and their old env-key chat tree is orphaned until the next launch. `appService.bootstrap.test.ts:129-132` asserts the two-argument call, so the test pins the asymmetry.

**27. [R2-M5] No retry after a transient restore failure, for the whole process lifetime.** `appService.ts:1063-1069`. `restoreSavedSessionOnce` keeps the token ("keeping token for retry") but `getSessionState()` blocks any retry while `savedSessionRestoreAttemptedEnvKey === getActiveEnvKey()`, and reselecting the same environment doesn't reset it. A one second `fetch failed` at startup forces a re-login even though a valid token is saved, and the test name ("does not repeat a failed startup restore before the login screen can paint") pins the non-retry.

**28. [R2-M17] `prunedBeforeCount` is dropped on write and has no consumer.** `chatStore.ts:1211-1220`. The only "there is older local history" marker the persistence layer produces never reaches the UI, so the renderer infers it from turn numbers instead.

### Test quality (flagging these as findings, not nits)

**29. [R2-M18] No property tests on the merge function.** `chatMessageMerge.test.ts` is 857 lines and 19 hand-written cases, on a pure merge function with precedence rules. A 4k-seed fuzz surfaced `R2-B1`, `R2-M2` and `R2-M3` in seconds, and the two cases that claim to cover idempotency don't fail on any of the three. Worth adding `fast-check` (dev dep, works with vitest, no runtime deps) with the five domain properties: total ordering, no duplicate ids, no unrepresented local dropped, idempotency, composability.

**30. [R2-M19] The body's staleness invariant has no test for the sidebar half.** "Stale async work cannot replace a newer sidebar or chat selection" has coverage for the chat half, none for the list. `R2-H3` is that invariant being violated.

**31. [R2-M20] `chatThread.chunking.test.ts` asserts the bucketing function, not the mounted transcript.** Its third case actually documents the positional-suffix behavior from `R2-M13` as correct, which is a good illustration of why the observable output is what needs asserting here.

**32. [R2-M21] Attachment fixture can't be produced by the real pipeline.** `chatMessageMerge.test.ts:6-50` writes `content: '[Image attached]'` with no attachments and asserts the local ones are kept. With the real `user_input` a turn with an image produces, `parseChatMessageDisplay` generates attachments and the merge takes the server's, so the observable result is the opposite of what the test pins.

**33. [R2-M22] Two repro tests don't demonstrate a red.** `57c4b22`'s 84 tests all pass against `ed7ff03` (ran it), so R1-L1 is a coverage close rather than a defect close. And `74bd85b`'s `reconcileChat.test.ts` fails on the previous head with `supersede is not a function`, which is a setup failure rather than an assertion failure. The companion `switchToChat.unified.test.tsx` case does fail properly, so that finding is genuinely covered.

## Smaller stuff

- **[R2-L1]** `supersede` skips the `finally` that dispatches `RECONCILE_FINISHED`, so "Syncing..." can stick. `reconcileChat.ts:280-285`, regression from `74bd85b`. Repro'd.
- **[R2-L2]** `replaceLocalWindow` is dead across three layers: declared in `types.ts:1054`, sanitized and forwarded in `ipc.ts:1183`, never read in `chatStore.ts`. Either implement it or drop it.
- **[R2-L3]** `pathSafety.ts` has no unit test of its own despite gating filesystem segments and RPC routes. I probed it against 38 hostile vectors and it rejected all of them, so this is about locking that in. Minor gaps: `CONIN$`/`CONOUT$` aren't in the reserved-name regex, no NFC/NFD normalization, and `maxLength` counts code units rather than bytes.
- **[R2-L4]** `pendingChatSelectionByAgentRef` isn't cleared in `clearList()` (`useChatListController.ts:118`), so a pre-logout selection can survive into the next user's session.
- **[R2-L5]** `useActivityController.ts:105` lost the `NaN` guard the deleted helper had, so an unparseable `updatedAt` gives an undefined sort order.
- **[R2-L6]** `sanitizeOptionalPositiveInteger` has a floor but no ceiling on three handlers (`ipc.ts:522,759,1331`), asymmetric with the siblings that clamp to 100/200/1000.
- **[R2-L7]** `--evenfire-desktop-dev-package` re-opens env-driven runtime config in a packaged build (`config.ts:489-492`). Bounded by `runtimeEndpointsMatch(..., localhostRuntimeConfig)`, so localhost only.
- **[R2-L8]** The symlink guard lstat's the leaf only and there's a TOCTOU window before `readdir`/`rm` (`chatStoreBinding.ts:302-344`). Same-UID only given the `0o700` dirs, so defense in depth.
- **[R2-L9]** New typecheck error at `useAgentsDataController.test.tsx:14` (the `AccessCatalog` fixture adds `mcpServerNames`/`contextNames` and omits five required fields). Invisible because `ui/tsconfig.json` isn't run by any script and `build:renderer` is vite.
- **[R2-L10]** "Show less" is unreachable with under 10 local chats and `hasMore` remote (`ChatThread.tsx:295-297`).
- **[R2-L11]** "Load more" re-enables on agent reselect while the request is still in flight, so the click is a silent no-op (`useChatListController.ts:172-175`).
- **[R2-L12]** `keytar.deletePassword(...).catch(() => {})` no longer covers a sync throw the way the old `try/catch` did (`tokenStore.ts:363`).

## What checked out clean

Verified rather than assumed, and there's a lot of it:

**Durability.** The fsync protocol is correct and no write path skips it: tmp -> `syncFile` -> `rename` -> fsync of both source and target dirs. `syncDirectory` only swallows on Windows and rethrows elsewhere. Snapshot promotion has no window where the only good copy is unreferenced. `adoptDurablePageWrites` rethrows non-ENOENT FS errors instead of treating them as corruption. `maybeWipeLegacyCache` only deletes on `version === 1`; corrupt, future-version and I/O-error catalogs are all preserved.

**Concurrency.** `serialize` is a correct single-flight, and lock acquisition is always chat then index, so no deadlock.

**Security.** `pathSafety.ts` rejected all 38 hostile vectors I threw at it and is strictly stronger than the `assertSafeSegment` it replaces. All 15 `join(` sites in `chatStore.ts` are covered, and all 14 `fs.rm`/`unlink` calls target validated paths. Files are `0o600` and dirs `0o700`, which is new here (base inherited the umask). The token temp file is opened with `fs.open(path,'w',0o600)` before the secret is written, so there's no permissive window. Zero `dangerouslySetInnerHTML`/`innerHTML`/`eval` in `ui/src`. `package-lock.json` is a version bump and nothing else, no new packages.

**Tokens.** Persist-before-delete is correct across all three backends and all six migration paths, with no `.catch()` wrapping the persist. `clearSessionToken` can't reach another environment's tokens.

**Races that are done right.** `stillRelevant` does propagate end to end and is rechecked *between* successive delta pages, so the dropped-4th-argument issue is genuinely closed. The delta loop is bounded three independent ways (page cap, `seenCursors`, explicit non-advancing-cursor bail), so a server answering `hasMoreAfter: true` forever can't spin the client. `loadOlderMessages` cancellation is correct in both A-B and A-B-A, and those tests use real out-of-order resolution and assert the rendered list. `scheduleAfterFirstPaint` has a proper fallback (`setTimeout(1000)` registered before the double rAF, idempotent) and is cancelled on logout and identity change.

**Contract.** `pathSafety.ts` mirrors the rpc-proxy rules exactly. `renderer.d.ts` shrinks because it swaps six inline types for four richer named imports, so bridge typing improves with no `any`/`unknown`/index signature introduced. `sessionAdapter.ts`'s 36 deleted lines are a pure extraction with the public signature intact. New client against old server degrades correctly (server ignores the params, returns everything, pagination fields are optional), so there's no merge-order dependency here.

**Renderer.** Streaming does not re-chunk the transcript per token. Chunk keys are stable across append and prepend, which is a real improvement over the previous positional keys that remounted everything.

**Suites and builds**, run against this head:

| | |
|---|---|
| `npx vitest run` | 1331/1331 across 134 files, matches the body exactly |
| `npm run build:main` | pass |
| `npm run build:renderer` | pass |
| net diff of `ui/src/pages/` (AuthPage restore claim) | empty, claim holds |
| `npx tsc --noEmit -p ui/tsconfig.json` | 5 errors: 4 pre-existing, 1 new (see R2-L9) |

## Suggested order

`R2-B1` first since the patch is one line and already verified. Then `R2-B3` (index quarantine + rebuild, closes `R2-M6`), `R2-H1` (trivial, and it's breaking against the real server today), `R2-H5`, and `R2-H3` + `R2-M1` together as one class-level commit.

On `chatMessageMerge.ts`: `R2-B1`, `R2-B2`, `R2-M2` and `R2-M3` all live there, and the file already carries 9 `fix:` commits on this branch with almost identical subjects ("preserve desktop chat history during reconciliation", "preserve live chat turns", "close desktop chat history residual gaps"). `R2-B2`, `R2-M2` and `R2-M3` don't read like missing cases to me, they read like the precedence and anchoring table itself. Half a page writing that out (for each input kind: what wins, which anchor it uses, what happens when the authoritative side doesn't supply one) plus the property tests would probably cost less than another round of patches. Happy to be wrong on that if you see a smaller cut.

josue-tech-94:
@claude Please review the latest PR #220 head after the final Codex fixes for Zach’s R2 blockers and the follow-up R3 blockers.

Review the current PR diff against `feat/desktop-loading-times-improvements`. Do not rely on stale intermediate comments or commit messages.

The latest local review found:

- Decision: Approve with follow-ups.
- No must-fix items found.
- Reviewed head: `e46b7e449238a3770ebdd083ce742397c4204d59`.
- Full desktop suite: 1,399/1,399 tests passed across 135 files.
- Focused review suites: 287/287 tests passed across 10 discovered files.
- `npm run build:main`: passed.
- `npm run build:renderer`: passed, with only the existing bundle-size warning.
- UI typecheck has only four known pre-existing fixture errors.
- `git diff --check`: passed.

Please verify the final fixes especially around:

1. R2-B2 legacy turnless cache replacement
   - Legacy local caches without `serverTurnNumber` must not duplicate wholesale.
   - Producer-shaped server turns 1/2/3 plus local q1/a1/q2/a2 should render and persist exactly six rows, not ten.
   - Durable local system, error, and failed-send messages should still be retained.

2. R3-B1 future-version index handling
   - A syntactically valid future-version `index.json` must remain byte-preserved and fail closed.
   - Ordinary create/append/replace writes must not downgrade it to v2 or mutate transcript bytes.
   - The old torn-index recursive cache wipe must not return.

3. R3-H1 durable v3 normalization
   - Known v3 indexes should remain readable.
   - v3-to-v2 normalization must happen only through the shared crash-durable writer:
     temp file, file fsync, rename, and parent-directory fsync.
   - No production binding path should bypass that durable writer.

4. Earlier R2 blockers and High findings
   - R2-B1 suppressed authoritative server messages cannot drop both local and server messages.
   - R2-B3/R2-M6 corrupt/torn indexes repair safely without bricking writes.
   - R2-B4 stale A→B→A hydrate cannot repaint or persist stale messages.
   - R2-H1 turn zero remains accepted while invalid turn values reject.
   - R2-H2 same-chat stream-loss recovery still reaches Resend / durable result lookup.
   - R2-H3/R2-M1 stale sidebar failure and create-chat paths remain generation guarded.
   - R2-H4 page metadata written by adoption remains reader-compatible.
   - R2-H5 invalid `tool_steps` cannot poison persisted state or crash the renderer.
   - R2-H6 activity read failures remain retryable.
   - R2-H7 saved runtime profiles remain selected outside explicit development launches.

5. Owner-decision follow-ups
   - Confirm R2-M3 does not block publication if the owner accepts the temporary live optimistic/server-window divergence.
   - Confirm R4-M1 is a follow-up/owner decision around uploaded-image chip metadata during settled legacy replacement.
   - Confirm repeated same-chat coalescing and startup restore retry can remain follow-ups.

Please report:

- whether any merge blocker remains;
- whether the final PR body validation text is accurate;
- whether any Codex classification is wrong or stale;
- whether PR #220 is merge-ready with follow-ups;
- the most important remaining owner decisions.

josue-tech-94:
@claude Please review the latest PR #220 head after the final Codex fixes and PR body update.

Current head:
`e46b7e449238a3770ebdd083ce742397c4204d59`

Review the current PR diff against `feat/desktop-loading-times-improvements`.
Do not rely on stale intermediate comments or old commit messages.

Please specifically verify the final fixes for:

1. R2-B2 legacy turnless cache replacement
   - Legacy local caches without `serverTurnNumber` must not duplicate wholesale.
   - Producer-shaped server turns 1/2/3 plus local q1/a1/q2/a2 should render and
     persist exactly six rows, not ten.
   - Durable local system, error, and failed-send messages should still be retained.

2. R3-B1 future-version index handling
   - A syntactically valid future-version `index.json` must remain byte-preserved
     and fail closed.
   - Ordinary create/append/replace writes must not downgrade it to v2 or mutate
     transcript bytes.
   - The old torn-index recursive cache wipe must not return.

3. R3-H1 durable v3 normalization
   - Known v3 indexes should remain readable.
   - v3-to-v2 normalization must happen only through the shared crash-durable writer:
     temp file, file fsync, rename, and parent-directory fsync.
   - No production binding path should bypass that durable writer.

Also confirm the earlier R2 blockers and High findings remain fixed:

- R2-B1 suppressed authoritative server messages cannot drop both local and server messages.
- R2-B3/R2-M6 corrupt/torn indexes repair safely without bricking writes.
- R2-B4 stale A→B→A hydrate cannot repaint or persist stale messages.
- R2-H1 turn zero remains accepted while invalid turn values reject.
- R2-H2 same-chat stream-loss recovery still reaches Resend / durable result lookup.
- R2-H3/R2-M1 stale sidebar failure and create-chat paths remain generation guarded.
- R2-H4 page metadata written by adoption remains reader-compatible.
- R2-H5 invalid `tool_steps` cannot poison persisted state or crash the renderer.
- R2-H6 activity read failures remain retryable.
- R2-H7 saved runtime profiles remain selected outside explicit development launches.

Validation reported in the PR body:

- Full desktop-app suite: 1,399/1,399 tests passed across 135 files.
- Focused merge, storage, RPC, controller, activity, config, and renderer review suites:
  287/287 tests passed across 10 discovered files.
- `npm run build:main`
- `npm run build:renderer` with only the existing bundle-size warning.
- UI typecheck has only four known pre-existing fixture errors.
- Prettier verification for all changed files.
- `git diff --check`

Please report:

- whether any merge blocker remains;
- whether the final PR body validation text is accurate;
- whether any Codex classification is wrong or stale;
- whether PR #220 is merge-ready with follow-ups;
- the most important remaining owner decisions.

zach88:
Took another pass on this after the R1-B1 / R2-H1 fixes landed. The merge work holds up well: the server-turn loss is genuinely gone (verified by reproducing the old failure against the parent commit) and the idle-echo collapse is loss-safe for text by construction. Two things worth flagging this round, plus one carryover that's now unblocked.

## High

1. **[R1-H1] Session-list parser drops malformed items silently**
`src/rpcProxyClient.ts:124-188`. `parseSessionsListResult` swallows each malformed catalog item (flatMap to `[]` in the catch, only throws if *every* item fails) and returns the partial list with no observable signal, while its sibling `parseSessionMessagesResult` throws the whole response on a single bad turn. Net effect: the server returns 50 sessions, 1 with a non-integer `turnCount`, the sidebar shows 49, and the missing chat is invisible to the consumer with no log and no trace. This was raised in round 1 and parked behind R1-B1; now that R1-B1 is closed it's unblocked. Suggested direction: keep the tolerant partial-drop policy (one corrupt session shouldn't empty the sidebar) but make the drop observable, e.g. a warn per dropped item with its index plus an additive `droppedItemCount` the renderer can read.

## Medium

2. **[R2-M1] Idle-echo collapse discards the local's non-text metadata**
`src/chatMessageMerge.ts` (`dropLocalEcho`). Reproduced. The collapse is drop-only: it discards the echoed bubble whole, including `toolSteps`/`attachments`, instead of merging them onto the surviving authoritative row the way `markLocalReplacement` does via `copyLocalMetadata`. When the optimistic bubble carried tool-steps/attachments that the reconciled server row didn't, those fields vanish from the durable transcript (they survived pre-fix as a visible duplicate). Text is never lost (the collapse only fires on a content match against the authoritative row, which survives verbatim), so it's Medium, but it's a real regression of the R2-H1 fix. Sub-case: strict `'' === ''` makes an empty / attachment-only bubble collapse on a trivial content match. Suggested direction: merge `attachments`/`toolSteps` onto the authoritative row (never `task_id`, that's echo identity) and require non-empty content to collapse.

3. **[R2-M2] The saturated-gap collapse branch (Q > P+1) has no test**
`src/chatMessageMerge.ts:277-279`. The collapse branch is deliberately kept general over `candidateTurns.size === 0` rather than restricted to adjacent turns, and the comment says why (restricting it would reintroduce a permanent duplicate in the saturated-gap case). But no test exercises `Q > P+1`, so a future "simplify to strict adjacency" refactor would silently reintroduce that duplicate. Worth a case with a saturated gap (all intermediate same-role slots already numbered) that asserts the echo still collapses.

## Regressions

- **[R2-M1]** was introduced by the R2-H1 content-gating fix (`40cc7afe`). It surfaced in the collateral pass this round rather than a round later.

## Carryover

- **[R1-M4]** `content-visibility: auto` + `contain-intrinsic-size` on off-screen chunks may break scroll anchoring on long chats. Not reproducible in jsdom, needs a visual check in Electron with a long chat before merge. Open since round 1, so it's on its second round now and needs a decision: verify in Electron or close with a note.
- **[R1-M2] / [R1-M3]** (chunking property tests, catalog-retry gate test) are coverage debt rather than live bugs; closing these out with a writeup rather than carrying them another round.

## What checked out clean

- R1-B1 closure verified by reproducing the old server-turn loss against the parent commit; the property test at 40k plus a named invariant test cover it, and the collapse can't reintroduce it (it only ever drops a turnless local, never a numbered server row).
- The idle collapse compares against the authoritative incoming row, not the possibly-stale on-disk neighbor, so the text drop is safe by construction.
- Security: no new surface. The one dependency added (`fast-check`) is dev-only, pinned, has no install scripts, and doesn't reach the production bundle.
- `chatMessageMerge` / `chatStore` / `rpcProxyClient` suites green and the main typecheck is clean. Full desktop-app suite is 1,411/1,412; the single failure is a known-flaky renderer anti-coupling test that passes in isolation.

zach88:
Quick follow-up on the two Mediums still open from my last pass: we're deferring both as out of scope for this PR and tracking them separately, so nothing is left gating this split.

- **[R1-M4]** `content-visibility: auto` + `contain-intrinsic-size` on off-screen chunks and the scroll-anchoring question. This one is unverified and not reproducible in jsdom; it needs a real visual check in Electron on a long chat, which is better done on its own than holding up the split from #173. Moving it to a follow-up with that QA attached.

- **[R2-M2]** the missing test for the saturated-gap collapse branch (`Q > P+1`). This is coverage debt, not a live bug: the branch behaves correctly today, the gap is that a future "simplify to strict adjacency" refactor could silently regress it. Tracking it as a follow-up test rather than blocking here.

Everything else from the round is closed and pushed. R1-H1 and R2-M1 are in, and the collateral pass on those fixes came back clean.

PR #205, opened 2026-07-28
fix(hcc): decouple readiness from initial fleet convergence

alfredolopez80:
> **✅ Deploy prerequisite satisfied.** This PR sets the `host-context-controller` Deployment to `strategy: Recreate`; the live HCC in clerum-dev was `RollingUpdate`, and Kubernetes rejects a `Recreate` Deployment while `spec.strategy.rollingUpdate` is present. The infra-side repair — adding the HCC to the Recreate repair targets so the field is cleaned before the server-side apply — landed in **[keyper-labs/evenfire-infra#47](https://github.com/keyper-labs/evenfire-infra/pull/47) (merged, `5d3b4470`)**. A server-side dry-run against clerum-dev confirmed the fix (the bare `Recreate` apply is rejected; the repair patch is accepted; no mutation). The clerum-dev/prod deploy of #205 now applies cleanly on the first attempt.

## Summary

This PR removes the startup coupling identified in #204: HCC can become Ready after its authoritative, fail-closed safety boundary completes, without waiting for the complete initial MCP/Context/Host fleet reconciliation.

The final safety follow-up closes the remaining same-name stale-policy and pass-currentness gaps:

- Compare canonical `NetworkPolicy.spec` values during authoritative revocation, rather than treating a matching name as proof that an allow is current.
- Revoke DNS-derived external egress allows before readiness **when the binding identity changed** (protocol/port/selector/ownership drift), because a divergent allow cannot be proven current without the deferred DNS lookup; recreate them only in the additive phase. **When the identity is unchanged (B3), the existing identity-stable allow is retained through readiness certification** — its already-resolved CIDRs are kept (no deny window) and refreshed by the bounded async resync (≤ observed TTL/2), a deliberate availability tradeoff. Readiness is therefore fail-closed against *divergent/unprovable* allows, not against a same-identity IP re-resolution.
- Fence the complete authoritative pass with Context and McpServer desired-state revisions, captured before any asynchronous default-policy setup.
- Keep the bounded, four-LIST inventory design: Context ingress, mcp-host egress, rpc-proxy egress, and external egress. Additive convergence remains asynchronous after the safety marker.
- Extend the real Minikube E2E to create a TCP external-egress policy through HCC, change the same binding to UDP while HCC is stopped, retain DNS, and prove stale-policy revocation happens before `/ready` is accepted.

[...]

PR #232, opened 2026-08-01
feat(mcp-host): authoritative fleet + MCP client hardening (split from #205)

alfredolopez80:
## What this is

The **mcp-host** slice of PR #205, extracted so it can merge on its own. zach88's round-3 review recommended splitting this out: mcp-host's three round-2 blockers are closed and pinned with mutation-goes-red, and round 3 only turned up test-quality / unreachable-code Mediums — nothing blocking.

## Extraction is diff-neutral

- Based on **current `origin/dev` (`1c8d2da`)**, which touches no `mcp-host/` file → clean fast-forward, no conflict.
- `mcp-host/` here is **byte-identical to `87c374b`** (the head zach88 reviewed) — `git diff 87c374b -- mcp-host/` is empty. This is exactly the code already reviewed, not a re-implementation.
- Touches only `mcp-host/` (14 files).

## What it contains (reviewed in #205 rounds 2–3)

- Authoritative fleet lifecycle (`authoritativeFleet.ts`): epoch capture, superseded-candidate rejection, synchronous revoke.
- MCP client/manager hardening (`client.ts`, `manager.ts`): transport close contract, timeout handling.
- `contextMapperClient` no longer turns an HTTP failure into an authoritative empty fleet.
- `authoritative?: boolean` contract — optional and skew-safe with HCC (absent field = legacy fail-closed).

## Verification (this branch)

- `tsc --noEmit`: clean.
- Full suite: **276 files / 3070 tests pass** (1 todo).

## Follow-ups (not blocking; land on this branch)

- **M9**: derive `main.mcpInitialization.test.ts` fixture from the real producer (`getStatus`/`toServerInfo`); re-label legacy-shape variants.
- **M10**: delete the unreachable `previousManager != null` branch in `replaceAuthoritativeMcpFleet` and rewrite its tests as cold-start.
- Grab-bag: dead `effects.push` after `detachServer`; dead `else replaceFleet([])` dev branch; `drainCleanups()` global ceiling.

Merge-order-free vs `dev`.

### DIFF
--- a/mcp-host/package-lock.json
+++ b/mcp-host/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-mcp-host",
-  "version": "0.1.211",
+  "version": "0.1.220",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-mcp-host",
-      "version": "0.1.211",
+      "version": "0.1.220",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.32.0",
--- a/mcp-host/package.json
+++ b/mcp-host/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-mcp-host",
-  "version": "0.1.211",
+  "version": "0.1.220",
   "description": "MCP Host - reads Host CRD and provides LLM access via OpenAI/Claude",
   "main": "dist/main.js",
   "engines": {
--- a/mcp-host/src/__tests__/main.mcpInitialization.test.ts
+++ b/mcp-host/src/__tests__/main.mcpInitialization.test.ts
@@ -1,0 +1,1934 @@
+import { describe, expect, it, vi } from 'vitest'
+import {
+  admitDevelopmentMcpServers,
+  createCoalescedPollRunner,
+  startContextMapperPolling,
+  stopContextMapperPolling,
+} from '../main'
+import {
+  AuthoritativeMcpFleetCoordinator,
+  pollAuthoritativeMcpSnapshotIfCurrent,
+  reconcileAuthoritativeMcpSnapshot,
+  replaceAuthoritativeMcpFleet,
+  runAuthoritativeMcpInitialization,
+} from '../mcp/authoritativeFleet'
+import { McpManager } from '../mcp/manager'
+import type { McpServerInfo } from '../types'
+
+function readyServer(overrides: Partial<McpServerInfo> = {}): McpServerInfo {
+  return {
+    name: 'secured-server',
+    contextRef: 'production',
+    transport: { type: 'streamableHttp', url: 'http://secured-server.test/mcp' },
+    auth: { type: 'bearer', secretRef: 'secured-server-auth' },
+    enabled: true,
+    status: { deployed: true, ready: true },
+    ...overrides,
+  }
+}
+
+function deferred<T = void>(): {
+  promise: Promise<T>
+  resolve: (value: T) => void
+  reject: (reason?: unknown) => void
+} {
+  let resolve!: (value: T) => void
+  let reject!: (reason?: unknown) => void
+  const promise = new Promise<T>((resolvePromise, rejectPromise) => {
+    resolve = resolvePromise
+    reject = rejectPromise
+  })
+  return { promise, resolve, reject }
+}
+
+function appliedAdmission(
+  effect?: (server: McpServerInfo, authToken?: string) => void | Promise<void>
+) {
+  return vi.fn(
+    async (
+      server: McpServerInfo,
+      authToken?: string,
+      control?: Parameters<McpManager['addServer']>[2]
+    ) => {
+      await effect?.(server, authToken)
+      if (control?.isCurrent?.() === false) return 'stale' as const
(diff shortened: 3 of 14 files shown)


zach88:
Took a careful pass through this one. Given it's the mcp-host slice of #205, I first confirmed the extraction claim holds: `git diff 87c374b HEAD -- mcp-host/` is empty, so this is byte-identical to the head reviewed in #205 round 3, and `dev` (`1c8d2da`) touches no `mcp-host/` file, so the fast-forward is clean. Nothing blocks the merge. `tsc --noEmit` is clean and the full suite is green (3070 passed / 1 todo). Everything below is **Low / optional** — none of it gates the merge; feel free to land as-is and pick these up as follow-ups.

## Optional (all Low, non-blocking)

1. **[R1-L1] Unreachable `previousManager != null` branch (this is the M9/M10 follow-up you flagged).** In `mcp-host/src/mcp/authoritativeFleet.ts:509-624`, the atomic-replacement path (`replaceAuthoritativeMcpFleet`) is dead in production. The only caller is `main.ts:786`, and `pollContextMapper` only calls `initializeMcpServers` when `!mcpManager` (`main.ts:859`); after a successful cold-start every later poll goes through `reconcileAuthoritativeMcpSnapshot`, so `previousManager` is always `null`. Repro: grep of call sites confirms it. Either drop it or gate it behind an assertion until a real caller exists.

2. **[R1-L2] Hand-built fixture in the init test (your M9).** `mcp-host/src/__tests__/main.mcpInitialization.test.ts:18` — the `readyServer` helper constructs `McpServerInfo` by hand instead of deriving it from the real HCC producer (`getStatus`/`toServerInfo`). The default shape happens to match what HCC emits today so it's not shielding a live contract bug, but it'll drift silently the day the producer changes.

3. **[R1-L3] Recovery precedence logic is under-described for its complexity.** `mcp-host/src/mcp/client.ts:509-586` introduces non-trivial epoch-based recovery adoption (a slower peer adopting an already-completed recovery, epoch fencing). It's correct per the tests, but it's more than "transport close + timeout handling" and has no decision table in the body. Worth a couple of property tests over the adoption rules rather than only example cases.

4. **[R1-L4] Timeout test mock doesn't abort in-flight I/O.** `mcp-host/src/__tests__/mcp/mcpClient.timeout.test.ts:47-65` — the mocked `close()` only records the call; it doesn't reject the pending `callTool`/`listTools`. The real contract test (`mcpTransportClose.contract.test.ts`) does prove the real transport aborts synchronously on close, but nothing chains that abort through to a `McpClient.callTool` rejection. Repro: the pending call resolves on its own schedule, not from the close. Low impact since `ensureCallCurrent` fencing is covered by the late-success/late-retry cases.

5. **[R1-L5] Dead delete branch for managers without `detachServer`.** `mcp-host/src/mcp/authoritativeFleet.ts:830-845` — the `effects.push` fallback is unreachable for the real `McpManager` (which always implements `detachServer`), so revoke always exits via the sync path. Fine as defensive code, but either add a test with a manager lacking `detachServer` or make the interface non-optional.

6. **[R1-L6] Dead devMode branch.** `mcp-host/src/main.ts:837` — the `else { await replaceFleet([]) }` in `initializeMcpServers` is effectively unreachable since the function is only invoked from production paths.

7. **[R1-L7] `addServer` early-return doesn't tear down an already-connected client.** `mcp-host/src/mcp/manager.ts` (the `status?.authoritative === false` branch) returns via `markNotReady` without disconnecting a live client. Not reachable today (it's always preceded by `removeServer` in `reconcileAuthoritativeMcpSnapshot`), but it's fragile defense-in-depth if that function ever gets called from another path.

8. **[R1-L8] Heads-up on a semantics change (intentional).** `mcp-host/src/contextMapperClient.ts:82-84` — `getAuthToken` now throws on HTTP 404 instead of returning `undefined`. This looks deliberate and consistent with the fail-closed model (the sole consumer already throws on a missing required token), just calling it out so it's on the record.

## What checked out clean

I verified the core of the PR against real code rather than assuming:

- **`authoritative?: boolean` skew-safety — correct, reproduced.** Absent field falls through to the legacy fail-closed path; consumers use `=== false` / `!== false` so `undefined` is never read as `authoritative=true`. Reproduced the old-server/new-host skew: legacy `ready:false` with no field → revokes (correct); explicit `authoritative:false` with unchanged desired → preserves. No mass-revoke DoS and no capability exposure. HCC doesn't emit the field today, so production is always the legacy path.
- **`contextMapperClient` no longer fabricates an empty authoritative fleet.** The `catch { return [] }` paths are gone; `listAllServers`/`listServersByContext`/`pollServers` now reject on failure and the caller preserves the prior fleet. Confirmed these tests fail against the base (which returned `[]`).
- **Synchronous revoke is idempotent** (`claimClientCleanup` → cleanup exactly once, 32/32 status tests green asserting observable state), **epoch fencing** keeps superseded candidates from winning, and the **transport close contract** is idempotent (double `disconnect()`/`retire()` → one `transport.close()`, in-flight calls reject rather than hang).
- Security-wise nothing new: no token values in logs, `encodeURIComponent` on URLs, timeouts use a fixed default rather than untrusted input.

Nice, tight extraction. Happy to see this land on its own.

zach88 (approved):

PR #233, opened 2026-08-01
fix(rpc-proxy): point HCC base-URL dev default at the real endpoint (split from #205)

alfredolopez80:
## What this is

The **rpc-proxy** slice of PR #205, extracted to merge on its own (zach88's round-3 recommendation). 4 files, one dev-default correction plus its regression fence.

## Extraction is diff-neutral

- Based on **current `origin/dev` (`1c8d2da`)**, which touches no `rpc-proxy/` file → clean fast-forward.
- `rpc-proxy/` is **byte-identical to `87c374b`** — `git diff 87c374b -- rpc-proxy/` is empty.
- Touches only `rpc-proxy/` (4 files).

## What it fixes

The `RPC_PROXY_HCC_BASE_URL` dev default pointed at `host-context-controller.mcp-server` — a Service that exists in no overlay, and one the rpc-proxy egress NetworkPolicy would block anyway (it only permits the api-gateway peer). Dead since the initial import. Points it at `host-context-controller-api-gateway.control-plane`, and adds a test pinning the dev default to the deployed configmap so the two cannot silently diverge again.

Latent, not live (NODE_ENV=production is baked into the image, so in-cluster the configmap always supplies the value), but an objectively wrong endpoint that would mislead anyone running rpc-proxy locally.

## Verification (this branch)

- `tsc --noEmit`: clean.
- Full suite: **32 files / 426 tests pass**.

## Follow-up (not blocking; lands on this branch)

- **M7**: assert the value against the Service YAML (`api-gateway.yaml` Service host+port), not only the configmap, so a Service rename is caught.

Merge-order-free vs `dev`.

### DIFF
--- a/rpc-proxy/package-lock.json
+++ b/rpc-proxy/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-rpc-proxy",
-  "version": "0.1.50",
+  "version": "0.1.51",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-rpc-proxy",
-      "version": "0.1.50",
+      "version": "0.1.51",
       "dependencies": {
         "cors": "^2.8.5",
         "express": "^4.21.2",
--- a/rpc-proxy/package.json
+++ b/rpc-proxy/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-rpc-proxy",
-  "version": "0.1.50",
+  "version": "0.1.51",
   "description": "External-facing RPC proxy for tenant-scoped MCP access",
   "main": "dist/main.js",
   "engines": {
--- a/rpc-proxy/src/config.hccBaseUrl.test.ts
+++ b/rpc-proxy/src/config.hccBaseUrl.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from 'vitest'
+import { readFileSync } from 'node:fs'
+import path from 'node:path'
+
+/**
+ * The dev default for RPC_PROXY_HCC_BASE_URL must name the same HCC endpoint the
+ * deployed configmap injects. They drifted once: the default pointed at
+ * host-context-controller.mcp-server (a Service that exists in no overlay, and
+ * one the rpc-proxy egress NetworkPolicy would block anyway — it only permits
+ * the api-gateway peer), while the real target is
+ * host-context-controller-api-gateway.control-plane. NODE_ENV=production is
+ * baked into the image so the default is unreachable in-cluster today, but a
+ * developer running rpc-proxy locally would inherit the wrong endpoint. Pin the
+ * two together so they cannot silently diverge again.
+ */
+describe('rpc-proxy HCC base URL default', () => {
+  const repoRoot = path.join(__dirname, '../..')
+
+  const configSource = readFileSync(path.join(__dirname, 'config.ts'), 'utf8')
+  const defaultMatch = configSource.match(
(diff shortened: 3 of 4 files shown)


zach88:
Took a careful pass through this one. The core fix is solid and I verified it end to end: the new `host-context-controller-api-gateway.control-plane...:8081` default matches the only egress lane the rpc-proxy NetworkPolicy actually permits, the target Service exists, the old `...mcp-server` one exists in no overlay (and egress would've dropped it anyway), and the deployed configmap already carried the right value. The regression test genuinely catches the drift — it passes on this head and fails against the old default. Security came back clean.

**Nothing here blocks the merge.** Everything below is optional / non-blocking test-quality polish — happy to see it land as follow-up rather than hold the PR.

## Optional (non-blocking)

1. **[R1-M1] Test asserts source text, not the produced value** — `rpc-proxy/src/config.hccBaseUrl.test.ts:20-33`. The test regex-parses the *source* of `config.ts` instead of the value `config` actually produces at runtime. If someone rewrites/normalizes the string inside `requiredOrDevDefault` or moves the default into another construct, the runtime value changes but the test stays green because the literal in source didn't. The values themselves do come from real files (not hand-written), so the T1 side is fine — it's the assertion target that's off. Suggestion: export the default as a constant (`HCC_BASE_URL_DEV_DEFAULT`) and assert that against the configmap value, dropping the regex over `config.ts` entirely.

2. **[R1-L1] Unit test reaches two levels up into `deploy/`** — `config.hccBaseUrl.test.ts:14,24-26`. `path.join(__dirname, '../..')` resolves to the repo root and reads `deploy/overlays/minikube/...`. Running rpc-proxy's tests in isolation (image build, partial checkout, extracting the service to its own repo) would `ENOENT` and go red for a reason unrelated to the SUT. Could move the consistency check to a manifest/CI check under `deploy/`, or tolerate the file's absence with an explicit `skip`.

3. **[R1-L2] Regexes are brittle to reformatting** — `config.hccBaseUrl.test.ts:19-21,27`. They assume single quotes + `//` comments in `config.ts` and double-quoted flow scalars in the YAML. A switch to a template literal or a YAML block scalar leaves the matches `null`. It's mitigated by the `parses both sources` test (you'd get a noisy red, not a false green), so this is fine as-is — just flagging.

4. **[R1-L3] PR body slightly oversells the test's guarantee** — the body says the test keeps the default and configmap from silently diverging "again", which is true for *drift between those two strings*. But the test doesn't verify the endpoint resolves to an existing Service or matches the egress peer — i.e. the original bug's *class* ("names a Service that exists nowhere") isn't covered: if both configmap and default were wrong but equal, it'd pass green. Worth trimming the wording, and it's exactly what the M7 follow-up would close. Tracking M7 as its own issue (assert against the Service YAML host+port) sounds right.

## What checked out clean

- New default matches the `allow-hcc-api-egress-rpc-proxy` lane exactly (control-plane, `app: host-context-controller-api-gateway`, port 8081); target Service exists in `deploy/base/control-plane/host-context-controller-api-gateway.yaml`.
- "Latent, not live" confirmed — `NODE_ENV=production` baked into the Dockerfile, so in-cluster the configmap always wins; only a local dev inherits the default.
- No collateral: nothing else references the old default.
- Regression test verified — 3/3 green on this head, and 2/3 fail against the previous default (it really catches the bug).
- Security review clean: aligns traffic with the existing egress lane, adds no lanes, URL is fixed (no SSRF/redirection surface), `http:8081` is unchanged from before, no authn/authz/token/limit touched.

zach88 (approved):

PR #236, opened 2026-08-02
self-hosted Marketplace UX redesign + org area

andypavia:
## feat(registry): self-hosted Marketplace UX redesign + org area

Reworks the Marketplace/registry surfaces around the narrowed distribution
strategy (private, org-scoped plugins; first-party connectors) and adds an org
management area for entries, images, and API keys.

### control-ui

**Marketplace catalog**
- Connectors/Plugins tabs with search + category filters, all capability-gated
  via a new `useRegistryCapability` hook (composes publish-scope + connection
  probe; module-cached and in-flight-deduped to avoid rate-limit storms).
- Public "Publish to Marketplace" surfaces commented out under the distribution
  narrowing (restore when public/org publishing returns).

**Org area (`/marketplace/org`)**
- Sub-tabs: **API Keys**, **Entries**, **Images**.
- **Images** tab lists real org image repos from a new control-api endpoint
  (read-only; loading/ready/unavailable/error states).
- **Entries** (`OwnedEntries`): collapse same-named private plugins into one row
  showing the latest version, expandable to install previous versions
  individually. Per-row **Install** CTA + **Installed** (disabled) state driven
  by the installed-resources catalog.

**Install wizard**
- Injects `spec.agent` for recipes whose `pluginWorkloadSdk.promptBridge`
  requires a resolvable agent (model picker from `allowedModels` +
  preventing the post-install "promptBridge requires a resolvable agent" failure.
- Private-plugin header/chip so an org-scoped install reads as private.
- **Egress review now covers non-transport workloads.** Previously only MCP
  *transport* workloads were inspected, so a plugin's real egress
  plain deployments/cronjobs — e.g. GitHub + Trivy CDN hosts) was invisible.
  The analyzer now reviews any workload that declares `egressBindi
  review renders the concrete hosts/ports.

**Misc**
- `PluginsEmptyState` explaining what a plugin is (recurring jobs, UI, webhooks).
- Spacing fix for the previous-versions list.

### control-api
- `GET /admin/registry/images` — lists the org's image repos via the
  voucher-exchanged per-admin user token (`orgApiKeyClient.listIma
  the registry's `repos` payload). No changes to the registry itself.

### Testing
- control-ui: full suite green (1176 tests), `tsc --noEmit` clean
  in this change.
- control-api: `tsc` clean for the touched files (`registry.ts`,
  `orgApiKeyClient.ts`).
- control-api image built + rolled out to minikube; `/admin/registry/images`
  verified returning real org repos.

### Notes
- Merged latest `origin/dev` into this branch (one conflict in `Ow
  around the Share-modal `opener` refactor, resolved to keep both dev's
  focus-return handling and this branch's version-collapsing).
- Pre-existing on `dev` (verified reproducing on a clean `origin/dev` checkout,
  **not** introduced here): 1 `tsc` error in `LlmDiscoveryPanel.te
  (`source: 'fallback'` vs a `'live' | 'vendored'` type) and 5 failing tests in
  `GfsBrowser.test.tsx` / `LlmModelTable.test.tsx`. Flagged for the dev owners.

### DIFF
--- a/control-ui/app/registry/install/page.tsx
+++ b/control-ui/app/registry/install/page.tsx
@@ -259,12 +259,18 @@ function RegistryRecipeInstallPreview({
     ? (validation?.issues.filter(issue => issue.severity === 'error') ?? [])
     : []
   const egressErrors = egressFindings.filter(finding => finding.severity === 'error')
+  // A promptBridge recipe must resolve an agent before install, or it fails
+  // after install. When the recipe's allowedModels[0] is not in the operator
+  // catalog, providerForModel() returns '' and no spec.agent is injected — so
+  // block the submit until a provider + model are actually chosen.
+  const agentReady = !agentReq.needsAgent || (Boolean(agentProvider) && Boolean(agentModel))
   const canInstall =
     recipeText.length > 0 &&
     !parseResult.error &&
     validationErrors.length === 0 &&
     egressErrors.length === 0 &&
     !egressEditError &&
+    agentReady &&
     !submitting
   const canContinue = step === 0 ? Boolean(recipeText.trim()) && !parseResult.error : canInstall
 
--- a/control-ui/app/registry/install/page.tsx
+++ b/control-ui/app/registry/install/page.tsx
@@ -13,6 +13,7 @@ import { EgressEditor } from '@components/EgressEditor'
 import { RegistryInstallForm } from '@components/RegistryInstallForm'
 import { IconStore } from '@components/Sidebar/icons'
 import { useToast } from '@components/Toast'
+import { SelectInput } from '@components/ui'
 import { CREATE_FLOW_LOADING } from '@constants/createFlowLoading'
 import { CONTROL_ROUTES } from '@constants/routes'
 import { DEFAULT_WORKFLOW_RECIPE_NAMESPACE } from '@constants/workflowRecipes'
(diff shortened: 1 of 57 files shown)


jozer-rami:

control-ui/app/registry/install/page.tsx
```diff
@@ -113,6 +114,87 @@ function applyWorkflowWorkloadEgress(
...
+  }
+  const spec = parsed.spec
+  const sdk = isPlainObject(spec.pluginWorkloadSdk) ? spec.pluginWorkloadSdk : null
+  const promptBridge = sdk && isPlainObject(sdk.promptBridge) ? sdk.promptBridge : null
+  if (!promptBridge) return { needsAgent: false, allowedModels: [] }
+  const agent = isPlainObject(spec.agent) ? spec.agent : null
+  const hasSpecAgent =
+    !!agent && typeof agent.provider === 'string' && typeof agent.model === 'string'
+  const steps =
+    isPlainObject(spec.workflow) && Array.isArray(spec.workflow.steps) ? spec.workflow.steps : []
```
jozer-rami: **Wrong path: `spec.workflow.steps` does not exist in the CRD, so `hasStepAgent` is dead code.**

Steps live at `spec.steps` (`charts/clerum-crds/crds/workflowrecipe.yaml:306`, with the step agent enum at `:549`). There is no `spec.workflow` property anywhere in the schema. The canonical resolver agrees:

```ts
// workflow-recipes/src/workflow/agentResolution.ts:22
const stepAgent = (spec.steps ?? []).find(step => step.agent?.provider && step.agent.model)?.agent
```

That file's own docstring says "Keeping both on this helper prevents the two from drifting apart" — this is a third copy that drifted.

**Trigger:** a recipe with `pluginWorkloadSdk.promptBridge` and `spec.steps[0].agent = { provider: 'claude', model: 'claude-sonnet-4-6' }`, no `spec.agent`. The reconciler accepts it (`pluginWorkloadSdkValidator.ts:85` passes because `hasResolvableAgent` finds the step agent). The wizard computes `needsAgent: true`, shows the false alarm "without one it will fail after install", and on submit writes `spec.agent`. Since `resolveMcpHostAgent` checks `spec.agent` **first**, the injected agent overrides the author's step agent and the plugin runs on a different provider/model than what was published.

**Fix:** read `spec.steps`, or better, import the shared resolver so this cannot drift again.

control-ui/app/registry/install/page.tsx
```diff
@@ -113,6 +114,87 @@ function applyWorkflowWorkloadEgress(
...
+  'xai',
+  'cerebras',
+  'deepinfra',
+  'perplexity',
+  'moonshot',
+  'nebius',
+  'novita',
+] as const
+
+// Best-effort provider guess from a model id; the user can override it.
```
jozer-rami: **The heuristic mis-derives the provider for about half the supported providers.**

Checked against `lib/llm.ts`'s own `LLM_DEFAULT_MODEL_BY_PROVIDER`, i.e. each provider's *own* default model:

| model | real provider | heuristic returns |
|---|---|---|
| `gpt-oss-120b` | cerebras | openai |
| `anthropic/claude-sonnet-latest` | openrouter | claude |
| `llama-3.3-70b-versatile` | groq | openai |
| `kimi-k2.6` | moonshot | openai |
| `sonar-pro` | perplexity | openai |
| `qwen3-coder-plus` | bailian | openai |
| `gemini-2.5-pro` | vertex | gemini |

This is not only a failed install. A wrong `spec.agent.provider` routes the plugin's prompts to a different LLM vendor, on whatever credential that provider holds. The user can override the select, but it arrives pre-filled and confident, and nothing flags the guess.

**There is already an authoritative source in this codebase.** `useLlmAllowedModels()` (`/admin/llm-models`) returns `{ provider, model, enabled }` rows — an exact model-to-provider map, and it also tells you which providers the operator actually enabled. `lib/llm.ts:10-12` states the convention outright: "derives its provider-level UI metadata from it instead of re-declaring the enum." `AGENT_PROVIDERS` just below re-declares it, and the picker offers 19 providers regardless of which have credentials configured.

Same spot: the new test fixture uses `['gpt-4o-mini', 'claude-3-5-sonnet']`, the one pair this gets right, so replacing the whole function body with `return 'openai'` keeps `RegistryInstallPage.test.tsx` green. I verified that.

control-ui/next.config.js
```diff
@@ -88,6 +88,22 @@ const nextConfig = {
...
+        permanent: true,
+      },
+      {
+        source: '/publisher/shared-with-me',
+        destination: '/marketplace/org/entries',
+        permanent: true,
+      },
+      { source: '/publisher/entries', destination: '/marketplace/org/entries', permanent: true },
+      { source: '/publisher', destination: '/marketplace/org/entries', permanent: true },
+      { source: '/publisher/:path*', destination: '/marketplace/org/entries', permanent: true },
```
jozer-rami, line 106: **These redirects orphan the Docker push credentials, and the new Images tab points users straight at the gap.**

- `DockerCredentialsPanel` is rendered **only** by `components/PublisherView/index.tsx:85`, reachable only through `app/publisher/**`, which these lines now redirect away wholesale.
- `/publisher/credentials` lands on `/marketplace/org/credentials` → `RegistryApiKeysPanel`, which has no `resolveDockerCredential`, no `dockerconfigjson`, and no `docker login` command. It is a plain key CRUD table.
- `components/MarketplaceOrgImages/index.tsx:73-77` tells the user "Authenticate with a push credential from the **API Keys** tab" and links there. That tab no longer offers one.

The design doc flags this as a phase-ordering problem rather than an oversight: `docs/features/self-hosted-registry-ux-redesign.md:165` marks the org Credentials sub-view "(unified in Phase 5)", and §5.1 at `:273-288` is where `DockerCredentials.tsx` gets absorbed into `RegistryCredentials/**`. The redirect shipped a phase early.

Meanwhile this PR spends +35/-2 improving that now-unreachable panel and +19 lines testing it, duplicating the mode-specific auth-disabled copy that `RegistryApiKeysPanel.tsx:162-176` already has on the reachable path.

Related stale comment: `MarketplaceOrgArea/index.tsx:84-85` says "DockerCredentialsPanel self-gates via the keys API, so it renders regardless" — it is not rendered there at all.

**Options:** hold the `/publisher/credentials` redirect until Phase 5, or render `DockerCredentialsPanel` alongside `RegistryApiKeysPanel` on the org Credentials tab in the interim.

jozer-rami (commented):
@andypavia deep review pass (second look, scoped to avoid re-covering the earlier @claude review). Three issues worth fixing before merge, left inline below.

1. `recipeAgentRequirement` reads `spec.workflow.steps`, which does not exist in the CRD. The step-agent guard never fires, so the wizard can inject a `spec.agent` that overrides an author's declared step agent.
2. `providerForModel` mis-derives the provider for roughly half the supported providers, including their own default models. The plugin's prompts then go to a different LLM vendor than the author intended.
3. Docker push credentials are unreachable after the `/publisher/*` redirects, while the new Images tab tells users to go get them from the API Keys tab.

Verification notes, since the earlier review could not run anything: `tsc --noEmit` gives exactly 1 error (`LlmDiscoveryPanel.test.tsx:149`) and the suite is 1186 passed / 5 failed across `GfsBrowser` + `LlmModelTable`. All six are in untouched files, so your "pre-existing on dev" note checks out. The body's "full suite green (1176 tests)" line is just stale.

Also worth knowing: I mutation-tested two of the new test groups. Replacing `providerForModel`'s body with `return 'openai'` keeps `RegistryInstallPage.test.tsx` green (the fixture uses `gpt-4o-mini` / `claude-3-5-sonnet`, the one pair the heuristic gets right), and deleting the module cache plus the in-flight dedup entirely keeps `useRegistryCapability.test.tsx` green. The rate-limit fix this PR is built around has no coverage.

The rest of the findings (org area locking into "unclaimed" on a transient probe blip, the hardcoded `registry.evenfire.ai` in the Images coordinates, the Install Plugin CTA landing on API Keys) are in the full write-up. Happy to paste those too if useful.

andypavia:

andypavia:

andypavia:
@jozer-rami I think this is ready to be reviewed again.

jozer-rami (approved):

Issue #223, opened 2026-07-31
Add credential rotation for existing MCP connectors


miguelo-999:
## Problem

  The connector UI can create a Kubernetes Secret while creating a new connector, but it cannot update credentials after the connector already exists.

  Today, connector-scoped secret support is limited to:

  - `POST /api/v1/admin/mcp-secrets` — create a Secret
  - `DELETE /api/v1/admin/mcp-secrets/:name` — delete a Secret

  There is no `PUT`/update endpoint for MCP secrets, and the connector edit screen has no credential-management control.

  This forces operators to rotate API keys manually with `kubectl`, then restart the connector deployment so its environment variables reload. That is inconvenient for normal operations and creates avoidable
  risk during credential rotation.

  ## Requirement

  Add safe credential rotation for existing MCP connectors through the Control UI and Control API.

  ### Backend

  Add an MCP-secret update endpoint:

  
  PUT /api/v1/admin/mcp-secrets/:name

  Requirements:

  - Always target config.mcpServersNamespace; ignore any caller-provided namespace.
  - Validate the Secret name and payload using the same guardrails as connector-secret creation.
  - Use merge semantics by default so updating one key does not remove unrelated keys from the same Secret.
  - Return the Secret name and namespace, but never return plaintext secret values.
  - Reuse the existing gateway.updateSecret() / merge-secret infrastructure where possible.

  ### UI

  Add an Update credentials action to the existing connector edit/details experience.

  Requirements:

  - Show the referenced Secret name and configured key-to-environment-variable mappings.
  - Never display stored secret values.
  - Allow an operator to enter replacement values only for keys they intend to rotate.
  - Mask all credential inputs and disable browser autocomplete.
  - Clearly state that saving rotates the selected credentials and restarts the connector.
  - Show success/failure feedback and the resulting deployment rollout status.

  ### Runtime behavior

  After a successful Secret update:

  1. Restart only the affected connector Deployment.
  2. Wait for rollout completion.
  3. Surface an actionable error if the rollout fails.
  4. Do not alter the connector CRD, Context allowlist, or unrelated Secrets.

  ## Why we need this

  Most remote MCP connectors authenticate with API keys, personal access tokens, or service-account tokens. Those credentials need rotation when they expire, are revoked, or are replaced for security reasons.

  Without this feature, even non-technical operators must use cluster tooling to update credentials. That makes the connector UI incomplete: it supports creating credentials but not maintaining them.

[...]

PR #219, opened 2026-07-30
Feat/ux provements 26

miguelo-999:

PR #137, opened 2026-07-22
Desktop App Deep Links

josue-tech-94:
## Summary

Adds reliable, shareable deep links for opening WorkflowRecipe apps directly in the Evenfire desktop app.

Key changes:

- Registers `evenfire://` as a desktop protocol in packaged Electron applications.
- Adds shareable Profile UI links using:
  `/open/apps/{recipeNs}/{recipeName}?path={path}&team={teamId}`.
- Serves the Profile UI handoff as an unauthenticated, no-cache redirect to the desktop app.
- Restores the styled Evenfire handoff page with a fallback “Open in Evenfire” button.
- Validates recipe names, namespaces, paths, and team IDs before generating desktop links.
- Handles deep links when the desktop app is:
  - Already open.
  - Running without a window on macOS.
  - Starting from a fully closed state.
  - Receiving simultaneous startup or second-instance events.
- Queues links until the renderer and authenticated session are ready.
- Recreates and focuses the desktop window when necessary without repeating session startup work.
- Opens the requested app, team, and nested route after the handoff.
- Preserves app conversation origin when navigating through the app picker and embedded UI.
- Makes dependency health checks non-blocking so they do not delay deep-link processing.
- Requires explicit post-login confirmation for app links received before auth.
- Acknowledges app links only after the native Sandbox UI mount succeeds.
- Retries starting/not-ready app handoffs instead of dropping them.
- Honors valid explicit `PROFILE_UI_BASE_URL` values for self-hosted deployments.

The resulting flow is:

`Profile UI link` → `evenfire://app/...` → `Desktop app` → `Requested app and route`

## Testing
- [x] Focused desktop deep-link/session/window tests passed: 83 tests.
- [x] `packages/desktop-app-links` tests passed: 6 tests.
- [x] `packages/desktop-app-links` build passed.
- [x] Desktop main build passed.
- [x] Desktop renderer build passed.
- [x] Profile UI tests passed: 39 tests.
- [x] Prettier check and `git diff --check` passed.
- [x] GitHub CI

## Checklist

- [x] This is not a new third-party MCP server / WorkflowRecipe (those go to the registry)

PR #242, opened 2026-08-03
promote: dev → main (registry pull-secret self-provisioning, connector credential rotation, mcp-host hardening, Marketplace UX, desktop deep links, HCC rollout budget)

jozer-rami:
Promotion of `dev` → `main` for the prod deploy. **204 commits, 416 files, +40638/−4353.**

## What ships (12 PRs)

**Registry pull-secret self-provisioning (#243)**
- control-api now **self-provisions** the `evenfire-registry-pull` dockerconfigjson Secret. Previously it only ever *referenced* the Secret by name and nothing in this repo created it, so on self-hosted clusters every private-image plugin died with `ImagePullBackOff: unauthorized`.
- `registryPullSecretService.ensureRegistryPullSecret()` is a read-before-mint state machine gated to `self-hosted` mode + active registry auth + a resolved org. It coexists with an externally-provisioned Secret via a `clerum.io/managed-by=control-api` ownership label (unlabeled ⇒ foreign ⇒ left untouched), repairs an empty Secret, and tolerates create races.
- The Secret lands in `targetNs` (`imagePullSecrets` are namespace-local) and its `.dockerconfigjson` is keyed on the **image host**, not the registry's token issuer.
- **Uses no new Kubernetes RBAC** — it rides the `secrets` grant in `deploy/base/mcp-server/rbac.yaml` that #226 extends with `patch` in this same promotion.
- Base `workflow-recipes` Deployment now projects `CLERUM_REGISTRY_URL` from `control-api-config` via `configMapKeyRef` (`optional: true`) so WRC and control-api cannot disagree about which images are ours.

**MCP connector credential rotation (#223 → #226)**
- New `PUT /admin/mcp-secrets/:name` rotation path using `gateway.mergeSecret()` (RFC 7396 merge-patch) so rotating one key can't wipe sibling credentials on the same Secret and concurrent rotations can't race.
- Control UI: `UpdateConnectorCredentials` flow, success banner scoped to the verified connector only, no longer latches failure on a transient `RolloutIncomplete`.
- Human-paced Playwright e2e for the full rotation journey.

**mcp-host / rpc-proxy hardening (#232, #233 — splits of #205)**
- mcp-host: authoritative fleet lifecycle (epoch capture, superseded-candidate rejection, synchronous revoke), MCP client/manager transport-close and timeout hardening, and `contextMapperClient` no longer converts an HTTP failure into an authoritative empty fleet. `authoritative?: boolean` is optional and skew-safe with HCC.
- rpc-proxy: the `RPC_PROXY_HCC_BASE_URL` dev default pointed at a Service that exists in no overlay (and that the egress NetworkPolicy would block anyway). Repointed at `host-context-controller-api-gateway.control-plane`, with a test pinning it to the deployed configmap. Latent, not live — `NODE_ENV=production` is baked in, so in-cluster the configmap always supplied the value.
- Both slices are byte-identical to the head reviewed in #205.

**Marketplace / registry UX (#219, #236)**
- Unified model catalog + discovery, refined management surfaces, provider-filter styling and stale-filter reset. API keys moved under the Marketplace menu.
- Self-hosted Marketplace redesign: Connectors/Plugins tabs with search + category filters, capability-gated through a new `useRegistryCapability` hook (module-cached and in-flight-deduped to avoid rate-limit storms), plus an org area at `/marketplace/org` with API Keys / Entries / Images sub-tabs.

**Desktop deep links (#137)**
- Registers `evenfire://` as a desktop protocol in packaged Electron builds; shareable Profile UI links at `/open/apps/{recipeNs}/{recipeName}`, served as an unauthenticated no-cache redirect.
- Handles the app already open, running windowless on macOS, cold-starting, and simultaneous startup/second-instance events. New `packages/desktop-app-links`.

**Desktop / UI performance (#253)**
- Keeps authenticated UI shells mounted and stabilizes registry scope (117 files).

**Deploy: HCC rollout budget (#271)**
- The clerum-dev deploy was failing at *Provision GFS runtime* with `host-context-controller exceeded its progress deadline`, which skipped the per-service pin bump and left `gcp-dev` pinned behind its own cluster.
- HCC serves 503 on `/ready` until initial fleet reconciliation completes, so time-to-Ready scales with the fleet. What made that fatal is that the wait had no real budget: `kubectl rollout status` aborts the instant the controller sets `ProgressDeadlineExceeded`, so its `--timeout` can never outlast `progressDeadlineSeconds` — and that field was never declared, so it was the Kubernetes default of 600s. The clock also starts earlier than the wait, at `apply-inter-service-tokens.sh`'s `rollout restart` several steps before.
- Declares `progressDeadlineSeconds: 1200` on the HCC Deployment and raises the wait to 900s, with a test pinning the relationship so it cannot regress to a timeout that does nothing.

**Release versioning spine (#270)**
- `desktop-app/package.json` was serving two incompatible roles: a per-commit counter auto-bumped by the pre-commit hook, *and* the release/compatibility version that `releaseManifest.ts`, the update gate, and every artifact filename read. Three consequences were live in the tree: the manifest had drifted (`0.1.252/0.1.56/0.1.36` against `0.1.333/0.1.60/0.1.51`), the update prompt linked to a 404 (tag shape `desktop-app-${version}` has never existed, against a `your-org` placeholder base URL), and neither validator was wired into CI, which is why the drift went unnoticed.
- Retires the per-commit counter for `desktop-app` only (`external-rest-api` and `rpc-proxy` keep theirs), decouples `minimumDesktopVersion` from `desktopVersion`, and implements `--minimum-desktop-version`, which the design assumed existed but never did.
- Adds the missing CI validation: release-manifest/package-version agreement and service version bumps.
- Also fixes image-change detection so `packages/desktop-app-links/**` rebuilds `profile-ui` (the package #137 introduced).

**Self-host onboarding (#258, #218)**
- `make prereqs` (alias `make doctor`): checks Docker (daemon reachable, ≥10 GB RAM / ≥6 CPUs), minikube ≥1.30, kubectl, node ≥24, python3, ruby, git, make **before** any cluster work. `make minikube-setup` runs it first (`SKIP_PREREQS=true` bypasses).
- All **21** LLM providers now reach the minikube secret; previously only openai/claude/zai/bailian were wired, so setting e.g. `GROQ_API_KEY` silently did nothing.
- New `docs/get-started/registry-quickstart.md` covering fresh cluster → installed → publishing, and fixes the stale `/registry/connect` → `/marketplace/connect` path.

**Hardening / CI**
- e2e runs `kubectl` via `execFileSync`, keeping context/namespace off the shell (CodeQL #827/#828).
- Vitest e2e gate now fails on zero executed tests.
- Secret key-name validation: reject `.`, `..`, and `..`-prefixed data keys at the `SecretService` chokepoint (full `IsConfigMapKey` parity); invalid names return `400`, not an opaque `500`.

Plus back-merges **#224** and **#267** (the latter cleared this PR from `mergeStateStatus=BEHIND` after #172 was merged straight into `main`).

## Ops notes for this promotion

- **No CRD changes.** `charts/clerum-crds/**` is untouched — no out-of-band `gcp-prod-deploy-crds`.
- **No DB migrations.** Nothing under `*migrations*` changed.
- **No new environment variables.** No new `process.env` reads in control-api / workflow-recipes / mcp-host / rpc-proxy / host-context-controller. The prod overlay is already prepared: infra PR #28 added the byte-identical `CLERUM_REGISTRY_URL` `configMapKeyRef` projection to `gcp-prod/patches/registry-env.yaml`. **Shape is load-bearing** — a literal `value:` there would strategic-merge with base's `valueFrom` into one EnvVar the API server rejects mid-apply.
- **RBAC changed → `gcp-prod-post-deploy-sync` is required after the apply.** `deploy/base/mcp-server/rbac.yaml` adds the `patch` verb on `secrets`. The prod deploy SA lacks `container.roles.update`, so CD strips RBAC kinds. Without the sync, rotation 403s **and** #243's pull-secret path breaks. Run it against an assembled tree (`bootstrap-rbac.sh` lives in this repo and scans `deploy/base/**/rbac.yaml`).
- **Pull-secret prerequisite is now satisfied on prod.** prod runs `REGISTRY_CONNECTION_MODE: managed`, where #243 writes nothing but *verifies* the operator's Secret and returns `operator_secret_missing` (409) if a required namespace lacks a usable one. `evenfire-registry-pull` was absent from every prod namespace; it has since been provisioned in `mcp-server`, `sandbox-recipes`, and `sandbox-ui`, and probes **live** against `registry.evenfire.ai`. The registry-side dependency (`evenfire-registry` #64, plus #65 / migration 021) was deployed to production 2026-08-04.
- **`progressDeadlineSeconds: 1200` also applies to prod's HCC** (#271). Effectively inert there — prod runs 7 McpServers to dev's 108 — and a longer deadline only ever delays a failure verdict.
- **Image pins: 14 of 15 move** (`workspace-files-controller` already matches at `sha-094c86f`). Prod promotes dev's already-mirrored digests — no rebuild.
- **`gcp-dev` pins now match the live dev cluster** (`sha-449d2ff` for control-api / control-ui / profile-ui, committed as infra `99ff29c`), so `promote-dev-tags.sh` will copy an accurate set.

## CI

All required checks green: 18/18 `Build / test / typecheck`, `Repo hygiene`, `Analyze (javascript-typescript)`, and every `Build & push`.

#270's four new CI validations (release coordinates, precommit manifest exemption, manifest/package agreement, service version bumps) are **steps inside existing jobs, not new jobs**, so they add no new status-check contexts and `main`'s ruleset needs no update.

The red `CodeQL` check is **not a required context** and reports no new risk. Open code-scanning alerts are identical on both refs — `main` and `dev` each carry 1 critical / 497 high / 129 medium / 130 unrated. The "new alerts" figure is main's existing backlog re-attributed against the merge ref.

## Merge instructions

**Merge commit — never squash.** Squashing diverges `dev` from `main` permanently and forces recurring back-merges.


jozer-rami:
### Infra note for this promotion — platform image-pin enforcement

Heads-up for whoever runs the prod deploy off this promotion; a recent infra-side change affects it (detailed operator steps live in the infra repo, not here):

- **The prod deploy now validates that first-party platform images are sha-pinned.** It's already green against the current prod overlay, so this promotion's deploy should pass — but if it flags an image on a non-sha tag (e.g. `:latest`), that's a real catch to fix before applying, not a flake.
- **`workspace-files-controller` finally rolls off a frozen `:latest`.** Its prod image had been stuck on a ~5-week-old build (a gap between where `:latest` is published vs what the cluster pulls, after the Phase-2 carve); its prod pin is now a real sha, so expect the WFC pods to change image on apply.
- **Two runtime items need separate manual attention during the prod window** (they are not carried by the config apply): the `airtable` MCP server is on `:latest` and should be re-pinned to a real sha, and a dead pre-carve `profile-api` leftover should be removed. Both were already handled on dev.

No CRD or secret changes come from this infra change.

PR #252, opened 2026-08-03
fix(control-api): constrain Secret type and annotation keys on write (No Merged Requiere Update CRD)

alfredolopez80:
## Scope

This PR hardens Secret metadata handling and Registry mutation safety across Control API and the required Control UI contracts:

- centralized Secret type and annotation constraints across create, update, merge, and delete;
- safe preservation of existing metadata without granting internal capabilities;
- Registry install, upgrade, rollback, and uninstall mutations bound to live UID and resourceVersion;
- fail-closed Secret reference discovery across all nine repository CRD schemas;
- PostgreSQL-backed, one-time MCP Secret rollback permits with hashed JTI storage, bounded TTL, claim lease, release, finalize, and stale-claim reissue;
- removal of the client-side rollback-proof cookie: no rollback bearer or proof value is stored in a browser cookie;
- Control UI identity/CAS handling, error recovery, recorder/E2E adaptations, runtime database ACLs, contracts, fault injection, and operational documentation.

The branch-owned Minikube harness, profile ownership, port-forward, image/setup synchronization, and T0/T1/T2 tooling were extracted to [PR #443](https://github.com/evenfire-ai/evenfire/pull/443). They are not part of this PR's diff.

## Current base and verifiable scope

- base: `origin/dev` at `ff8bb922f9abd4df5737c1a23c72f7fbc4306b58`;
- head: `fadd179ca006e4c2dd4d59508b1e3871e0d885b8`;
- merge-base: `ff8bb922f9abd4df5737c1a23c72f7fbc4306b58`;
- the PR branch contains the latest `origin/dev`;
- diff: 78 files, 11,608 insertions, 917 deletions;
- affected surfaces: Control API, Control UI, nine CRD schemas, runtime database access profile, shell contracts, E2E recorder flows, and the rollback-permit runbook;
- worktree and remote branch are synchronized and clean.

## Historical findings revalidated

All prior Claude, Jozer, Zach, CodeQL, and independent-report findings were re-triaged against the current head. Every surviving in-scope blocker/high finding was fixed and covered by tests.

Key closures include:

- rollback restoration and production/MockGateway parity;
- compatibility with preserved or custom Secret types;
- fail-closed CRD Secret-reference coverage;
- UID/resourceVersion fencing and ambiguous readback behavior;
- compensation and repair behavior across Registry sagas;
- UI identity/error handling and cleanup;
- server-side rollback permit lifecycle and cancellation paths;
- removal of clear-text rollback proofs from cookies;
- runtime PostgreSQL role/ACL provisioning;
- CodeQL alert #1198 fixed in code without dismissal.

Two independent GPT-5.6 SOL MAX adversarial reviews challenged the final implementation through three rounds. Both returned `CODE PASS`, with P0=0 and P1=0. The remaining cross-resource reference-scan/delete TOCTOU is pre-existing, reduced by this PR, and belongs to a separate admission/finalizer design.

### GitHub

- CI: [34/34 checks successful](https://github.com/evenfire-ai/evenfire/actions/runs/33219738663);
- CodeQL: [successful](https://github.com/evenfire-ai/evenfire/actions/runs/33219738628);
- CodeQL alert #1198: latest PR instance is `fixed`, with no dismissal;
- PR state: OPEN, merge state `CLEAN`.

### Canonical local T0/T1/T2 certification

Exact-head certification:

- `T0=PASS`;
- Control API: 374 files passed, 4,847 tests passed, 108 skipped;
- Host Context Controller: 86 files and 1,877 tests passed;
- Control UI: 167 files, 1,778 tests passed, 1 skipped;
- Desktop: 195 files and 1,953 tests passed;
- `T1=PASS`: 23/23 Real PostgreSQL files, 113/113 tests, zero skipped;
- `T2=PASS`: exact marker, images, PostgreSQL, namespaces, Services, and deployments ready;
- `NP08_HCC_AUTHORIZATION=PASS`;
- all 25 required local images verified;
- all core workloads Ready.

Certification identity:

- run ID: `9dad2b5e46c5465ebc9c9d0c2ea352a7`;
- profile/context: `clerum-test-pr-252-t2-certification-20260828-b60f6d0c`;
- head: `fadd179ca006e4c2dd4d59508b1e3871e0d885b8`;
- origin/dev: `ff8bb922f9abd4df5737c1a23c72f7fbc4306b58`.

`Health=NOT_RUN` and `Playwright=NOT_RUN`: no profile-owned user-facing health or applicable Control UI/Desktop Playwright command was supplied to this full-bootstrap certification. No browser-runtime PASS is claimed.

## Deployment requirement

Deploy Control API before Control UI. A rollback must roll back both together so clients and API retain the same Secret identity/CAS contract.

## Closing status

The branch is synchronized with the latest `origin/dev`, pushed, and green in CI, CodeQL, T0, T1, T2, and NP08.

Automatic or manual merge is not authorized by this update. The next step is the requested complete closing re-review on this exact head.


alfredolopez80:
@claude Please perform a **final validation review** of this PR before merge.

This PR (`fix/secret-type-annotation-constraints`, base: `dev`) has been through two review rounds and an adversarial review with Fable 5. All findings have been addressed. Here is the complete commit history:

**Commits in scope (oldest → newest):**
1. `ed4a8815` fix(control-api): constrain Secret type and annotation keys on write
2. `a4c150e7` fix(control-api): harden annotation denylist, rollback sanitization, and error extraction
3. `f5c745a7` fix(control-api): address review findings (import type, catch logging)
4. `9b286839` fix(control-api): eliminate redundant computation, align extractHttpStatus, add edge-case test
5. `e0a8bdb3` fix(control-api): make recipeExists fail-closed instead of fail-open
6. `f0db54c5` test(e2e): migrate integration tests to cookie-based admin auth
7. `1f08272c` fix(control-api): resolve 6 review findings (F1-F6) from PR #252
8. `1c4b61e9` fix(e2e): replace execSync with execFileSync to prevent command injection (CWE-78)
9. `4796dacc` fix(control-api): clear stale annotations on rollback, deduplicate tests
10. `29be8e42` fix(e2e): remove literal quotes from jsonpath args for execFileSync
11. `9abd2b89` fix(e2e): remove quoted jsonpath in workflow-lifecycle (missed in 29be8e42)
12. `bedc8cce` fix(e2e): replace silent-skip guards with vitest ctx.skip() (F8)
13. `d533aee1` fix(e2e): complete F8 migration in channel-reader-mcp-host, add kubectl() docs

**Key areas to validate:**
1. **Secret type/annotation constraints** (F1-F6): `assertValidSecretConstraints` on every write path (`createSecret`, `updateSecret`, `mergeSecret`), no bypass via `allowPlatformAnnotations`
2. **Rollback annotation cleanup**: `stripBlockedAnnotationKeys(...) ?? {}` in `registry.ts` — no stale annotations survive rollback
3. **CWE-78 mitigation**: all `execSync` → `execFileSync` in 3 e2e files (no shell invocation)
4. **F8 complete**: all 3 e2e test files now use `skipEachIfClusterDown()` + `ctx.skip()` instead of `if (!flag) return` — including `channel-reader-mcp-host.test.ts` (fixed in `d533aee1`)
5. **recipeExists fail-closed**: errors propagate, no silent swallow
6. **No unused variables**: `initialDeployUid` removed (was TS6133)

**What was adversarially verified by Fable 5:**
- F8 gap in `channel-reader-mcp-host.test.ts` → CONFIRMED and fixed (7 guards removed)
- `kubectl()` whitespace split risk → REFUTED (28 call sites checked, none use spaces; WARNING doc added)
- All 5 "clean" items from prior review → independently verified as correct

Please confirm this PR is merge-ready or flag any remaining issues.

zach88:
Took a deep pass through this one across the constraints core, the admin routes, the e2e migration, and a dedicated security sweep. Good news up front: nothing blocks the merge. The security side came back **clean and verified** — the allowlist + annotation denylist hold, and the one bypass worth worrying about (sneaking reserved annotations in through `mergeSecret`) got ruled out with evidence: neither HTTP caller passes annotations and the merge body never writes `req.annotations`. Everything below is Medium/Low: a defense-in-depth gap, a couple of test-quality issues, two stale claims in the body, and some scope creep on the e2e side.

## Medium

1. **[R1-M1] `mergeSecret` doesn't validate annotations and `_opts` is dead code** — `control-api/src/services/secretService.ts:116`. It validates the type (`assertValidSecretType`) but skips `assertValidSecretAnnotations`, and the options param comes in as `_opts` unused. Not exploitable today (traced both callers: `communicationChannelCredentials.ts:72` and `secrets.ts:451` don't pass annotations, and the merge-patch body never writes them), but this is exactly the class of hole this PR closes — the day someone adds `annotations` to the merge body without the guard, the vector reopens. Cheap fix: call `assertValidSecretConstraints(req, opts)` like create/update do (rename `_opts`→`opts`), or leave an explicit comment on why merge intentionally never touches annotations.

2. **[R1-M2] The recipe-secrets ownership test asserts against a fabricated error handler (T1/T4)** — `control-api/test/routes.recipeSecrets.ownership.test.ts:44-48,104`. The test mounts a local handler that maps `err.message → res.body.error`, but the real handler (`app.ts:328-348`) collapses a `.status`-less `Error` into `{ error: 'Internal Server Error' }` (and drops the message entirely in prod). So `expect(res.body.error).toContain('apiserver unavailable')` only passes against the stub — the app never emits that body. The other two assertions (`.expect(500)` + `createSecret` not called) do catch the regression and fail against the merge-base, so the test isn't worthless — just drop the content assertion or wire up the real handler.

3. **[R1-M3] The "sanitize annotation snapshots in rollback" invariant has no test on the real path (T5)** — `stripBlockedAnnotationKeys` is well covered as a unit, but the rollback tests (`routes.registryInstall.test.ts:824,868`) exercise the *install* path (Secret deleted), not the *upgrade* path where the snapshot flows through `updateSecret(..., stripBlockedAnnotationKeys(...) ?? {})` in `routes/admin/registry.ts:~1912`. Nothing verifies a snapshot carrying `kubectl.kubernetes.io/...` gets restored already stripped.

4. **[R1-M4] Lost coverage of the `rotated:false` branch** — `tests/e2e/integration/channel-reader-via-api.test.ts:268-292`. Adding inline `credentials` to the POST means the channel is born with a `credentialsSecretRef`, so the first `PUT /credentials` returns `rotated:true`. The retitle + assert flip to `rotated:true` is correct, but now nothing in the suite exercises the creation branch (`communicationChannelCredentials.ts:88-111`). A regression in the create-on-PUT path would slip through. Worth adding a `rotated:false` case back. (Couldn't run e2e here — no cluster.)

5. **[R1-M5] The "returns 400 for invalid JSON" test got hardened in a fragile way** — `tests/e2e/integration/channel-reader-mcp-host.test.ts:120-129`. It used to send a valid token; now it drops the auth header and still expects 400 straight against mcp-host. That only holds if mcp-host parses the body (→400) *before* the runtime-auth guard (→401). If auth runs first, this now returns 401 and fails. Worth confirming the middleware order in mcp-host; if auth precedes parsing, keep a valid token or assert `[400,401]`. (Not verifiable without a cluster.)

6. **[R1-M6] Scope creep on the e2e side (D1)** — the PR is about constraining the Secret write-path in control-api, but the e2e changes drag in a broad refactor across two services and several unrelated concerns: cookie-auth migration (`channel-reader-via-api`, `workflow-lifecycle`), rpc-proxy rerouting (`channel-reader-mcp-host`, nothing to do with Secrets), `execFileSync`/CWE-78 hardening, and the F8 skip migration + new helpers. Only the `channel-reader-via-api` credential bits are actually coupled to the write-path. Mitigating factor: those suites were silently broken (fail-open with the cluster down) and do exercise the hardened path, so fixing them alongside is defensible — your call whether to split the e2e hygiene into its own PR or keep it here.

## Smaller stuff

- **[R1-L1] Body claims two allowed types, code allows three (R4)** — the summary says "only `Opaque` and `kubernetes.io/dockerconfigjson` are accepted", but `secretConstraints.ts:3-7` also includes `kubernetes.io/tls` (exercised at `services.secretService.test.ts:259`). All three checked out as safe and strictly more restrictive than the base, so just sync the body (or confirm `tls` is intended).
- **[R1-L2] Body claims the edge-case returns `undefined`, but it asserts `{}` (R4)** — body says "test for `stripBlockedAnnotationKeys({})` returning `undefined`", but the code returns `{}` and the test asserts `{}` (`services.secretService.test.ts:533-535`). Body contradicts the test.
- **[R1-L3] `?? {}` in the rollback flips the "no annotations" case from preserve→wipe** — `routes/admin/registry.ts:~1922`. When `snapshot.annotations` is `undefined`, the base passed `undefined` → `updateSecret:86` (`req.annotations || existing...`) fell through and preserved live annotations; now `stripBlockedAnnotationKeys(undefined) ?? {}` = `{}` (truthy) wipes them. Low impact (registry secrets always carry `clerum.io/`, which is preserved), but it's a behavior change — worth confirming it's intended.
- **[R1-L4] `console.error(..., err)` logs the full error object** — `routes/admin/registry.ts:2183,2206,2294,2321`. No leak today (verified: fetch/undici errors and the hand-built `Error`s carry status + OAuth body, never the secret or Bearer), but it's latent and it bypasses pino's `redact.ts` by hitting stderr raw. Prefer `err instanceof Error ? err.message : String(err)`.
- **[R1-L5] `requireControlApiUp` and `requireServiceUp` are near-identical** — `tests/e2e/integration/mcpCredentialRotation.helpers.ts:272-309`; the former could delegate to the latter.
- **[R1-L6] The POST test doesn't assert the Secret was created from inline `credentials` (T4)** — `channel-reader-via-api.test.ts:227`; only covered indirectly via the later `rotated:true`.

## What checked out clean

- **Security: SAFE, zero findings.** `mergeSecret` bypass ruled out with caller evidence; denylist is complete (platform keys live under flat `clerum.io/`, Helm covered by `meta.helm.sh/`, and the apiserver forces lowercase DNS-subdomain keys so no case/whitespace evasion); the type allowlist is strictly more restrictive than the base (which allowed the dangerous `kubernetes.io/service-account-token`); fail-closed introduces no new DoS surface; rollback snapshot is read live from the cluster, not from anything attacker-persistable; the new logs don't leak credentials.
- **`recipeExists` fail-open → fail-closed** verified: the caller (`secrets.ts:711`) is inside `asyncHandler`, so the thrown error propagates to the global handler → a controlled 500, not a crash. Intended tradeoff. Optional nicety: map the `listResource` failure to 503 so clients can tell transient from permanent.
- **control-api suites all green:** `services.secretService` 145/145, `routes.recipeSecrets.ownership` 11/11, `routes.registryInstall` bundled 164/164 (run in a real worktree, symlinked node_modules, cleaned up after).
- **T3 satisfied by construction:** `secretConstraints.ts` doesn't exist at the merge-base, so the type/annotation tests would fail there; the fail-closed test asserts observable output (500 + `createSecret` not called) and fails against the base (which returned `true` → 201).
- **`extractHttpStatus`/`extractK8sError` with `.status`** are guarded by `typeof === 'number'` + the 400–600 range, so a `.status` that's a k8s resource object falls out — no false positives.
- **package.json:** no dependency changed, just the `version` field (release bump) and a coherent lock.

alfredolopez80:
@claude Please perform a final closing review of PR #252 for merge readiness.

Review target and synchronization

- Remote PR head is fae0634b3b6e91e3598aef492c8f79b74328af24.
- Base is dev at 3b9ff0f42dfbca5708f9d53d9c1f799748556d1d.
- The candidate was synchronized with origin/dev; local merge-base equals the base and the pushed worktree is clean.
- Please review the exact pushed head, not the historical PR snapshot.

What was reviewed and corrected

1. Secret annotation policy and public mutation boundaries. Platform annotations remain explicit opt-in only. Public Secret writes stay strict, unknown metadata and infrastructure annotations remain rejected, and the Registry-managed credential path preserves only the exact catalog-id/catalog-version pair when both trusted Registry annotations already exist. The Registry rotation fix is in b219318d.
2. Install-hook and guardrails blast radius. The opt-in writer contract, upgrade path, pull-secret path, CRD/RBAC behavior, HCC consumption, and unrelated credential writes were traced. No new production conflict with the Andy Pavia install-hook/guardrails feature was found.
3. Minikube T0/T1/T2 harness. Same-run full image baselines now prevent duplicate image acquisition, incremental sync falls back conservatively, T1 requires complete green JSON plus Vitest exit zero, GFS Real PostgreSQL files run in isolated processes, and NP-08 waits for asynchronous runtime-token convergence.
4. NP-08 fresh bootstrap. A genuinely fresh bootstrap whose Secret revision already matches the Ready updated Host Deployment is accepted without a redundant restart. Long-lived Hosts still use the restart plus Secret/Deployment revision convergence contract.
5. Profile and runtime safety. The previous unhealthy profile was not bypassed; its fail-closed GFS writer guard remained active. A dedicated branch-owned profile supplied the final runtime evidence.

Validation evidence

- Local T0: ShellCheck PASS, E2E Guardian static audit PASS for 12 files, control-api 4620 passed / 103 skipped, mcp-host 3531 passed / 1 todo, affected builds and fixture build PASS.
- Local T1: control-api 20 files / 100 tests plus gfs-controller 2 files / 8 tests; aggregate 108/108, zero pending, Vitest exit zero.
- Local T2: MINIKUBE_T2_PASS, T0=PASS, T1=PASS, T2=PASS, NP08_HCC_AUTHORIZATION=PASS. Positive and negative caller-binding journeys passed.
- Electron runtime verification passed with Node 24.18.0 and the installed Electron executable.
- Remote CI/CD: 33/33 checks SUCCESS, including control-api, mcp-host, HCC, GFS, Real PostgreSQL, Registry contract, desktop, manifests, leak guards, shell syntax, and CodeQL.
- CI run: https://github.com/evenfire-ai/evenfire/actions/runs/32655613301
- CodeQL run: https://github.com/evenfire-ai/evenfire/actions/runs/32655613319

Known limits and requested closing checks

- Health and Playwright were not run in local T2 because no opt-in commands were configured; they are separate evidence lanes.
- A live Registry install → MCP credential-rotation journey was not available locally because the sibling evenfire-registry checkout is absent. The remote Registry contract job is green, but please assess whether any additional live integration evidence is required.
- Please inspect for any remaining P0/P1 correctness, security, concurrency, rollback, or production blast-radius issue; verify that the exact annotation preservation is non-forgeable and that the fresh-bootstrap NP-08 exception cannot accept stale or revoked runtime material; and confirm that the T0/T1/T2 harness changes do not mask failures.

Please report any blocker or required change against this exact head, with severity and file/line references. If no blocker remains, provide the final approval recommendation for merge.

alfredolopez80:
@claude Adversarial follow-up for the exact pushed head `51bffdbf`.

I challenged the assumptions behind your closing review:

- N1 survives as an operational compatibility concern, but not as a security defect. The correct invariant is fail-closed: a connector Secret with partial or unknown `clerum.io/*` metadata must not be treated as Registry-managed. I kept that behavior, added a regression proving the merge rejects with HTTP 400 before `patchNamespacedSecret`, and documented the operator remediation. No annotation-preservation exception was broadened.
- N2 does not survive. CodeQL alert #847 most recent instance is already `fixed`; the new CodeQL run is also green. No dismissal or security-state mutation was needed.
- N3 does not survive. The current PR description already documents all three allowed Secret types and the `{}` snapshot contract.

Validation after the adversarial adjustment:

- Local `control-api`: 363 test files passed, 21 skipped; 4621 tests passed, 103 skipped.
- Focused regressions: `services.secretService.test.ts` 156/156 and `routes.mcpSecrets.test.ts` 41/41.
- GitHub CI/CD: `33/33 SUCCESS`, including `control-api`, Desktop, Postgres 16, leak guards, shell syntax, and CodeQL.
- PR head: `51bffdbf`; base remains `dev@3b9ff0f`.

Please perform the final review against this exact head. The PR is merge-ready subject to the required human/repository approval; automatic merge is intentionally not being used.

alfredolopez80:
@claude Please perform the final approval review for PR #252 at exact HEAD `df74dc356bcb01b28229ccf4f45a1f2412913c27` over base `3b9ff0f42dfbca5708f9d53d9c1f799748556d1d`. All changes are committed and pushed.

Validation: GitHub CI is 33/33 green; focused RED→GREEN `full-setup` harness passes; exact-head local T2 passes T0, T1 (108/108), T2, and NP-08 Host-to-HCC authorization. Playwright is `NOT_RUN` by the opt-in contract because no command is configured.

Please review the complete diff and specifically confirm: Registry install-hook/guardrails blast radius; public versus internal annotation writers and SSA ownership; UID/resourceVersion-safe upgrade/rollback; NP-08 Secret/Deployment/ReplicaSet/Ready-pod token rotation; Docker endpoint/probe/pull/build boundaries and egress allowlists; T1 fail-closed JSON plus process-exit semantics; and the surgical CI fixture fix.

The independent GPT-5.6 SOL MAX review found no confirmed P0/P1/high-severity defect and confirmed the Registry/guardrail boundaries are constrained. Known evidence limits: the sibling `evenfire-registry` checkout was unavailable for live registry runtime seeding, and Playwright was not configured. Please report any surviving blocker; if none remains, approve the PR / mark it ready for merge. Do not merge automatically.

zach88:
Took a deep pass through this one for round 2. The secret-constraint hardening itself is solid and security came back clean, but two things stand out: a fix from round 1 introduced a real rotation regression, and the PR ballooned from 15 to 100 files by absorbing a big chunk of unrelated minikube/scripts/mcp-host work.

### Verdict
🔴 **Not ready to approve**
- **Must resolve before approval: 2** (1 High regression + 1 Medium regression, same root class)
- **Optional / non-blocking: 4**
- Heads up: this PR is well over the size budget (100 files, ~4700 lines of it unrelated to the secret feature). Strongly suggest splitting, see the process note at the bottom.

1. **[R2-H1] Validating existing annotations breaks secret rotation** `regression from 38ebf5e9`
   `mergeSecret`/`updateSecret`/`removeSecretKey` in `control-api/src/services/secretService.ts` now do an extra read and validate the *already-persisted* annotations of the target Secret. Any Secret carrying an infra annotation (`kubectl.kubernetes.io/last-applied-configuration`, which every client-side `kubectl apply` stamps, plus `meta.helm.sh/*` and `kubernetes.io/*`) now makes the mutation throw `DangerousAnnotationError` (400) instead of rotating. Production callers that hit this with no exemption: `PUT /admin/mcp-secrets/:name` (`control-api/src/routes/admin/secrets.ts:501`), `PUT /admin/channel-secrets` save and key-delete (`control-api/src/routes/admin/communicationChannelCredentials.ts:107,169`, no opts and no test), `PUT /admin/secrets` (`control-api/src/routes/admin/secrets.ts:306,892`), and `control-api/src/routes/admin/recipes.ts:1702`.
   Repro: unit test on SecretService with an existing secret carrying `kubectl.kubernetes.io/last-applied-configuration`. 3/3 of mergeSecret/updateSecret/removeSecretKey succeed at the round-1 head, 3/3 throw at the current head.
   It's fail-closed so there's no data loss, but it denies a legit rotation for a very common condition (kubectl-applied/helm/GitOps secrets). The `b219318d` follow-up only exempts the exact `clerum.io/catalog-id` + `catalog-version` pair, nothing for infra annotations or non-registry secrets. Fix direction (your call on intent): either validate only request-supplied annotations on the merge/removeKey path (the merge-patch never rewrites the preserved ones), or if fail-closed is intended, wire an actionable remediation message on both rotation routes and add a test for the channel-secret path. The upgrade route already got that treatment; its sibling rotation routes didn't.

2. **[R2-M1] Upgrade refuses outright on a kubectl-applied credential secret** `regression from 38ebf5e9`
   `POST /admin/registry/upgrade` now runs a fail-closed pre-check (`nonRestorableSecretSnapshotReason`, `control-api/src/routes/admin/registry.ts:2937`): if the existing credentials Secret carries infra metadata it returns 409 `credentials_secret_state_not_restorable` and blocks the whole upgrade, even when no rollback would ever be needed. Same root class as R2-H1. It's partly by design and at least degrades to a 409 with an actionable `LEGACY_SECRET_REMEDIATION` string, so it's softer than the bare 400 above, but a legit upgrade of a kubectl-applied cred secret is blocked. Reproduced at the service layer. Fixing the R2-H1 class should cover this one too.

## Optional (non-blocking)

1. **[R2-M2] PR body describes a design that no longer exists.** The summary says annotation snapshots are sanitized via `stripBlockedAnnotationKeys()` and that the opt-out is `{ allowPlatformAnnotations: true }`. Neither exists in the code anymore (0 references); the design was reworked to a `{ capability: 'registryCredential' | 'registryPullSecret' }` model. Worth rewriting the body to match before merge, especially since the automated final-validation comment above validated against the old description, so its confirmations about those two symbols don't reflect the current code.

2. **[R2-L1] `removeSecretKey` can't preserve even the whitelisted pair** `regression from 38ebf5e9` (class). It uses `assertValidSecretConstraints` with no opts, so unlike `mergeSecret`/`updateSecret` it can't keep the `clerum.io/catalog-*` pair. Latent today (no caller removes keys from a registry-managed secret), but it's the sibling left open.

3. **[R2-L2] invalid-JSON test dropped its auth header** (still open since round 1). `tests/e2e/integration/channel-reader-mcp-host.test.ts` removed the `Authorization: Bearer` header from the "400 for invalid JSON" test and still asserts 400. That assumes mcp-host's body parser runs before auth; if auth runs first it's a 401 and the test fails. Restore the header or assert the middleware order explicitly. This is a Medium that's now sat 2 rounds, so it should either get fixed or be closed out with a written reason rather than riding another round.

4. **[R2-L3] misleading param name.** `waitForTriggeredRunPhase(token, ...)` in `tests/e2e/integration/workflow-lifecycle.test.ts` takes a run id, not a token (it's ignored for auth). Rename `token` to `runId`.

## What checked out clean
- Security came back clean. The allowlist/denylist isn't bypassable (case/whitespace variants are rejected by the apiserver's own key rules), the `capability` opt-out isn't reachable from user input, the SA-token type and annotation are both blocked, and making `recipeExists` fail-closed is strictly more restrictive with no new DoS or bypass. Error-log sanitization drops upstream response bodies.
- control-api suite: 4626 passed, 103 skipped, 0 failed. `tsc --noEmit` clean.
- The e2e assertions actually got stronger (fail-loud `ctx.skip()` instead of silent early-return, hard `toBe('completed')`, real taskId checks, L6 secret existence + base64 content). No hand-written fixtures standing in for a producer.
- mcp-host changes are prettier reflow only, no coverage lost. The python audit tool and the minikube shell fixes check out too.

## One process note
The actual feature here is ~9 control-api src files (~820 lines). The other ~90 files (scripts/minikube ~4200 lines, mcp-host formatting-only churn, tools, docs) aren't related to the secret-constraint intent. They came back clean, but they make every review pass re-scan 100 files to find bugs concentrated in a handful. Strongly recommend pulling the minikube/scripts tooling and the mcp-host reflow into their own PR so this one is just the secret feature.

alfredolopez80:
@claude Please perform the final closure review for PR #252 at the exact current head `ae54324e05ad1d5d3d484117768f2e36759342c2`, which is synchronized with `origin/dev` `684005e0f28ad7a68badc2940b75345864302b7f` through merge commit `ae54324e`.

Please review the complete diff and explicitly validate:

- Secret constraint invariants: allowed Secret types, strict request-side annotation validation, unchanged infrastructure metadata preservation, and `clerum.io/*` capability/allowlist boundaries.
- The central `SecretService` behavior for full update, merge, and key removal, including CAS/resourceVersion behavior and no mutation on rejected injection.
- Registry upgrade preflight, rollback fidelity, UID/resourceVersion ownership binding, and preservation of legacy Kubernetes/Helm metadata.
- Blast radius of `install-hook`, Registry credential capabilities, guardrails, pull-secret paths, MCP rotation, channel credentials, and recipe secrets; confirm no public route can use the internal capability to inject metadata.
- Adversarial Zach88 findings R2-H1/R2-M1/R2-M2/R2-L1/R2-L2/R2-L3 and whether any material regression survives.

Implemented in the current head:

- One shared policy boundary in `secretConstraints.ts`; no route-specific exception for the regression.
- TDD RED tests first reproduced the pre-fix `DangerousAnnotationError` on legacy existing metadata; GREEN tests now prove update/merge/key-removal preservation, Registry mutation+rollback, and rejection of capability-side metadata injection.
- The obsolete remediation and stale PR-body claims were removed.
- The PR body now documents the actual design and evidence.

Validation evidence:

- Local full `control-api` suite before the dev sync: 4,630 passed / 103 skipped; TypeScript build passed.
- Post-sync focused SecretService/Registry suite: 325/325 passed.
- Current GitHub CI run `32774103856`: all required jobs passed, including control-api, mcp-host, desktop-app, Real Postgres, Registry contract, leak guards, manifest and shell checks.
- Current CodeQL run `32774104096`: `javascript-typescript` passed.

Known limits: prior local T2 evidence belonged to an older exact HEAD and is not being presented as current after the dev synchronization; the local evidence also lacks a live `evenfire-registry` sibling deployment and configured Playwright run. Please treat those as evidence boundaries, not as silently passed lanes. Do not merge automatically; report any remaining blocker or approve closure explicitly.

zach88:
Took a deep pass through round 3. Good news up front: the design blocker from round 2 is genuinely closed, not just relocated, and the relaxation didn't open a security hole. Most of what's left is test-coverage polish plus one stale nit that's overdue for a decision. One heads-up before the details: this is still a 100-file PR and the actual secret fix is ~19 of those files, so there's a strong case for splitting (see the note at the bottom).

**Status: 🟡 Approve with nits**
- Optional / non-blocking: 6

Nothing blocking. The round-2 blocker and both mediums are closed and verified:
- **[R2-H1]** closed by `1a8026b5`. The effective-state rule now applies consistently across `updateSecret` / `mergeSecret` / `removeSecretKey`, introducing infra/platform metadata by request is still rejected, and the regression tests genuinely fail against the fix's parent (`df74dc35`).
- **[R2-M1]** closed. Registry upgrade preflight no longer 409s a kubectl/helm-applied credential Secret; it still rejects unauthorized `clerum.io/*` platform metadata, and CAS (uid + resourceVersion) is retained on both the forward write and the rollback.
- **[R2-M2]** closed. The body now matches the capability model in the code.

## Optional (non-blocking)

The PR can merge without these. They're follow-ups or your call.

1. **[R3-M1] e2e coverage of the preservation fix is incidental.** `tests/e2e/integration/mcp-secret-rotation-preservation.test.ts` (and siblings) seed the pre-existing Secret with `kubectlApply(...)`, and `kubectl apply` happens to stamp `kubectl.kubernetes.io/last-applied-configuration`, so the rotation exercises the R2-H1 path only as a side effect. No e2e ever reads `secret.metadata.annotations` after the rotation, so the "infra annotations are preserved" half of the invariant lives only in the `services.secretService.test.ts` unit tests, not in the e2e the body sells as R2-H1 coverage. If someone switches those fixtures to `kubectl create` / server-side apply, the coverage silently disappears. Suggestion: seed an explicit infra annotation (e.g. `meta.helm.sh/release-name`) in the `secretYaml`, and assert the Secret's annotations before/after the PUT.

2. **[R3-M2] mockGateway doesn't validate constraints, so a route test leans on a permissive mock.** The real `K8sGateway` delegates to `SecretService` (which runs the validators); `mockGateway.updateSecret`/`createSecret` (`control-api/test/mockGateway.ts:327,372`) just store whatever they receive. The new upgrade+rollback test asserts the resulting Secret keeps its legacy infra metadata (`control-api/test/routes.registryInstall.test.ts:3411`), but that's produced by the mock. So if a future `SecretService` regression started rejecting `kubectl.kubernetes.io/*` on a registry upgrade, the unit test would go red while this route test stays green and asserts preservation that isn't actually happening. Suggestion: have `mockGateway.{create,update,merge}Secret` run the real `secretConstraints` validators, or add a contract-test case that requires the mock to reject what the real gateway rejects.

3. **[R3-L1] TOCTOU between upgrade preflight and the write path (regression from `1a8026b5`).** `nonRestorableSecretSnapshotReason` (`control-api/src/routes/admin/registry.ts:2905`) treats an infra annotation key as restorable without comparing its value, while `updateSecret` (`control-api/src/services/secretService.ts:123`) requires the value to match a fresh read. So a Secret can pass preflight but fail the write with a 400 if a concurrent writer (kubectl/helm) changes that annotation's value between the snapshot read and the update read. It happens after `ensureRegistryPullSecret`, which weakens the "fail closed before creating side effects" guarantee. It's race-gated and non-deterministic (fails closed, no data loss), which is why it's Low, but it's the one thing this round's fix introduced.

4. **[R2-L1] `removeSecretKey` is the strict sibling.** `control-api/src/services/secretService.ts` `removeSecretKey` takes no `opts`, so it can't preserve the whitelisted `clerum.io/catalog-*` pair the way `mergeSecret` can. No live caller hits this today (the only caller is the channel-credentials key delete, and channel secrets carry no `clerum.io/*` annotations), so it's latent. Suggestion: document the asymmetry, or thread `opts` for symmetry if a future route deletes keys from a registry-managed Secret.

5. **[R2-L2] invalid-JSON test dropped its Bearer header (still open since round 1, and this round made it worse).** `tests/e2e/integration/channel-reader-mcp-host.test.ts` "returns 400 for invalid JSON" used to send `Authorization: Bearer` + a bad body, so the 400 was deterministic. This round removed the header, so the 400 now depends entirely on the body parser running before the auth middleware; if auth ran first, a request with no token would 401 and the test would fail. This is the one finding that's now survived three rounds, so it needs a decision rather than another pass: either restore the Bearer header (cheap, recovers the original intent) or pin the middleware order as a contract with a mcp-host unit test and close it in writing.

6. **[R2-L3] misleading param name.** `tests/e2e/integration/workflow-lifecycle.test.ts` `waitForTriggeredRunPhase(token, ...)`: the first param is actually a `runId` (it's used as the run identity in the URL, not for auth). Rename `token` to `runId`.

A few smaller nits worth a boy-scout pass if you're in there: the annotation key in `routes.registryInstall.test.ts:3398` is rebuilt from a charCode array when the literal already sits in the same test's seed; the `canPreserve` branch in `mergeSecret` is effectively dead (RFC 7396 merge never sends `req.annotations`); and the `hasInfrastructureAnnotations` block is duplicated verbatim in `updateSecret` and `mergeSecret`, which is the kind of shared business rule worth extracting and testing as composable.

## What checked out clean

- **Security: safe.** The relaxation doesn't open a bypass. Across every mutation path the source of truth for prior state is always a live apiserver read, never a client-provided snapshot, so a request can't pass off injected infra or `clerum.io/*` metadata as "already there." Type stays deny-by-default over the effective type, ownerReferences/finalizers are covered separately by the 409 fail-closed path, and error logging got tighter (upstream bodies no longer interpolated; audit moved to the redacting logger). `recipeExists` went from fail-open to fail-closed.
- **Collateral: clean.** All call sites of the changed signatures were updated; for Secrets with no infra annotations the behavior is bit-identical to before; no still-valid security invariant lost its coverage.
- **Suites green** in the worktree: 823 passing across the core secret/registry suites, 437 across the 7 test files the fix touched, `tsc --noEmit` clean, no dangling references to renamed symbols. The regression tests were confirmed red against the fix's parent.

## A note on scope

The real diff against the merge-base is 100 files. The control-api secret feature is ~19 of them and it's converged (no blockers, nothing should-fix in the core logic). The rest is unrelated: ~36 minikube/scripts files, 15 e2e, 12 mcp-host files that are pure prettier-format reversion from dev (functional no-op, just diff noise that'll come back as a revert/conflict when this merges to dev), plus tools/docs. My strong suggestion is to extract the control-api feature into its own PR and merge it, move the minikube/scripts bulk to a separate PR, and drop the mcp-host churn with a `git checkout dev -- mcp-host/...`. That keeps the next round from re-reviewing 100 files to chase bugs concentrated in about 9.

alfredolopez80:
Scope split completed. PR #252 now points to ae44cf544a07ef78dd2a0e8c35cd12f4f197406e and contains exactly the 19 control-api files for the Secret metadata policy; Minikube/scripts/E2E/docs/tools were removed from this diff, and all mcp-host formatting churn was dropped. The extracted harness/support work is in PR #443 (https://github.com/evenfire-ai/evenfire/pull/443), based on origin/dev at 684005e0f28ad7a68badc2940b75345864302b7f, commit c3b993fc, with no control-api or mcp-host paths. Local control-api validation passed: 4630 passed, 103 explicit skipped (4733 total), and TypeScript build passed. CI/CodeQL are being recomputed for the new heads.

alfredolopez80:
@claude Please perform a fresh, independent closing review of the current PR head `ae44cf544a07ef78dd2a0e8c35cd12f4f197406e` against base `dev` / `684005e0f28ad7a68badc2940b75345864302b7f`.

This PR was intentionally rewritten after the previous 100-file review. Its scope is now exactly 19 files, all under `control-api/`, for the Secret metadata-policy feature and its tests. Minikube/scripts, E2E harness changes, tools/docs, and all `mcp-host` formatting churn were removed and split into PR #443.

Please review the actual current diff, not the historical commit discussion, and answer whether this PR is ready to merge. In particular, validate:

1. The central Secret policy is enforced consistently at the public boundary for create, update, merge, and remove paths, including strict type/annotation handling and preservation of existing external metadata.
2. Internal capability/full-snapshot paths cannot inject infrastructure metadata or bypass the intended allowlist/denylist semantics.
3. Registry install/upgrade/rollback behavior preserves UID/resourceVersion CAS guarantees, avoids destructive metadata loss, and handles failure compensation correctly.
4. The route/service/repository contracts, tests, package metadata, and TypeScript behavior agree with the production design; identify any dead or misleading test coverage.
5. The extraction itself introduced no hidden dependency on the removed Minikube/E2E/mcp-host files and does not rely on PR #443 being merged first.
6. Re-evaluate blast radius across production Secret consumers, especially registry pull-secret rotation and existing platform annotations.

Evidence available for calibration: the isolated local `control-api` suite passed 4,630 tests with 103 explicit skips (4,733 total), the TypeScript build passed, and CI plus CodeQL are green for this exact head. Treat those as evidence to verify, not as a substitute for the independent review.

Please report every surviving finding with severity, file/line, concrete failure mode, and required fix. Explicitly state if there are no blockers or should-fix findings and whether you certify the PR as merge-ready.

jozer-rami:
## Review findings: 2 blockers, 3 major

Reviewed against head `ae44cf54`. Full detail below; happy to discuss any of these.

### 🔴 Blockers

**1. `mergeSecret`/`removeSecretKey` reintroduce the multi-owner write race they exist to prevent**
[`control-api/src/services/secretService.ts#L225`](https://github.com/evenfire-ai/evenfire/blob/ae44cf544a07ef78dd2a0e8c35cd12f4f197406e/control-api/src/services/secretService.ts#L225)

Both methods now unconditionally embed the read `resourceVersion` into the JSON merge-patch, turning what was documented as a conflict-free merge into a CAS-protected write. `mergeSecret`'s own docstring says it exists specifically for multi-owner Secrets where one owner updates its keys without wiping another owner's keys (e.g. per-Host channel-reader credentials alongside HCC-owned keys). If a second owner touches the Secret between this function's read and its patch, the patch now 409s even though the two writes touch completely disjoint keys — reintroducing the exact race the merge-patch design was built to avoid.

**2. A single stray annotation permanently locks a Secret out of all future writes**
[`control-api/src/services/secretConstraints.ts#L119`](https://github.com/evenfire-ai/evenfire/blob/ae44cf544a07ef78dd2a0e8c35cd12f4f197406e/control-api/src/services/secretConstraints.ts#L119)

`existingSecretAnnotationKeyReason` validates *every* annotation already stored on the Secret, not just the ones the current write touches, and there's no self-service remedy once it fails. Any credentials Secret that ever picked up a `clerum.io/`-prefixed annotation outside the exact allowlist for its capability (e.g. only `catalog-id` set without `catalog-version`, or a future annotation key not yet added to any allowlist) will permanently 400 on every future `mergeSecret`/`updateSecret`/`removeSecretKey` call against it — **including a routine credential rotation via `PUT /admin/mcp-secrets/:name` that never touches annotations at all** — until an operator manually strips the annotation with `kubectl`. Given this validates pre-existing state rather than the delta, this can brick rotation for Secrets that predate this PR.

### 🟠 Major

**3. Ambiguous CRD write + failed classification read skips rollback entirely**
[`control-api/src/routes/admin/registry.ts#L3048`](https://github.com/evenfire-ai/evenfire/blob/ae44cf544a07ef78dd2a0e8c35cd12f4f197406e/control-api/src/routes/admin/registry.ts#L3048)

When the mcp-server upgrade's CRD write outcome is ambiguous (status-less/5xx error) and the follow-up classification read also fails, the handler logs and returns 503 (`registry_upgrade_outcome_ambiguous`) before ever reaching the credential-rollback block. If credentials were already rotated for this upgrade, they're left in their new state with zero rollback attempt — the pre-PR code always attempted a best-effort rollback here.

**4. The PR's own rollback hardening only covers 1 of 3 upgrade sagas**
[`control-api/src/routes/admin/registry.ts#L3342`](https://github.com/evenfire-ai/evenfire/blob/ae44cf544a07ef78dd2a0e8c35cd12f4f197406e/control-api/src/routes/admin/registry.ts#L3342) (upgrade-recipe), also upgrade-hook (~L2518-2547)

The PR description calls out "preflight de Registry con identidad/CAS (uid y resourceVersion) y rollback no destructivo," but the upgrade-recipe route has no try/catch at all around its CRD replace, and upgrade-hook only has a bare status-forward/rethrow. Neither gained the ambiguous-outcome (read-after-error) classification added to the mcp-server upgrade route. A 5xx/timeout during a workflowrecipe or llmhook upgrade CRD write that actually committed at the apiserver is reported to the client as a hard failure with no read-after-error to prove otherwise — the bug class this PR sets out to fix remains live in the other two sagas, and there are no tests for the 5xx/ambiguous path on either.

**5. Secret write endpoints silently change response shape**
[`control-api/src/routes/admin/secrets.ts#L158`](https://github.com/evenfire-ai/evenfire/blob/ae44cf544a07ef78dd2a0e8c35cd12f4f197406e/control-api/src/routes/admin/secrets.ts#L158) (also `PUT` at ~L306-307)

`POST`/`PUT /admin/secrets` echo `gateway.createSecret`/`updateSecret`'s return value straight to the client. This PR silently changes that value's shape from a nested K8s `V1Secret` (`.metadata.name`/`.metadata.namespace`/`.metadata.labels`) to the new flat `SecretSnapshot` (`.name`/`.namespace` at top level), with no test covering the real response shape. Any external consumer reading `.metadata.name`/`.metadata.namespace` off these documented admin endpoints silently gets `undefined` instead of erroring. Currently harmless because `control-ui` discards the body on both call sites, but it's an untested, unannounced breaking response-shape change.

---
*(5 additional medium/low findings — empty-string secret type validation, an unrecognized new error code, a retry-misclassification edge case, a latent capability-scoping smell, a stale test double, and some duplication/efficiency cleanups — happy to share those too if useful.)*

zach88:
Took a deep pass through this one for round 4. Good news up front: the metadata-policy design converged back in round 3, and the control-api code here is byte-for-byte identical to what got verified then (`git diff <r3-head>..HEAD -- control-api/` is empty). What actually changed this round is the re-scope: the branch got rebased and squashed down from ~100 files to 19, all under `control-api/`, with the minikube/mcp-host/scripts/e2e churn moved out to #443. That's exactly the split we'd been asking for, and it lands the PR inside the size budget. So no new findings this round, and nothing in the policy logic itself is broken.

The one thing keeping me from a clean thumbs-up is a fix-introduced regression that's still sitting open rather than dispositioned.

**🔴 Not ready to approve**
- Must resolve before approval: 1
- Optional / non-blocking: 2

1. **[R3-L1] Preflight is more permissive than the write-path for infra-annotation *values* (regression from `1a8026b5`)** — `control-api/src/routes/admin/registry.ts:2905` / `control-api/src/services/secretService.ts:123`. The preflight in the upgrade path checks annotation *keys* against `existingSecretAnnotationKeyReason`, but the write-path re-validates the full constraint set including the annotation *value*. So there's a race window where a request passes preflight, we start doing side-effecting work (`kubectl apply`), and then the write-path throws a 400 on the value, which weakens the "fail closed before side effects" guarantee the fix was supposed to hold. It's Low severity and race-gated (non-deterministic, no clean repro), but per our process an open regression shouldn't ride into merge without a decision. Two ways to close it: either tighten the preflight to validate the annotation value with the same check the write-path uses (so a request that will 400 never mints side effects), or, if you consider the window acceptable, close it by writing in the ledger with the justification signed. Right now it's neither, which is why it gates.

## Optional (non-blocking)

The PR can merge without these; they're follow-up or author's-call.

2. **[R3-M2] `mockGateway` doesn't enforce the real constraints, so the upgrade+rollback test asserts preservation over a permissive mock** — `control-api/test/mockGateway.ts:327,372`, load-bearing at `control-api/test/routes.registryInstall.test.ts:3411`. The mock accepts writes the real gateway would reject, so the test proves the idea of the contract rather than the contract. Worth having the mock route through the real `secretConstraints` validators so the green means what it says. Still open since round 3.

3. **[R2-L1] Strict sibling in `removeSecretKey`: can't preserve even the catalog-whitelisted pair** — `control-api/src/services/secretService.ts` (`removeSecretKey`). Latent, no live caller exercises it today (confirmed core + collateral). Either document the asymmetry or thread an `opts` through so it matches the other two paths. Still open since round 2.

## What checked out clean

- The control-api policy code is identical to the round-3-verified head (empty diff), so all four fixes that were verified then (R2-H1, R2-M1, R2-M2, and the R1 mediums) survive the squash intact.
- Scope is back in budget: 19 files, 100% `control-api/`, single service.
- No semantic conflict with current `dev` under the touched files. The only overlap is the version bump in `package.json`/`package-lock.json`, a trivial textual resolve at merge.
- One coverage note, not a blocker: the e2e rotation/preservation tests left with the re-scope, so this merges on unit coverage (which is substantial: +679 in `services.secretService.test.ts`, +316 in `routes.registryInstall.test.ts`, +118 in `mockGateway.contract.test.ts`, +139 in `services.pullSecretConstraints.test.ts`). Would be good to bring that e2e back in #443.

zach88:
Took a focused pass on this one for round 5. Rather than re-review the whole head again, I scoped it to two things: what the 6 new commits since the last reviewed head (`ae44cf54`) might have broken, and a design look at the new CAS module that landed (`registryMutation.ts` / `secretReferenceService.ts`). Good news up front: the collateral pass came back clean — no hard regressions, the `identity-safe` refactor got propagated to `control-ui` and the unit tests in the same commit, and nothing re-introduced a Secret-value leak (the channel-credentials route even dropped the raw `result` field, nice). So nothing here blocks on correctness.

The one thing I'd genuinely push on isn't a bug, it's process: this is now the module's **second** `identity-safe` pass, and `registryMutation.ts` is a real decision module (CAS readback classification with a `committed | not-committed | ambiguous` taxonomy). The route boundary is fail-closed and safe on the live paths, but the effective decision table got discovered through patches rather than designed up front — and there's already dead decision code + one fail-open spot to show for it. My honest recommendation is a short decision-table spec to close out the items below, instead of iterating a round 6.

## Verdict
🟡 **Approve with nits**
- **Optional / non-blocking: 5** (this round) + 3 carried over from earlier rounds

None — nothing open in the blocking tier, and the collateral pass found no hard regressions.

## Optional (non-blocking)
This PR can merge without these on correctness grounds. That said, `R5-M1` + `R5-L3` + `R5-L4` are the substance behind my "write a small spec first" note — they're the shape of thing that comes back a round later if left as-is.

1. **[R5-M1] `secretReferenceService` fails open on unknown reference shapes** — `Medium`. `REFERENCE_KEYS` / `REFERENCE_SCOPES` (`control-api/src/services/secretReferenceService.ts:16-44`) are hand-maintained lists with no link to the CRD schema. It's the only classifier that degrades fail-*open*: a new CRD or a reference field with a different name falls into `'not-referenced'`, and that path deletes a Secret that's actually in use. It's latent today (everything known is enumerated), but the safety of the delete guard depends on someone remembering to update two sets. Worth either deriving these from the schema or adding a test that fails when a new CRD reference field isn't enumerated.

2. **[R5-L3] Dead secret-readback classifier** — `Low`. `classifySecretMutationReadback` (`registryMutation.ts:80-131`) has no route caller — production goes through `readSecretMutationOutcome` (`routes/admin/registry.ts:1052`), which collapses to `ambiguous` without calling it. So the subtlest ~50 lines (metadata race, delete/recreate) plus their 5 tests certify a contract prod never runs. Prune it or wire it.

3. **[R5-L4] Unreachable `not-committed` branch** — `Low`. The `not-committed` arm of `classifyRegistryMutationReadback` (`registryMutation.ts:216-225`) is dead: its only consumer (`registry.ts:1338`) reads `=== 'committed'` only, and production's `not-committed` arrives via the fence, not the classifier. Half the tri-state output is unread.

4. **[R5-L2] Rotation lost its `resourceVersion` fence** — `Low`, and really a design call for you. `mergeSecretRaw` (`services/secretService.ts:301-327`) now passes `uid` only, no `resourceVersion`. Two concurrent rotations of the *same* key used to 409-and-retry; now it's silent last-writer-wins. It's documented in-code as intentional (multi-owner / disjoint keys), but that contradicts the earlier comment claiming this partial mutation "cannot overwrite a concurrent rotation." Just want you to confirm the per-key last-writer-wins is the intended trade, since the code says two different things.

5. **[R5-L1] e2e cleanups desynced with the new CAS delete** — `Low`. `DELETE /admin/mcp-secrets/:name` and `/admin/recipe-secrets/:name` now require `uid`+`resourceVersion` in the body (428 without it). A few e2e cleanups still delete by name with no body — `control-ui/e2e/registry-install.spec.ts:2670`, `qa-recorder-secret-connector.spec.ts:77`, `qa-recorder-connector-edit.spec.ts:312`, `qa-recorder-secret-recipe.spec.ts:85`. They're best-effort in `finally`/`afterAll` so they won't turn the suite red, but they'll leave orphan secrets in the e2e cluster.

Also worth noting, since I didn't re-audit them this round: **R3-M2** (mockGateway doesn't run the real constraint validators), **R3-L1** (preflight/write-path TOCTOU), and **R2-L1** (`removeSecretKey` strict sibling) are still open from prior rounds per the ledger. R3-M2 is a Medium that's now been open ≥2 rounds — it should either move up or get closed in writing rather than ride another round.

## What checked out clean
- Collateral (`ae44cf54..7251309d`, the 6 branch-own commits + the dev merge): **no hard regressions**. Internal `SecretService` signature change (`SecretSnapshot` → `SecretSummary`, with new `*Snapshot` variants) is safe — the only direct caller is `k8s.ts:415-457` using the `*Snapshot` forms, and route callers go through the gateway.
- `ambiguous` / missing-identity classification is **fail-closed** everywhere (503 `*_outcome_ambiguous` / `repair_required`) — no "ambiguous treated as success" silent data loss on the registry paths.
- The two `test: mark ... synthetic` commits only rename credential literals for secret-scanning hygiene; no assertions touched, no coverage lost.
- No Secret-value leak reintroduced — audited every `res.*.json` on the changed routes; `communicationChannelCredentials.ts` dropped the raw `result` field in favor of `rotatedKeys`.
- Dev merge (`8e6dd317`) carried no mis-resolved conflicts in `control-api/src/**` or `control-ui/**`.

Note: unit/integration suites weren't executed this round (worktree had no `node_modules`); the above is read-verification except where a path was traced end-to-end.

zach88 (approved):

alfredolopez80:
@claude

Please perform a complete independent closing review of Evenfire PR #252. Do not merge automatically. Review the exact pushed head:

- PR: https://github.com/evenfire-ai/evenfire/pull/252
- HEAD: `e9876bbb71665e7c1625c313abc8f7457605c6f4`
- base: `origin/dev` / `d90f564daba7283d12961e0c1a764874c7adb12b`
- branch: `fix/secret-type-annotation-constraints`
- current PR state: open, mergeable, `mergeable_state=clean`

## Scope boundary

The synchronized diff is 42 files and is limited to `control-api/**` and `control-ui/**`. This PR intentionally does not contain Minikube/T0/T1/T2 harness changes, `scripts/minikube/**`, `mcp-host/**`, or the E2E/harness work extracted to PR #443. Please do not reintroduce #443 findings as blockers for this PR; do verify that #252 has no hidden dependency on that work.

## What was reviewed and corrected

1. Secret effective-state constraints are centralized in `control-api/src/services/secretConstraints.ts` and enforced by create, update, merge, and remove-key paths. Existing opaque infrastructure/platform metadata is preserved; caller-introduced or changed reserved metadata is rejected. The capability boundary is internal and key-scoped (`registryCredential` / `registryPullSecret`); public routes cannot opt into it.
2. The multi-owner merge race is closed: data-only merges do not inject an implicit resourceVersion, while UID fencing is retained where identity ownership is required.
3. Existing `clerum.io/*` metadata no longer bricks ordinary future writes. Partial/unknown metadata remains fail-closed for registry ownership decisions; it is not silently treated as a valid capability.
4. Registry mcp-server, recipe, and hook upgrades use the common identity/CAS mutation protocol in `control-api/src/routes/admin/registry.ts`. Upgrade, rollback, and uninstall are bound to UID/resourceVersion; replacement Secrets cannot be mistaken for the original object.
5. Ambiguous CR write/readback outcomes never become a false success and never trigger unsafe blind compensation. They return the repair-required/ambiguous outcome and preserve the live dependency state for operator repair.
6. MockGateway now exercises the relevant validators/identity/CAS contract rather than accepting an unrealistically permissive write.
7. The earlier response-shape concern is not a PR regression: the names-only public Secret response contract is already the current `origin/dev` contract and is covered by the existing route tests.
8. Zach88's process/scope finding was resolved by extracting the unrelated Minikube/scripts/mcp-host/E2E work to #443 and synchronizing this PR's body with the actual capability design and current evidence.

## R3-L1 fault-injection proof

The last regression was stale preflight metadata being replayed during an upgrade. The implementation no longer forwards annotation metadata from the preflight snapshot into the credential write (`registry.ts` around the forward upgrade request). The new test in `control-api/test/routes.registryInstall.test.ts` deliberately changes the live Secret between preflight and write: UID remains the same, resourceVersion changes from 1 to 2, and an infrastructure annotation changes. The expected result is HTTP 409, not success. The test proves that the current live Secret retains UID, RV=2, annotation, data, and the existing McpServer CR; there is no false commit and no overwrite.

## Review disposition to challenge

- Jozer-rami: multi-owner CAS regression, existing metadata lockout, ambiguous readback compensation, incomplete saga hardening, and response-shape concern have been resolved or safely disposed as described above. The additional observations in the original comment were not specified with reproducible detail; please re-evaluate them explicitly and report any material survivor rather than assuming they are closed.
- Zach88: R2-H1/R2-M1 metadata-rotation regressions, R2-M2 stale PR-body/design description, and R3-L1 are addressed. R2-L2/R2-L3 and the Minikube/mcp-host/E2E observations belong to the extracted #443 scope. R2-L1 is latent/non-blocking because no live registry caller removes keys from a registry-managed Secret; please call it out if the current diff proves otherwise.
- Prior Claude findings: effective constraints, capability non-forgeability, partial platform metadata, MockGateway behavior, CAS/rollback/uninstall identity protection, and CodeQL status were revalidated against this exact head.

## Evidence for this exact head

- TDD RED/GREEN was performed for R3-L1; the fault injection is green with the fail-closed 409 assertion.
- Focused Registry suite: 208 passed.
- SecretService + metadata-preservation + MockGateway contract suites: 198 passed.
- Full `control-api`: 4,787 passed, 103 explicitly skipped, across 392 files.
- `npm --prefix control-api run build`: passed.
- Prettier check and `git diff --check`: passed.
- GitHub Actions workflow run `32909380327` for this head: completed successfully; required PR checks report success, including control-api, Control UI, Real Postgres, Registry contract, leak guards, manifest, shell, and CodeQL.
- CodeQL check: success with no PR-specific open alert reported.
- Local worktree is clean; no automatic merge is requested.

Please read the complete current diff, challenge the concurrency/identity/rollback claims and the capability blast radius, and return a fresh closing verdict. If no blocker or material should-fix survives, please approve/mark this PR merge-ready subject to the required human approval. If anything survives, give the exact path, scenario, severity, and the smallest complete correction. Do not merge automatically.

alfredolopez80:
@claude

Please perform a complete independent closing re-review of Evenfire PR #252. Do not merge automatically.

Review target:
- PR: https://github.com/evenfire-ai/evenfire/pull/252
- Exact remote HEAD: `be27d3bc15a74b05e0ab7199dfa98d68b75743fe`
- Base: `origin/dev` `6b7292cacdaeb106b5fbc83a9c8c82873f6f5ccd`
- Branch: `fix/secret-type-annotation-constraints`
- The merge-base matches origin/dev and the pushed worktree is clean.

Mandatory historical-finding validation:
1. Read every prior issue comment, review submission, and inline review comment on this PR, including findings from `claude[bot]`, `zach88`, `jozer-rami`, and `github-advanced-security[bot]`.
2. Build and validate a finding-by-finding matrix: original claim, affected path/scenario, prior disposition, current exact-head verification, and final status. Treat every prior fixed, closed, non-blocking, or out-of-scope disposition as untrusted until independently verified.
3. Re-challenge the historical Secret constraints/capability boundary, Registry mutation and CAS/UID/resourceVersion handling, stale metadata/rollback/uninstall behavior, MockGateway fidelity, response shape, error handling, security findings, and the prior E2E/harness observations.
4. Validate the scope split in detail. The companion Minikube harness work is PR #443 (`chore/minikube-harness-t0-t2`); PR #433 does not exist. Confirm that #252s synchronized diff is intentionally limited to `control-api/**` and `control-ui/**`, and that Minikube, `mcp-host`, E2E, and harness concerns moved to #443 do not leave a direct dependency or regression in #252. Do not reintroduce #443-only findings as #252 blockers unless you prove a concrete dependency or failure path in this PR.

Changes since the prior closing review:
- `ebe44327`: test fixture correctly models the no-Host-reference path.
- `be27d3bc`: tracked Next route declaration aligns with the production build output.

Validation for this exact head:
- Local T0: PASS.
- Local T1: PASS, 108/108 Real PostgreSQL tests across 22 expected files.
- Local T2: PASS; NP-08 deployed Host-to-HCC authorization: PASS.
- CI/CD: all 34 reported checks succeeded, including CodeQL, control-api, Control UI, Real PostgreSQL, Registry contract, Desktop, manifests, leak guards, and shell checks.
- Local T2 Health and Playwright are NOT_RUN because no opt-in commands were configured; they are separate lanes and must not be treated as green evidence.

Please review the exact current diff and return a fresh closing verdict. For every surviving finding, provide severity, path, concrete failure mode, and the smallest complete fix. Explicitly state whether every prior finding was independently validated and whether this PR is merge-ready subject to human approval. Do not merge automatically.

alfredolopez80:
@claude

Please perform a final closing re-review of PR #252 at exact HEAD e4f246f1c2873394d8779cbccadce4ea8871aae4 against origin/dev d8fe82d02c7a0c97e00e0a163c44feeabc2609ec. Do not merge automatically.

Re-validate every prior issue/review/inline finding, including Claude, Zach88, Jozer-rami, and CodeQL, rather than trusting earlier dispositions. An adversarial pass plus an independent GPT-5.6 SOL MAX review found and corrected the surviving items:
- R5-M1: Secret reference detection now covers explicit current CRD fields, including array-shaped imagePullSecrets and OAuth clientIdRef/clientSecretRef; deletion is blocked while referenced.
- R5-L3/L4: removed the unused credential readback classifier and made an unchanged Registry readback remain ambiguous until the caller has an identity fence.
- Direct E2E compatibility: seven Control UI cleanup callers now preserve/send uid + resourceVersion, assert cleanup status, and no longer rely on positional selectors or terminal-route navigation. E2E Guardian static audit passes for all five changed specs, and Playwright loads all changed specs successfully.
- The branch was re-synchronized with current origin/dev, including HCC changes; direct E2E callers remain in #252 because they consume #252s DELETE identity contract. PR #443 remains the separate Minikube harness scope.

Exact-head evidence:
- Local T0=PASS.
- Local T1=PASS: 108/108 Real PostgreSQL tests across 22 files.
- Local T2=PASS; NP-08=PASS; required targeted-sync Health=PASS.
- CI/CD: 34/34 checks success, including CodeQL, Control API, Control UI, HCC, Real PostgreSQL, Registry contract, and Desktop.
- Worktree clean; PR remains open.
- Local Playwright runtime was NOT_RUN because no opt-in command was configured. This is not claimed as green; static audit and Playwright collection passed.

Please return a fresh closing verdict. For every surviving issue, provide severity, exact path, scenario, and minimal complete fix. Explicitly say whether the PR is merge-ready subject to human approval. Do not merge automatically.

alfredolopez80:
@claude Please perform the final close-out re-review of PR #252 at `afcbd030b9563ea5030cb82acc97ea8ca9b1e9a1`. This is a review request only; do not merge.

Please re-check every prior finding and the scope split:

- #252 is limited to `control-api`/`control-ui`. The Minikube T0–T2 harness split is PR #443 (not #433; #433 does not exist in this repository).
- R5-M1 is closed with CRD-derived coverage for every declared Secret reference, including the previously omitted LlmHook `authHeadersSecret`; a new schema contract fails if an unscoped field/CRD is added.
- N1/R3-M2 is validated as resolved: MockGateway create/update/merge invoke the shared validators.
- R3-L1 is validated fail-closed: the concurrent metadata test returns 409 and preserves the competing change/CR.
- R2-L1 has an explicit catalog-id/catalog-version key-removal regression test.
- R5-L2 remains the documented intentional multi-owner last-writer-wins behavior, not a silent regression.
- The Control UI E2E was updated against the current visible wizard/actions, added accessible labels for egress/context controls, and makes credential rotation require `Credentials rotated.` (timeouts/failures are red).

Exact-head evidence:

- T0: control-api 4,790 passed (103 skipped); control-ui 1,745 passed (1 skipped); both builds passed.
- T1: 108/108 real PostgreSQL tests passed (22/22 expected files).
- T2 runtime on `afcbd030`: NP-08 PASS, user-facing health PASS, Playwright PASS.
- Playwright: both connector-edit journeys passed together (egress edit and credential rotation), with verified cleanup.
- E2E Guardian static audit: PASS.
- GitHub CI: 34/34 checks green, including CodeQL.

Please report only evidence-backed remaining blockers, if any, and confirm whether this PR is ready to close. Again: do not merge it.

alfredolopez80:
@claude Please perform a complete closing re-review of PR #252 at exact head `fadd179ca006e4c2dd4d59508b1e3871e0d885b8`, based on `origin/dev` `ff8bb922f9abd4df5737c1a23c72f7fbc4306b58`.

This is a review request only. It is not authorization to merge the PR.

Please execute the closing review in three explicit phases:

1. **Historical convalidation:** re-read every prior review comment and validate every reported finding and every corrective change against the current exact head. Produce a complete table with one row per finding/change, its origin, current status (`CLOSED`, `OPEN`, or `NOT APPLICABLE`), current-head evidence, test evidence, and any residual risk.
2. **Independent re-review:** ignore prior conclusions as authority and perform your own fresh, line-by-line review of the complete `origin/dev...HEAD` diff, including correctness, security, concurrency, failure recovery, compatibility, migration/deployment safety, tests, and operational blast radius.
3. **Closing audit:** reconcile both phases, identify any missed or newly introduced issue, and return an explicit final `PASS` or `FAIL`, P0/P1 blockers, blast-radius result, deployability result, and whether the PR is ready for human merge authorization. Do not merge it.

Please re-read and validate every historical finding and review thread, including:

- https://github.com/evenfire-ai/evenfire/pull/252#issuecomment-5451648912
- https://github.com/evenfire-ai/evenfire/pull/252#issuecomment-5452999987
- all earlier Jozer, Zach, Claude, and CodeQL findings on this PR;
- the independent 2026-08-28 review, which reported B1, H1-H2, M1-M13, and L1-L12.

Please explicitly validate the following historical finding families against the current code, not against an earlier head:

1. B1: Registry upgrade rollback restoration and production/MockGateway parity.
2. H1: preserved or custom Secret types must not become permanently unwritable.
3. H2: CRD Secret-reference discovery must fail closed across all nine CRD schemas.
4. M1-M13:
   - production/mock validator parity;
   - API-before-UI deployment and paired rollback;
   - ambiguous mutation fencing and compensation;
   - unreachable or tautological compensation branches;
   - committed/readback audit reachability;
   - last-writer-wins and truthful rotation auditing;
   - Kubernetes `stringData`/`data` mock fidelity;
   - UI mappings for new error identities;
   - comments matching actual validation behavior;
   - ambiguous hook/recipe saga coverage;
   - swallowed cleanup failures;
   - inert validators/dead policy code;
   - retry masking and zero-test risks.
5. L1-L12:
   - dead exports and ineffective branches;
   - duplicated Registry operation annotation constants;
   - swallowed resource-deletion errors;
   - apiserver failures converted to 404;
   - capability-scope smells;
   - duplicated or weaker E2E identity helpers;
   - inconsistent 428/404 ordering;
   - post-create identity-unavailable compensation;
   - Registry response diagnostic detail;
   - UID-only fencing/comments and useless conditionals;
   - recipe PUT/DELETE identity asymmetry;
   - SecretsTable identity-branch coverage.
6. Every prior JZ-*, R2/R3/R5-*, CQ-*, and approval-obsolescence observation.

Also perform a detailed review of the newer security architecture added after that report:

- PostgreSQL-backed one-time rollback permits;
- domain-separated SHA-256 JTI storage;
- 120-second permit TTL and 15-second claim lease;
- atomic claim, transient release, terminal finalize, and stale-claim reissue;
- name-bound CAS cleanup;
- cancellation of rejected tokens and failed Registry bundle paths;
- removal of the rollback-proof cookie and resolution of CodeQL #1198 without dismissal;
- migration ordering, runtime ACL/profile changes, and all cancellation/rollback failure paths.

Scope split must also be validated: all branch-owned Minikube harness/profile/port-forward/image/setup/T0-T2 tooling was extracted to PR #443:
https://github.com/evenfire-ai/evenfire/pull/443

PR #433 is not a pull request in this repository; #443 is the verified split PR. Confirm that PR #252 does not reintroduce that extracted Minikube harness scope.

Current evidence:

- GitHub CI/CodeQL: 34/34 SUCCESS;
- CodeQL #1198 latest instance: fixed, not dismissed;
- canonical local certification: T0=PASS, T1=PASS, T2=PASS;
- Real PostgreSQL: 23/23 files and 113/113 tests, zero skipped;
- NP08_HCC_AUTHORIZATION=PASS;
- all 25 images verified and all core workloads Ready;
- two independent GPT-5.6 SOL MAX adversarial reviews: CODE PASS, P0=0, P1=0;
- Health=NOT_RUN and Playwright=NOT_RUN; no browser-runtime PASS is claimed.

Please return:

- an explicit overall PASS or FAIL;
- the complete historical convalidation table requested above;
- a separate findings table from your fresh independent audit;
- any remaining P0/P1 blocker;
- a blast-radius verdict covering PostgreSQL migration, Control API, Control UI, nine CRDs, runtime ACL/deployment, and E2E contracts;
- a migration and rollout-order verdict;
- a final deployability verdict;
- a clear statement of whether every prior finding and every corrective change has been independently revalidated.

Do not merge the PR.

alfredolopez80:
@claude Please audit this pull request.

Head is `40d22871f`; all 41 CI checks are green on it.

Context, so the audit does not repeat work already done:

- Two review rounds have been adjudicated: the CodeQL inline findings and the
  open review comments from @zach88, @jozer-rami and @claude[bot]. Survivors
  were fixed in `9fe8f2ff7` and `4f91d758c`.
- A blast-radius analysis against `dev` has been done. Its conclusions: a stale
  `control-ui` bundle loses Recipe Secret delete (428) and its MCP-create
  rollback outside the 120 s permit window; an old-checkout redeploy is refused
  by the runtime-access privilege verifier; `control-api` is `Recreate` with one
  replica, so the DB migration Job must complete before the image rolls.

Most useful areas to look at:

1. The MCP Secret rollback permit lifecycle — issue, claim, finalize, release —
   under concurrent requests and replica restarts.
   `control-api/src/services/mcpSecretRollbackPermitService.ts` and
   `control-api/src/routes/admin/secrets.ts`.
2. Whether any delete path can reach Kubernetes without both `uid` and
   `resourceVersion` preconditions.
3. The registry saga changes in `control-api/src/routes/admin/registry.ts`:
   whether an `ambiguous` outcome can leave a resource in a state the operator
   cannot diagnose from the response body alone.
4. Anything in the diff that would break a caller outside this repository.

Please skip style and naming. This branch has already been through several
formatting rounds, and findings there are not what is missing.

PR #253, opened 2026-08-03
perf(ui): keep auth shells mounted and stabilize registry scope

josue-tech-94:
## Problem

Protected routes in both Control UI and Profile UI owned their own authentication and navigation shells. Client-side route changes therefore unmounted and rebuilt the sidebar, causing flicker and resetting local navigation state.

Profile UI access data and Control UI publish scope also needed shared, per-user provider contracts so permission-dependent navigation and pages observe the same loading, error, cache-reset, and refresh semantics.

## Changes

- Added persistent authenticated application frames to Control UI and Profile UI.
- Kept public login, invitation, confirmation, reset, and setup routes outside protected frames.
- Preserved root `inviteToken` links as public so `/?inviteToken=...` redirects before protected auth gating.
- Centralized publish scope across Sidebar, Publisher, and the publish form.
- Made publish-scope cache resets safe against late in-flight responses.
- Centralized Profile UI manageable teams and approval targets in a provider.
- Added fixed, visible sidebar placeholders while access entries resolve.
- Made loading skeletons and spinners respect reduced-motion preferences.
- Preserved sidebar DOM identity across protected client-side navigation.
- Kept Control UI and Profile UI together because their persistent-frame, access-provider, and loading contracts are mirrored.
- Refreshed provider-backed publish scope after direct registry mutations and authoritative registry status re-syncs.
- Hardened publish-scope refresh regression tests to require forced provider refreshes.
- Restored repeated session-expiration 401 de-duplication coverage.
- Makes backend registry identity, credential, org API-key user-token, catalog/category, machine-token, and publish-scope cache entries unavailable after successful registry connection changes by advancing the shared identity generation.
- Prevents stale in-flight registry identity work from being cached or returned after an identity change, while preserving same-generation request deduplication.
- Replaces Publisher, registry, workflow recipe permission, and Profile auth loading text with accessible skeleton fallbacks.
- Preserves root invitation-token redirects before Profile UI auth gating.
- Keeps Profile UI active access navigation stable during access refreshes.
- Adds Profile UI build, typecheck, and test coverage to the public CI gate.
- Restores global not-found pages outside authenticated application frames.
- Restores shell-safe skeleton loading for Control UI route fallbacks.
- Bounds backend publish-scope cache lifetime while preserving generation invalidation.

## Validation

- Control API registry-client and registry-connection focused tests passed.
- Control UI publish-scope, registry panel, auth, sidebar, Publisher, and loading-style focused tests passed.
- Profile UI test suite passed.
- Profile UI typecheck passed.
- Control UI production build passed.
- Profile UI production build passed.
- Prettier and `git diff --check` passed.


josue-tech-94:
@claude Please perform a deep review of this PR.

Focus on App Router lifecycle behavior, persistent shell identity, public-route classification, nested auth boundaries, cache invalidation, concurrent request deduplication, and loading-state accessibility. Please also trace protected navigation in both UIs for any route that could remount the sidebar or issue a full document navigation.

Review the Suspense fallbacks for the required contract: titles and disabled actions remain visible while every body section uses skeletons. Flag race conditions, stale per-user access data, missing cleanup, or regression-test gaps.

josue-tech-94:
@claude Please perform a final merge-readiness review of the current PR head
587fdb46a5a85cc0f8b42719e697b947ad1cbd7e against the PR base branch.

Verify the final code as it exists now, not intermediate commit history. Focus on
the registry identity-generation contract, stale in-flight retry behavior,
cross-organization token/catalog isolation, persistent UI shell behavior, and
whether the current PR description accurately describes the final behavior.

Please identify only current actionable issues. Do not repeat findings that are
already fixed unless they still reproduce on this exact head.

jjwakn:
@claude Follow-up from your latest merge-readiness review on `587fdb46a5a85cc0f8b42719e697b947ad1cbd7e`:

Please address these current actionable items only:

1. Move `ProfileShell.persistence` coverage into `profile-ui`'s own test harness, and remove the `control-ui` cross-project alias / `tsconfig` exclusion if they become unnecessary.
2. Fix the malformed PR-description backend bullet and phrase the backend cache behavior as generation-scoped invalidation rather than literal map deletion.
3. Replace the root `profile-ui` `Suspense fallback={null}` with a lightweight fallback or narrower boundary so `useSearchParams()` suspense does not blank the whole shell.

Keep the registry identity/cache implementation unchanged unless this follow-up work reveals a concrete defect. Please keep the PR behavior final-state only and avoid unrelated cleanup.

josue-tech-94:
@claude check again

josue-tech-94:
@claude review the latest commit please

jozer-rami:
@josue-tech-94 — deep review of this PR. Four independent lenses (UI correctness, security/tenancy, test integrity with live mutation testing, backend cache concurrency), every finding re-verified against the code before it made this list. Ran the suites and mutation-tested the load-bearing assertions rather than trusting the green checkmarks.

**Summary:** one blocker (a crash on the invite-link path), one real backend cache defect, and the PR's headline publish-scope fix has no test that can fail. The control-api half is the strongest work here.

---

## What's solid

The registry identity-generation design is careful. Every cache family named in the description — machine token (`control-api/src/services/registryClient.ts:167,182`), catalog/category (`:394`), publish scope (`:705,720`), org API-key user token (`control-api/src/services/orgApiKeyClient.ts:107,114`) — gates on both read and write, re-reads the generation *at write time* rather than trusting the captured value, and the bump plus every check-then-set are synchronous with no intervening `await`, so there's no read-modify-write window. Same-generation dedup is preserved, errors aren't negatively cached, and `markConnected`'s `if (wrote)` guard is right.

`control-api/test/registryClient.test.ts` is the best test file in the PR — all five identity-generation mutations I applied went red, including a no-op `invalidateRegistryIdentityCaches()`.

The frame refactor itself is sound: no auth-gate regression (the AuthGate-presence set across all 100 control-ui and 14 profile-ui pages is byte-identical to `dev`), the re-entrancy guards make the retained per-page shells correct no-ops, `isDetailPage` lifting survives detail→detail navigation and Suspense fallback swaps, and `next build` passes for both apps.

---

## Blocker

### 1. `/?inviteToken=…` hard-crashes for a signed-in user

`profile-ui/lib/profileAppFrame.ts:12`

The rule that keeps root invite links public makes `ProfileAppFrame` (`profile-ui/app/components/ProfileAppFrame/index.tsx:16-17`) return bare children with **no `ProfileAccessProvider`**. But `profile-ui/app/page.tsx:29,62` still renders `<AuthGate><ProfileShell>`; `ProfileShell` sees no `ProfileShellContext`, so it mounts `PersistentProfileShell` → `profile-ui/app/components/Sidebar/index.tsx:13` → `useProfileAccess()` → throws at `profile-ui/app/components/ProfileAccessContext/index.tsx:160`.

Reproduced with a throwaway render (deleted afterwards):

```
Error: useProfileAccess must be used within a ProfileAccessProvider
 ❯ useProfileAccess app/components/ProfileAccessContext/index.tsx:160:21
 ❯ Sidebar app/components/Sidebar/index.tsx:13:5
```

**Trigger:** an already-authenticated user opens an invite link. `getMe()` starts on layout mount, *before* the page effect issues `router.replace('/invitations/<token>')`, and `profile-ui/app/invitations/[token]/page.tsx:40` is an async server component — so auth commonly resolves first, `HomeContent` renders on the still-`/` pathname, and the render throws before the redirect commits. There is no `error.tsx` anywhere in either app, so this is Next's blank "Application error" screen — on the exact flow commit `426cf541` was written to protect.

**Why the suite is green:** `profile-ui/app/components/__tests__/ProfileShell.persistence.test.tsx:26-30` mocks `@components/Sidebar` away.

**Fix (cleanest):** drop `ProfileShell` from `HomeContent` — the frame owns it now, and `currentRoute="home"` is already ignored — and move the `inviteToken` redirect into `ProfileAppFrame` ahead of the public check. **Cheapest:** make `useProfileAccess()` return a fail-closed default instead of throwing, matching what `usePublishScope` already does with `EMPTY_STATE`.

---

## Major

### 2. The connection row is the one registry cache left ungated

`control-api/src/services/registryConnectionDb.ts:73-106`

`getRegistryConnection()` writes `cached = {...}; cachedAt = Date.now()` unconditionally after its `await db.query`, with no generation captured before the await and none checked at the write — while `markConnected` (`:171-173`) clears the cache and bumps the generation. A SELECT that straddles the commit reinstalls the **pre-connect** row with a fresh 15 s TTL (`CONNECTION_CACHE_TTL_MS`, `:59`).

Verified consequence, for 15 s immediately after the connect succeeded:

- `isRegistryAuthActive()` (`:236-239`) reads the stale row and returns false → `/admin/registry/keys` returns `409 registry_auth_disabled` (`control-api/src/routes/admin/registry.ts:2176`).
- `mintToken()` returns `''` (`control-api/src/services/registryClient.ts:146`) → catalog reads go out with **no `Authorization` header at all** (`authedFetch:252` only sets it when the token is truthy).

This is exactly the invariant the description claims to establish, so it's worth closing before merge.

**Conditional amplification:** `control-api/src/routes/admin/registryConnect.ts:341` guards only on `status === 'connected'`, so a stale `'pending'` row lets a re-clicked Connect run `upsertPendingConnection` → `DELETE FROM registry_connection`, destroying just-claimed credentials and the keypair (the comment at `:447-453` calls that unrecoverable). `:349` separately blocks `'approved'` rows, and with `OPEN_REGISTRATION` on the pre-claim row is `'approved'` — so this is only reachable against a registry that does not auto-approve.

**Fix:** capture the generation before the query and gate both the write and the read on it:

```ts
const gen = getRegistryIdentityCacheGeneration()
const res = await db.query(...)
if (isRegistryIdentityCacheGenerationCurrent(gen)) { cached = row; cachedGen = gen; cachedAt = Date.now() }
// read side: require cachedGen === current too
```

### 3. The headline publish-scope fix has no test that can fail

`control-ui/lib/hooks/usePublishScope.ts:114`

I changed `options.force ? undefined : scopeCache.get(cacheKey)?.scope` to always read the cache — making `refresh({force:true})` return the stale scope forever — and **all 13 tests stayed green**.

That line is the mechanism behind "Refreshed provider-backed publish scope after direct registry mutations", commit `a1aab0e1` "make publish-scope refresh authoritative", and "Hardened publish-scope refresh regression tests to require forced provider refreshes". The existing tests cover `force` only inside `requestPublishScope`, which the provider short-circuits before ever reaching. The one production caller — `control-ui/components/RegistryConnectPanel.tsx:118` — is tested against a wholesale mock of `usePublishScope` (`control-ui/components/__tests__/RegistryConnectPanel.test.tsx:21-30`), so its ~8 assertions verify the mock was *called* with `{force:true}`, never that `force` bypasses anything.

### 4. A transient refresh failure erases access and hides the nav entry you're standing on

`profile-ui/lib/profileAccessState.ts:49,60`

The error reducers reset `manageableTeams: []` / `approvalTargets: []`, applied in the provider's catch blocks (`profile-ui/app/components/ProfileAccessContext/index.tsx:69-75,104-110`).

- A team lead (non-admin, non-inviter) on `/members` hits Refresh; `profile-ui/app/members/page.tsx:46`'s `refreshManageableTeams({force:true})` 500s → `canManageMembersForAccess` goes false → `profile-ui/app/components/Sidebar/index.tsx:48` marks `members` denied and the entry vanishes underneath them.
- Same shape on settings: a failed `refreshApprovalTargets` (`profile-ui/app/settings/SettingsContent.tsx:149`) empties `approvalTargets`, so `visibleSocialTabs` (`:123-127`) renders nothing — every social tab disappears — along with the Approval channels nav entry.

Recovers only by navigating away and back. Pre-PR each page held its own state and left it untouched on a failed reload.

**Fix:** keep `previous.manageableTeams` / `previous.approvalTargets` on error and flip only the `*Error` flag — drop the `: []` resets.

### 5. The new cross-user cache resets are untested

`control-ui/components/AuthContext.tsx:62,100`

Deleting `resetPublishScopeCache()` + `invalidateRegistryCapabilityCache()` from the 401 handler leaves `AuthContext.test.tsx` at 6/6 green. Same for the pair in `login`. These guard against user A's `@acme` publish scope being served to user B after a same-tab login, and nothing would notice their removal.

### 6. Neither UI's tests gate this PR

- profile-ui appears **nowhere** in `.github/workflows/ci-public.yml` — not in the matrix (`:48-66`), no build job, no typecheck — and is absent from `Makefile` `TEST_SERVICES`, so `make test` misses it too. Its only workflow is `.github/workflows/build-publish.yml:226-228`, which is `on: push: branches: [dev]` — a post-merge image build, not a PR gate.
- control-ui gets `next build` alone (`ci-public.yml:193-208`, deliberate per the comment at `:188-192`).

Of ~20 new/changed test files, only the 3 control-api ones are CI-enforced. This PR rewrites profile-ui's auth, shell, and access layer with no pre-merge gate at all. Not caused by this PR, but this PR is where it starts to hurt — adding profile-ui to the matrix would be a good companion change.

---

## Minor

- **`_scopeCache` has no TTL** (`control-api/src/services/registryClient.ts:690`). I could not construct a durable poisoning — every chain ends at a second generation bump in `markConnected` that clears it — so this is fragility rather than a live bug. Worth giving it the bounded TTL the connection row already has.
- **Refreshing skeletonises the active sidebar entry** (`profile-ui/app/components/Sidebar/index.tsx:44-46`). The `members` branch of `accessLoading` is guarded by `&& !canManageMembers`; `approvalChannels` has no equivalent, so any forced refresh greys out the link the user is on.
- **Both 404 pages regress** (`control-ui/app/not-found.tsx:7`, `profile-ui/app/not-found.tsx:7`). They render `<main className="cu-app cu-app--auth">` (`globals.css:682-690`: `min-height:100vh`, centered). Unknown paths aren't public, so they now render `<main>` nested inside `DashboardLayout`'s `<main className="cu-main">` — duplicate landmark, 100vh block inside the content column — and a logged-out user hitting a bad URL is redirected to login instead of seeing the 404.
- **Deleting `control-ui/app/loading.tsx`** leaves `/settings`, `/hosts/[name]`, `/contexts/[name]`, `/secrets/[scope]`, `/outputs/[tab]`, `/traces/**` and others with no navigation feedback — on-demand segments with no `loading.tsx` in themselves or any ancestor. A root `loading.tsx` rendering a body-level skeleton would now land *inside* the persistent `DashboardLayout`, so it no longer blanks the shell.
- **Vacuous curator assertion** (`control-ui/lib/hooks/usePublishScope.ts:168`). Deleting `!scope.curator` keeps 13/13 green — the only curator case (`usePublishScope.test.tsx:69-71`) passes `{scope: null, curator: true}`, already false via `scope.scope !== null`. No `{scope: '@x', curator: true}` case exists.
- **`profile-ui/lib/profileAccess.ts:43-47` lacks the liveness check its control-ui twin has** (`usePublishScope.ts:70`). Benign today (the stale write lands on a detached object), but the description says these two contracts are mirrored and they aren't.
- **`mintToken` is the only identity path with no in-flight dedup** (`control-api/src/services/registryClient.ts:102-107`), unlike `getUserToken` / `withReadThroughCache` / `resolvePublishScope`. N concurrent requests on a cold token cache each POST `/oauth/token`.
- **Invalidation doesn't zero the secrets it invalidates** (`control-api/src/services/registryIdentityCache.ts:10-12`). After a disconnect the process still holds the old org's machine bearer and every admin's user token in heap until natural expiry. Not servable — the generation gate blocks every read path — so heap-residency only.
- **Description is stale after the `dev` merge.** The merge correctly took dev's Sidebar, which deleted the Publisher tab entirely, so "Centralized publish scope across Sidebar…" and "Added fixed, visible sidebar placeholders while access entries resolve" no longer describe control-ui at HEAD. `control-ui/app/globals.css:9003` `.cu-sidebar__item--loading` now has no consumer.

---

## Notes

- **profile-ui test wiring:** the new `test/*.test.ts` files do run, but only because `profile-ui/test/invitations.test.ts:4-9` side-effect-imports them. `vitest.config.mts:25` includes only `app/components/**/__tests__/**` and `package.json:8` names a single file, so any future `test/*.test.ts` not added to that barrel silently never runs. Worth a comment on the barrel.
- **Persistence tests** (`DashboardLayout.persistence.test.tsx`, `ProfileShell.persistence.test.tsx`) mock the Sidebar and use `rerender` on an identical tree. They're not vacuous — forced-remount and route-misclassification mutations both go red — but they can't reproduce a real App Router layout remount. `control-ui/e2e/qa-recorder-navigation.spec.ts` is the only real-navigation check, and Playwright isn't in CI.
- **Pre-existing, not from this PR:** 5 control-ui vitest failures (`LlmModelTable.test.tsx:207`, `GfsBrowser.test.tsx`) and 1 tsc error (`LlmDiscoveryPanel.test.tsx:149`), all already on `dev`.

Happy to take the blocker and the `registryConnectionDb` fix if that's useful.

josue-tech-94:
@claude Please perform a final merge-readiness review of the current PR head
`2b132307fc2072822b64d3dd19cd76b106f2ed2e` against the latest PR base branch.

Please review the final code as it exists now, not intermediate commit history.

The latest change set addresses Jozer's review findings in the collateral diff:

`59ff941220244eea4844e0d2583ebb81fcd02d83..2b132307fc2072822b64d3dd19cd76b106f2ed2e`

Please independently verify the complete PR and focus particularly on:

* signed-in and logged-out `/?inviteToken=...` behavior, including provider and
  shell placement;
* registry connection-row cache generation safety when a database read overlaps
  connect, disconnect, or credential replacement;
* provider-level `{ force: true }` publish-scope refresh behavior;
* preservation of same-user Profile access after transient refresh failures,
  without retaining prior-user access;
* login, logout, and session-expiration cache invalidation;
* direct Profile UI test discovery and the new Profile UI pull-request CI gate;
* stable permission-controlled sidebar navigation during background refreshes;
* authenticated and unauthenticated not-found behavior in both UIs;
* Control UI route loading fallbacks remaining inside the persistent shell;
* the backend publish-scope cache TTL and its interaction with identity-generation
  invalidation;
* interactions between the eleven new focused commits that may not be visible
  when reviewing each fix separately.

Please verify that the current tests would fail if the load-bearing fixes were
removed, especially:

* the signed-in invitation crash regression;
* stale connection-row cache repopulation;
* forced publish-scope refresh cache bypass;
* Profile access preservation after refresh failure;
* cross-user auth cache invalidation;
* Profile UI test discovery and CI coverage.

The known lack of real Profile UI App Router/browser shell-persistence coverage
is an accepted non-blocking follow-up unless you find a current runtime defect.

The root local `make test-unit-all` omission for Profile UI is also currently
tracked as a Low follow-up; please flag it only if it creates a present
merge-blocking correctness or CI gap.

Please report only issues that currently reproduce on this exact head. Do not
repeat older findings that are fixed unless the defect still exists in the final
code.

Conclude with a clear verdict:

* merge-ready;
* merge-ready with follow-ups; or
* blocked, with the exact current actionable issues.

jozer-rami (approved):

PR #264, opened 2026-08-04
Followup/pr 137 deep link findings

josue-tech-94:
## Summary

Follows up on the deep-link review findings identified after PR #137.

This PR strengthens the complete deep-link flow from a shared Profile UI link to the requested app, team, and route in the desktop application. It improves routing testability, retry behavior, route normalization, security assertions, confirmation handling, and cross-version compatibility.

### Deep-link routing and delivery

* Makes the central deep-link router independently testable without triggering Electron startup side effects.
* Covers startup, second-instance, renderer-readiness, queued delivery, window recreation, and ordered draining.
* Preserves exactly-once delivery while links wait for the renderer and authenticated session.

### Retry and team-context behavior

* Keeps the target team active while a starting app is retried.
* Keeps the target team active while starting apps, transient team-context
  failures, or failed native mounts are retried.
* Keeps authorization, membership, and other permanent failures terminal.
* Uses bounded retries and avoids duplicate mounts or acknowledgements.
* Adds coverage for retry exhaustion, successful recovery, and permanent failures.

### User confirmation

* Requires confirmation for app links received while the user is already authenticated.
* Preserves the existing confirmation flow for links received before authentication.
* Prevents duplicate prompts and duplicate navigation for repeated delivery of the same link.
* Cancelling a link acknowledges it without opening the requested app.

### Route normalization and sharing

* Establishes one canonical route-normalization contract for shared and default app routes.
* Prevents double encoding when links containing Unicode, spaces, encoded values, or literal percent characters are opened and shared again.
* Keeps traversal, encoded-separator, control-character, query, fragment, and authority protections intact.
* Documents an additive compatibility policy for Profile UI and desktop versions released at different times.

### Profile UI handoff security

* Verifies the real handoff response is not cacheable.
* Verifies the response includes a Content Security Policy.
* Confirms the CSP nonce matches the nonce used by the emitted redirect script.
* Preserves referrer-policy and content-type protections for successful and invalid requests.

### Test coverage

* Adds property-based coverage for pending-link ordering, uniqueness, idempotency, retention limits, and interleavings.
* Adds direct coverage for the central routing and delivery seam.
* Adds behavioral tests for confirmation, cancellation, retries, route round trips, team switching, and security headers.
* Uses real application-generated link and queue shapes rather than synthetic contract fixtures.

## Validation

* Desktop application tests and production build
* Profile UI tests and production build
* Shared deep-link package tests and build
* TypeScript type checking
* Repository style-rule checks
* Changed-file formatting checks
* `git diff --check`
* GitHub Actions
* CodeQL

## Not run locally

Desktop E2E and platform-specific protocol tests require service and Kubernetes configuration that
was not available in the local review environment.

## Checklist

* [x] This is not a new third-party MCP server or WorkflowRecipe.
* [x] Resolve the remaining retry lifecycle finding.
* [x] Synchronize the branch with current `dev`.
* [x] Rerun CI after synchronization.
* [x] Complete the final PR re-review.


josue-tech-94:
@claude Please review the **entire PR diff against the current `dev` base**, not only the latest commit.

Review all changes across the desktop main process, renderer deep-link lifecycle, Profile UI handoff, and shared deep-link contract. Focus on:

* correctness and regressions across the full PR;
* security and untrusted deep-link handling;
* cross-process and cross-version contract compatibility;
* retry, team-switch, confirmation, acknowledgement, and exactly-once behavior;
* route normalization and encoding safety;
* test quality, including false-green tests and missing observable assertions;
* whether the previously reported `R1-M1` manual-team-switch retry regression is fully closed without introducing collateral behavior changes.

Please separate blocking from non-blocking findings, cite `file:line`, give a concrete failure scenario, and state what validation you ran or could not run.

josue-tech-94:
@claude Please perform a final review of the **entire PR diff against the current `dev` base**, not only the latest commits.

The current PR head is `d9d02fb3dadd47b954140f67c6302278539e83ca`.

Please verify that the findings from your previous review are fully resolved:

* The cross-team failed-mount retry no longer treats an app-initiated restore as a manual team switch.
* A real manual switch to a third team still wins during any pending retry.
* Failed mounts remain bounded and retryable without premature acknowledgement, duplicate mounts, or stale restoration.
* Required desktop-environment and desktop-setup parameters fail closed and have direct tests.
* The previously weak or vacuous deep-link test assertions now verify observable behavior.
* Route documentation accurately describes canonical semantic round trips rather than byte-exact encoded round trips.

Also re-review the complete PR for:

* correctness and collateral regressions;
* security of untrusted deep links;
* retry, confirmation, team-switch, mount, and acknowledgement lifecycle behavior;
* cross-process and cross-version contract compatibility;
* route normalization and encoding safety;
* false-green tests or missing invariant coverage.

Please clearly state:

1. Whether the PR is safe to merge.
2. Any blocking findings.
3. Any remaining non-blocking findings genuinely worth fixing before merge.
4. What tests or validation you ran and what you could not run.

zach88:
Took a deep pass through this one across the main process, renderer, the shared `desktop-app-links` package, and the profile-ui handoff. Overall it's solid work — security actually gets *stronger* here (every authenticated app link now requires confirmation, retry is fail-closed, the openExternal base-URL is validated), the router extraction faithfully reproduces the old `handleEvenfireUrl`, and the route-normalization contract is genuinely unified across all three layers. One real blocker though: the new canonicalization broke the idempotency invariant the PR body promises.

My suggested sequencing: fix the blocker + its missing test together in one commit (the property test must fail against the pre-fix head), and leave the three Low items as follow-ups.

## Blockers

**1. [R1-B1] `normalizeSandboxUiRoute` isn't idempotent — the share round-trip loses data**
`packages/desktop-app-links/index.cjs:75` (+ `:57-72`). The new canonicalization *decodes* percent-escapes, but the `.trim()` on line 75 runs on the whole input string, not per-segment after decoding. So when the last segment decodes to a value with trailing whitespace (e.g. a trailing `%20`), the first pass keeps the space and a second pass strips it: `normalize(x) !== normalize(normalize(x))`.

Reproduced:
- `normalizeSandboxUiRoute('/report%20')` → `'/report '`
- `normalizeSandboxUiRoute(normalizeSandboxUiRoute('/report%20'))` → `'/report'` (space gone)
- `buildSandboxUiDeepLink({recipeNs:'ns', recipeName:'app', path:'/report '})` → `evenfire://app/ns/app?path=%2Freport`
- `sandboxUiDeepLinkTargetsEqual({appRef:'ns/app', path:'/report '}, parse(link))` → `false`

That last one is the real bite: a canonical link stops being equal to its own re-share, which breaks the dedup this PR relies on (`targetsEqual` is the dedup key in the deep-link state machine, so a re-delivered "same" link can double-prompt / double-navigate). There's also a dot-segment variant: `normalizeSandboxUiRoute('/.%20')` → `'/. '` is accepted on the first pass, but re-sharing collapses it to `/.` and `buildSandboxUiDeepLink` throws.

This contradicts the body's own invariant ("one canonical route-normalization contract … idempotent share/open/share round trips" and "Prevents double encoding … when links are opened and shared again"). It's introduced by this PR — at the merge-base `normalizeSandboxUiRoute` didn't decode, so `normalize∘normalize` was idempotent. The practical trigger is narrow (routes ending in whitespace), but it's an executable repro against a promised invariant, so worth closing rather than waiving.

Fix: make canonicalization explicitly idempotent — either reject (`return null`) any segment whose decoded form has leading/trailing whitespace (consistent with "control characters remain rejected"), or apply the trim per-segment after decoding, before the `join`. The input-level `.trim()` isn't enough because it only touches the ends of the whole string.

## Medium

**2. [R1-M1] The contract test doesn't cover the promised invariant and green-lights the bug above**
`packages/desktop-app-links/index.test.cjs:51-88`. The added cases only check *interior* spaces (`/space here`) and happy-path mappings — there's no `normalize(normalize(x)) === normalize(x)` assertion and no `build→parse` round-trip over a canonical form with trailing whitespace, so the suite passes green with R1-B1 present. `fast-check` is already a devDependency; a property test over segments with `%20`/`+`/`%2520`/unicode asserting (a) idempotency and (b) `targetsEqual(canonical, reshare(canonical)) === true` would catch this. Pair it with the R1-B1 fix, and make it fail against the pre-fix head so we know it actually covers the bug.

## Smaller stuff (follow-ups, non-blocking)

**3. [R1-L1] New `tsc` error in a test file**
`desktop-app/ui/src/__tests__/App.deepLinks.test.tsx:827-828` — `currentController.pushToast.mock.calls` errors under `tsc` (`pushToast` is typed as a plain function) plus an implicit-any on the destructured `[message, tone]`. Doesn't gate CI (the renderer builds with `vite build` and tests run under vitest; neither typechecks these files, and other untouched test files already error), but it's the only new type error the diff adds. `vi.mocked(currentController.pushToast).mock.calls` cleans it up.

**4. [R1-L2] Confirmation dialog selector doesn't filter stale links**
`desktop-app/ui/src/App.tsx:843`. With the widened predicate (`confirmedIdentity !== currentIdentity`, no longer looking at `receivedIdentity`), on an A→B account switch there's a one-render window where B's session still holds A's pending link and the dialog momentarily renders "Open <appRef>" for A's link. The purge effect clears it within a frame and the processing path can't actually launch it (it does filter stale at `:592`), so impact is a sub-frame flash. Optional: gate the dialog with `!isPendingSandboxUiDeepLinkStale(item, identity)` for symmetry with the processing effect.

**5. [R1-L3] Cross-version nuance with literal `%` routes**
`packages/desktop-app-links/index.cjs:74` + `desktop-app/src/sandboxUiDriver.ts:427`. The on-wire `path` moves from byte-preserving to decoded-canonical, so a route whose literal segment is `literal%percent` now travels decoded as `/literal%percent`. A pre-PR desktop opening that link mounts it without the new `encodeSandboxUiRoutePath`, so Chromium leaves the bare `%` and it requests a byte-different server URL than a new client would (`literal%percent` vs `literal%25percent`). Needs the conjunction of a literal-`%` route + a link from the new stack + an un-updated desktop + a server strict about invalid `%xx`. Normal inputs (unicode, spaces) all converge across versions, and `encodeSandboxUiRoutePath` is the forward fix. Fine to accept as-is; maybe note the literal-`%` case in the README migration section.

## What checked out clean

- **Security — no new issues, several hardenings.** Authenticated app links now require confirmation (closes the drive-by gap); retry classifies transient/permanent denylist-first with default=permanent, keeps authz/membership terminal, and only mounts after a successful team switch + active `resolution.status` (no authz bypass); `requireProfileUiBaseUrlForBrowserAction` validates the base URL before `shell.openExternal`; canonicalization rejects traversal / encoded separators / control chars at every decode depth; the profile-ui handoff sets `no-store` + CSP with a 144-bit per-response nonce, and the new test extracts the nonce from the *real* response (CSP == script), so it's asserting the real producer.
- **Unified contract.** One `normalizeSandboxUiRoute`/`canonicalizeRouteSegment`/`isUnsafeRouteSegment` in the shared package, consumed by both profile-ui and desktop; the driver helpers delegate route safety to it and only add default-path metadata + re-encoding. `path`/`team` param parity holds; "same contract for default and shared routes" checks out.
- **Main process.** The `evenfireDeepLinkRouter` extraction matches the old behavior (host dispatch, ready-window gate, enqueue-dedup, drain splice-once); exactly-once holds under the "renderer ready + duplicate link before drain" race since the drain runs synchronously in the IPC handler.
- **Renderer.** Authenticated-confirmation, preserved pre-auth flow, prompt/nav dedup, cancel=ack-without-open, and the queue property tests are derived from the real `SandboxUiDeepLinkQueue` producer.
- **Suites/typecheck:** desktop-app `npm test` 1543/1543, `tsc -p tsconfig.json` exit 0; package `node --test` 7/7; profile-ui `desktopAppRoute` 54/54; touched renderer vitest suites 33+5 pass.

josue-tech-94:
@claude Please perform a final **full-PR review** of PR #264 against the current `dev` base, not only the latest commits.

Current head:

`531b226e15d4f035d1a43ce80996d4302fb3744c`

Please specifically verify the latest review fixes from:

https://github.com/evenfire-ai/evenfire/pull/264#issuecomment-5220615542

Confirm that:

* `normalizeSandboxUiRoute` is now idempotent for every accepted route;
* trailing/leading encoded whitespace no longer breaks share/open/share behavior;
* build → parse preserves the canonical target and `sandboxUiDeepLinkTargetsEqual` remains true;
* `%20`, `%2520`, literal `%`, `+`, Unicode, interior spaces, dot segments, traversal and encoded separators behave according to the documented contract;
* the new property/regression tests would actually catch the previous canonicalization bug;
* the new renderer test TypeScript error is resolved;
* stale links from user A never appear in user B's confirmation dialog during an identity switch;
* literal-percent cross-version behavior is either compatible or accurately documented.

Then re-review the **entire PR** for:

* correctness and collateral regressions;
* security of untrusted deep links;
* exactly-once/dedup behavior;
* retry, team-switch, confirmation, mount and acknowledgement lifecycle;
* Profile UI → shared link → Electron main → renderer → Sandbox UI contract compatibility;
* route canonicalization and version skew;
* false-green tests or missing invariant coverage;
* PR-description accuracy.

Please clearly state:

1. **Safe to merge: yes/no**
2. **Blocking findings**
3. **Any non-blocking finding genuinely worth fixing before merge**
4. **Whether the previous review findings are fully closed**
5. **What validation you ran and what you could not run**

zach88:
Took another full pass on this after the round-1 fixes landed. This round is clean: no new findings and no regressions from the fixes. All five round-1 findings are closed, each verified against the real code (not just read), and the three that changed behavior have regression tests that go red against the pre-fix head. From my side this is ready to merge.

I scoped this round to what actually changed since round 1 (the five fix commits, 11 files) plus a "what did these fixes break?" pass over that diff, rather than re-reviewing the untouched files.

## Round-1 findings, all closed

1. **[R1-B1] Route normalization is now idempotent** (`packages/desktop-app-links/index.cjs:34,75`, fix `b4df2062`). Dropping the outer `.trim()` and rejecting any segment with edge whitespace (`segment.trim() !== segment`) fixes the share round-trip. Reproduced: a 20k-case fuzz over `%`, ` `, `%20`, `%2520`, `%2e`, `%2f`, `..`, `%00` gives 2102 accepted routes, all idempotent, all with `sandboxUiDeepLinkTargetsEqual(canonical, reshare(canonical)) === true`. The original repro cases (`/report%20`, `/report `, `/.%20`, `/%20report`) now return `null` stably, and interior whitespace like `/report summary` still normalizes fine. The executable repro from round 1 is genuinely fixed, not closed on paper.

2. **[R1-M1] Contract tests now cover idempotency and round-trip** (`packages/desktop-app-links/index.test.cjs`, fix `fb7ec39c`). Property tests on `fast-check` (dev-only dep), and the arbitrary actually generates edge-whitespace segments, so it exercises the bug rather than dancing around it. Reproduced T3: the new tests fail by assertion against the pre-fix head.

3. **[R1-L2] Stale cross-identity confirmations are filtered** (`desktop-app/ui/src/App.tsx:844,855,870` + `sandboxUiDeepLinkState.ts:58`, fix `88a97d00`). The new `findPendingSandboxUiDeepLinkAwaitingConfirmation` adds `!isStale` to the display selector and both the confirm and dismiss handlers. I checked the "stuck link" concern: a stale+awaiting link that the handlers now hide still gets cleaned up by the processing effect at `App.tsx:592`, which auto-acks stale links before the in-flight guard, so nothing leaks. The renamed purge test keeps all three original synchronous assertions and adds the no-render one, so it's strictly stronger. T3 reproduced (red against the pre-fix head).

4. **[R1-L1] Toast assertions typecheck** (`desktop-app/ui/src/__tests__/App.deepLinks.test.tsx:838`, fix `4d78723c`). `vi.mocked(...)` removes the three new `tsc` errors (reproduced: present before, gone now) without changing the assertion.

5. **[R1-L3] Literal-percent version skew documented** (`packages/desktop-app-links/README.md`, fix `531b226e`). The note matches what the code actually does on the wire; it describes the skew without promising a protection the code doesn't provide.

## What checked out clean

- Security: SAFE. The route change is strictly more restrictive (everything previously rejected stays rejected, and it now also blocks variants the base accepted like `/..%20` and a leading-BOM segment); disguised dot-segments stay blocked; `normalize*` never throws. The confirmation filter closes the cross-identity leak in the dialog without opening a reverse drive-by, and the anti-drive-by processing path is untouched.
- Callers of the changed functions were traced: share links come from already-canonical paths, `profile-ui/lib/desktopAppLinks.ts:15` wraps the builder in try/catch (a malformed `?path=` now 400s instead of silently trimming), and the recipe default-path path falls back to `/` with a warn. No real producer breaks.
- Tests: `packages/desktop-app-links` 11/11, `desktop-app` `sandboxUiDriver` 27/27, renderer `App.deepLinks` + `sandboxUiDeepLinkState` 34/34.

Nice work on the fixes, the route contract is in good shape now.

zach88 (approved):

PR #266, opened 2026-08-05
feat(desktop): offer "switch to Localhost" when login hits a dead backend

andypavia:
## Problem

A desktop login aimed at a backend that isn't answering — most often the app is on a saved environment profile while a local minikube cluster is what's actually up — surfaced only as a generic **"Login failed"**. Nothing told the user that the *runtime profile*, not the password, was the problem. (This is the "login silently appears broken against minikube unless you're on the Localhost profile" trap.)

## What this does

On a **generic** login failure (not a 401, not an invitation error), the app diagnoses the backend and, when a local Evenfire is reachable but the active one isn't, surfaces a one-click **"Switch to Localhost & retry"** banner.

A clean **401 means the backend answered**, so it's a password problem, not a wrong-server problem — no switch is offered there. That gate keeps the affordance from firing on ordinary wrong-password attempts against a healthy backend.

## Design

- **`passwordLogin` keeps throwing** — the existing invitation / 401 / generic error messaging is untouched. The diagnosis is a separate, self-contained IPC the renderer calls only after a failure, so it never has to reconstruct those messages.
- **`AppService.diagnoseLoginBackend()`** returns a hint only when all hold: not already on localhost, localhost `/health` answers, and the active backend's `/health` does **not**. The two probes run in parallel with a 1.5s bound, so a diagnosis never makes a failed login feel slower.
- **`AuthClient.healthAt(baseUrl, signal)`** probes an explicit backend without mutating the active runtime config.
- Renderer: `useAuthController` stores `backendSwitchHint` and adds `handleSwitchLoginBackend` (select the target profile, then retry); `AuthPage` renders a `StatusBanner` + button. The hint clears on new login attempts and manual environment switches.

## Why not auto-switch / not a result-union

- **Auto-redirect to localhost** would hide which backend you're talking to and could silently reroute real credentials meant for a remote/staging server — surprising and mildly alarming. An explicit "switch?" keeps the user in control, matching the rest of the profile model.
- Converting `passwordLogin` to a `{ ok, hint }` result-union would have forced reconstructing all the existing thrown-error messaging in the renderer — more surface area, more regression risk, for no user-visible gain.

## Scope

All changes are under `desktop-app/` — main-process logic + renderer UI. No API, control-plane, or infra changes.

## Tests

- **`appService.diagnoseLoginBackend`** (4 cases): unreachable→hint, reachable→null, localhost-down→null, already-localhost→null-and-no-probe.
- **`AuthPage`** (2 cases): banner renders and its button fires the switch; absent without a hint.
- Full main-process suite green (333 tests); `tsc` clean on both projects with zero new errors vs. baseline; prettier clean.

## Manual verification

With minikube up and `make minikube-pf-desktop` running (localhost reachable):
1. Add an environment pointing at a dead URL and select it.
2. Sign in → after the ~1.5s probe, the **"Switch to Localhost & retry"** banner appears.
3. Click it → switches to Localhost, retries, signs in.

Guardrails confirmed: wrong password on a reachable backend → no banner (clean 401); already on Localhost with port-forwards down → no banner (nothing to switch to).


andypavia:
@claude do a deep review

Issue #204, opened 2026-07-28
bug(hcc): decouple readiness from initial fleet reconciliation


alfredolopez80:
## Observed cause and evidence

HCC starts its HTTP server before provider startup, but it does not mark the server ready until `await provider.start()` completes.

`McpServerWatcher.start()` awaits the initial reconciliation sequence, including the initial McpServer/NetworkPolicy work, CommunicationChannel snapshot/watch setup, the full Host fleet reconciliation, SharedFileSystem reconciliation, and GlobalFileSystem reconciliation. The Host fleet pass uses bounded cross-Host concurrency, currently defaulting to two workers.

A deterministic in-process reproduction holds only the initial Host full reconciliation:

1. HCC's HTTP server starts.
2. The initial Host fleet reconciliation remains pending.
3. `provider.start()` remains pending.
4. HCC readiness remains unavailable.
5. Releasing that reconciliation allows provider startup to complete and readiness to become available.

A focused provider-start test independently confirms that a pending initial Host full reconciliation is sufficient to keep `McpServerWatcher.start()` unresolved. This establishes lifecycle coupling between HCC readiness and complete initial fleet convergence. It does not claim a live-cluster latency measurement or implicate the separate `mcp-web-research` egress/protocol defect.

## User impact

On startup or rollout with a large MCP-server, Context, and Host fleet, HCC can remain unready for the whole initial convergence interval. This makes its discovery/readiness surface unavailable and can cause the deployment rollout deadline to expire even when reconciliation is progressing normally.

## Temporary mitigation

PR #202 (`fix(deploy): extend HCC post-overlay rollout timeout`) increases only HCC's post-overlay rollout timeout from 240 seconds to 480 seconds. It preserves the 240-second GFS writer/reader contracts and deliberately does not change HCC readiness, provider startup, reconciliation semantics, NetworkPolicies, CRDs, or GFS behavior.

That mitigation gives the current serialized startup more time; it does not remove the coupling described above.

## Proposed durable design

Split HCC bootstrap into two explicit phases:

1. **Safety-critical bootstrap:** establish the HTTP listener, required authoritative cache/watch continuity, and baseline safety prerequisites. HCC must remain unready if these essential prerequisites are unavailable.
2. **Background initial convergence:** schedule the full initial fleet reconciliation without making API readiness wait for every Host, McpServer, Context, SFS, or GFS operation to settle.

The implementation must preserve the existing safety semantics:

[...]

PR #202, opened 2026-07-28
fix(deploy): extend HCC post-overlay rollout timeout

alfredolopez80:
## Summary
- Extend the post-overlay `host-context-controller` rollout timeout from 240s to 480s.
- Preserve the 240s rollout timeouts for `gfsc-writer` and `gfsc-reader`.
- Lock the distinct timeout contract in the focused runtime provisioning test.

## Validation
- `bash -n deploy/scripts/provision-gfs-runtime.sh`
- `bash scripts/tests/test-provision-gfs-runtime.sh`
- `git diff --check`

## Scope
This is the controlled step 1 mitigation only. It does not change HCC readiness, `provider.start()`, MCP-server reconciliation, NetworkPolicies, CRDs, or GlobalFileSystem behavior.

PR #271, opened 2026-08-05
fix(deploy): make the HCC rollout wait a budget that can actually elapse

jozer-rami:
## The failure

The clerum-dev deploy for `449d2ff` (#253) failed at step 23, *Provision GFS runtime and dev CRD instances*:

```
[provision-gfs-runtime] Waiting for the post-overlay HCC rollout
Waiting for deployment "host-context-controller" rollout to finish: 1 old replicas are pending termination...
error: deployment "host-context-controller" exceeded its progress deadline
```

`Commit per-service pin bump` was skipped as a result, so `gcp-dev` is still pinned at `sha-75e44f8` while the cluster runs `sha-449d2ff`. Since `promote-dev-tags.sh` copies the **overlay pins**, that also silently blocks a correct dev → prod promotion.

## Root cause

Two separate things, and only together are they fatal.

**1. HCC takes a long time to become Ready.** `/ready` serves 503 until initial fleet reconciliation completes, so time-to-Ready scales with the fleet rather than with process start (#204; #205 decouples it). clerum-dev is at **108 McpServers / 22 Contexts / 11 Hosts** and took **651s**. clerum-prod is at 7/6/5 and was never close.

**2. The wait had no real budget.** `kubectl rollout status` aborts the instant the Deployment controller sets `Progressing=False`/`ProgressDeadlineExceeded`, so its `--timeout` can never outlast `progressDeadlineSeconds` — and that field was **never declared** on the HCC Deployment, so it was the Kubernetes default of 600s. Worse, the clock starts earlier than the wait: `apply-inter-service-tokens.sh:311` does a `rollout restart` of HCC several deploy steps before `provision-gfs-runtime.sh` starts waiting.

The 2026-08-05 numbers:

| event | time |
|---|---|
| new ReplicaSet created (step 19 restart) | 10:34:02 |
| step 23 begins waiting | 10:36:42 |
| controller aborts (`10:34:02 + 600s`) | **10:44:03** |
| `--timeout=480s` would have fired | 10:44:42 |
| HCC actually Ready | 10:44:53 |

The effective budget was `600 − 160 = 440s` against a 480s timeout. **The timeout raised in #202 has never once bound** — it was dead code. And the deploy gave up 50s before HCC came Ready.

## The fix

Both halves, so the intended budget is the real one:

- `deploy/base/control-plane/host-context-controller.yaml` declares `progressDeadlineSeconds: 1200` (the 900s wait plus a 300s head-start allowance; observed head start was 160s).
- `deploy/scripts/provision-gfs-runtime.sh` waits `900s`, comfortably over the observed 651s.

Against the numbers above this would have passed: HCC's 651s lands inside both the 900s wait and the 1200s deadline.

## Test

`scripts/tests/test-provision-gfs-runtime.sh` now pins the *relationship* between the two numbers, which is the thing #202 missed. Verified it fails for both regressions, not just that it passes:

```
FAIL: ... does not declare progressDeadlineSeconds, so it defaults to 600s
      and silently caps the rollout-status timeout above.

FAIL: progressDeadlineSeconds=900s cannot cover a 900s wait that
      starts 300s after the restart — the deadline would abort first.
```

## Scope and verification

- `bash -n` clean on both scripts. Shellcheck reports only a pre-existing SC2034 (present on `origin/dev` at line 33), left alone.
- `kubectl kustomize deploy/overlays/minikube` builds and renders `progressDeadlineSeconds: 1200`.
- No infra overlay sets `progressDeadlineSeconds`, so the base value applies unmodified to both `gcp-dev` and `gcp-prod`.

**This is a mitigation, not a cure.** #205 is the real fix — it decouples readiness from fleet convergence — after which both numbers should come back down. #205 is currently `CONFLICTING` and unreviewed, so this unblocks the dev deploy in the meantime.

`deploy/base` is shared with prod, so this rides along in the next promotion. On prod it is effectively inert: 7 McpServers, and a more generous deadline only ever delays a failure verdict.

### DIFF
--- a/deploy/base/control-plane/host-context-controller.yaml
+++ b/deploy/base/control-plane/host-context-controller.yaml
@@ -13,6 +13,22 @@ metadata:
     app: host-context-controller
 spec:
   replicas: 1
+  # HCC's /ready stays 503 until its initial fleet reconciliation finishes, so time-to-Ready
+  # scales with the McpServer/Context fleet rather than with process start. clerum-dev took
+  # 651s on 2026-08-05 at 108 McpServers / 22 Contexts.
+  #
+  # Unset, this defaults to 600s, and that default is what actually governed the deploy:
+  # `kubectl rollout status` aborts the instant the controller sets ProgressDeadlineExceeded,
+  # so provision-gfs-runtime.sh's --timeout could never exceed the deadline remaining. The
+  # clock also starts several steps earlier, at apply-inter-service-tokens.sh's `rollout
+  # restart` of HCC, so the wait inherits only what is left. On 2026-08-05 that was 600-160
+  # = 440s against a 480s timeout: the timeout raised in #202 never bound, and the deploy
+  # failed 50s before HCC came Ready.
+  #
+  # 1200s covers the 900s wait plus a 300s head-start allowance. This is a mitigation, not
+  # the fix — evenfire-ai/evenfire#205 decouples readiness from fleet convergence, after
+  # which both numbers should come back down.
+  progressDeadlineSeconds: 1200
   selector:
     matchLabels:
       app: host-context-controller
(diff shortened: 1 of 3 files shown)


alfredolopez80 (approved):
looks good approved

PR #272, opened 2026-08-05
feat(images): multi-arch publishing and release promotion

jozer-rami:
Workstream C. Publishes every image for both amd64 and arm64, and promotes
dev-proven digests to a release tag, so a following workstream can make
`make minikube-setup` pull images instead of building all 28 locally.

## What changes

- **`deploy/images.json`** is the one image list. Three previously disagreed:
  `build-publish.yml`'s 28-entry matrix, `ALL_IMAGES` in `build-images.sh` (28),
  and the minikube overlay's `images:` block (18). They are overlapping subsets,
  so the manifest carries per-image flags. `pull_in_ghcr_mode` is derived from
  `published && deployed_to_minikube`, never stored.
- **Every image is now built multi-arch**, on native runners rather than QEMU:
  the build job gained an `arch` dimension, and a new `merge` job assembles the
  manifest list. Previously only `mcp-host` and `channel-reader` were multi-arch.
- **`release-images.yml`** promotes on a `v*` tag by `crane copy`, never rebuilding.
- **`image-guard.yml`** asserts every published image stays anonymously pullable
  and multi-arch.

## Why the revision annotation rather than git history

Builds are change-detected, so each image's newest `sha-` tag sits at a different
commit, and the release tag is cut on `main` where the dev→main merge was never
built. Computing "the last commit touching this image's source paths" lands on
feature-branch merge commits that have no published image at all. So the image
records what it was built from, and the resolver reads it, checks it is an
ancestor of the release, and refuses if source moved since.

## Required before the next release cut, not just before the guard goes green

**Dispatch `build-publish.yml` with `build_all=true` on `dev` after this merges.**

Today no image carries a revision annotation — verified live, `mcp-host:latest`
returns `annotations: null`. Until that dispatch runs, **every** image fails the
release resolver, and 26 of 28 additionally fail the arch check. Cutting a tag
before then promotes nothing and leaves the following workstream pinning a tag
that does not exist.

That dispatch also fires the infra mirror for all 28 images and applies
`clerum-dev`, so it is a deploy-affecting step rather than a formality.

## `image-guard.yml` will be red on dev until that dispatch

Deliberate. The red is accurate — those images genuinely are not multi-arch yet —
and the guard triggers on `push: dev` only, never `pull_request`, so it cannot
block a PR. It was shipped enforcing rather than warn-only because a guard that
cannot fail is the exact defect this workstream exists to remove.

## Blast radius worth knowing

`merge` depends on `build-push` wholesale, so **one image's failing leg skips the
merge and the infra dispatch for the whole run**. That is pre-existing, but the
radius now includes 28 arm64 legs that have never actually run `docker buildx
build`. A first red run is far more likely to be one arm64 leg than a broken
pipeline. Base images, Dockerfiles and lockfiles were all checked for arm64
support and are clean, but the toolchain itself is unexercised until this merges.

## Not in this PR, deliberately

The kustomize component, `pull-images.sh`, `IMAGE_SOURCE`, and the pre-gate
changes are the next workstream. `make minikube-setup` still builds locally after
this merges.

## Verification

Five bash harnesses, all wired into `shell-syntax`: images-manifest 10, resolve
6, promote 7, plus workstream A's release-coordinates 24 and precommit 14. The
matrix expansion was simulated independently against GitHub's include-merge rules
(56 legs, zero field bleed) rather than trusted from a comment — `actionlint` does
not evaluate matrix expansion and passes either way.

## Known follow-ups

- The `shell-syntax` job is not in dev's required status checks, so a red harness
  does not block the merge button. Worth adding to ruleset 19597343.
- A partial promotion failure leaves some images tagged and others not; re-running
  is idempotent since `crane copy` overwrites.
- `scripts/tests/` holds 40 harnesses and CI wires 10.

https://claude.ai/code/session_01HqtLT5fQeqbc3m8DsUPgRT


jozer-rami:
@claude review this PR

PR #275, opened 2026-08-06
fix(minikube): make first-run deploy fail-loud and race-free

alfredolopez80:
> **Base branch: `dev`. This PR is built on `origin/dev` and is INDEPENDENT of PR #205.**
> It shares no commits with #205 and touches none of its readiness-decoupling code —
> these are preexisting first-run deploy-script bugs that merge on their own, without
> waiting on #205 review. #205 is a separate PR pending review.

## Summary

Three preexisting first-run bugs in the minikube deploy scripts, surfaced while
running the PR #205 T2 e2e rotation gate but **independent of #205** (none touch
the readiness-decoupling work; blame points to older commits). Each is a
fail-open or too-tight timeout that masked a real failure or starved a healthy
pod. All three now fail loud / fail fast.

These ship on their own branch off `dev` so they can merge without waiting on
#205 review.

### 1. `sync-auth-key.sh` — fail loud on gfsc rollout instead of masking the wait

After an auth-key drift the sync restarts the gfsc writer+reader, then waited on
a single label-wide `rollout status deployment -l <selector> ... --timeout=90s || true`.

- **Root cause:** the 90s budget was *shared* across both deployments, and the
  result was swallowed by `|| true`. The writer's init-container `chown` consumed
  the budget, starved the reader (which never reached Ready), and `|| true` hid
  it. The failure then resurfaced 60–150s later in `verify-gfs`, killing Step 8 of
  `full-setup` with no local signal of the real cause.
- **Fix:** wait per-deployment (writer AND reader) with an independent 180s budget
  each and no `|| true`. A genuinely stuck gfsc now fails *here*, named, instead
  of minutes downstream.
- **Blame:** the `|| true` came in with the initial squashed commit `5d663a49`
  (Jose Ramirez).

### 2. `verify-gfs.sh` — raise verify rollout timeout default 60s → 240s

- **Root cause:** on a cold cluster a freshly-restarted gfsc-reader needs ~145s to
  reach Ready; the 60s default killed the verify prematurely.
- **Fix:** default `VERIFY_ROLLOUT_TIMEOUT` to `240s` — still bounded, still
  fail-loud, still overridable via the env var. Belt-and-suspenders alongside
  fix #1. Empirically confirmed during the #205 T2 run: `VERIFY_ROLLOUT_TIMEOUT=240s`
  cleared the race.

### 3. `full-setup.sh` — fail the e2e seed loudly instead of exiting 0 on partial fixtures

- **Root cause:** under `SEED_PROFILE=minimal` a failed test-user seed aborts
  (`err` + `exit 1`), but under `e2e` it only `warn`ed and let `full-setup` exit 0
  over a failed Step 10. The e2e gates depend on those fixtures, so a partial seed
  shipped broken journeys that only surfaced later in an unrelated gate.
- **Fix:** make the e2e branch fail loud too (`err` + `exit 1`). Setup is
  idempotent, so re-running after the underlying fix recovers cleanly
  (`full-setup.sh --seed-profile=e2e --skip-build --keep-db`).
- **Blame:** the fail-open `warn` came in with `c2b8a7247` (Jose Ramirez, 2026-07-17).

> ⚠️ **Behavior change:** `full-setup` now exits non-zero on a partial e2e seed.
> Lanes that previously tolerated partial e2e fixtures (silently green over a
> failed Step 10) will now fail the setup. This is intentional — a partial e2e
> seed is a broken install, not a warning.

## Validation

- `bash -n` passes on all three scripts.
- `shellcheck` clean on the changed lines (only preexisting `SC1091`
  "not following source" infos remain, on unrelated `source` lines).
- Fix #2 was validated empirically in the #205 T2 run.
- Fix #1 is validated statically here; a clean from-zero deploy to confirm it
  removes the verify-gfs race is a recommended follow-up before merge.

## Scope notes

- Built on `origin/dev`, **not** on the #205 branch — independent, no shared commits.
- Not included: the `seed-e2e-data.sh` `provider` fix (already in `dev`) and the
  relative-marker concern (not a bug in the public-repo lane where CWD is the
  repo root).

### DIFF
--- a/scripts/minikube/full-setup.sh
+++ b/scripts/minikube/full-setup.sh
@@ -1083,6 +1083,11 @@ step_header 9 $TOTAL_STEPS "Seed Registry Catalog"
 # else still works.
 EVENFIRE_DIR_DEFAULT="$(cd "${PROJECT_DIR}/.." && pwd)/evenfire-registry"
 EVENFIRE_DIR="${EVENFIRE_REGISTRY_DIR:-${EVENFIRE_DIR_DEFAULT}}"
+# Track whether the catalog was actually seeded so the Step 12 summary tells the
+# truth: the three skip branches below are legitimate (sibling repo absent in this
+# distribution — non-fatal by design), so this stays a warn, not an abort, and the
+# summary ✓ is gated on this flag instead of printed unconditionally.
+REGISTRY_SEEDED=false
 if [[ ! -d "${EVENFIRE_DIR}" ]]; then
   warn "evenfire-registry not found at ${EVENFIRE_DIR} — skipping registry seed"
 elif ! $KC get deployment registry-api -n registry &>/dev/null; then
@@ -1093,6 +1098,7 @@ else
   log "Running evenfire-registry minikube-seed target..."
   if (cd "${EVENFIRE_DIR}" && make minikube-seed 2>&1 | tail -25); then
     ok "Registry catalog seeded"
+    REGISTRY_SEEDED=true
   else
     warn "Registry seed encountered errors — check output above"
   fi
@@ -1177,7 +1183,12 @@ else
      bash "${SCRIPT_DIR}/seed-workflow-triggers-test-data.sh" 2>&1 | tail -20; then
     ok "Workflow-trigger E2E recipes seeded"
   else
-    warn "Workflow-trigger E2E seed encountered errors — desktop/control-ui workflow E2E may fail until fixed"
+    # Same fail-loud rationale as Step 10 (P5): under e2e a partial fixture seed
+    # ships broken journeys that only surface in a later gate, and the summary
+    # below prints an unconditional ✓ for this step — so a warn-and-continue here
+    # would leave a lying green checkmark. Abort; setup is idempotent.
+    err "Workflow-trigger E2E seed failed under SEED_PROFILE=e2e — aborting. Downstream desktop/control-ui workflow E2E gates depend on these fixtures. Fix the error above and re-run setup (idempotent: --skip-build --keep-db)."
+    exit 1
   fi
 
   if [ "$SKIP_UIS" = true ]; then
@@ -1188,7 +1199,11 @@ else
        bash "${SCRIPT_DIR}/seed-sandbox-ui-test-data.sh" 2>&1 | tail -25; then
       ok "Sandbox-ui local test app seeded"
     else
-      warn "Sandbox-ui test app seed encountered errors — Desktop Apps may not list local sandbox-ui fixtures"
+      # Fail loud like the workflow-trigger seed above (P5-consistent): an e2e
+      # install with incomplete Desktop Apps fixtures is broken, not a warning.
+      # The legitimate opt-out is --skip-uis (handled above), not a failed seed.
+      err "Sandbox-ui test app seed failed under SEED_PROFILE=e2e — aborting. Desktop Apps validation fixtures would be incomplete. Fix the error above and re-run setup (idempotent: --skip-build --keep-db)."
+      exit 1
     fi
   fi
 fi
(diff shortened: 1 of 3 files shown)


alfredolopez80:
@claude please do a complete review of this PR for final sign-off before merge.

**Context.** Three preexisting first-run minikube deploy bugs, all made fail-loud:
- **P1** `sync-auth-key.sh` — replace the shared label-wide `rollout status ... --timeout=90s || true` with a per-deployment wait (independent 180s budget each, no `|| true`).
- **P2** `verify-gfs.sh` — raise `VERIFY_ROLLOUT_TIMEOUT` default 60s → 240s (still bounded, still `|| fail`).
- **P5** `full-setup.sh` — the e2e seed-failure branch now `err` + `exit 1` instead of `warn` + exit 0. **Behavior change:** setup now exits non-zero on a partial e2e seed.

Plus a code-review cleanup commit `db95e5a5` (no success-path behavior change): removes the now-dead `SEED_USER_OK` machinery + four unreachable `✗` summary branches that P5 orphaned, rewrites a stale comment that still referenced the deleted "e2e soft-fail path", and de-races `sync-auth-key.sh` by capturing the gfsc deployment list once instead of re-querying inside the loop (closes a TOCTOU window).

**Please focus on:**
1. Correctness of the fail-loud changes under `set -euo pipefail` (does P1 actually abort on a stuck rollout, given the invocation context of `sync_gfs`).
2. **Blast radius** of the P5 behavior change on deploy / local-setup / CI lanes — any lane that previously tolerated a partial e2e seed (silently green) will now fail the setup. Is that acceptable everywhere `full-setup --seed-profile=e2e` is invoked?
3. Any shell pitfalls (word-splitting on `for d in $gfs_deploys`, empty-list handling, `set -e` neutralization).

Validation so far: `bash -n` clean; `shellcheck 0.11.0` clean except preexisting `SC1091` infos. Thanks!

alfredolopez80:
@claude please re-review — the branch has moved well past your last review (which covered `db95e5a`). Current head is `79d3f7e`. New since then:

**`752067d5`** — align the P1 gfsc rollout wait `180s → 240s` to match `verify-gfs.sh` and `deploy/scripts/provision-gfs-runtime.sh` (a freshly-restarted gfsc-reader needs ~145s cold).

**`95a22fc0`** — adversarial-review follow-up closing the remaining fail-open paths you flagged (and two you did not):
- `sync-auth-key.sh:137` — dropped `|| true` / `2>/dev/null` on the gfsc deployment-list capture. A label-selector LIST returns exit 0 + empty stdout on zero matches (legit first-run → else branch), so only a genuine API error now aborts. Closes your Note 1.
- `full-setup.sh` Step 11 **workflow-trigger** seed — `warn` → `err` + `exit 1` under e2e (P5-consistent), which makes the previously-unconditional `✓ Workflow-trigger E2E recipes seeded` summary line truthful. Closes your Note 2.
- `full-setup.sh` Step 11 **sandbox-ui** seed — same `err` + `exit 1` under e2e, preserving the legitimate `--skip-uis` opt-out (a parallel you did not flag).
- `full-setup.sh` Step 9 **Registry catalog** — the summary `✓` is now gated on a `REGISTRY_SEEDED` flag (NOT an abort — the sibling-repo-absent skips are legitimate), so it stops printing a false ✓ (a parallel you did not flag).

**`79d3f7e`** — merge of `dev` (control-api changes, unrelated to the minikube scripts).

**Please focus on:**
1. The `sync-auth-key.sh:137` change — confirm dropping `|| true` cannot abort a legitimate zero-deployment first-run (LIST semantics: exit 0 + empty), and only fails on a real API error.
2. The two new `exit 1` hard-aborts in Step 11 — consistent with P5? Is `--skip-uis` still a safe warn-and-continue?
3. The `REGISTRY_SEEDED` gating — any path where the flag is misread, or where the summary now lies in the other direction?
4. Anything the previous review missed on blast radius (note: `sync-auth-key.sh` also runs in the private deploy path via `provision-gfs-runtime.sh`; `verify-gfs.sh` and `full-setup.sh` are local-minikube-only).

Validation: `bash -n` clean; `shellcheck 0.11.0` clean except preexisting `SC1091`. Thanks!

PR #276, opened 2026-08-06
feat(minikube): pull published images instead of building them

jozer-rami:
**Draft — tasks 7-10 of 10 still to land.** Opened early to get CI signal on the parts that cannot be tested locally.

Workstream D of the image program: a new contributor runs `make minikube-setup` and **pulls published images instead of building them**. Follows C (#272), which published the multi-arch images this consumes.

## What CI is being asked to prove

Every implementer verified on local kubectl v1.34.1 / kustomize v5.7.1. **CI pins kubectl v1.31.3 (kustomize v5.4.2)**, and the design's claim that kustomize *components* build at that pinned version has been asserted but never exercised in CI. That is the single most valuable signal here. If `deploy-manifest-check` goes red, the pin needs bumping and the plan's premise was wrong.

## Landed so far

| Task | What |
| --- | --- |
| 1 | `deploy/components/ghcr-images/` — the image-coordinate rewrite, plus its own byte-identical FieldSpec copy (kustomize forbids referencing a `configurations:` file outside the component) |
| 2 | `minikube-ghcr` / `minikube-no-uis-ghcr` overlays + render guard |
| 3 | `scripts/minikube/pull-images.sh` |
| 4 | Mode-aware `build-images.sh --verify-only` |
| 5 | `full-setup.sh` — `IMAGE_SOURCE=ghcr` default, `MINIKUBE_IMAGE_TAG` override, banner |
| 6 | Makefile targets |

Still to come: pre-gate shadow builds, CI wiring, the release-coordinate pointer row, and the docs.

## Behaviour change with real blast radius

`IMAGE_SOURCE` defaults to **ghcr**, so `make minikube-setup` stops building. `IMAGE_SOURCE=local` restores today's behaviour exactly, and `make minikube-setup-local` is the named path.

`MINIKUBE_IMAGE_TAG=latest` is required until `v0.6.0` exists — the component pins a tag that has not been cut yet, which is inherent to pinning in-tree and is why the puller fails with a message naming the tag and the override rather than a raw `MANIFEST_UNKNOWN`.

## What is NOT proven

- **No live cluster run.** Every harness stubs docker/minikube/kubectl. The Apple Silicon arm64 path — the reason this workstream exists — has not been exercised end to end. Task 3 did pull and alias real images against `clerum-test`, but a full fresh-cluster setup has not run.
- `shell-syntax` is still not a required check on `dev`, so the 40+ harnesses guarding this cannot block a merge.

## Defects found in the plan while implementing

Each surfaced only because implementers mutation-tested instead of trusting the text:

- `"$MINIKUBE_MULTI_NODE"` unbound under `set -u` — would have aborted the new ghcr default for every developer on first run
- an unguarded `rm -rf "$(dirname …)"` in a mandated test that could delete the repo's parent
- a regex char class missing `:`, making a correct implementation read as failing
- the design doc's third render-guard exception does not exist as an image field; the real one is `CONTEXT_MAPPER_ALLOWED_IMAGE_PREFIXES`
- pre-existing on `dev`: `((fail_count++))` aborting the verify loop under `set -e`, so only the first missing image was ever reported

https://claude.ai/code/session_01HqtLT5fQeqbc3m8DsUPgRT


jozer-rami:
@claude deep review

PR #277, opened 2026-08-06
fix(control-ui): let operators set a connector's missing credentials instead of dead-ending on a 404

jozer-rami:
## Problem

Installing an MCP connector from the registry without filling in its credentials is a supported path ("install pending"), but the two halves of it disagree:

- `registry.ts:1123-1128` sets `spec.envSecret` on the McpServer whenever the plugin declares required credentials.
- `registry.ts:1216-1218` creates the backing Secret only when credentials were actually supplied.

An empty credential form therefore produces a connector pointing at a Secret that was never created. HCC reports it correctly and the connector never starts. Two connectors on `clerum-dev` are in this state today: `evenfire-mcp-web-research` (`SEARCH_API_KEY`) and `evenfire-mcp-github-clerum` (`GITHUB_TOKEN`).

The recovery path was a dead end. The "Missing Secret" badge (`McpServerTable.tsx:54`) routes the operator to the connector edit page, whose only credential control calls `PUT /admin/mcp-secrets/:name` — a route that deliberately 404s when the Secret does not exist, because creating one is POST's job:

```
Rotation failed: 404 - Secret "evenfire-mcp-web-research-credentials" not found in mcp-server
```

The UI flagged the problem, sent the operator to a screen, and that screen offered only an operation that could not succeed. The workaround was to leave and hand-type the exact Secret name under Secrets → New.

## What this does

The connector screen now resolves one of three surfaces and acts accordingly:

| Surface | When | Behavior |
|---|---|---|
| `set` | `SecretResolved=False` / `reason=SecretNotFound`, managed connector | Requires **every** declared key, POSTs to create |
| `rotate` | anything else | Unchanged — today's merge-patch rotation |
| `recipe-owned` | same missing-Secret condition on a `managed:false` connector | Explanation pointing at the recipe secret surface, no form |

Three design points worth calling out, each established by an adversarial review of the spec:

**Matching on `reason` is load-bearing.** `SecretMissingKey` means the Secret exists but lacks a key, which the existing PUT merge-patch already handles. Only `SecretNotFound` changes surface. Matching just type+status would push the missing-key case into a create that hits AlreadyExists.

**The create retry is deliberately not status-gated.** control-api's global error handler now forwards a sanitized Kubernetes 4xx, so a duplicate create can reach the browser as **409** rather than a bare 500. The client still retries once as PUT on *any* POST failure (set mode already holds every key, so the merge is complete). Gating the retry on 409 would strand operators on older control-api builds that still collapse AlreadyExists to 500.

**The rotate direction is deliberately asymmetric.** A 404 there does *not* fall back to POST: rotate mode requires only one value, so creating from partial data would yield a Secret missing a declared key — trading one broken state for another while the screen reported success. It flips to set mode and demands every key instead.

**Ownership is derived from `spec.managed` alone**, never from `status.conditions`. A `managed:false` connector with absent, `Unknown`, stale `True`, or `SecretMissingKey` status has no `SecretNotFound` condition, so any condition-coupled ownership check would silently disengage exactly where it is needed. `mode` can never resolve to `set` while recipe-owned, and a late PUT 404 lands on explanation-only.

**When both the create and its retry fail**, a PUT `404` keeps the POST error (the 404 only means "still absent" — noise); any other PUT error is surfaced instead, because it describes the current state more precisely. So POST 500 + PUT 409 shows the recipe-owned message including the `/admin/recipe-secrets` pointer.

**The authoritative `SecretResolved` condition** is selected by strictly-validated `lastTransitionTime` — RFC3339 syntax *plus* a calendar round-trip, since `new Date('2026-02-30T…')` silently normalizes into March. Missing, empty, malformed and invalid-calendar timestamps share one "oldest" rank; ties resolve to last-in-array; the caller's array is never mutated.

**Create is audited.** `POST /admin/mcp-secrets` emits `mcp-secret-created` after a successful write and `mcp-secret-create-failed` when the write throws. The record carries the operation, Secret name, namespace, and sorted key names only — never credential values. Kubernetes ApiException messages are sanitized so apiserver headers cannot land in the log.

## Scope

control-ui, plus a small control-api audit on `POST /admin/mcp-secrets` (R3-M2). No HCC, CRD, or RBAC changes.

## Testing

- CI `Build (control-ui)` now runs **vitest then `next build`**. Latest run on this branch: **173 files / 1910 tests** passed (1 skipped).
- `tsc --noEmit` is green; `next build` remains the production typecheck.
- control-api create audit: `routes.mcpSecrets.test.ts` covers success, failed write, validation (no audit), and ApiException header sanitization.
- Load-bearing tests are **mutation-verified**, not just observed green. Each of these was individually confirmed to turn the intended test red: reverting the retry; the mode derivation; the 404 predicate; each resolver clause; coupling ownership to `status.conditions`; dropping ownership entirely; gating the retry on `status === 500`, on a non-500, or on `status === undefined`; inverting the dual-error rule; replacing strict timestamp validation with a bare `new Date()`; reverting the resolver to `.some()`; flipping the tie-break; sorting the input array in place; and focusing the first field rather than the first *invalid* field.
- Page-level tests drive the real `EditMcpServerPage` through a mocked `getMcpServer` — nothing hand-injects `surface` or `recipeOwned` — so the page → resolver → component wiring is genuinely covered.
- Condition fixtures come from one shared builder matching the HCC producer byte-for-byte (`status:'True', reason:'SecretFound'`); deliberately impossible fixtures are labelled as such.

## Known follow-ups (not in this PR)

Two pre-existing control-api defects found while designing this, both reachable from Secrets → New today and neither introduced here:

1. `POST /admin/mcp-secrets` has no WorkflowRecipe-ownership guard, while the PUT route does. The `recipe-owned` surface added here reduces how easily this is reached but is a UI gate, not a security boundary.
2. `POST /admin/mcp-secrets` stamps **no provenance labels**. The Secret created during live testing has `labels: None`, whereas the registry install path stamps `clerum.io/managed-by: control-api`. That compounds follow-up 1: an unlabeled Secret is indistinguishable from a foreign one.

A duplicate create can now reach the browser as a sanitized Kubernetes **409** via the global error handler (FIX-B on `dev`); the old "always a bare 500" mapping is no longer current. Create is audited (`mcp-secret-created` / `mcp-secret-create-failed`).

Also open: `InstallFromRegistryResponse` does not model `pendingCredentials`, so the install flow still leaves a pending connector to be discovered later via a red badge rather than prompting up front.

Pre-existing UI papercut, unrelated to this change: rotating a credential to the *same* value writes nothing, so HCC computes an identical `credentials-revision`, no rollout fires, and `DeploymentReady` never transitions. The screen therefore spins its full 180s budget and reports "the rollout did not finish" for what was a harmless no-op — it cannot distinguish "nothing needed to change" from "stuck".

Two infrastructure gaps found while verifying this end to end, filed separately against `keyper-labs/evenfire-infra`: **#35** (RBAC changes never reach `clerum-dev` — the deploy strips RBAC and the out-of-band sync is invoked by no workflow; `secrets: patch` from #223 was missing for five days, so every rotation on dev 500'd) and **#36** (`bootstrap-rbac.sh` classifies `clerum-dev` as prod, training reflexive `CONFIRM=yes`).


josue-tech-94 (changes requested):
## Decision

**Decision:** Blocked

**Why:** Five in-scope findings must be fixed before merge: one ownership-sensitive recovery bug,
two reproduced false-green test gaps, one misleading retry-error path, and one stale-condition
resolution bug. The full test suite and production build pass, but the current tests would remain
green after breaking two load-bearing PR invariants.

**Author next step:** Keep connector ownership available through the late PUT-404 transition, then
add producer-backed page and retry tests before addressing the remaining error and condition cases.

## Review summary

| Area | Result | Notes |
| --- | --- | --- |
| Correctness | Issues found | Managed-false 404 recovery, retry errors, and duplicate conditions are unsafe. |
| Security | Issues found | `R1-H1` crosses the recipe-owned UI boundary; no privilege escalation was found. |
| Contracts | Issues found | Real POST 500, PUT 404, and PUT 409 semantics expose `R1-H1` and `R1-M1`. |
| Tests | Gaps found | Two load-bearing mutations remain green (`R1-H2`, `R1-H3`). |
| PR scope/body | Needs update | Eight files in one service is within D1; behavior/testing claims need qualification. |

## Review context

- **Round:** 1. No `pr-277` ledger, `branch-feat-connector-set-credentials` ledger, or legacy
  report existed, so no history was reset and no collateral range applies.
- **Reviewed head:** `20d34cf23d1b6e1c1e81b6f7bb883e9f9274d4e8`.
- **Base:** `dev` at `a223a8a07a6c4fefd99ca087c27d3541899dbf7b`.
- **Merge base:** `a223a8a07a6c4fefd99ca087c27d3541899dbf7b`.
- **Scope:** 8 changed files, 632 additions, 38 deletions, control-ui only.
- **GitHub state:** open, not draft, review required; no issue comments, inline comments, submitted
  reviews, latest reviews, or review threads were present when resolved.
- **Checks:** all resolved GitHub checks were successful, including Build (control-ui) and CodeQL.

## Action items

| Priority | ID | Severity | Action | Owner decision |
| --- | --- | --- | --- | --- |
| Must fix | `R1-H1` | High | Prevent managed-false PUT 404 from ever exposing POST/set mode. | None |
| Must fix | `R1-H2` | High | Test real page wiring with producer-backed multi-condition status. | None |
| Must fix | `R1-H3` | High | Prove retry remains status-agnostic, including a future POST 409. | None |
| Must fix | `R1-M1` | Medium | Surface the actionable retry error instead of always discarding it. | None |
| Must fix | `R1-M2` | Medium | Resolve one authoritative SecretResolved condition under duplicates. | None |
| Should fix | `R1-L1` | Low | Expose set-mode requiredness and validation to assistive technology. | May defer |

## Must fix before merge

### R1-H1 — Managed-false 404 recovery exposes the managed Secret form

- **Severity:** High
- **Evidence:** `control-ui/components/UpdateConnectorCredentials/index.tsx:249`
- **What is wrong:** For absent, Unknown, or stale non-missing status, a `managed:false` connector
  starts in rotate mode. A real PUT 404 then sets `recreateRequired` without retaining ownership,
  which exposes set mode and POST even though recipe-owned creation does not belong on this screen.
- **User/operator impact:** An operator can create an unlabeled Secret outside the WorkflowRecipe
  flow, collide with later recipe provisioning, and wait for an HCC deployment that managed-false
  connectors never receive.
- **Why this blocks merge:** The PR introduces this POST reachability and violates its
  explanation-only ownership invariant on a realistic initial-status race.
- **Required fix:** Pass ownership or an explicit `canCreate` invariant independently of the
  observed condition. A managed-false PUT 404 must transition to explanation-only behavior and
  never enable POST.
- **Required test:** Render the real page for managed-false connectors with absent, Unknown, stale
  True, and exact missing status; make PUT return the real 404; assert explanation, no set form, and
  no POST after partial input.
- **Confidence:** code-verified

### R1-H2 — Mode tests bypass the producer and page wiring

- **Severity:** High
- **Evidence:** `control-ui/components/__tests__/resolveCredentialSurface.test.ts:5`
- **What is wrong:** Resolver tests use isolated, hand-written single conditions, while component
  tests inject `surface` directly. They never prove that the edit page maps a real McpServer response
  into the correct UI.
- **User/operator impact:** Breaking condition selection or removing the page's surface wiring can
  ship with a fully green changed-test suite, including ownership-sensitive regressions.
- **Why this blocks merge:** Two realistic mutations were reproduced green: first-condition-only
  resolution passed 8/8, and removing the page's `surface` prop passed both changed suites 36/36.
  These are false-green tests over load-bearing behavior.
- **Required fix:** Add producer-backed multi-condition fixtures and page-level tests through
  `getMcpServer`, keeping malformed defensive cases labeled separately.
- **Required test:** Cover managed missing -> set, SecretMissingKey -> rotate, unmanaged missing ->
  explanation, unrelated condition reordering, and the late managed-false PUT-404 path.
- **Confidence:** reproduced

### R1-H3 — Retry tests do not prove status-agnostic fallback

- **Severity:** High
- **Evidence:** `control-ui/components/__tests__/UpdateConnectorCredentials.test.tsx:816`
- **What is wrong:** Retry success is tested only with POST status 500. The both-fail 400 test does
  not assert that PUT was attempted, so it does not protect the unconditional retry invariant.
- **User/operator impact:** A future control-api that correctly returns 409 AlreadyExists could
  strand operators in set mode if the client later gates fallback on 500.
- **Why this blocks merge:** Gating retry on `postError.status === 500` left all 28 component tests
  green, contradicting a load-bearing PR-body invariant and the requested version-skew safety.
- **Required fix:** Strengthen tests rather than changing the currently correct unconditional retry.
- **Required test:** POST 409 plus PUT success must reach user-visible set progress/success; a
  non-500 POST failure must prove one PUT attempt; both-fail cases must verify useful error choice.
- **Confidence:** reproduced

### R1-M1 — A failed retry hides the actionable error

- **Severity:** Medium
- **Evidence:** `control-ui/components/UpdateConnectorCredentials/index.tsx:358`
- **What is wrong:** When POST and fallback PUT both fail, the code always throws the original POST
  error. It discards the retry error even when that error describes the current state more precisely.
- **User/operator impact:** POST AlreadyExists currently appears as a generic 500; if PUT then
  returns the real recipe-owned 409, the UI hides the instruction to use recipe secrets and shows
  only an opaque failure.
- **Why this blocks merge:** This is the primary recovery path's error contract, and the current test
  pins it with a POST-400 fixture the UI cannot produce because all submitted values are strings.
- **Required fix:** Preserve both failures or choose semantically. At minimum, keep POST when PUT is
  404, but surface actionable ownership, authorization, validation, or server errors from PUT.
- **Required test:** Cover POST 500 + PUT 409, POST 500 + PUT 404, and future POST 409 + successful PUT
  with user-visible assertions.
- **Confidence:** code-verified

### R1-M2 — A stale duplicate SecretNotFound condition wins

- **Severity:** Medium
- **Evidence:** `control-ui/components/UpdateConnectorCredentials/resolveCredentialSurface.ts:28`
- **What is wrong:** `.some()` treats any duplicate `SecretResolved=False/SecretNotFound` entry as
  authoritative, even when a newer duplicate says True/SecretFound or False/SecretMissingKey.
- **User/operator impact:** The UI can require every credential and POST to an existing Secret; for
  managed-false connectors, it can also hide a valid rotate surface behind recipe-owned copy.
- **Why this blocks merge:** The CRD schema is an ordinary conditions array and does not enforce
  uniqueness. HCC normally rewrites a type uniquely, but legacy/manual duplicates can survive,
  including its identical-condition no-op path.
- **Required fix:** Select one authoritative SecretResolved condition by transition time with a
  deterministic tie-breaker; ambiguity must not let stale absence proof win.
- **Required test:** Contradictory duplicates with newer True, newer SecretMissingKey, and newer
  SecretNotFound, plus unrelated reordering and equal/malformed timestamps.
- **Confidence:** code-verified

## What looks good

- **Security:** Credential values remain masked, are sent only in JSON bodies, are cleared after a
  successful write, and are not placed in URLs, confirmation text, banners, toasts, logs, or API
  responses. Server-side global-admin authentication and namespace enforcement remain intact.
- **Contracts:** The real POST route currently turns Kubernetes AlreadyExists into production 500;
  PUT returns exact 404 for absence and 409 for recipe ownership. The current unconditional fallback
  remains safe if POST later returns a proper 409.
- **Tests/CI:** The full control-ui suite passes 129 files / 1,288 tests. The production build passes.
  Claimed retry-removal, latch, widened-404, and resolver-clause mutations turn intended assertions
  red. Regression tests copied to `68c4b88` and `27a5332` fail on the intended assertions.
- **Behavior:** Exact single-condition resolution handles SecretNotFound, SecretMissingKey, Unknown,
  absent conditions, and unrelated condition types correctly. Rotate 404 never POSTs partial data;
  non-404 errors retain rotate mode. Busy controls disable during writes/polling, and the page shows
  a skeleton until its initial server response is complete.

## Detailed findings

### High

#### R1-H1 — Managed-false 404 recovery exposes the managed Secret form

- **Priority:** Must fix
- **Severity:** High
- **Evidence:** `control-ui/components/UpdateConnectorCredentials/index.tsx:249`
- **Confidence:** code-verified
- **Problem:** Late 404 recovery loses connector ownership and makes POST reachable.
- **Scenario:** A new managed-false connector has no status yet; rotate PUT 404s; set mode appears.
- **Impact:** Credentials are created outside the recipe flow and rollout status cannot converge.
- **Fix:** Preserve ownership/can-create state through all transitions.
- **Test needed:** Real-page managed-false absent/Unknown/stale status plus PUT 404.
- **Notes:** The generic POST route predates this PR, but connector-edit reachability does not.

#### R1-H2 — Mode tests bypass the producer and page wiring

- **Priority:** Must fix
- **Severity:** High
- **Evidence:** `control-ui/components/__tests__/UpdateConnectorCredentials.test.tsx:82`
- **Confidence:** reproduced
- **Problem:** Changed tests can pass without the real page-to-resolver-to-component connection.
- **Scenario:** Remove the surface prop or use only the first condition; all changed tests stay green.
- **Impact:** The key surface-selection regression reaches CI undetected.
- **Fix:** Exercise the edit page using producer-backed McpServer output.
- **Test needed:** Page integration tests plus multi-condition resolver cases.
- **Notes:** Duplicates `R1-H1` and `R1-M2` only as test evidence, not as another production defect.

#### R1-H3 — Retry tests do not prove status-agnostic fallback

- **Priority:** Must fix
- **Severity:** High
- **Evidence:** `control-ui/components/__tests__/UpdateConnectorCredentials.test.tsx:816`
- **Confidence:** reproduced
- **Problem:** A 500-only retry test does not guard future 409 compatibility.
- **Scenario:** Gate fallback on POST 500; every component test still passes.
- **Impact:** A corrected server response can break the recovery flow unnoticed.
- **Fix:** Add non-500 retry-success and retry-attempt assertions.
- **Test needed:** POST 409 -> PUT success and POST 400 -> one PUT attempt.
- **Notes:** Current production code is status-agnostic; the defect is the false-green safeguard.

### Medium

#### R1-M1 — A failed retry hides the actionable error

- **Priority:** Must fix
- **Severity:** Medium
- **Evidence:** `control-ui/components/UpdateConnectorCredentials/index.tsx:358`
- **Confidence:** code-verified
- **Problem:** The original POST error always replaces the retry error.
- **Scenario:** POST returns opaque 500 AlreadyExists; PUT returns actionable recipe-owned 409.
- **Impact:** The operator loses the only useful recovery instruction.
- **Fix:** Preserve both or select the error that best describes the final state.
- **Test needed:** Producer-shaped 500/409 and 500/404 combinations.
- **Notes:** The existing POST-400 fixture is not reachable from this UI's string-only payload.

#### R1-M2 — A stale duplicate SecretNotFound condition wins

- **Priority:** Must fix
- **Severity:** Medium
- **Evidence:** `control-ui/components/UpdateConnectorCredentials/resolveCredentialSurface.ts:28`
- **Confidence:** code-verified
- **Problem:** Any matching duplicate proves absence regardless of recency.
- **Scenario:** Old False/SecretNotFound coexists with newer True/SecretFound.
- **Impact:** Existing Secrets are handled as missing and partial rotation is unnecessarily blocked.
- **Fix:** Resolve the authoritative condition deterministically.
- **Test needed:** Contradictory duplicates, timestamps, and order permutations.
- **Notes:** HCC normally deduplicates on writes, but the API schema does not enforce uniqueness.

### Low

#### R1-L1 — Required fields lack programmatic required semantics

- **Priority:** Should fix
- **Severity:** Low
- **Evidence:** `control-ui/components/UpdateConnectorCredentials/index.tsx:479`
- **Confidence:** code-verified
- **Problem:** Set mode communicates requiredness only through placeholder text and an aggregate alert.
- **Scenario:** A screen-reader user cannot identify required or invalid inputs before/after submit.
- **Impact:** The all-keys recovery is harder to understand and repair with assistive technology.
- **Fix:** Use `Field required`, native/ARIA required state, associated errors, and first-invalid focus.
- **Test needed:** Required semantics only in set mode and focus/error association after partial submit.
- **Notes:** Existing Field/TextInput primitives already support the needed attributes.

jozer-rami:
@josue-tech-94 thanks — this was a genuinely useful review. All five must-fix items and the should-fix are addressed. Two of them (`R1-H1`, `R1-M1`) were findings my own pre-PR review had surfaced and I downgraded to "minor, self-limiting"; you were right and I wasn't, so I've said why below rather than quietly fixing them.

Head is now `200e0019`. Full control-ui suite **1321 tests green** (was 1288), production build green.

## R1-H1 — managed-false 404 recovery exposed the create form → `3c58b794`

You found the case I missed. My earlier reasoning assumed the ownership guard had already engaged — but for a `managed:false` connector with **absent/Unknown/stale** status there is no `SecretNotFound` condition to observe, so it resolves to `rotate` and the guard never runs at all. The late PUT 404 then flipped it into set mode.

Ownership is now derived independently of the observed condition:

- `isRecipeOwned(spec)` → `spec?.managed === false`, threaded from the page as a `recipeOwned` prop (`edit/page.tsx:205`), documented in `types.ts` as explicitly **not** derivable from `surface`
- `mode` can never resolve to `'set'` when `recipeOwned` (`index.tsx:285`)
- a late-discovery early return covers `recipeOwned && secretMissing` (`index.tsx:309`), so the create form is not even mountable in that state

`spec.managed` is safe to test for exactly `false`: the CRD declares `managed: {type: boolean, default: true}`, so the API server always persists it and absence can never mean recipe-owned.

**Proof it was real:** on the reverted code, a test driving the whole operator journey (404 → fill both keys → "Set & start") actually calls `createMcpSecret` against the recipe-owned Secret name — `expected "vi.fn()" to not be called at all, but actually been called 1 times`. That reproduces the defect end to end, not just a copy assertion. Reverting turns 8 tests red across three files.

One disclosure: the `&& !recipeOwned` clause at `index.tsx:285` is an *equivalent mutant* — reverting it alone leaves the suite green, because the early return already makes the state unrenderable. They're deliberate belt-and-braces; the load-bearing half (the early return) is independently pinned by 8 tests.

## R1-H2 — tests bypassed the producer and page wiring → `3c58b794`

Both mutations you reproduced now fail. Added `EditMcpServerPageCredentials.test.tsx`, which drives the **real** `EditMcpServerPage` through a mocked `getMcpServer` — nothing hand-injects `surface` or `recipeOwned` — covering managed+missing → set, `SecretMissingKey` → rotate, unmanaged+missing → explanation, unrelated-condition reordering, and the late managed-false PUT-404 path. Resolver fixtures are now multi-condition.

- removing the page's `surface` prop → RED (was 36/36 green)
- first-condition-only resolution → RED (was 8/8 green)

## R1-H3 — retry not provably status-agnostic → `519624cf`, `200e0019`

Took your instruction literally: **the retry logic is unchanged**, only the tests. Adding a status branch would resurrect exactly the dead code the design review killed (control-api collapses AlreadyExists to a bare 500 and cannot emit 409 today).

Added: POST 409 + PUT success reaching visible success (labelled **forward-compatibility**, so nobody reads it as current server behaviour); a non-500 POST failure asserting exactly one PUT attempt; and — beyond your ask — a POST rejection carrying **no `.status` at all** (transport failure), which kills the `status === undefined` gate that still survived. Gating on `status === 500` is now RED.

## R1-M1 — failed retry hid the actionable error → `519624cf`

Implemented your rule: a PUT **404** keeps the POST error (the 404 is just "still absent" — noise); any other PUT error surfaces instead, since it describes the current state more precisely. So POST 500 + PUT 409 now shows the recipe-owned message including the `/admin/recipe-secrets` pointer. Both combinations you listed are pinned; inverting the rule turns them red.

## R1-M2 — stale duplicate `SecretNotFound` won → `ff6f21b2`

`.some()` replaced with selection of one authoritative `SecretResolved` condition by `lastTransitionTime`, with a documented deterministic tie-break and defined handling for equal/malformed/absent timestamps — an unparseable timestamp cannot let stale absence win. Covers newer-True, newer-SecretMissingKey, newer-SecretNotFound, reordering. Reverting to `.some()` is RED.

Fixtures now use the producer's real strings (`status:'True', reason:'SecretFound'` per `reconciler.ts:1979-1986`), with malformed/adversarial fixtures separately labelled.

## R1-L1 — required semantics → `3be7e8c2`, `200e0019`

Set-mode inputs carry real required semantics, the validation error is programmatically associated with the offending inputs, and focus moves to the first invalid input.

Worth flagging: my first version of that focus test was itself a false-green. It filled the second key and left the first empty, so "first invalid" was indistinguishable from "first field". Hardened to cover both-empty (focus must land on the first) and first-filled/second-empty (focus must land on the **second**, proving it targets the first *invalid* rather than the first field). Focusing unconditionally on `keys[0]` is now RED. The production logic was correct all along — only the test was weak.

## Caveats

- `control-ui` is excluded from the CI test matrix (`ci-public.yml`, `control-ui-build` job), so its only CI gate is `next build`. Every mutation result above is **local** proof; the green checkmark on this PR does not cover these tests.
- `tsc --noEmit` still reports one error in `LlmDiscoveryPanel.test.tsx` — pre-existing, present at the reviewed head `20d34cf2`, and untouched by this branch.
- The R1-M2 fixture change is fidelity, not coverage: no fixture edit can go red on its own, it just stops the tests asserting against strings the producer never emits.

Separately, the create flow and the rotate flow are both now verified end to end against real data on `clerum-dev` (Secret created → connector `Ready=True` with the key in use; then a rotation → new resourceVersion, new `credentials-revision`, pod rolled, fresh `DeploymentReady`). Chasing that also turned up a dev-cluster RBAC drift unrelated to this PR — `deploy-clerum-dev.yml` strips RBAC and the out-of-band sync is invoked by no workflow — which I'm filing separately.

Ready for another look.

josue-tech-94:
@jozer-rami

**Decision:** Blocked

Two round-1 findings remain merge-blocking. `R1-H2` still permits a false-green page-wiring
regression that loses recipe ownership for non-empty status arrays. `R1-M2` still lets a malformed
calendar timestamp outrank a valid condition because JavaScript normalizes the invalid date. The
production ownership fix, retry behavior, dual-error selection, and accessibility behavior are
otherwise verified.

**Author next step:** Because four files have crossed F3, do not add another local patch directly.
Write a small mini-spec for the ownership contract and strict condition-authority policy, then add
producer-faithful page scenarios and strict timestamp parsing as one reviewed redesign.

## Existing finding status

| ID      | Status                       | Closing commit                     | Round-2 result                                                                             |
| ------- | ---------------------------- | ---------------------------------- | ------------------------------------------------------------------------------------------ |
| `R1-H1` | Closed                       | `3c58b794`                         | Ownership is independent of status and late 404 cannot expose POST.                        |
| `R1-H2` | **Open**                     | —                                  | A condition-dependent ownership mutation remains green; page fixtures also drift from HCC. |
| `R1-H3` | Closed                       | `519624cf`, hardened by `200e0019` | 500-only, non-500, 409, and statusless retry gates are killed.                             |
| `R1-M1` | Closed                       | `519624cf`, hardened by `200e0019` | PUT 404 keeps POST error; other retry errors are shown.                                    |
| `R1-M2` | **Open; re-triage required** | —                                  | Invalid calendar timestamps are treated as authoritative dates.                            |
| `R1-L1` | Closed                       | `3be7e8c2`, hardened by `200e0019` | Required, invalid, described-by, and first-invalid focus behavior pass.                    |

## Must fix before merge

### R1-H2 — Page coverage still permits ownership-sensitive false greens

- **Priority:** Must fix
- **Severity:** High
- **Evidence:**
  `control-ui/components/__tests__/EditMcpServerPageCredentials.test.tsx:60`, `:188`, `:208`,
  `:225`
- **Failure scenario:** Change the page's ownership mapping to pass `recipeOwned` only when
  `status.conditions` is empty. All 69 changed tests remain green. A `managed:false` connector with
  `Unknown`, stale `True`, or `SecretMissingKey` status starts in rotate mode; a real PUT 404 then
  exposes the all-keys set form, and a later submit can POST the operator's partial draft outside the
  recipe flow.
- **Impact:** The exact round-1 ownership regression can return while CI remains green, creating an
  unlabeled Secret for a recipe-owned connector and waiting on a rollout that HCC does not own.
- **Required fix:** Add page-level journeys for managed-false missing status, Unknown, stale True,
  and SecretMissingKey. Each must submit partial rotate data, receive a status-bearing 404, and land
  on explanation-only behavior with no credential form and no POST.
- **Required test:** Kill the reproduced condition-dependent ownership mutation. Replace the
  hand-written page fixture with an established producer-backed builder, or derive and verify every
  condition triple against the HCC producer. Correct the impossible clean reason
  `SecretResolved` to the producer's `SecretFound`.
- **Confidence:** reproduced

The current production implementation is correct: `page.tsx:194-205` passes ownership separately;
`index.tsx:276-285` prevents set mode while recipe-owned; and `index.tsx:298-317` renders the late
404 explanation before any form can remount. This finding remains open because its required
false-green protection is incomplete, not because the current head still executes the bad path.

### R1-M2 — Malformed timestamps can still make stale absence authoritative

- **Priority:** Must fix
- **Severity:** Medium
- **Evidence:** `control-ui/components/UpdateConnectorCredentials/resolveCredentialSurface.ts:65`
- **Failure scenario:** A duplicate `SecretNotFound` condition has
  `lastTransitionTime: 2026-02-30T00:00:00.000Z`, while a clean `SecretFound` condition has the valid
  `2026-02-28T23:59:59.000Z`. The documented policy says malformed timestamps rank oldest, but
  `new Date()` normalizes February 30 into March and the resolver returns `set`.
- **Impact:** The UI can require every key and reopen POST based on malformed legacy status even
  though the valid condition says the Secret exists.
- **Required fix:** Validate RFC3339 syntax and calendar validity strictly before comparing. Missing,
  empty, invalid-calendar, and otherwise malformed values must share the documented oldest rank.
- **Required test:** Assert the rendered credential surface for invalid calendar dates, missing
  timestamps, contradictory equal timestamps in both orders, identical duplicates, unrelated
  reordering, and input-array immutability.
- **Confidence:** reproduced

**Medium expiry:** `R1-M2` has remained open for two rounds. Before any round 3, explicitly escalate
it into the mini-spec/redesign or close it in writing with a justified contract change. It may not
silently carry forward.

## New round-2 finding (Non blocking)

### R2-L1 — The PR description contradicts the reviewed head

- **Priority:** Should fix
- **Severity:** Low
- **Evidence:** PR body design/testing sections compared with
  `control-ui/components/UpdateConnectorCredentials/index.tsx:417-431` and the verified test output
- **Failure scenario:** A maintainer relies on the body and preserves the old rule that always shows
  the POST error, or assumes the branch still contains 129 files / 1,288 tests and only the original
  mutation set.
- **Impact:** The merge contract documents behavior the code no longer implements and overstates
  the completeness of producer-backed and mutation coverage.
- **Required fix:** Replace the behavior and testing text with the copy-ready current-head text in
  this report, including the remaining blocked cases.
- **Required test:** Re-run and state the exact focused, full-suite, build, typecheck, and mutation
  results; do not carry counts from round 1.
- **Confidence:** reproduced
- **Regression of:** none; this is a PR-contract finding, not a behavior regression in the bounded
  fix diff.

## Verified positives

- **Credential surfaces:** Only authoritative `False/SecretNotFound` enters set; SecretMissingKey,
  Unknown, absent status, unrelated conditions, and non-authoritative duplicates stay rotate.
- **Ownership:** Exact missing status renders recipe explanation immediately; absent/stale status
  preserves `managed:false` through a late 404. No recipe-owned path can reach POST at current head.
- **Set mode:** Every declared key is required; POST is first; every POST failure retries PUT once;
  a future 409 and a statusless transport failure are both covered; no loop exists.
- **Dual failure:** PUT 404 preserves the create error; 409, 403, validation, and other retry errors
  surface the second operation's current-state diagnosis.
- **Rotate mode:** Existing merge-patch behavior is unchanged. A real 404 never POSTs partial data,
  changes to all-keys set mode only for managed connectors, and other statuses remain failures.
- **Accessibility/UI:** Set fields expose required and invalid state, share associated error text,
  and focus the first invalid key. Rotate fields are not required. Loading uses a skeleton; busy
  controls disable during writes/polling; errors and timeouts retain mode-correct copy.
- **Sensitive data:** Values remain password inputs and JSON request bodies, never URLs, logs,
  success copy, or server responses. Successful writes clear draft state.
- **Dependencies:** package.json and package-lock remain synchronized; the PR changes only the app
  version from 0.1.386 to 0.1.400 and introduces no dependency delta. A clean merge-tree with the
  newer dev PostCSS patch was produced.

## Security review

No new security vulnerability was found. Server-side `/admin` authentication and fixed namespace
enforcement are the authorization boundary; AuthGate and recipe-owned UI handling are safety/UX
controls. The base-existing POST ownership/linkage gap remains out of scope. A recipe-owned late 404
leaves the now-hidden partial value in React state until unmount; clearing it is reasonable data
minimization hardening, but the PR does not widen the base/round-1 retention window.

zach88:
Took a deep pass through this one, including verifying the three contract claims in the description against the real control-api and host-context-controller code. It holds up well: nothing blocks the merge, the state machine is consistent across every transition I traced, and the load-bearing `reason` matching lines up byte-for-byte with what the HCC actually writes. Only two optional nits from me, plus a couple of pre-existing things worth flagging (not caused by this PR).

## Optional (Low)

1. **[R1-L1] Test fixture uses a `reason` the producer never emits** — `control-ui/components/__tests__/resolveCredentialSurface.test.ts:57`. The "resolves cleanly" case uses `reason: 'SecretResolved'`, but the HCC never emits that string as a *reason*; it writes `reason: 'SecretFound'` for the resolved condition (`host-context-controller/src/reconciler.ts:1674,1742`). `SecretResolved` is the condition *type*, not the reason. It's harmless today (any reason ≠ `SecretNotFound` falls to `rotate`, and the case also varies `status:'True'`), but it's a hand-written fixture rather than one derived from the producer. If some future clause ever inspected the resolved-case reason, this would mask it. Trivial swap to `reason: 'SecretFound'`. Every load-bearing string in the suite (`SecretNotFound`, `SecretMissingKey`, `status:'False'`, `type:'SecretResolved'`) does match the producer and is mutation-pinned, so this is cosmetic.

2. **[R1-L2] No end-to-end test for the page→resolver wiring** — `control-ui/app/mcp-servers/[name]/edit/page.tsx:194`. `resolveCredentialSurface` and the `surface` prop are each unit-tested, but nothing asserts the page passes the right `status.conditions` and `spec` through to it. Low risk since both pieces are trivial and covered individually, just calling it out.

## Pre-existing (not introduced here, non-blocking)

Flagging these for visibility — they reproduce on the merge-base without this diff, and I confirmed this PR does not make them worse. Natural home is a follow-up issue, not this PR.

1. **`POST /admin/mcp-secrets` has no WorkflowRecipe-ownership guard** — `control-api/src/routes/admin/secrets.ts:293-339`. The POST only applies `enforceNamespace` + RFC1123 validation; no ownership guard (unlike what the PUT path assumes). An operator with control-ui access could create a Secret under any RFC1123 name in `mcpServersNamespace`, including one belonging to a recipe-owned connector whose Secret is missing. Reproduced: identical on merge-base `a223a8a0`, already reachable from the Create MCP Server form and by direct API call before this PR. This PR does **not** worsen it — the `set` flow is deliberately gated off for recipe-owned connectors (`resolveCredentialSurface` returns `recipe-owned` → copy, no form, never calls the POST). This is follow-up #2 in the description; the mirror gap (POST not mapping AlreadyExists to 409) is the same debt.

2. **Pre-existing tsc error in an untouched test** — `control-ui/components/__tests__/LlmDiscoveryPanel.test.tsx:149`, `TS2322 '"fallback"' is not assignable to LlmDiscoverySource`. Not touched by this diff, reproduces on `a223a8a0`. Already noted in the description.

## What checked out clean

- **Contract, all three claims verified against real code.** The `reason` strings match the HCC producer byte-for-byte (`reconciler.ts:226,239`, written into the `SecretResolved`/`False` condition on both the managed and WRC-owned paths). POST AlreadyExists really does collapse to a bare 500 (the error middleware only forwards numeric `err.status` in `[400,500)`, and the k8s client throws with `.code`, not `.status`), so a 409 branch would be dead code and the blind POST→PUT retry is the right call. The PUT `.status===404` check lands correctly via `formatApiError`.
- **State machine.** `secretCreated` / `recreateRequired` / `submittedMode` stay consistent across create→delete→retry, rotate-404→recreate, and stale-`surface`-after-create. `setSubmittedMode(mode)` runs synchronously before any `await`, so there's no visible window where the captured `mode` and the banner's `submittedMode` diverge. The `20d34cf2` fix (clearing `secretCreated` on the 404 branch) correctly closes the one-way latch.
- **Tests.** T3 verified — copying the new tests against the fix's parent (`68c4b88b`), the two latch tests fail there, so they actually cover the bug. Error fixtures match the real `apiSend` shape (T1), asserts are on operator-visible banners rather than intermediate effects (T4). 36 tests green, `tsc --noEmit` clean on the touched files.
- **Security.** No credential leakage (password inputs, `setDraft({})` clears on success, no message/toast reflects a credential *value* — only key names). No new XSS, no path injection (POST sends `name` in the body, server validates `isValidDNSSubdomain`), no new dependencies (the package.json/lock diff is a version bump only).

Nice work on the mutation-verification discipline in the description, made the review a lot faster to trust.

jozer-rami:
@josue-tech-94 both remaining blockers are closed at `513a3862`, and I followed your process note — mini-spec first, then one change rather than another incremental patch.

@zach88 thanks for the deep pass, especially verifying the three contract claims against the real control-api and HCC source. Both of your nits were already closed by `200e0019`: the `reason: 'SecretResolved'` fixture is gone (it's now the producer's `SecretFound`, and `grep` confirms the string survives only inside explanatory comments), and the page→resolver wiring test you flagged as missing now exists as `EditMcpServerPageCredentials.test.tsx`.

Full suite **130 files / 1339 tests** green (was 1321). `tsc --noEmit`: only the pre-existing `LlmDiscoveryPanel.test.tsx` error. Build green.

## The mini-spec

`docs/superpowers/` is gitignored in this repo, so it can't ride along in the diff — reproducing the two contracts here so they're reviewable.

**Ownership contract.** `recipeOwned == spec.managed === false`, and *nothing else*. It must not depend on `status`, `status.conditions`, that array's length, any condition field, or the resolved `surface`. Stated as an invariant precisely because the round-1 defect was an ownership decision that only engaged when a particular condition was observed — which silently disengages on absent/`Unknown`/stale/`SecretMissingKey` status, exactly where it's needed. Consequence: for any recipe-owned connector, in every reachable state, `mode` never resolves to `set`, no credential form renders, and `createMcpSecret` is never called — including on the late 404 path. (`spec.managed` is safe to test for exactly `false`: the CRD declares `default: true`, so absence can never mean recipe-owned.)

**Condition-authority policy.** Exactly one authoritative `SecretResolved` condition decides the surface. Authority ranks by `lastTransitionTime` validated *strictly* — RFC3339 syntax **plus** a calendar round-trip. Missing, empty, non-RFC3339 and invalid-calendar values all share one "oldest" rank and can only win when no validly-timestamped condition exists. Ties resolve to last-in-array, stably under permutation. The caller's array is never mutated.

## R1-H2 — closed

Your mutation now fails. Applying `recipeOwned = (status?.conditions ?? []).length === 0 && managed === false` at `page.tsx:205` turns the page suite red on exactly the three non-empty shapes it was meant to expose:

```
× keeps a managed:false connector on explanation-only — SecretResolved / Unknown / Reconciling
× keeps a managed:false connector on explanation-only — stale SecretResolved / True / SecretFound
× keeps a managed:false connector on explanation-only — SecretResolved / False / SecretMissingKey
   Tests  3 failed | 14 passed (17)
```

The printed DOM confirms it's the real defect rather than an incidental break — under the mutant the panel contains `<input type="password" id="mcp-cred-api-key" value="partial-value" required>` and `<button type="submit">Set credentials</button>`, i.e. the operator's partial rotate draft sitting in the create form of a recipe-owned connector. All six `managed:true` controls stay green, so the guard is ownership-specific, not a blanket "never recover".

**One correction to my own spec, worth stating plainly.** Shapes 1 and 2 (absent `status`, `conditions: []`) are *equivalent mutants* under your mutation — it computes the identical boolean there, so no test can go red for them, and my first draft of the spec demanded an impossible result. They still earn their place under a second mutation, `recipeOwned = false`, which turns all five journeys red. Both mutations are now documented so a future round doesn't chase a red that cannot exist.

Fixtures now come from one shared producer-faithful builder verified against `reconciler.ts` (managed path and WRC-owned path both write `type: SecretResolved, status: 'True', reason: 'SecretFound'`), with impossible/adversarial shapes explicitly labelled.

## R1-M2 — closed

You were right, and this one was a real bug rather than a coverage gap: I wrote "malformed timestamps rank oldest" into the policy and then implemented it with a bare `new Date()`, which happily accepts February 30 and normalizes it into March.

Ranking now requires an RFC3339 syntax match **and** a round-trip check that the parsed instant returns the same calendar components it was given. `[SecretNotFound@2026-02-30T00:00:00.000Z, SecretFound@2026-02-28T23:59:59.000Z]` now resolves `rotate` in both array orders, and `rotate` for a `managed:false` spec too.

Verified against 11 malformed forms sharing the oldest rank: absent field, empty string, `not-a-timestamp`, Feb-30, month 13, hour 25, non-leap `2027-02-29`, offset-less local time, date-only, `Aug 7, 2026`, and a space separator.

Mutations, each red: bare `new Date()` comparison (3 failed), revert to `.some()` (14 failed), flipped tie-break (3 failed), in-place `sort()` (1 failed, on both order and element identity).

## R2-L1 — description updated

You were right that the body had drifted from the head. Updated: the dual-error rule now documents PUT-404-keeps-POST rather than the old always-POST rule, ownership and condition-authority are described, counts are the current 1339, the full mutation set is listed rather than round 1's, and I've added the caveat that `control-ui` is outside the CI test matrix so all of this is local proof.

Ready for round 3.

josue-tech-94:
**Decision: Blocked**

The current implementation is materially safer than the round-1 head, and the six-commit collateral range introduced no runtime or UI regression. Merge readiness is still blocked by two carried test findings and one current-base integration finding:

1. `R1-H2` still uses manually mirrored HCC/API fixtures rather than a producer-backed contract.
2. `R1-M2` still has a reproduced false-green for three or more relevant duplicate conditions.

## Existing finding status

| ID | Status | Closing commit | Round-2 result |
| --- | --- | --- | --- |
| `R1-H1` | Closed | `3c58b794` | Ownership is independent of status; every managed-false late 404 becomes explanation-only and cannot POST. |
| `R1-H2` | **Open** | — | Page wiring is covered, but the GET resource and HCC conditions remain manually mirrored rather than producer-backed. |
| `R1-H3` | Closed | `519624cf`, hardened by `200e0019` | 500-only, 409, other non-500, and statusless retry gates are killed; one PUT attempt is enforced. |
| `R1-M1` | Closed | `519624cf`, hardened by `200e0019` | PUT 404 keeps the POST error; all other PUT failures surface the current-state error. |
| `R1-M2` | **Open; re-triage pending** | — | Production handles reviewed examples, but a last-two-candidates mutation leaves all 87 changed tests green. |
| `R1-L1` | Closed | `3be7e8c2`, hardened by `200e0019` | Set-only required/invalid semantics and first-invalid focus are verified. |

## Carried must-fix findings

### R1-H2 — Cross-layer fixtures remain manually mirrored

- **Priority:** Must fix
- **Severity:** High
- **Evidence:**
  `control-ui/components/__tests__/EditMcpServerPageCredentials.test.tsx:81`,
  `control-ui/components/__tests__/fixtures/secretResolvedConditions.ts:48`, and
  `host-context-controller/src/reconciler.ts:456`.
- **Concrete scenario:** HCC changes a `SecretResolved` status/reason/message or response shape while the UI test builder remains unchanged. The page tests continue to pass because both the asserted surface and the mocked producer output encode the UI author's old assumption.
- **Impact:** A producer/client contract drift can reintroduce the missing-Secret dead end or expose the wrong credential surface with a green control-ui suite.
- **Required fix:** Produce the page fixture through the real producer, an exported shared producer function, or an executable producer-contract fixture. Do not maintain a second handwritten copy of the HCC contract in control-ui.
- **Required test:** Drive producer output through `getMcpServer -> page -> resolver -> component` and prove that a producer contract change turns the test red. Keep impossible defensive conditions clearly separated from producer fixtures.
- **Confidence:** code-verified

The new page tests are otherwise strong. They drive the real page and component, cover absent status, empty conditions, Unknown, stale True, SecretMissingKey, exact SecretNotFound, and managed controls, and assert visible explanation/form state plus PUT/POST behavior. Coupling ownership to a non-empty conditions array now fails exactly the three intended journeys. The remaining issue is T1 producer provenance, not page wiring.

### R1-M2 — Decision logic still permits an arbitrary-cardinality false green

- **Priority:** Must fix
- **Severity:** Medium
- **Evidence:**
  `control-ui/components/UpdateConnectorCredentials/resolveCredentialSurface.ts:142` and
  `control-ui/components/__tests__/resolveCredentialSurface.test.ts:128`.
- **Concrete scenario:** Mutate authority selection to rank only the last two `SecretResolved` conditions. All 3 changed files / 87 tests remain green. For `[True/SecretFound@09:00, False/SecretNotFound@07:00, False/SecretNotFound@08:00]`, the mutant drops the newest condition and returns `set` instead of `rotate`.
- **Impact:** A bounded or prematurely reduced selector can reopen POST against an existing Secret while all changed tests and the current CI build remain green.
- **Required fix:** Treat the resolver as a pure precedence/reconciliation function and add a reference-model or property/invariant suite across arbitrary relevant cardinality, irrelevant condition insertion/permutation, strict timestamp classes, the intentional last-in-array tie, deterministic repeated resolution, and input immutability.
- **Required test:** The reproduced last-two-candidates mutation and the three-condition counterexample must fail for the intended surface assertion. Generated cases must preserve the explicit order-sensitive tie rule rather than assuming all permutations are equivalent.
- **Confidence:** reproduced

**R2 Medium expiry:** this Medium has remained open for two rounds. Before any further review it must be explicitly re-triaged. Recommendation: escalate its test-integrity risk to High/Must-fix, or close it in writing only if the owner replaces the T2 invariant requirement with a justified bounded-cardinality contract. It may not silently carry into another round.

jozer-rami:
@josue-tech-94 both carried findings are closed at `f26286d8`. Also: CI was reporting nothing on this PR for a reason unrelated to the code — details at the bottom.

**132 files / 1362 tests** green locally, **32/32 checks green** in CI, build green, `tsc` unchanged (the one pre-existing `LlmDiscoveryPanel` error).

## R1-M2 — re-triage decision: ESCALATED to must-fix

Per your Medium-expiry rule, recording the decision explicitly rather than carrying it: **escalated**, not closed with a bounded-cardinality contract. You reproduced two distinct false-greens in this one resolver across two rounds, which says example-based tests keep missing its edges — arguing for a bounded contract would be defending the weaker position.

Implemented the invariant suite you asked for (`resolveCredentialSurfaceInvariants.test.ts`):

- an independent reference model plus a generator over arbitrary cardinality (3, 4, 5+ relevant `SecretResolved` conditions), irrelevant types inserted at random positions, and every timestamp class (valid, missing, empty, malformed syntax, invalid calendar);
- a **seeded** PRNG, so any failure reproduces rather than being a heisenbug;
- the order-sensitive last-in-array tie encoded in the reference model too, not assumed away;
- deterministic repeated resolution and input immutability;
- your three-condition counterexample as a named regression test.

Mutations, each red: **last-two-candidates** (yours), `.some()`, flipped tie-break, in-place `sort()`, bare `new Date()`.

## R1-H2 — closed, and your literal required test now passes

First attempt did *not* satisfy you and I want to be straight about it. I added a contract test that reads `reconciler.ts` and goes red on drift — but the fixtures still restated the literals by hand, so drift was **detected in one file without propagating**. Renaming `SecretNotFound` → `SecretAbsent` left `EditMcpServerPageCredentials`, `resolveCredentialSurface` and the invariants suite all green. Your required test — *"drive producer output through `getMcpServer → page → resolver → component` and prove that a producer contract change turns the test red"* — was unmet. My own verifier caught it before I pushed.

Fixed by making the fixtures **derive** from the extracted producer contract (`fixtures/producerContract.ts`) instead of restating it. Same mutation now:

```
EditMcpServerPageCredentials.test.tsx — 3 failed | 14 passed
  × renders the CREATE form when a managed connector reports SecretNotFound
      Unable to find an element with the text: /needs credentials before it can start/i
  × renders the recipe explanation, and no form, for an unmanaged connector with a missing Secret
      Unable to find an element with the text: /managed by its WorkflowRecipe/i
  × honors a NEWER SecretNotFound over a stale clean resolution
      Unable to find an accessible element with the role "button" and name "Set credentials"
```

Real page-level assertion failures, not import errors — the end-to-end path is now what goes red.

The extractor classifies `validateSecret` failures by the **branch guard** they sit under (k8s 404 / 401-403 / declared-key check) rather than by reason string, which is what makes a rename propagate rather than merely throw.

I chose the executable-contract option over exporting from HCC deliberately: making control-ui import a controller package at runtime seemed worse coupling than the drift it fixes. Say the word if you'd rather have the shared export and I'll do that instead.

**Anti-vacuity**, since a contract test that silently matches nothing would be worse than none — all of these fail loudly at import (`Tests no tests`, exit 1), verified live: producer sites renamed away; path repointed at a real file with no producer sites; path missing; file below a byte floor; each extraction regex broken individually, including the subtle partial-match case.

## Still open, and I'd rather name them than let them ride

1. **These guards never run in CI.** `ci-public.yml` excludes control-ui from the test matrix, so every mutation above dies only on a developer's machine. Its stated rationale — *"its vitest suite has known pre-existing failures"* — no longer holds: the suite is 1362/1362 green on this branch. Wiring control-ui vitest into the matrix is now viable and is the honest way to make R1-H2 and R1-M2 enforceable rather than aspirational. I've kept it out of this PR because adding a CI job in round 3 would open a new review surface, but I think it should be the immediate follow-up.
2. **The GET-resource half of R1-H2** is untouched: `mcpServer()` still hand-builds `metadata`/`spec`/`status` against control-ui's loose `McpServerResource`. Lower severity than it reads, since control-api passes the CRD through verbatim, so the residual is CRD field-name drift (`spec.managed`, `spec.envSecret.keys`) rather than API-transform drift. Happy to extend the contract to the CRD schema if you want it closed here.
3. The RFC3339 **syntax** regex is byte-identical between implementation and reference model, so a shared misreading of the grammar would cancel out. The calendar half is independently implemented, so this is confined to syntax.

## Why CI showed nothing

Not a workflow problem. The PR had drifted into `CONFLICTING` against `dev` (289 commits ahead), and GitHub cannot build a merge ref for a conflicting PR — so no `pull_request` runs were created at all, and two consecutive pushes reported zero checks.

Merged `dev` in. Both conflicts were pre-commit-hook version churn plus one real change: kept our higher `0.1.403`, and took dev's `overrides.postcss` `^8.5.25` over our `^8.5.12` — the newer PostCSS patch you flagged in round 2. `node_modules` was reinstalled from the merged lock and the suite re-run against postcss 8.5.25 rather than the stale tree. CI came back immediately and is 32/32 green.

josue-tech-94:
@jozer-rami ## Decision

**Blocked — 1 must-fix item remains.**

Round 3 closed the remaining carried correctness/test findings and the `dev` synchronization issue. The collateral audit is clean: **0 review-fix regressions and 0 sync/semantic-integration regressions**.

### Must fix before merge

#### `R3-M1` — Prototype-named Secret keys cannot complete the set flow

**Severity:** Medium
**Priority:** Must fix

The new set/create flow breaks for Kubernetes-valid Secret keys that collide with `Object.prototype`, such as `__proto__`, `constructor`, or `toString`.

For example, with `__proto__`, assigning the entered value into the current plain-object credential state/payload does not create a normal own property. Validation then reads an inherited/prototype value instead of the credential string, and the set flow fails before POST.

**Required fix:** use a prototype-safe representation for dynamic credential keys throughout draft state, updates, validation, payload construction, and submission. Use own-property checks rather than inherited truthiness.

**Required test:** cover at least `__proto__` end to end in set mode and class-check the other accepted `Object.prototype` names in both set and rotate behavior.

Because `UpdateConnectorCredentials/index.tsx` is already beyond the F3 per-file fix budget, this should be handled through the required small mini-spec/redesign rather than another incremental patch.

## Good / verified

* `R1-H1`, `R1-H2`, `R1-H3`, `R1-M1`, `R1-M2`, `R1-L1`, and `R2-M1` are closed.
* Recipe-owned late-404 behavior remains safe and cannot reach POST.
* Create retry remains status-agnostic and bounded to one PUT retry.
* Dual-failure error selection is covered by visible-behavior tests.
* Condition authority now has producer-backed and invariant/reference-model coverage.
* Required/invalid accessibility semantics and first-invalid focus are covered.
* Round-3 collateral audit is **CLEAN**.
* Full tests pass: **132 files / 1,362 tests**.
* Production build passes.
* Current GitHub Actions passed.

## Follow-up

`R3-M2` — The newly promoted credential-create path has no audit record. This is a valid **non-blocking follow-up**: the endpoint gap predates this PR, but the PR makes that POST path a normal connector-recovery operation.

## Nice to have

`R2-L1` — Updating the PR description would improve accuracy, but **I would not block merge on it**. The stale test counts, follow-up numbering, and older dual-error wording can be cleaned up before merge if convenient.

Once `R3-M1` is resolved and re-verified, I have no other current merge blocker from this review.

josue-tech-94 (approved):
## Approved — follow-ups only

Round 4 is complete on `11eec150a9b38ca4ab59ffff4c075a8707eb3bb1`.

I’m approving this PR. There are **no remaining merge-blocking findings**, no new round-4 findings, and the collateral audit was clean:

`R3-M1` is closed: prototype-named Secret keys such as `__proto__`, `constructor`, `toString`, and `valueOf` are now handled through prototype-safe credential maps, with current-head, mutation, and previous-head regression proof.

### Non-blocking follow-ups

1. **`R3-M2` — add auditing for Secret creation**

   * `POST /admin/mcp-secrets` still creates credentials without an audit record.
   * The audit should record the operation, Secret name, namespace, and key names only — never credential values — with coverage for successful and failed creation.

2. **`R2-L1` — PR description cleanup**

   * Nice to have; does **not** block merge.
   * Update the stale test counts.
   * `tsc --noEmit` is now green.
   * `Build (control-ui)` now runs both tests and build.
   * Current control-api can forward a sanitized Kubernetes `409`, so the old “duplicate create always becomes a bare 500” wording is outdated.
   * Clean up the follow-up numbering/count.

jozer-rami:
@josue-tech-94 both remaining follow-ups from the round-4 approval are closed at `6a2e1f64`.

**R3-M2 — Secret-create audit.** `POST /admin/mcp-secrets` now emits `mcp-secret-created` after a successful write and `mcp-secret-create-failed` when the write throws. The record carries the operation, Secret name, namespace, and sorted key names only — never credential values. A Kubernetes `ApiException` (AlreadyExists) is sanitized so apiserver headers cannot land in the log. Validation 400s do not emit a create audit, because no write was attempted.

Covered in `control-api/test/routes.mcpSecrets.test.ts`: success, failed write, validation no-audit, and the ApiException header case.

**R2-L1 — PR description.** Test counts now match CI (`173` files / `1910` tests). `Build (control-ui)` is described as vitest + `next build`. The stale “duplicate create always becomes a bare 500” wording is gone — current control-api forwards a sanitized Kubernetes 409, and the client retry stays status-agnostic for version-skew. Follow-up numbering is 1–2 (ownership guard, provenance labels); audit is no longer listed as open.

PR #278, opened 2026-08-06
fix: reconcile Plugin Workload SDK runtime contracts

alfredolopez80:
## Summary

Adds a forward-only `0090_plugin_workload_sdk_runtime_contract_reconciliation` migration for long-lived databases where Plugin Workload SDK migrations 0061, 0065, 0080, and 0081 were already recorded before their runtime contract bodies evolved.

The migration:

- installs a stricter, superseding `plugin_workload_sdk` trace-metadata validator (canonical UUIDs, `provider_attempt_index`/`attempt_count` in 1-4, nested-key allowlist for `prompt_bridge`, `cache_tokens_reported`, and fail-closed on NULL `event_kind`) — this strict body is new in 0090; migration 0065's recorded body remains the laxer historical one;
- restores function `search_path` and exact runtime ACLs;
- reconciles only v2 invocations that are still `in_progress` without a lease, aborting if a physical provider attempt makes the state ambiguous;
- fences only the captured invocation-attempt rows, preserves terminal v2 records, restores the v2 lease check, and accepts the provider-attempt `skipped` terminal state;
- keeps the migration idempotent and does not edit or rerun historical migration versions.

## Validation

- `control-api` build: passed
- `control-api` typecheck: passed
- focused migration/finalization tests: 14 passed, 1 real-PostgreSQL test skipped locally because `CONTROL_API_REAL_PG_ADMIN_URL` is not configured
- full control-api suite: 3,873 passed, 59 skipped
- CI real-PostgreSQL job is required to execute the upgrade simulation and finalization path against PostgreSQL 16

No cluster or database was mutated directly; deployment remains through the normal CI and infra flow.


alfredolopez80:
@claude pls give a full review of this PR

control-api/src/db.ts
```diff
@@ -2286,6 +2286,301 @@ async function applyGfsPermissionStoreSchema(db: DbClient): Promise<void> {
...
+    -- cannot hold row locks and deadlock against the ACCESS EXCLUSIVE that the
+    -- ALTERs below acquire; plain readers (ACCESS SHARE) are still allowed. This
+    -- closes the rolling-deploy race in which a writer could change a row between
+    -- the ambiguity check and the repair UPDATE.
+    LOCK TABLE
+      plugin_workload_sdk_invocations,
+      plugin_workload_sdk_invocation_attempts,
+      plugin_workload_sdk_provider_attempts,
+      agent_run_events
+      IN EXCLUSIVE MODE;
```
jozer-rami: **MAJOR (blocks deploy)** — this lock strategy can abort the migration Job, and the retry the comment relies on does not exist.

`EXCLUSIVE` conflicts with `ROW EXCLUSIVE`, which every ordinary `INSERT`/`UPDATE`/`DELETE` takes. I reproduced this on PostgreSQL 16: with a single open writer transaction, the lock request dies with `ERROR: canceling statement due to lock timeout` after 5s.

**Concrete trigger.** `trace_maintenance_runtime` is a separate live deployment running `governed_trace_prune_expired_events`, which does `DELETE FROM agent_run_events` (`db.ts:3997`, `:4194`; called from `services/tracing/maintenance/retention.ts:29`). A prune batch open longer than 5s during the migration window blocks this lock. Old control-api pods appending traces do the same, and the migration Job runs *before* the control-api rollout (`deploy/scripts/run-control-api-db-migration.sh:1077` → `:1099`).

**Failure chain.** Lock timeout → the whole `initDb` transaction rolls back, taking **every** pending migration with it (`db.ts:5603-5611`) → `dist/migrate.js` sets exit 1 (`migrate.ts:16-19`) → the Job has `backoffLimit: 0` and `restartPolicy: Never` (`run-control-api-db-migration.sh:353,363`) → `kubectl wait --for=condition=complete` fails and the script hard `exit 1`s (`:1099-1110`). The deploy aborts.

The comment on line 2300-2301 says "Fail loud and let the migration Job be retried instead of silently stalling the deploy." There is no retry. `backoffLimit: 0` makes one failure terminal.

**Fix** — pick one:
- raise `lock_timeout` well above the prune-batch duration;
- drop `agent_run_events` from the `LOCK TABLE` list (the validator swap plus constraint rebuild takes its own `ACCESS EXCLUSIVE` and does not need the SDK invocation tables held alongside it);
- or give the Job `backoffLimit > 0` so the comment's premise becomes true.

control-api/src/db.ts
```diff
@@ -2286,6 +2286,301 @@ async function applyGfsPermissionStoreSchema(db: DbClient): Promise<void> {
...
+    ALTER TABLE agent_run_events
+      DROP CONSTRAINT IF EXISTS agent_run_events_check;
+    ALTER TABLE agent_run_events
+      DROP CONSTRAINT IF EXISTS agent_run_events_payload_metadata_check;
+    ALTER TABLE agent_run_events
+      ADD CONSTRAINT agent_run_events_check
+      CHECK (governed_trace_safe_agent_run_metadata(event_type, payload_metadata) IS TRUE)
+      NOT VALID;
+    ALTER TABLE agent_run_events
+      VALIDATE CONSTRAINT agent_run_events_check;
```
jozer-rami: **MAJOR** — the `NOT VALID` + `VALIDATE` split buys nothing here and doubles the scan inside the reader-blocking window.

The comment claims the ordering "keeps the expensive full scan out of the reader-blocking window." That holds for the `DO`-block scan above, but not for `VALIDATE`.

Because `applyPendingMigrations` runs every migration inside one transaction (`db.ts:5596-5601`), the `ACCESS EXCLUSIVE` taken by `ADD CONSTRAINT` is held until `COMMIT`. Verified on PostgreSQL 16: a plain `SELECT count(*)` against a table with an open `ADD CONSTRAINT ... NOT VALID` transaction times out waiting for the lock.

So `VALIDATE CONSTRAINT` on line 2466 performs a **second** full scan, invoking the validator on every row again, while readers are already locked out. Net effect versus a plain `ADD CONSTRAINT`: identical lock profile, 2x the scan cost.

**Failure scenario.** On a large `agent_run_events`, this doubles the total table lockout and pushes the Job closer to the deploy script's 300s `kubectl wait` timeout.

**Fix.** Use a plain `ADD CONSTRAINT` (the `DO`-block scan above already proved every row passes), or move the reconciliation out of the shared transaction so `VALIDATE` can genuinely run under `SHARE UPDATE EXCLUSIVE`.

control-api/src/db.ts
```diff
@@ -2286,6 +2286,301 @@ async function applyGfsPermissionStoreSchema(db: DbClient): Promise<void> {
...
+            WHERE attempts.invocation_id = invocations.id
+              AND attempts.attempt_generation = invocations.attempt_generation
+              AND attempts.status IN ('in_progress', 'complete')
+         );
+
+      GET DIAGNOSTICS fenced_count = ROW_COUNT;
+      IF fenced_count <> expected_count THEN
+        RAISE EXCEPTION
+          'plugin workload SDK invocation state changed during reconciliation; retry safely';
+      END IF;
```
jozer-rami, line 2557: **MINOR** — this guard is unreachable.

The temp table is populated with the same predicate as the `UPDATE` minus the `NOT EXISTS(provider attempts)` clause, and the ambiguity check at `:2476-2492` has already raised if any captured row has a physical attempt in `('in_progress','complete')`. Under the `EXCLUSIVE` lock nothing can change in between, so `fenced_count` always equals `expected_count`.

Harmless defense-in-depth, but it reads as protection that is not doing any work and it cannot be covered by a test. Worth either a comment saying it is belt-and-braces, or dropping it.

control-api/src/db.ts
```diff
@@ -2286,6 +2286,301 @@ async function applyGfsPermissionStoreSchema(db: DbClient): Promise<void> {
...
+        OR status <> 'in_progress'
+        OR lease_expires_at IS NOT NULL
+      );
+
+    ALTER TABLE plugin_workload_sdk_provider_attempts
+      DROP CONSTRAINT IF EXISTS plugin_workload_sdk_provider_attempts_status_check;
+    ALTER TABLE plugin_workload_sdk_provider_attempts
+      ADD CONSTRAINT plugin_workload_sdk_provider_attempts_status_check
+      CHECK (status IN ('reserved','in_progress','complete','failed','provider_unavailable','skipped'));
+  `)
```
jozer-rami, line 2601: **MINOR** — `SET LOCAL lock_timeout = '5s'` (line 2302) leaks past this migration.

`SET LOCAL` is transaction-scoped, and `applyPendingMigrations` runs all pending migrations in one transaction (`db.ts:5574-5584`). 0090 is currently last, so this is latent today, but any future 0091+ silently inherits a 5s lock timeout on fresh installs and on the next upgrade.

**Fix.** `RESET lock_timeout;` at the end of this statement block.

control-api/test/db.pluginWorkloadSdkTraceValidatorMigration.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    expect(migrationSql).toContain('plugin_workload_sdk_invocations_v2_lease_check')
+    expect(migrationSql).toContain('plugin_workload_sdk_provider_attempts_status_check')
+    expect(migrationSql).toContain("'skipped'")
+    expect(migrationSql).toContain('cannot reconcile % v2 invocations without leases')
+    expect(migrationSql).toContain('LOCK TABLE')
+    expect(migrationSql).toContain('agent_run_events_check')
+    expect(migrationSql).toContain('IS DISTINCT FROM TRUE')
+    expect(migrationSql).toContain('IS TRUE)')
+    expect(migrationSql).toContain('state changed during reconciliation')
+    expect(migrationSql).toContain("attempts.status IN ('reserved', 'in_progress')")
```
jozer-rami, line 72: **MAJOR (test-integrity)** — this test cannot fail on any behavioral regression.

All 99 lines are `toContain()` assertions against the migration's own SQL string.

**Mutation test.** Change `^[1-4]$` to `^[1-9][0-9]*$`, swap `EXCLUSIVE MODE` for `SHARE ROW EXCLUSIVE`, delete the `SET LOCAL lock_timeout`, or invert `IS DISTINCT FROM TRUE`. Every assertion here still passes. Line 59 asserts the SQL contains `LOCK TABLE` without asserting the mode; line 62 asserts it contains `IS TRUE)` without asserting what it guards.

This matters because it is the only test covering 0090 in the `Build / test / typecheck (control-api)` job. The real-Postgres suite runs as a separate job (`ci-public.yml`, `npm test -- --run realPostgres`), so for the default job this migration is effectively unguarded.

**Fix.** Assert behavior, not text. The real-PG suite is the right home for most of it; this test should at minimum assert the lock **mode** and the timeout **value**, since those are the two knobs that decide whether the deploy survives (see the finding on `db.ts:2302`).

control-api/test/db.pluginWorkloadSdkRuntimeContract.realPostgres.integration.test.ts
```diff
@@ -0,0 +1,706 @@
...
+      RETURNS BOOLEAN
+      LANGUAGE sql
+      IMMUTABLE
+      AS $$
+        SELECT event_kind <> 'token_usage'
+          OR (
+            jsonb_typeof(value) = 'object'
+            AND value->>'source_kind' IN ('channel', 'desktop', 'workflow', 'cron', 'unknown')
+            AND NOT (value ? 'prompt_bridge')
+          );
```
jozer-rami: **MAJOR (test-integrity)** — this restores a strawman validator, not migration 0065's actual body, so the suite cannot prove a real long-lived DB survives 0090.

The real pre-0090 body is `db.ts:4633-4697`: a full top-level key allowlist, `request_ref ~ '^[0-9a-f]{64}$'`, `prompt_bridge` **permitted**, `^[0-9a-f-]{36}$` for the UUIDs, unbounded `provider_attempt_index`, unbounded `target_ref`, and no nested-key allowlist.

The rows that actually threaten this migration are the ones 0065 accepted and 0090 rejects:
- `prompt_bridge.invocation_id` of 36 hex/dash chars that is not a canonical UUID (new: `db.ts:2389`)
- `provider_attempt_index >= 5` (new: `db.ts:2392`)
- `target_ref` longer than 256 chars (new: `db.ts:2394`)
- any extra nested key (new: `db.ts:2373`)

This fixture forbids `prompt_bridge` outright, so none of those rows can even be seeded and the suite never tests one. The abort test at `:284-322` seeds `{source_kind:'unknown', input_tokens:1, output_tokens:1}`, a row the *real* 0065 body would also have rejected. It proves the abort plumbing works, not that production data survives.

**In fairness:** I traced the write path and the app layer bounds all four today. `provider_attempts.attempt_index` has `CHECK (attempt_index BETWEEN 1 AND 4)` (`pluginWorkloadSdkSchema.ts:483`) and events must match it exactly (`usageEvents.ts:463`); `target_ref` is capped at 256 (`usageEvents.ts:156`); `invocation_id` must pass `UUID_REGEX` (`usageEvents.ts:263`). Production is very likely clean. But the migration bets the whole deploy on that, and per the `db.ts:2302` finding a false bet is unrecoverable without manual intervention.

**Fix.** Install the actual `db.ts:4633-4697` body here, then seed one row per newly-tightened rule and assert 0090 aborts on each.

jozer-rami (commented):
## Deep review

Reviewed across five lenses. Every candidate finding was verified or refuted against the real code and, where it mattered, against a live PostgreSQL 16 instance. Findings that did not survive verification are listed at the bottom so you can see where the review pushed and came up empty.

| Lens | Verdict |
|---|---|
| correctness (SQL/data) | **Pass.** I attacked this hard and every candidate defect refuted. Fencing predicates, status values and column references all check out against the real schema. |
| security / tenancy | **Pass.** No authz or cross-org surface. The grant changes are a net narrowing and match existing state. |
| deploy-safety | **Fail.** The locking strategy can abort the deploy, and the retry it explicitly relies on does not exist. |
| migration-safety | **Concerns.** The `NOT VALID`/`VALIDATE` split is a no-op that doubles a full scan inside the reader-blocking window. |
| test-integrity | **Fail.** The unit test cannot fail on any behavioral regression, and the real-Postgres "upgrade simulation" restores a strawman historical validator rather than 0065's actual body. |

### What is genuinely solid

The reconciliation SQL itself is careful work. Capturing the target set into a temp table before mutating, aborting on ambiguous state rather than guessing, refusing to rewrite or delete append-only history, preserving terminal v2 records, and adding a new migration instead of editing a recorded one are all the right calls. The abort-path integration test (asserting `schema_migrations` stays empty and the offending row survives) is the kind of assertion most migration PRs skip.

### Findings

Inline, ranked:

1. **MAJOR** `db.ts:2302,2313-2318` — `lock_timeout` + `EXCLUSIVE` lock aborts the migration Job, and the retry it depends on does not exist.
2. **MAJOR** `db.ts:2453-2466` — `NOT VALID` + `VALIDATE` buys nothing here and doubles the scan under a reader-blocking lock.
3. **MAJOR** `db.pluginWorkloadSdkTraceValidatorMigration.test.ts:48-64` — the unit test cannot fail on any behavioral regression.
4. **MAJOR** `...RuntimeContract.realPostgres.integration.test.ts:140-150` — the upgrade simulation restores a strawman validator, not 0065's actual body.
5. **MINOR** `db.ts:2581` — `SET LOCAL lock_timeout` leaks to every later migration.
6. **MINOR** `db.ts:2541` — the `fenced_count <> expected_count` guard is unreachable.

### Refuted (checked, dropped)

Stating these so you know where the review actually pushed:

- **`agent_run_events_check` collides with the `occurred_at <= ingested_at` clock-skew guard.** Refuted empirically. Built the real table on PostgreSQL 16: because the `payload_metadata` check references two columns, Postgres drops the column name and it becomes `agent_run_events_check`; the clock-skew guard is `agent_run_events_check1` and is untouched. The `DROP`/`ADD` at `db.ts:2457-2462` is correct.
- **`authorization_decision = 'migration_interrupted'` violates a CHECK.** Refuted. The column is plain `TEXT NOT NULL` with no CHECK (`pluginWorkloadSdkSchema.ts:58`).
- **`REVOKE ... FROM ... workflow_recipes_runtime` fails when the role is absent.** Refuted. All three roles are created unconditionally in an earlier migration (`db.ts:4262-4273`).
- **`provider_attempt_index ~ '^[1-4]$'` is stricter than the runtime.** Refuted. The schema caps it at 4 (`pluginWorkloadSdkSchema.ts:483`) and `usageEvents.ts:463` enforces equality with the stored row.
- **`credential_slot` breaks the new nested-key allowlist.** Refuted. `usageProjection.ts:79-87` emits exactly the seven allowlisted keys; `credential_slot` stays in the raw usage ledger and never reaches `agent_run_events`.
- **Revoking EXECUTE from `workflow_recipes_runtime` breaks its writes.** Refuted. Already revoked at `db.ts:4389`; 0090 preserves the existing state.

### Bottom line

The data-reconciliation logic is correct and I could not break it. What I would not merge as-is is the locking strategy: finding 1 is a deploy-abort with no retry path, and its stated mitigation does not exist in the Job spec. Findings 3 and 4 mean CI being green here proves less than it looks like it does.

Fixing 1 and 2 is a small edit to the migration body. Fixing 4 is the higher-value test change, because it is the only thing that would tell you whether real cluster data survives before you find out during a prod deploy.

<sub>Left as a comment rather than a change request so it does not gate the merge queue; findings 1-4 are the ones I would want landed first.</sub>

control-api/src/db.ts
alfredolopez80: Addressed in 099485da. Raised `lock_timeout` 5s → 60s — well above the bounded prune wake-transaction and well below the deploy's 300s `kubectl wait` — so a transient concurrent writer is tolerated while a genuinely stuck lock still fails fast, and corrected the comment that wrongly implied the Job self-retries.

Dropping `agent_run_events` from the `LOCK TABLE` wouldn't help: the `DROP`/`ADD CONSTRAINT` take their own `ACCESS EXCLUSIVE` on that table under the same timeout, so the timeout is the real knob (and dropping it would reopen the scan→ADD race the comment documents). The complementary `backoffLimit > 0` change lives in the deploy script (`run-control-api-db-migration.sh`), out of this control-api PR's scope; it's called out in the code comment as a follow-up. Happy to open a tracking issue for it.

control-api/src/db.ts
alfredolopez80: Addressed in 099485da. Replaced the `NOT VALID` + separate `VALIDATE` with a single validating `ADD CONSTRAINT`. You're right that in one transaction the split buys nothing — the `ADD`'s `ACCESS EXCLUSIVE` is held to `COMMIT` — so this drops the redundant second scan and the comment that wrongly claimed `VALIDATE` runs under `SHARE UPDATE EXCLUSIVE`. The fail-closed `DO`-block scan above still runs under the `EXCLUSIVE` lock (plain readers proceed) and yields the triage sample.

control-api/test/db.pluginWorkloadSdkTraceValidatorMigration.test.ts
alfredolopez80: Addressed in 099485da. The unit test now asserts the two deploy-critical knobs — `IN EXCLUSIVE MODE` and `SET LOCAL lock_timeout = '60s'` — plus the `^[1-4]$` cap, so mutating the lock mode or timeout now fails this default-job test instead of passing. The behavioural coverage stays in the real-Postgres suite (see the fixture finding below).

control-api/test/db.pluginWorkloadSdkRuntimeContract.realPostgres.integration.test.ts
alfredolopez80: Addressed in 099485da. `restoreHistoricalRuntimeState()` now installs 0065's actual validator body (copied verbatim from the recorded migration) instead of the strawman, and the suite seeds one row per newly-tightened rule — a non-canonical 36-char `invocation_id`, `provider_attempt_index = 5`, a `target_ref` longer than 256 chars, and an extra nested key — each accepted by the real 0065 body and rejected by 0090, asserting the migration aborts on each with `schema_migrations` staying empty and the offending row intact. Added an anti-strawman fidelity check so a future drift of the fixture fails loudly.

control-api/src/db.ts
alfredolopez80: Addressed in 099485da. Added `SET LOCAL lock_timeout = '0';` as the last statement of the 0090 block so a later 0091+ applied in the same `applyPendingMigrations` transaction batch doesn't inherit the 60s timeout.

control-api/src/db.ts
alfredolopez80: Addressed in 099485da. Kept the guard as explicit defence-in-depth but documented it as belt-and-braces that cannot fire under the `EXCLUSIVE` lock, and dropped the misleading `retry safely` wording (there is no auto-retry with `backoffLimit: 0`) in favour of a diagnostic `fenced % of % captured; aborting the migration` message. Happy to drop the guard entirely instead if you'd prefer.

control-api/src/db.ts
alfredolopez80: Tracking the complementary `backoffLimit > 0` deploy-script change in #293 so it isn't lost.

alfredolopez80: Update: rather than deferring this, the complementary `backoffLimit: 2` change is now included in this PR (56a936ab) alongside the `lock_timeout` 5s→60s raise, so both halves of the finding land together. #293 will close on merge.

zach88:
Took a thorough pass through this one across the SQL logic, the tests, security, and the validator-vs-runtime contract. Good news first: nothing here blocks the merge. The migration is pure hardening, the lock ordering matches finalization, idempotency is guaranteed by the single-transaction rollback model, and the stricter validator lines up exactly with what the runtime actually produces today, so it doesn't break the live write path. Everything below is test-quality and defense-in-depth polish, not correctness bugs.

## Medium

1. **[R1-M1] The `db.ts` comment contradicts the backoffLimit change in this same PR.** `control-api/src/db.ts:2305-2307` still says *"the migration Job runs with backoffLimit 0 … raising the Job's backoffLimit (deploy script, out of this PR's scope) would let it self-retry"*, but this same diff bumps it to 2 in `deploy/scripts/run-control-api-db-migration.sh:353`. So the comment is doubly wrong now: the Job no longer runs at 0, and the change is very much in scope. Someone reading `db.ts` to reason about retry/idempotency will conclude a timeout aborts the deploy with no retry, when there are actually up to 3 attempts. Cheap to fix and it's right here in the PR — I'd close this one before merging.

2. **[R1-M2] The "idempotency" test only proves the runner dedups, it never re-runs the migration body.** `control-api/test/db.pluginWorkloadSdkRuntimeContract.realPostgres.integration.test.ts:815-825` runs `initDb` a second time and asserts `COUNT(schema_migrations WHERE version='0090') == 1`. But since 0090 is already recorded, `applyPendingMigrations` hits `if (appliedVersions.has(...)) continue` (`db.ts:5593`) and skips the body entirely — the SQL never executes again. So it's testing "the runner doesn't re-register", not "the body is idempotent". The body genuinely *is* idempotent (the transaction guarantees it), so this is a coverage gap rather than a bug, but the body-idempotency invariant from the PR description ends up with no dedicated test. Suggest re-running the body over already-migrated state (`DELETE FROM schema_migrations WHERE version='0090'` + re-`initDb`, same trick `restoreHistoricalRuntimeState` already uses) and asserting the second pass doesn't throw and leaves state identical.

3. **[R1-M3] Two invariants the PR body promises have no test.** In the only non-aborting 0090 run (`...realPostgres.integration.test.ts:570`) the sole `plugin_workload_sdk_invocations` row present is the legacy v2-no-lease one. There's nothing exercising "preserves terminal v2 records" (a v2 row in a terminal state, present during the run, that 0090 should leave untouched) or "reconciles only … without a lease" (a v2/in_progress row *with* a lease that the `lease_expires_at IS NULL` clause shouldn't touch). If the fencing `WHERE` ever dropped `status='in_progress'` or `lease_expires_at IS NULL`, it'd downgrade a legit in-flight invocation to v1/failed and the suite would stay green. Suggest seeding both rows in the same run and asserting they keep their original `contract_version=2`, status and lease.

4. **[R1-M4] `validPayload()` is a hand-written fixture standing in for another layer's output.** `...realPostgres.integration.test.ts:292-313` re-types the `token_usage` `payload_metadata` by hand, but that shape is produced by `tokenUsagePayload()` in `control-api/src/services/tracing/usageProjection.ts:62`, which is pure and importable. They're aligned today, but if the producer's shape drifts (say a new key in `prompt_bridge`), the validator would reject it in prod while this contract test stays green on the stale shape. Suggest building `validPayload()` by calling `tokenUsagePayload(<real LlmUsageEvent>)`.

## Minor (non-blocking grab-bag)

5. **[R1-L1] The "real historical validator" in the integration test is transcribed by hand.** `...realPostgres.integration.test.ts:142-230` — the validator body itself checks out verbatim against 0065, but the pre-0090 GRANTs/constraints/default are hand-assumptions about the accumulated 0066-0089 state, not derived from `CONTROL_API_MIGRATIONS`. If some migration in that range already revoked `workflow_recipes_runtime` or changed the default/constraint, the "historical" premise diverges from a real long-lived cluster. At minimum, an assertion anchoring the transcription to the real body would help.

6. **[R1-L2] The `fail-closed` unit test asserts SQL text, not behavior.** `control-api/test/db.pluginWorkloadSdkTraceValidatorMigration.test.ts:80-106` is all `expect(migrationSql).toContain(...)`, so a subtly broken regex (`^[1-4]$` typo) or an inverted allowlist branch would still pass. The real behavioral coverage lives in the integration test (which does run in CI), so net risk is low — but the name oversells what it checks. Either rename it to what it actually asserts, or lean on the integration test for the fail-closed proof (it's already there at :599-672). No need to duplicate the behavior in a mock.

7. **[R1-L3] App validation doesn't mirror the stricter 0090 CHECK bounds, so malformed input would fail late (500 at INSERT) instead of early (400 at ingest).** Three instances: `provider_attempt_index` — CHECK wants `^[1-4]$` (`db.ts:2398`) but `promptBridgeMetadata` only checks `>= 1` with no upper bound (`control-api/src/services/usageEvents.ts:166`); the UUIDs — app validates case-insensitive and stores un-lowercased (`usageEvents.ts:92,171,177`) while the CHECK is lowercase-only; `iteration` — app takes `intOrNull` without a `>= 0` guard (`usageEvents.ts:288`). Nothing malformed reaches the INSERT today — the ingest cross-check, the `attempt_index BETWEEN 1 AND 4` column CHECK, and mcp-host always setting `channel_type='plugin_workload_sdk'` all save it — but that safety is indirect coupling, not local validation. Worth mirroring the bounds in the app so a future change to the cross-check surfaces as a clean 400.

8. **[R1-L4] `backoffLimit: 2` × `lock_timeout 60s` can crowd the `kubectl wait` budget.** `deploy/scripts/run-control-api-db-migration.sh:359` with `restartPolicy: Never` gives 3 attempts, each waiting up to 60s for the lock plus pod startup. Against a persistently stuck writer, 3 sequential attempts can eat much of the 300s `kubectl wait --for=condition=complete` (`TIMEOUT` at :126, wait at :1105), so the wait expires before the Job exhausts its retries. The comment's attempt arithmetic is correct; it just doesn't account for the wait interaction. Not a correctness bug — maybe a note in the comment or bump `TIMEOUT` accordingly.

## What checked out clean

- **Body idempotency** under pod retry — the whole batch runs in one `BEGIN…COMMIT` with full rollback, and the `ALTER … SET DEFAULT` / `DROP … IF EXISTS + ADD` are re-appliable.
- **Lock ordering** matches finalization's `FOR UPDATE` acquisition order (`pluginWorkloadSdkFinalization.ts:228/248/268`), so no deadlock against a concurrent finalization.
- **Fail-closed scan** — `COUNT(*) OVER () … LIMIT 10` totals over the post-WHERE set, `invalid_count` stays 0 on empty, and the `DROP CONSTRAINT IF EXISTS` covers both possible constraint names.
- **v2 reconciliation** is internally consistent — after fencing there's no v2/in_progress-without-lease left, so the reinstated `v2_lease_check` holds, and no attempts dangle under a failed invocation.
- **Validator vs real producer** are aligned today: 9 required top-level keys + the 7 `prompt_bridge` keys, `[1-4]` ranges, identical `source_kind` set, no extra keys. The migration does not break the runtime write path.
- **contract_version/lease** — the real writer always passes an explicit `contract_version` and a non-null `lease_expires_at`, and the new CHECK is already guaranteed by the state-transition guards.
- **Security** — ACLs match the fresh-install state (no widening, no denial), `search_path` is pinned on every touched function, NULL `event_kind` is genuinely rejected, no dynamic SQL.

Suite status: the touched unit tests pass (11), and the real-Postgres integration test does execute in CI (`ci-public.yml`, `--run realPostgres`) against Postgres 16; it's `describe.skip` only when `CONTROL_API_REAL_PG_ADMIN_URL` is unset locally.

jozer-rami (approved):
## Approved

Re-reviewed at `56a936ab`. All four blocking findings are fixed, and I verified each against the code rather than taking the commit messages at face value.

| # | Finding | Resolution |
|---|---|---|
| 1 | `lock_timeout` + `EXCLUSIVE` aborts the deploy, no retry | **Fixed.** 5s → 60s, and `backoffLimit` 0 → 2 so the comment's retry premise is now true. |
| 2 | `NOT VALID` + `VALIDATE` no-op, double scan | **Fixed.** Single validating `ADD CONSTRAINT`; the new comment states the real reason. |
| 3 | Unit test cannot fail on regressions | **Fixed to the bar I set.** The three mutations I named now fail the test. |
| 4 | Strawman historical validator | **Fixed, and well.** See below. |
| 5 | `lock_timeout` leaks to later migrations | **Fixed.** Reset added at the end of the block. |
| 6 | Unreachable guard | **Fixed.** Documented as defence-in-depth, and the error now carries both counts. |

### What I verified, not just read

- **The 60s timeout is sized against the real trigger.** The trace-retention prune I cited is bounded to 1-1000 rows per batch (`db.ts:4175-4187`), so the concrete writer that would have blown the old 5s budget cannot hold a lock near 60s.
- **Retry safety actually holds.** `applyPendingMigrations` and `recordMigration` run in one transaction, so any failure rolls back atomically and each retry starts from identical state. The "migration body is idempotent" claim in the Job comment is sound.
- **The fixture is genuinely the real 0065 body.** I diffed it mechanically against `db.ts`: byte-identical, 65/65 lines. Not "close enough".
- **The hostile seeds are real.** All four hardened rules are covered (non-canonical 36-char UUID, `provider_attempt_index: 5`, 257-char `target_ref`, unknown nested key). The design detail worth calling out: `seedTraceEvent` inserts while the historical CHECK is installed, so if this fixture ever regresses into a strawman the INSERT throws and the test fails loudly. That is structurally self-validating rather than merely commented, and it is the single best change in this round.
- **The dropped assertion in `ProtocolMigration.test.ts` costs nothing.** It gave up a "0089 is last" invariant, but monotonicity is independently covered by `control-api/test/db.migrationsOrder.test.ts:32`.

### Follow-ups, none blocking

1. `deploy/scripts/run-control-api-db-migration.sh:126` — `TIMEOUT` stayed at 300s while the retry budget grew. Worst case is now 60s + 10s backoff + 60s + 20s backoff + 60s = 210s of lock waits and backoff, plus three pod startups, plus the scan. A deploy that would have succeeded on attempt 3 can be killed by the `kubectl wait`. Raising `--timeout` to ~600s would match the new arithmetic.
2. `db.ts:2593` — `SET LOCAL lock_timeout = '0'` means *disabled*, not *restore*. It is equivalent here only because nothing in `deploy/` configures `lock_timeout`, so this is not a defect today. `RESET lock_timeout` is the idiom that stays correct if a server or role level value is ever set.
3. The new reset statement is unasserted in the unit test, and `toContain("~ '^[1-4]$'")` matches either of its two occurrences, so loosening only `attempt_count` would still pass.

Good turnaround on this. The reconciliation logic was already correct; what landed here closed the deploy-safety hole and turned the test suite into something that would actually catch a regression.

PR #280, opened 2026-08-06
fix(gfs): make large GFS uploads work end-to-end (web + desktop) + raise cap to 16MB

### LINKED ISSUE #281: GFS: add timeout + fail-loud logging to the desktop upload proxy (external/gfs.ts)
Follow-up from #280 (Fable-5 audit finding **H1**).

The operator proxy `control-api/src/routes/gfs/proxy.ts` got an `AbortSignal.timeout` (`CONTROL_API_GFSC_PROXY_TIMEOUT_MS`) plus 4xx/5xx log-and-forward. The **desktop** upload path — `control-api/src/routes/external/gfs.ts` `proxyReadToGfsc` (~L530) and `proxyMutationToGfsc` (~L578) — still `fetch(target, …)` to gfsc with **no AbortSignal, no try/catch, no logging**. Desktop uploads (`desktop-app/src/gfs/uriHandler.ts`) land here.

Impact: a hung gfsc pins these requests indefinitely; a gfsc 5xx is invisible at this hop.

Acceptance: mirror the `routes/gfs/proxy.ts` A4 pattern (timeout via the same env; log+forward gfsc 4xx/5xx; 504 on timeout / 502 on unreachable) in both external/gfs.ts proxy helpers.

### LINKED ISSUE #284: GFS: desktop-app upload cap parity (10MB -> 16MB)
Follow-up from #280 (audit finding **M4**).

Web control-ui raised `GFS_FILE_UPLOAD_MAX_BYTES` to 16MiB; `desktop-app/ui/src/constants/gfsFileUpload.ts` is still `10 * 1024 * 1024`. Not breakage (10 < 16 < all hop caps), but an undocumented feature-parity gap.

Acceptance: decide + document — either raise the desktop cap to 16MiB to match web, or record why it stays at 10MB.

alfredolopez80:
## Summary

Large uploads to **Global File System** failed on **both** paths — the web operator path (Control UI → control-api → gfsc) and the desktop me-path (desktop-app → external-rest-api → profile-control-funnel → control-api → gfsc). In prod a 7.5MB PDF showed **"Request timed out. Check that Control API is reachable."**; the file never landed. This PR makes GFS uploads work end-to-end on **both paths** and raises the cap to **16MB**, proven live on minikube (upload **201** + download **200** + sha256 byte-match) for the web path *and* the desktop me-path.

**Closes #281** (H1 — desktop me-path timeout + fail-loud logging) and **Closes #284** (M4 — desktop upload cap 10→16MB): both were filed as follow-ups and are now implemented and live-validated in this PR, so they auto-close on merge.

It is **not one bug** — a fresh-cluster T2 uncovered **six stacked defects**, each masking the next, none catchable by unit tests (all were environmental: Next.js proxy config, pod memory, undici behavior, socket pooling, and an nginx body cap). This is why local dev "passed": nobody had pushed a file large enough (>3.2MB hits RC1, >7.5MB hits RC2, >1MB hits RC6 on the desktop path) — the defects are size-triggered and live in every deployment (dev included), not prod-only.

> **Branch sync:** up to date with `origin/dev` — merged via `Merge branch 'dev'` (head `5facff5e`), **0 commits behind dev**. Re-validated post-merge.

## Root causes (all reproduced live, exact thresholds)

| # | Defect | Fix |
|---|---|---|
| **RC1** | gfsc base64 validator regex used a grouped quantifier `(?:…{4})*` that overflows V8's regexp backtrack stack (`RangeError`) above ~4.47M base64 chars (~3.2MB raw) → bogus 500 `internal` | single character-class validation (linear, no backtrack) — `gfs-controller/src/api/serve.ts` |
| **RC2** | `next.config.js` rewrite routed `/control-api/*` through Next's internal proxy (`proxyClientMaxBodySize` 10MiB → silent truncation + 30s `proxyTimeout`) = the prod "Request timed out" | remove the rewrite so the app-router handler (streaming, no cap, honours `CONTROL_API_PROXY_TIMEOUT_MS`) owns the path |
| **RC3** | control-api **OOMKilled (exit 137)** — a base64 upload holds ~5 body copies; 4 concurrent crashed the pod | **768Mi/128Mi** in base **and** the minikube overlay (the overlay pinned 256Mi, so the base fix was inert in minikube — caught in audit) |
| **RC4** | the control-ui proxy forwarded `Expect: 100-continue` (curl/SDKs auto-add it >1MB) to undici's fetch, which refuses it (`NotSupportedError`) → 502 on every large upload | add `expect` (+ `proxy-connection`) to `HOP_BY_HOP_HEADERS`; log `err.cause` (fail-loud). Browsers can never send `Expect`, so this only bit curl/SDK clients — fixed regardless |
| **RC5** | a mid-stream 413 from gfsc poisoned control-api's pooled socket → the **next** request got 502 `ECONNRESET` | gfsc sends `Connection: close` when the response is written before the body is fully read (`req.complete === false`); red-first test |
| **RC6** | **desktop me-path only:** the `profile-control-funnel` nginx (external-rest-api → control-api) never set `client_max_body_size`, so nginx's **1m default** 413'd every desktop upload above ~1MB with an HTML body, surfaced as "Control API returned non-JSON body (status=413)". The operator path does not cross this funnel, which is why it already reached 16MB | set `client_max_body_size 25165824` (24MiB, gfsc write-cap parity) in the funnel ConfigMap — `deploy/base/profiles/configmaps.yaml`; manifest-level intent test added |

## Also in this PR
- **Homologous, runtime-configurable timeouts** across every hop (all default 300s): client `GFS_UPLOAD_TIMEOUT_MS`, control-ui proxy `CONTROL_API_PROXY_TIMEOUT_MS`, control-api→gfsc fetch `CONTROL_API_GFSC_PROXY_TIMEOUT_MS` (was **none** — a hung gfsc would pin the connection forever). Each validated (integer, ≤2³¹−1) with a fail-safe default.
- **Desktop me-path resilience (was #281 / H1)**: `external/gfs.ts` `proxyMutationToGfsc`/`proxyReadToGfsc` now bound the gfsc fetch with `AbortSignal.timeout(config.gfscProxyTimeoutMs)` and fail loud — 504 `gfsc_timeout`, 502 `gfsc_unreachable`, and verbatim forwarding of gfsc 4xx/5xx (logged). Mirrors the operator-path proxy.
- **Fail-loud everywhere**: gfsc logs 5xx with stack; control-api logs+forwards gfsc 4xx/5xx verbatim (both planes); control-ui logs the proxy `err.cause` (all were silent before).
- **Upload cap raised 10MB → 16MB on both clients**: web control-ui cap + `GFS_MAX_WRITE_BODY_BYTES` (24MiB, base64 of 16MB ≈ 22.4MB) wired through the host-context-controller gfsc template + gfsc memory 512Mi, **and** the desktop-app client cap (was #284 / M4), now that the desktop me-path accepts 16MB end to end.

## Limits after this PR
- **Upload:** 16MB raw / file on both paths (gfsc body cap 24MiB; funnel cap 24MiB). Beyond ~16MB is memory-bound with base64 (~5× body/hop). **50MB+ → Phase 2 (multipart streaming), deferred** (design in the plan doc).
- **Download:** streamed end-to-end at constant memory; the control-ui proxy exempts GET from its timeout (SSE-safe). The control-api→gfsc read proxy still applies the 300s `CONTROL_API_GFSC_PROXY_TIMEOUT_MS` to GET, so a single download that streams for >300s aborts mid-stream (logged; 200 truncated) — generous for ≤16MB files. Raise the knob or exempt GET there if very large/slow downloads are expected.

## Post-review hardening (multi-agent Fable 5 review)
- **CWE-117 log injection**: `gfsc` `sanitizeForLog` now uses the empty-string CR/LF replace CodeQL recognizes as a barrier, and the error message/stack is sanitized too. Alerts 888/889 resolved.
- **Pre-auth DoS (control-ui)**: removing the Next rewrite left `route.ts` buffering the request body with no cap, before auth. Added a runtime cap `CONTROL_UI_PROXY_MAX_BODY_BYTES` (default 24MiB, homologous to gfsc): 413 on oversize `content-length` + stream-read that never buffers past the cap. Also CR/LF-stripped the logged upstream URL.

## Validation (T0 / T1 / T2 per AGENTS.md)
- **T0**: unit — gfsc **473**, control-api **3878** (+ funnel intent test), control-ui **1274**, desktop-app gfs suites **34**; `tsc --noEmit` ✓; builds ✓; kustomize ✓.
- **T1**: `make build-preflight` ✓.
- **T2 — operator path** (fresh minikube profile): **16MB** upload → **201** (3.5s) + download → **200** (1.2s) + **sha256 byte-match**; **4 concurrent 16MB uploads → 201, control-api restarts=0 (no OOM) at 768Mi**; 10MB → 201; negatives fail loud (>cap → 413, invalid base64 → 400, gfsc timeout → 504).
- **T2 — desktop me-path** (real auth cookie → external-rest-api → funnel → control-api → gfsc, after the RC6 fix + `rollout restart`): **1MB / 8MB / 16MB** upload → **201** + download → **200** + **sha256 byte-match** on all three (was a silent 1MB 413 before).
- **Audited** by an independent adversarial pass which caught the RC3 overlay-override blocker (fixed here) and verified RC1 (200k-case fuzz), RC2/RC4/RC5, the HCC passthrough (live), and all suites. RC6 root-caused + fixed + live-validated by a follow-up investigation pass.

## Scope (multi-service)
`gfs-controller` (RC1/RC5 + body-cap env), `control-api` (RC3 mem + gfsc-fetch timeout/logging on both planes), `control-ui` (RC2/RC4 + client cap + timeout knob), `host-context-controller` (gfsc body-cap passthrough), `desktop-app` (client cap parity), plus `deploy/base/control-plane/*` and `deploy/base/profiles/*` (funnel cap) manifests.

## Follow-ups (tracked as issues — not required for the 16MB cap)
- #282 — H2: route-scoped 24MiB body cap ahead of the 150mb global `express.json` (bounded admin DoS).
- #283 — M1: `Connection: close` on gfsc's 404/health responses (finish RC5).
- #285 — test gaps: RC4 expect-strip unit test, timeout-signal assertion, **large-file e2e** (the gap that let this ship).
- #286 — Phase 2: multipart/streaming uploads (flat memory, enables 50MB+).
- #287 — other nginx gateways (`nginx-workflow-approval-gateway`, `control-api-rpc-gateway`, `host-context-controller-api-gateway`) also lack `client_max_body_size` → same latent 1MB 413 (found while fixing RC6).

**Resolved here (see `Closes` above):** #281 (H1) and #284 (M4).


alfredolopez80:
@claude pls give a full review of this PR

gfs-controller/src/api/serve.ts
```diff
@@ -210,7 +250,17 @@
...
       const { status, body } = toResponse(err);
-      if (!res.headersSent) sendJson(res, status, body);
+      if (status >= 500) {
+        // A 5xx must never be invisible: an unknown throw (e.g. the base64
+        // RangeError that produced a bogus `internal` 500) previously left no
+        // trace anywhere in the system. Log code + stack (never the body/content).
+        console.error(
+          `[gfsc] ${method} ${req.url ?? "?"} -> ${status} ${body.error.code}: ${
+            err instanceof Error ? (err.stack ?? err.message) : String(err)
+          }`
```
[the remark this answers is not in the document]
alfredolopez80: Fixed in a7a93ce3 (initial fix 9947c312). The 5xx fail-loud log line now passes the two user-controlled fields (`req.method`, `req.url`) through a new `sanitizeForLog()` that replaces every C0 control char and DEL — including CR/LF — with `?` and bounds length to 512, neutralizing CRLF log-forging and ANSI-escape injection (CWE-117). The error stack is intentionally multi-line and is not user-tainted, so it is left intact. Added `serve.sanitizeForLog.test.ts` (control-char strip, CRLF forge payload, printable/unicode passthrough, 512-char bound); full gfsc suite 477 passed. The sanitizer uses the `String.replace()` line-break-removal form so CodeQL recognizes the barrier — the alert should auto-close on the next scan of the branch.

gfs-controller/src/api/serve.ts
```diff
@@ -210,7 +263,17 @@
...
       const { status, body } = toResponse(err);
-      if (!res.headersSent) sendJson(res, status, body);
+      if (status >= 500) {
+        // A 5xx must never be invisible: an unknown throw (e.g. the base64
+        // RangeError that produced a bogus `internal` 500) previously left no
+        // trace anywhere in the system. Log code + stack (never the body/content).
+        console.error(
+          `[gfsc] ${sanitizeForLog(method)} ${sanitizeForLog(req.url ?? "?")} -> ${status} ${body.error.code}: ${
+            err instanceof Error ? (err.stack ?? err.message) : String(err)
+          }`
```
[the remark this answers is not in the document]
alfredolopez80: Fixed in 2edf638d. Two changes:

1. **Recognized barrier form.** The prior `sanitizeForLog` used `replace(/[C0-controls]/g, "?")` — functionally correct (it does strip CR/LF), but *not* a sanitizer CodeQL's `js/log-injection` query recognizes: the query requires an empty-string replace whose regex enumerates the newline. `sanitizeForLog` now removes CR/LF first via `replace(/[\n\r]/g, "")` (the recognized barrier), then scrubs any remaining C0 controls + DEL to `"?"`, then bounds length.

2. **The error stack/message is now sanitized too.** This alert anchored on the `err.stack ?? err.message` interpolation (serve.ts:~277), which the previous fix left unsanitized. An underlying throw can echo a decoded path (`serve.ts` calls `decodeURIComponent`, turning `%0a` into a real newline), so it is a log-injection source just like `method`/`req.url`. It now passes through `sanitizeForLog`.

Both the flagged sources are behind the barrier now, so the alert should clear on the next scan of the branch. Tests in `serve.sanitizeForLog.test.ts` updated (CR/LF removed rather than substituted).

jozer-rami (commented):
## Deep review — correctness / deploy-safety / security-tenancy / test-integrity

Every finding below was verified against the code (or refuted and dropped) before it shipped here. Where a claim in the PR body was checkable, I checked it rather than inheriting it.

### What holds up

- **RC1 is real, and its test is a real mutation test.** I ran the pre-PR regex against the exact 5 MiB fixture from `serve.test.ts`: `RangeError: Maximum call stack size exceeded`. The ~3.2 MB threshold reproduces precisely (3.0 MB passes, 3.2 MB throws). The replacement (`length % 4` + `/^[A-Za-z0-9+/]*={0,2}$/`) is semantically equivalent on padding placement and accepts the fixture. Revert the fix and the test goes red — that is the bar.
- **RC5's predicate is correct.** I stood up a real `http.Server` that throws mid-`for await`: `req.complete === false`, the 413 *is* still delivered despite `req.destroyed === true`, and `Connection: close` reaches the client. I could not reproduce the socket-poisoning half on loopback (it needs the in-flight bytes the live repro had), so I neither confirm nor refute the 502 — but the fix is correct and harmless either way.
- Timeout-env parsing (`Number.isInteger && > 0 && <= 2^31-1`) is consistent across all three parsers, and the `timeoutMs: 0` → default case is tested rather than assumed.
- Suites pass locally: 4 control-ui files (42 tests), 2 gfsc files (42 tests).

---

## 🔴 Blocker 1 — RC3 (the OOM fix) never reaches dev or prod

`deploy/base/control-plane/control-api.yaml:136` raises control-api to **768Mi**, and the PR's own comment says *"The minikube overlay must not reduce this."* The GCP overlays live in `keyper-labs/evenfire-infra` and both pin it lower:

- `evenfire-infra/deploy/overlays/gcp-prod/patches/resource-limits.yaml:15` → `memory: 512Mi`
- `evenfire-infra/deploy/overlays/gcp-dev/patches/resource-limits.yaml:19` → `memory: 512Mi`

Kustomize strategic merge: the overlay wins. This is the **exact defect the audit caught for minikube**, reproduced in the environment where the bug was actually reported. Per this PR's own arithmetic (~86 MiB baseline + ~125 MiB transient per in-flight 16 MB upload), 512Mi holds roughly 3 concurrent — so prod keeps the OOMKill (exit 137) → the 502/timeout failure mode this PR claims to close.

**Fix:** a companion `evenfire-infra` PR raising both overlays to 768Mi, merged before or with the promotion. control-ui is fine (base 512Mi == overlay 512Mi), and the base env additions do flow through — the GCP overlays only patch `resources` for these two Deployments.

## 🟠 Major 2 — the desktop me-path is capped at 60 s, not 300 s

The PR claims *"homologous, runtime-configurable timeouts across every hop (all default 300s)."* The desktop client is not one of them:

`gfs:createFile` (`desktop-app/src/ipc.ts:413`) → `appService.ts:1098` `createGfsFile` → `GfsClient.createResource` → `httpClient.ts:107` `signal: withTimeout(...)` → `AbortSignal.timeout(config.requestTimeoutMs)`, with `desktop-app/src/config.ts:558` defaulting `REQUEST_TIMEOUT_MS` to **60000**. `AbortSignal.timeout` is a total deadline, not an idle timer.

**Trigger:** 16 MB raw is a 22.4 MB base64 body ≈ 179 Mbit. On a ~3 Mbit/s uplink the transfer alone is ~60 s, so the desktop client aborts while every downstream hop is still willing to wait 300 s. Invisible in T2 because minikube is loopback (3.5 s). There is no desktop equivalent of control-ui's `GFS_UPLOAD_TIMEOUT_MS`.

Companion, lower risk because they are idle timers rather than total deadlines: the funnel got `client_max_body_size` but not `proxy_read_timeout` / `proxy_send_timeout` / `client_body_timeout`, and nginx defaults all three to 60 s.

**Fix:** a per-call generous timeout on the GFS mutations in `GfsClient`, mirroring `GFS_UPLOAD_TIMEOUT_MS`; add the three `300s` directives to the funnel `http` block.

## 🟠 Major 3 — the cap chain the whole PR rests on is comment-enforced

The invariant is `client cap × 4/3 < GFS_MAX_WRITE_BODY_BYTES <= funnel cap ≈ CONTROL_UI_PROXY_MAX_BODY_BYTES`, spread across 5 files and 2 repos. Only one leg is machine-checked — `control-api/test/network.gateway.intent.test.ts:341` asserts the funnel literal. Nothing asserts:

- that `deploy/base/control-plane/host-context-controller.yaml:73` sets `GFS_MAX_WRITE_BODY_BYTES` at all. Delete that env and gfsc silently falls back to `DEFAULT_MAX_WRITE_BODY_BYTES = 16777216` (`gfs-controller/src/api/serve.ts:99`) → a 16 MB file's 22.4 MB body hard-413s, **while the client-side cap says it should have worked**;
- that control-ui's and desktop-app's `GFS_FILE_UPLOAD_MAX_BYTES` are equal;
- that `ceil(cap × 4/3) < 25165824`.

**Fix:** one intent test asserting the chain, right next to the funnel test that already exists.

## 🟡 Minor 4 — GET is exempt from the timeout in one proxy but not the other two

`control-ui/app/control-api/[...path]/route.ts:163` deliberately exempts GET (*"this timeout can never cut a streaming response"*), while `control-api/src/routes/gfs/proxy.ts:74` and `control-api/src/routes/external/gfs.ts:535` arm `AbortSignal.timeout` on GET too — and the abort applies to the body, not just the headers. On timeout the operator proxy lands in the stream catch with `headersSent === true` and calls `res.end()` (`proxy.ts:134`): a 200 with a short body.

Mitigating: gfsc sets `Content-Length` (`serve.ts:461`) and the proxy passes it through, so Node aborts on the short write and the client sees a failed transfer rather than a silently truncated file. Still, before this PR GET had no timeout at all, so this is a *new* failure mode and the two planes now disagree. **Fix:** exempt GET as control-ui does, or `res.destroy()` instead of `res.end()`.

## 🟡 Minor 5 — the newly-live handler collapses multiple `Set-Cookie` headers

Removing the rewrite promotes `route.ts` from dead code (array-form rewrites are `afterFiles`, which beat catch-all dynamic routes) to the **sole** handler for all `/control-api/*` traffic. `copyResponseHeaders` (`route.ts:126`) uses `Headers.forEach` + `set`, and `forEach` yields `set-cookie` as one comma-joined value. Today control-api sets at most one cookie per response (`control-api/src/routes/admin/auth.ts:103,192,612`), so nothing breaks — but the day a second cookie is added to any response, admin login breaks silently, because express `res.cookie` with `maxAge` emits `Expires=Wed, 21 …` whose comma makes the joined value unsplittable. **Fix:** `upstream.headers.getSetCookie()` and append each.

Same promotion also makes `content-length` stripping apply to *responses* now (it is in `HOP_BY_HOP_HEADERS`), so every control-api response is chunked — GFS download progress in the browser is gone.

## 🟡 Minor 6 — RC2's construct is still live in profile-ui

`profile-ui/next.config.js:7` rewrites `/external-rest-api/:path*` through Next's internal proxy — the identical pattern RC2 removed, with the same 10 MiB `proxyClientMaxBodySize` truncation and 30 s `proxyTimeout`. No large-body route reaches it today (profile-ui has no base64/file upload), so it is latent, not broken. Worth a comment there or folding into #287.

## 🟡 Minor 7 — `err.message` / `err.cause` bypass `stripCrlf`

`route.ts:193` sanitizes `url` but interpolates `message` and `cause` raw into the same sink. Low reachability (WHATWG URL parsing strips CR/LF before a message could echo them), but inconsistent with `gfs-controller/src/api/serve.ts:928`, which does sanitize the error text.

---

## Test-integrity verdict

Sound where it counts — RC1 and RC5 both fail on revert — with two gaps:

- **control-ui is excluded from the CI test matrix**; its only check is `next build`. The new `controlApiProxyRoute.test.ts` and `apiSendTimeout.test.ts` therefore guard nothing in CI: the 413 body cap, the GET exemption, and the timeout parsing can all regress green. Pre-existing, but this PR sharply increases what rides on it.
- **RC4 (the `expect` strip) has zero coverage** — no test references the header, in the one file whose job is now header handling. Tracked as #285, but it is three lines and belongs here.

**Bottom line:** the six root causes are real and correctly fixed in-repo. What blocks merge-to-value is Blocker 1 — without the `evenfire-infra` companion, the memory fix is inert exactly where the bug was reported, the same way it was inert in minikube before the audit caught it.

deploy/base/control-plane/control-api.yaml
```diff
@@ -120,10 +125,18 @@ spec:
...
-            memory: 256Mi
+            # GFS uploads arrive base64-in-JSON and are fully buffered: one upload
+            # holds ~5 transient copies (express raw buffer, JSON.parse input, parsed
+            # contentBase64 string, JSON.stringify output, undici write buffer). The
+            # gfsc write cap is 24MiB (GFS_MAX_WRITE_BODY_BYTES), i.e. a 16MB raw file
+            # → ~22.4MB body → ~125MiB transient over the ~86MiB idle baseline. At 256Mi
+            # 2 concurrent uploads OOMKilled (exit 137) the pod; 768Mi holds ~5 concurrent
+            # worst-case uploads with margin. The minikube overlay must not reduce this
+            # (see deploy/overlays/minikube/patches/resource-limits.yaml).
+            memory: 768Mi
```
jozer-rami, line 141: 🔴 **Blocker — this 768Mi never reaches dev or prod.**

The comment right above says the minikube overlay must not reduce this, and that overlay was correctly fixed. But the GCP overlays live in `keyper-labs/evenfire-infra` and both pin control-api lower:

- `gcp-prod/patches/resource-limits.yaml:15` → `memory: 512Mi`
- `gcp-dev/patches/resource-limits.yaml:19` → `memory: 512Mi`

Kustomize strategic merge: the overlay wins, so RC3 is inert in the exact environment where the 7.5 MB PDF failure was reported. By the arithmetic in this comment (~86 MiB baseline + ~125 MiB per in-flight 16 MB upload), 512Mi holds ~3 concurrent uploads — the OOMKill (exit 137) survives.

Needs a companion `evenfire-infra` PR raising both overlays to 768Mi, landed before or with the promotion. control-ui is fine: base 512Mi matches the overlay's 512Mi.

deploy/base/control-plane/host-context-controller.yaml
```diff
@@ -66,6 +66,17 @@ spec:
           value: clerum/mcp-host:0.9.5
         - name: CONTEXT_MAPPER_EGRESS_PROXY_IMAGE
           value: clerum/nginx-egress-proxy:0.1.0
+        # GFS upload ceiling: gfsc write-body cap (bytes), passed through to the gfsc
+        # pod. Raised from the 16MiB default to 24MiB so a 16MB raw file (base64 body
+        # ≈ 22.4MB) is accepted with margin. Keep in step with the client cap
+        # (GFS_FILE_UPLOAD_MAX_BYTES in control-ui) and gfsc memory below.
+        - name: GFS_MAX_WRITE_BODY_BYTES
+          value: "25165824"
```
jozer-rami, line 74: 🟠 **Major — nothing asserts this env exists.**

The whole 16 MB cap depends on `client cap × 4/3 < 25165824`, but the only machine-checked leg of that chain is the funnel literal in `network.gateway.intent.test.ts:341`.

Delete this env (or have an overlay drop it) and gfsc silently falls back to `DEFAULT_MAX_WRITE_BODY_BYTES = 16777216` (`gfs-controller/src/api/serve.ts:99`). A 16 MB file's 22.4 MB body then hard-413s **while the client-side cap says it should have worked** — the exact silent-mismatch class this PR set out to kill.

Worth one intent test next to the funnel one, asserting: this env is set to 25165824, control-ui and desktop-app `GFS_FILE_UPLOAD_MAX_BYTES` are equal, and `ceil(cap * 4/3) < 25165824`.

deploy/base/profiles/configmaps.yaml
```diff
@@ -24,6 +24,13 @@ data:
 
       resolver kube-dns.kube-system.svc.cluster.local valid=10s ipv6=off;
 
+      # nginx defaults client_max_body_size to 1m, which 413s me-path GFS
+      # uploads (base64-in-JSON) at ~1MB. Homologate with the gfsc write cap
+      # (GFS_MAX_WRITE_BODY_BYTES = 25165824 bytes, 24MiB) so the me-path
+      # accepts the same payloads the operator path already does; gfsc remains
+      # the authoritative size gate downstream.
+      client_max_body_size 25165824;
```
jozer-rami, line 32: 🟠 **Major (companion) — the body cap is homologated here, the timeouts are not.**

nginx defaults `proxy_read_timeout`, `proxy_send_timeout` and `client_body_timeout` to 60s. This block sets none of them, so the funnel is the one hop on the me-path that never got the 300s homologation the PR summary claims for "every hop".

These are idle timers rather than total deadlines, so they are lower risk than the desktop client's hard 60s deadline (see the comment on `desktop-app/ui/src/constants/gfsFileUpload.ts`) — but a control-api that takes >60s to answer after the body is proxied still 504s here.

Suggest adding `proxy_read_timeout 300s; proxy_send_timeout 300s; client_body_timeout 300s;` alongside this line.

control-api/src/routes/gfs/proxy.ts
```diff
@@ -59,14 +59,55 @@ export function registerGfsProxyRoute(router: Router): void {
...
+          method: req.method,
+          headers,
+          body:
+            req.method === 'GET' || req.method === 'HEAD'
+              ? undefined
+              : JSON.stringify(req.body ?? {}),
+          // Homologous with the browser client and control-ui proxy (all 300s by
+          // default). Without this the fetch had NO timeout: a slow/large upload or
+          // a hung gfsc would pin the connection indefinitely.
+          signal: AbortSignal.timeout(config.gfscProxyTimeoutMs),
```
jozer-rami: 🟡 **Minor — this timeout also applies to GET, and the abort covers the response body, not just the headers.**

`control-ui/app/control-api/[...path]/route.ts:163` deliberately exempts GET so the timeout "can never cut a streaming response". This hop does the opposite: a download still streaming at 300s aborts, lands in the stream catch below with `headersSent === true`, and `res.end()` (line 134) emits a 200 with a short body. Same in `routes/external/gfs.ts:535`.

It is not silent in practice — gfsc sets `Content-Length` (`serve.ts:461`) and it is passed through, so Node aborts on the short write — but before this PR GET had no timeout at all, so this is a new failure mode, and the two planes now disagree on the rule.

Either exempt GET the way control-ui does, or `res.destroy()` instead of `res.end()` in that catch so truncation can never read as a clean EOF.

control-ui/next.config.js
```diff
@@ -12,7 +12,13 @@ const nextConfig = {
     root: path.join(__dirname, '..'),
   },
   async rewrites() {
-    const controlApiDestination = process.env.CONTROL_API_INTERNAL_URL || 'http://127.0.0.1:8090'
+    // NOTE: `/control-api/:path*` is intentionally NOT rewritten here. It is served by
```
jozer-rami, line 15: 🟡 **Minor — removing this rewrite promotes `route.ts` from dead code to the sole handler for *all* `/control-api/*` traffic, so two of its pre-existing behaviours are now live.**

Array-form rewrites are `afterFiles`, which beat catch-all dynamic routes — so the handler never ran before this change. Two consequences worth a look:

1. `copyResponseHeaders` (`route.ts:126`) uses `Headers.forEach` + `set`, and `forEach` yields `set-cookie` as one comma-joined value. Today control-api sets at most one cookie per response (`routes/admin/auth.ts:103,192,612`) so nothing breaks — but the day a second cookie lands on any response, admin login breaks silently, because express `res.cookie` with `maxAge` emits `Expires=Wed, 21 …` and that comma makes the joined value unsplittable. `upstream.headers.getSetCookie()` + append is the fix.
2. `content-length` is in `HOP_BY_HOP_HEADERS` for the *response* direction too, so every control-api response is now chunked. Harmless except that GFS download progress in the browser is gone.

Separately: `profile-ui/next.config.js:7` still has the identical rewrite construct to `/external-rest-api/:path*`, with the same 10 MiB truncation and 30s `proxyTimeout`. Latent today (profile-ui has no base64/file upload), but worth a comment there or folding into #287.

desktop-app/ui/src/constants/gfsFileUpload.ts
```diff
@@ -1,2 +1,4 @@
-export const GFS_FILE_UPLOAD_MAX_BYTES = 10 * 1024 * 1024
-export const GFS_FILE_UPLOAD_MAX_MEGABYTES = 10
+// Kept in parity with the web control-ui cap (control-ui/app/constants/gfsFileUpload.ts).
+// base64 inflates ×1.34 → 16MiB raw ≈ 22.4MB body, under gfsc's 24MiB write cap.
+export const GFS_FILE_UPLOAD_MAX_BYTES = 16 * 1024 * 1024
```
jozer-rami, line 3: 🟠 **Major — the cap is now 16 MB but the desktop client still gives up at 60s.**

The upload path is `gfs:createFile` (`desktop-app/src/ipc.ts:413`) → `appService.ts:1098` → `GfsClient.createResource` → `httpClient.ts:107` `signal: withTimeout(...)` → `AbortSignal.timeout(config.requestTimeoutMs)`, and `desktop-app/src/config.ts:558` defaults `REQUEST_TIMEOUT_MS` to **60000**. That is a total deadline, not an idle timer.

16 MB raw is a 22.4 MB base64 body ≈ 179 Mbit; on a ~3 Mbit/s uplink the transfer alone is ~60s, so the client aborts while every downstream hop is still willing to wait 300s. T2 could not surface this because minikube is loopback (3.5s).

There is no desktop equivalent of control-ui's `GFS_UPLOAD_TIMEOUT_MS`. Suggest a per-call generous timeout on the GFS mutations in `GfsClient` so the desktop path is genuinely homologous with the web one.

deploy/base/control-plane/control-api.yaml
alfredolopez80: **Confirmed — addressed with a companion infra PR + a base-comment fix.**

Raising both GKE overlays to 768Mi: **keyper-labs/evenfire-infra#41** (control-api `limits.memory` 512Mi → 768Mi in `gcp-prod` and `gcp-dev`; must merge before/with the next GKE deploy or RC3 stays inert in dev/prod).

Also fixed the base comment (5c3f67ac): it previously warned only about the minikube overlay — the exact omission that let this gap through — and now names EVERY overlay (minikube + the two private infra ones) as a hard ≥768Mi requirement.

deploy/base/control-plane/host-context-controller.yaml
alfredolopez80: **Fixed in 77830f9f.** Added an intent test (`control-api/test/network.gateway.intent.test.ts`) that machine-checks the whole chain: this env is set to `25165824`; the control-ui and desktop-app `GFS_FILE_UPLOAD_MAX_BYTES` are equal; and `ceil(cap × 4/3) < 25165824`. Delete this env now and CI fails loud instead of gfsc silently falling back to its 16 MiB default and 413-ing a 16 MB file the client said would fit.

deploy/base/profiles/configmaps.yaml
alfredolopez80: **Fixed in f080b833.** Added `proxy_read_timeout 300s; proxy_send_timeout 300s; client_body_timeout 300s;` to the funnel `http` block, so it is no longer the one me-path hop that 504s a large upload/download at nginx's 60s default while every downstream hop still waits 300s.

control-api/src/routes/gfs/proxy.ts
alfredolopez80: **Fixed in ed21f51f.** Both control-api gfs proxies (`routes/gfs/proxy.ts` and the `routes/external/gfs.ts` read path) now use a **header-only** deadline: an `AbortController` bounds only the wait for gfsc to START responding and is cleared (in `finally`) once headers arrive, so a streamed download is never aborted mid-body. Uploads keep the total deadline, and the `TimeoutError → 504` mapping is preserved (`signal.aborted || name === 'TimeoutError'`). This matches control-ui's GET exemption while keeping the hung-gfsc protection #281 added.

control-ui/next.config.js
alfredolopez80: **Fixed in ed21f51f** (in `control-ui/app/control-api/[...path]/route.ts`, the handler this rewrite removal promotes to live). `copyResponseHeaders` now uses `upstream.headers.getSetCookie()` + `append` (each `Set-Cookie` preserved separately, so a 2-cookie response no longer collapses), and `content-length` is kept on the **response** direction (browser download progress restored — it is still stripped on the request). New assertions in `controlApiProxyRoute.test.ts` cover both, plus an `Expect`-strip (RC4) guard. Dead-code→live is confirmed: the array-form rewrite was `afterFiles` and beat the catch-all, so the handler was inert before this change.

desktop-app/ui/src/constants/gfsFileUpload.ts
alfredolopez80: **Fixed in f080b833.** Added `GFS_UPLOAD_TIMEOUT_MS` (default 300000, parity with control-ui) and threaded a per-call `timeoutMs` through `httpClient.withTimeout`/`requestJson` and the `GfsTransport` interface; `GfsClient` create/replace uploads now pass `config.gfsUploadTimeoutMs` instead of the 60s `REQUEST_TIMEOUT_MS`, so a slow-uplink 16 MB upload no longer aborts at 60s while downstream waits 300s. `gfsUriHandler.test.ts` asserts the upload deadline reaches the transport. The funnel's idle timers were raised to 300s in the same commit.

alfredolopez80:
## Response to @jozer-rami's deep review — all 7 findings addressed

Thanks for the depth — every finding reproduced against the code, and an independent adversarial pass confirmed **none were false positives**. All 7 are now fixed in-repo or clarified, with a companion infra PR for the one out-of-repo blocker.

| Finding | Verdict | Resolution |
|---|---|---|
| 🔴 **B1** RC3 inert on GKE (512Mi overlays) | Confirmed (verified in the private infra repo) | Companion **keyper-labs/evenfire-infra#41** raises control-api to 768Mi in `gcp-dev`/`gcp-prod`; base comment now names every overlay as a ≥768Mi requirement (`5c3f67ac`). Merge #41 before/with the next GKE deploy. |
| 🟠 **M2** desktop capped at 60s + funnel idle timers | Confirmed | `GFS_UPLOAD_TIMEOUT_MS` (300s, parity) threaded through the desktop `GfsClient` upload path; funnel `proxy_read/send`+`client_body_timeout` = 300s (`f080b833`). |
| 🟠 **M3** cap chain comment-only | Confirmed | Intent test asserts the HCC env = 25165824, both client caps equal, `ceil(cap×4/3) < cap` (`77830f9f`). |
| 🟡 **m4** GET timeout in 2 proxies | Confirmed | Header-only deadline on both control-api read proxies — bounds the header wait, never the streamed body; uploads keep the total deadline (`ed21f51f`). |
| 🟡 **m5** Set-Cookie collapse + content-length | Confirmed | `getSetCookie()`+append; content-length kept on responses (download progress restored); tests added (`ed21f51f`). |
| 🟡 **m6** RC2 construct live in profile-ui | Confirmed (latent) | Warning comment tied to #287 — do not add an upload route behind that rewrite without replacing it first (`5c3f67ac`). |
| 🟡 **m7** `err.message`/`cause` bypass stripCrlf | Confirmed | Both now routed through `stripCrlf` in the route.ts fail-loud log (`ed21f51f`). |
| **Test-integrity** control-ui not in CI + RC4 no coverage | Confirmed | control-ui vitest now runs in CI (`5c3f67ac`); RC4 `Expect`-strip test added (`ed21f51f`). The GFS body-cap / timeout / Expect tests now actually guard in CI. |

**Local validation (all green):** control-api 3879 · control-ui 1279 · desktop-app 1517 (0 unhandled) · gfsc 478 · typechecks clean.

One note on m4: the pre-#280 state (GET with no deadline at all) was also a defect — the header-only fix keeps #281's hung-gfsc protection while removing the mid-stream truncation you flagged, rather than reverting to unbounded.

All six inline threads are answered and resolved. The only remaining merge gate is the required review approval + merging the infra companion.

zach88:
Took a deep pass through this one across all six services + manifests. Nothing here blocks the merge: the happy path is correct on both routes, the cap chain is coherent end to end (checked the base64 arithmetic explicitly — client caps raw at 16MiB, every base64-bearing hop sits at 24MiB=25165824, ~2.67MiB of headroom, so nothing dies in no-man's-land), suites are green and security came back clean.

The findings cluster in two places: one consumer the PR should have updated but didn't, and a batch of tests that look like they guard the headline fixes but don't. None are correctness bugs on the success path — but a few of them mean a green suite wouldn't catch a regression in exactly the code this PR is built around.

## Medium

**1. [R1-M1] Desktop me-path masks gfsc's 502/504 as a generic 500**
`external-rest-api/src/routes/gfs.ts:17` (+ `external-rest-api/src/app.ts:59`). This PR adds `504 gfsc_timeout` / `502 gfsc_unreachable` on the me-path in `control-api/src/routes/external/gfs.ts`, and the body promises "verbatim forwarding of gfsc 4xx/5xx". But the next hop, `external-rest-api`, has `PROPAGATED = {400,401,403,404,409,410,412,422,429,503}` — no 502/504 — so `forwardControlApiError` falls through to `next(error)`, and the global handler collapses every 5xx to `500 Internal Server Error`. `external-rest-api/src/routes/gfs.ts` isn't in the diff, but it's the consumer this PR should have updated. Traced each hop: a desktop upload with a wedged gfsc returns `504 {error:"gfsc_timeout"}` from control-api and lands as `500` at the desktop. The request does fail loud (no hang — the "no longer pins the upload" part of #281/H1 holds), but the documented codes are unobservable, which makes the body's "verbatim 4xx/5xx" claim false at the actual client. One-line fix: add 500/502/504 to `PROPAGATED` (or forward the upstream status unconditionally). Note the plain 413 does reach the desktop — the global handler forwards 4xx — only the funnel's non-JSON 413 becomes a 502→500.

**2. [R1-M2] `proxyMutationToGfsc` uses a different timeout mechanism than its two siblings, with `text()` outside the try/catch**
`control-api/src/routes/external/gfs.ts:624,640`. The operator proxy and the me-path read proxy both use an `AbortController` + `setTimeout` cleared once headers arrive (header-only deadline). The mutation proxy uses `AbortSignal.timeout(...)` — a total deadline, never cleared, and it also bounds `await upstream.text()` at :640, which sits outside the try/catch. The comment says "Homologous with the operator proxy" but the mechanism differs. If gfsc sends headers on a mutation and then stalls the body, `text()` rejects uncaught → generic 500 instead of the 504 fail-loud (feeds back into #1). Low practical impact since mutation responses are small JSON, but worth unifying onto one shared helper.

**3. [R1-M3] The timeout-classification test exercises a dead branch; the real `AbortError` path is never tested**
`control-api/test/routes.gfsProxy.test.ts:96`. Reproduced: the real proxy emits `AbortError` (`signal.aborted===true`) on timeout, but the test injects `err.name='TimeoutError'`, which these proxies never produce — it passes through `|| err.name === 'TimeoutError'` (dead branch) and never exercises `deadline.signal.aborted`. Break the real branch and a production timeout silently degrades to 502 with the test still green. Trigger a real timeout instead (tiny `gfscProxyTimeoutMs` + a fetch that respects the signal).

**4. [R1-M4] The me-path read proxy (`proxyReadToGfsc`) has zero coverage**
`control-api/src/routes/external/gfs.ts:530-597`. The `GET /external/gfs/resources/:rid/content` path this PR rewrote (header-only deadline, 504/502, verbatim forward, streaming) isn't exercised by any fetch mock — grep is clean in `routes.externalGfs.test.ts`. The "#281/H1 me-path fails loud" intent is unverified on the read side, even though its operator-path twin is tested.

**5. [R1-M5] The desktop timeout regression test passes against the previous head**
`desktop-app/test/gfsUriHandler.test.ts:421-460`. Reproduced green against `abb5ef49`: it asserts `timeoutMs: config.gfsUploadTimeoutMs`, but on the parent that value is `undefined` and vitest treats `{timeoutMs: undefined}` as equal to a call without the key — so the assertion is vacuous w.r.t. the change and doesn't prove the fix covers the reported 60s-abort bug. Assert the concrete literal (`timeoutMs: 300000`).

**6. [R1-M6] RC3's memory invariant has no executable test**
`control-api/test/network.gateway.intent.test.ts`. The manifest comment at `deploy/base/control-plane/control-api.yaml:135` literally says *"EVERY overlay must keep >=768Mi or the OOM returns"* — which is exactly the RC3 class (minikube overlay pinning 256Mi made the base fix inert). Nothing asserts base memory or that an overlay doesn't reduce it. Re-pin minikube to 256Mi in a future edit and the suite stays green while the 7.5MB-PDF OOM comes back. Add an assertion that control-api base `limits.memory == 768Mi` and the minikube patch doesn't drop below it.

**7. [R1-M7] The new `GFS_MAX_WRITE_BODY_BYTES` passthrough has no test, though its siblings do**
`host-context-controller/src/k8s/gfsFactory.ts:266`. It was added to the same `passthroughEnvEntries` array whose siblings (`GFS_SYNC_COPY_*`) are covered in both the omitted-when-undefined and passed-verbatim cases, but left out of the assertions. A typo in the env name or a dropped passthrough would leave gfsc on its 16MiB default (413 at 16MB) with no test noticing. Extend the "passes configured synchronous copy limits verbatim" test to include `maxWriteBodyBytes → GFS_MAX_WRITE_BODY_BYTES`.

**8. [R1-M8] control-ui proxy timeout code default (30s) isn't homologous with the rest (300s)**
`control-ui/app/control-api/[...path]/route.ts:7`. `DEFAULT_CONTROL_API_PROXY_TIMEOUT_MS = 30_000`, while the desktop client, control-ui client, and control-api→gfsc all default to 300000. Mitigated by the shipped manifest (`control-ui.yaml:38` sets 300000) and the comment acknowledges it, but any environment that doesn't set the env (new overlay, dev, a test harness) caps operator uploads at 30s while downstream waits 300s — the upstream cuts first and the user sees a spurious 504. Make the code default 300000 and keep the env override.

## Minor

- **[R1-L1] gfsc's fallback default (16MiB) is smaller than a 16MB upload's body.** `gfs-controller/src/api/serve.ts:99-108` — base64(16MB) ≈ 21.33MiB > 16MiB, so with `GFS_MAX_WRITE_BODY_BYTES` absent/invalid a 16MB upload 413s. Now mitigated by the intent test asserting the manifest env, so this is defense-in-depth; consider bumping the default to ≥24MiB so a missing env can't reintroduce the 413.
- **[R1-L2] external-rest-api→control-api fetch has no timeout of its own.** `external-rest-api/src/controlApiClient.ts:67` — doesn't hang forever only because control-api bounds itself at 300s; a control-api that stalls before arming its timer would pin the desktop request. Belt-and-suspenders: give it a deadline ≥ control-api's.
- **[R1-L3] Pre-auth memory amplification in the control-ui proxy (security residual, mitigated).** `route.ts:169-173` + the funnel `client_max_body_size` bump — dropping the Next rewrite left full-heap buffering of up to 24MiB per request pre-auth; ~20 concurrent 24MiB conns could pressure the 512Mi pod. Mitigated by the hard 24MiB cap, the 512MiB ceiling, and the explicit memory sizing. Standard reverse-proxy tradeoff; a follow-up could stream the body upstream instead of `arrayBuffer`.
- **[R1-L4] The "forwards 5xx/413 verbatim" tests pass against the previous head.** `control-api/test/routes.gfsProxy.test.ts:126,145` — only the 504/502 tests fail on the parent (genuine regression coverage); the 5xx/413 ones pass because the parent already forwarded verbatim, so the actually-new buffering + fail-loud logging on the `>=400` branch is unasserted.
- **[R1-L5] Env-parse invariant untested + hand-written RC5 fixture.** `gfs-controller/src/api/serve.ts:100` (the reject-`"24MiB"`/`-1`/`0` fallback lives only in prose) and `serve.test.ts:534` (RC5 sets `req.complete` by hand on a `Readable.from` that isn't an `IncomingMessage`, not derived from the real producer). Low risk, noting for completeness.

## What checked out clean (verified, not assumed)

- **Cap chain coherent end to end** across both routes — 3 independent passes with explicit base64 math; max body 22.37MB clears every 24MiB hop; MB vs MiB units consistent everywhere.
- **RC1** (linear base64 validator): 3M random strings → 0 divergences vs the old regex; the old one overflows (`RangeError`) around 6.99M chars, the new one doesn't. Red-first test confirmed.
- **RC5** (`Connection: close` on `req.complete===false`): predicate correct, doesn't over-close (404/GET-without-body keep keep-alive), honored by undici, control-api's own pool isn't poisoned.
- **RC3** memory: base **and** minikube overlay both at 768Mi (no longer inert in minikube). GKE overlays live in the private infra repo (companion #41), documented in the manifest comment.
- **RC4** (`expect`/`proxy-connection` hop-by-hop strip): right direction, case-insensitive.
- **RC6** funnel `client_max_body_size 25165824` matches the gfsc write cap, with a manifest intent test.
- **Pre-auth body cap** (control-ui proxy): actually cuts via streaming (`readBodyCapped`), doesn't buffer everything first; the `Transfer-Encoding: chunked` / no-content-length bypass is covered by a test.
- **Security: clean** — no NetworkPolicies touched, no securityContext relaxed, no SSRF (fixed host), no log injection (sanitizeForLog/stripCrlf on every new sink), no header smuggling.
- **GET timeout exemption coherent** on both planes (header-only deadlines → downloads never truncate). Closes out the earlier m4 note.
- **Desktop upload timeout wiring** correct: create/replace use `gfsUploadTimeoutMs` (300s), not the app's 60s default.

Suites: control-ui 1279, control-api (touched subset) 96/1 skip, gfsc 42, desktop 28+45, hcc 1041 — all green; typecheck clean.

On size: this is 44 files / 6 services, over the soft budget. The stacked root causes explain most of the breadth, just flagging it.

jozer-rami (approved):
## Approved — all 7 findings verified fixed at `5c3f67ac`

I re-checked each fix against the code and **mutation-tested every new assertion** rather than reading the summary. All six new tests fail when the fix they cover is reverted, which is the bar.

| # | Fix | How I verified it |
|---|---|---|
| 🔴 **B1** | GKE overlays | `keyper-labs/evenfire-infra#41` diff confirmed: `512Mi → 768Mi` on the **control-api** container in both `gcp-dev` and `gcp-prod` — correct target, not control-ui. |
| 🟠 **M2** | desktop + funnel 300s | `withTimeout(signal, timeoutMs)` threaded through `requestJson` to both upload call sites. **Mutation:** dropping `timeoutMs` from the two call sites → 2 tests fail. A malformed `GFS_UPLOAD_TIMEOUT_MS` is safely caught by the `Number.isFinite` guard rather than reaching `AbortSignal.timeout`. |
| 🟠 **M3** | cap chain | **Mutation:** deleting the HCC env → test fails with the intended message; drifting the desktop cap to 32 MB → also caught. Real chain test. |
| 🟡 **m4** | header-only deadline | Verified end-to-end against a real trickling HTTP server: the old total-deadline pattern `THREW TimeoutError — aborted mid-stream`; the new pattern returns `status=200` with the **complete** body. `clearTimeout` in `finally` lands after headers and before body streaming — correct pattern, and it keeps #281's hung-gfsc protection as you noted. |
| 🟡 **m5** | Set-Cookie + content-length | **Mutation:** reverting `copyResponseHeaders` to `forEach`+`set` → test fails. |
| 🟡 **m6** | profile-ui | Warning comment tied to #287; confirmed latent-only (no base64/upload route there today). |
| 🟡 **m7** | log sanitizing | Both `message` and `cause` now routed through `stripCrlf`. |

**On the CI gate** — this was the change I was most wary of, since turning on a previously-excluded suite can turn every other PR red. It's clean:

- PR branch: **129 files / 1279 tests green**
- `origin/dev`: **128 files / 1267 tests green** — so the gate won't break anything already in flight
- gfsc full suite: 478 passed / 4 skipped · desktop `gfsUriHandler`: 28 passed
- **Mutation on the new RC4 test:** removing `'expect'` from `HOP_BY_HOP_HEADERS` → test fails. Worth calling out because that test *could* have been vacuous — undici does not strip `expect` at `NextRequest` construction, so it genuinely proves the proxy strips it.

Nice work on the turnaround, and on picking the header-only deadline over reverting to unbounded on m4 — that was the right call.

---

### Two non-blocking notes (not conditions — approving regardless)

**1. The 504 timeout tests no longer exercise the path production takes.** Switching to `AbortController` changed which error the runtime emits:

```
AbortController.abort() -> name=AbortError  | signal.aborted=true    <- production now
AbortSignal.timeout()   -> name=TimeoutError                          <- what the tests simulate
```

So in `proxy.ts:83` / `external/gfs.ts:543`, only the `deadline.signal.aborted` clause can ever fire in production — but `routes.gfsProxy.test.ts` and `routes.externalGfs.test.ts` still throw a synthetic `TimeoutError`. I deleted the `deadline.signal.aborted ||` clause and **57/57 tests stayed green**. Concrete regression it would miss: gfsc hangs → 502 `gfsc_unreachable` instead of 504 `gfsc_timeout`, CI green. Two-line fix — throw a `DOMException(msg, 'AbortError')` from the fetch mock in each 504 test. Happy for this to be a follow-up.

**2. Deploy ordering.** Merging this PR changes nothing on GKE, but infra #41 needs to land before the next GKE deploy or RC3 is inert in dev/prod exactly as it was in minikube. You already have it open — just flagging the ordering.

Minor nits, no action needed: `proxyMutationToGfsc` (`external/gfs.ts:616`) still keeps the total deadline armed during `await upstream.text()`, which sits outside the try/catch, so an abort there surfaces as a 500 rather than the 504 the branch above returns. And the M3 regex `GFS_MAX_WRITE_BODY_BYTES[\s\S]{0,80}?value:` would also match a comment naming the var within 80 chars of any `value:` — fine today, mildly brittle.

PR #292, opened 2026-08-07
feat(desktop-app): Plugin UI SDK — consented context bridge + gfs.open

andypavia:
## Summary
Introduces the **Plugin UI SDK**: an untrusted plugin (a web component in an isolated Electron `WebContentsView`) can request host capabilities through a **consented context bridge** (preload → IPC → broker), gated by explicit user consent and persisted per `(envKey, userId, pluginId, capability)`.

- Consented context bridge + capability catalog, consent gate/store, per-plugin rate limiting and audit log
- `gfs.open` capability: open `gfs://` links (and `<a href>` clicks) in the Desktop viewer using the **user's** session
- Plugin consent modal now centers over the plugin embed rather than the whole desktop window
- Plugin-author guide + spec docs

## Security review
Ran a full-branch review (4 partitioned reviewers + verification). The trust boundary is sound and fails closed — identity is not plugin-asserted, consent cannot be bypassed, no path traversal / SSRF / DOM XSS. **One Low-severity finding:**

- `gfs.open` returns `{ opened: true|false }` to the plugin **without consent**, acting as an authorization/existence oracle for user-supplied `gfs://` paths (bounded by rate limits, visible positive hits, and unguessable identity rids). Recommend making it `requiresConsent: true` or returning a constant result.


andypavia:
@claude please do a deep review

zach88:
Took a deep pass through this one across four area reviews plus a security and a contract-consistency sweep. The trust boundary is genuinely well built and fails closed — identity is host-derived (never plugin-asserted), consent can't be bypassed, the `(envKey, userId, pluginId, capability)` key has no collision surface, and the embed's WebContentsView isolation is intact. No blockers, nothing High. Everything below is **optional** — none of it needs to land before merge. The Mediums are worth a look, but your call on all of them.

## Medium (all optional)

1. **[R1-M2] Rate limiter over-throttles** — `src/pluginRateLimiter.ts:73-83`. A request denied by the hour bucket still spends a token from the minute bucket (same for the global pair), so the limiter ends up stricter than the spec. Repro: with `perMinute:5, perHour:2`, two allowed calls drain the hour bucket; the next three get denied by the hour bucket but each still decrements minute, and a sixth call gets denied by *minute* (`retryAfterMs:12000`) after only 2 successful requests. This is the sibling of the guard you already wrote for the cap↔global boundary — the intra-pair case just got missed. Fix: take from both buckets of a pair tentatively and refund all consumed tokens on any denial, or check-all-then-commit.

2. **[R1-M1] `gfs.open` hands `reason` back to the plugin** — `src/pluginSdkCapabilities.ts:409-437` / `src/pluginSdkRuntime.ts:299-320`. The descriptor says "the plugin gets back `{ opened }` and nothing else" (and the PR body repeats it), but the failure path returns `{ opened:false, reason:'not_found' }` and the broker forwards it verbatim via `sdkOk(value)` — with no consent. So the plugin learns `not_found` (absent / no access) vs `unavailable` (transient), which sharpens the existence oracle the body already flags as Low. Cheap fix: normalize to a strict `{ opened }` on the plugin path (drop `reason`), or collapse `reason` to one opaque value. Also worth fixing the comment/body line since it's currently inaccurate.

3. **[R1-M5] Consent gate has an ordering window that can hang a prompt for 120s** — `src/pluginConsentGate.ts:261-282`. `runPrompt` assigns `this.pending` only *after* `await isGranted` (reads the consent file) and `await setSurfaceVisible(false)`. If the plugin unmounts in that window (`unpinAllSandboxUiSurfaces → releasePlugin → resetPlugin`), `resetPlugin` sees `pending === null`, cancels nothing, and the awaited prompt sits until the 120s timeout. Reproduced as a flaky test — `src/__tests__/pluginConsentGate.test.ts:163` ("cancels a pending prompt when its plugin goes away") fails ~1/3 under I/O load because it only waits a single `setTimeout(0)`. Fix: register the pending *before* `await setSurfaceVisible(false)` (or re-check `this.pending?.promptId` after the await), and have the test poll for `prompts.length === 1` instead of one tick.

4. **[R1-M4] Revoke/clear-activity failures are silent** — `ui/src/hooks/domain/usePluginPermissionsController.ts:679-681` + `ui/src/components/PluginPermissions/index.tsx:480,507,553`. `controller.revoke(...)` returns `mutateAsync(...)`, which rejects on failure, but the callers are `onClick={() => void controller.revoke(...)}` — the `void` attaches no `catch`, and `controller.error` only reflects query errors, never mutation errors. If `pluginSdk:revoke` rejects, the user sees no change and believes the permission was revoked when it wasn't (plus an unhandled rejection). Surface the mutation error and attach a catch.

5. **[R1-M3] Consent modal backdrop only covers the embed rect — your call on whether that's intended.** `ui/src/components/PluginConsentModal/index.tsx:242-251` + `ui/src/App.tsx:127-132`. With `slotRect`, `backdropStyle` overrides `inset:0` with `top/left/width/height`, so the dimmer/click-blocker covers only the embed. The rest of the app (sidebar, header, other views) stays un-dimmed and clickable, with no focus trap — yet the component comment says "the only thing on screen while it is up" and sets `aria-modal="true"`. It's **not** plugin-exploitable (main hides the plugin's WebContentsView while the prompt is up), which is why it's not High. The real question is a design one and it's yours to make: does the threat model require full-window modality for the consent dialog, or is hiding the embed enough? If full modality is intended, restore the full-window backdrop (matches the `ConfirmDialog` pattern); if embed-scoped is deliberate, worth dropping the `aria-modal`/comment so they don't overstate it.

## Test-quality notes (also optional, but they're why a couple of the above slipped)

6. **[R1-M6] Rate limiter has no dedicated unit test** — `src/pluginRateLimiter.ts`. It's only exercised transitively by broker/surface tests that assert allow/deny, not the two-window refund accounting, which is exactly why #1 got through. Given it's a trust-boundary component, a direct test for refund symmetry / `retryAfter` / continuous refill would be worth it.

7. **[R1-M7] `gfs.open` test fixture can't emit the real shape, so it masks #2** — `src/__tests__/pluginSdkCapabilities.test.ts:294-299`. The `openGfsResource` fake always returns `{ opened:true }`, but the real producer returns `{ opened, reason }`. The "returns only whether a viewer opened" test asserts `toEqual({ opened:true })` against a fixture that by construction never carries `reason`, so it certifies the intended contract, not the real one, and never exercises the failure path where the leak happens.

8. **[R1-M8] The untrusted-navigation dispatch isn't tested** — `src/sandboxUiPartitionPolicies.ts:151-189`. The refactor added the `handlers` shim (function-vs-object), the `gfs_open` branch, and routed `gfs_open`/`oauth_authorize` through the `window.open` handler too, but nothing asserts `onGfsOpen`/`onOauthAuthorize` actually fire, that the legacy positional callback still gets oauth, or that `window.open('gfs://…')` now dispatches. This is the seam that routes untrusted navigation, so a test driving a fake WebContents (`will-navigate`/`setWindowOpenHandler`) would be good insurance.

## Smaller stuff (grab-bag)

- **[R1-L1]** `clearActivity` wipes the whole-environment audit file, not just the current user's lines — `src/ipc.ts:1637` / `src/pluginSdkRuntime.ts:254`. `listGrants`/`revoke`/`activity` all scope to `currentUserId()`, but `clearAudit()` `fs.rm`s the entire JSONL, so one local user clearing their activity erases another's audit trail. Filter-and-rewrite keeping other `userId` lines.
- **[R1-L2]** The `gfs.open` existence/authorization oracle you already call out — `src/pluginSdkCapabilities.ts:409-438`. Bounded by the rate limit, a live session, and only the user's own access; if you want to close it, RID-only input or an opaque `reason` does it (overlaps with #2).
- **[R1-L3]** The permission channels (`listCapabilities`/`permissionState`/`requestPermissions`) skip the limiter — `src/pluginSdkBroker.ts:96-183`. Abuse is already bounded by the prompt budget + cooldown + cache, so it's just defense-in-depth.
- **[R1-L4]** `cacheKey` omits `envKey`/`userId` — `src/pluginSdkBroker.ts:245`. Grants scope by all four, the provider cache doesn't; an env switch that doesn't trigger `invalidateCache()` could serve the previous team's data for up to the 30–60s TTL. `notifySessionChanged` covers login/logout, so it depends on the env-switch wiring.
- **[R1-L5]** `ui/src/components/PluginConsentModal/types.ts:3-16` hand-redeclares the protocol wire types from `src/pluginSdkProtocol.ts:104-118` with no drift guard — three reviewers flagged this independently. Re-export from `pluginSdkProtocol` like `renderer.d.ts` already does.
- **[R1-L6]** No tests for the consent modal UI invariants (Escape/Deny → `onResolve(id, [])`, partial grant on uncheck, "all text except `pluginTitle` comes from the catalog") or the permissions controller invalidation after revoke — these are the security invariants from the body and none are executable yet. The new `SandboxUiPage.test.tsx` does nicely cover "the plugin doesn't name itself" (host-owned title).
- **[R1-L7]** Activity list uses an array-index `key` and shows the raw `pluginId` while Grants/modal show `pluginTitle` — `ui/src/components/PluginPermissions/index.tsx:541,544`. Two identities for the same plugin.
- **[R1-L8]** `closePluginGfsPreview` calls `setVisible(true)` unconditionally — `ui/src/App.tsx:85-88`. If the preview closes while off `SandboxUiPage`, it may re-show the WebContentsView over another page or leave it unbalanced.
- **[R1-L9]** `onGfsOpen` re-derives the target via `getActiveSandboxUi()` instead of the `webContents` that fired the navigation — `src/appService.ts:3113-3117`. Harmless with a single active embed, but couples attribution to "active surface" rather than "source of the click" for a future multi-embed world.

## What checked out clean

- **Identity is host-derived, never plugin-asserted** — `pluginId`/`pluginTitle` come from `activeView.appRef` + `args.title` on the `assertTrustedSender`-guarded `sandboxUi:open` path; `userId`/`envKey` from `getCachedUserId()`/`getActiveEnvKey()`. The plugin only controls its `webContents.id`, used as a lookup into a host pin map — unpinned sender gets a hard reject.
- **Consent fails closed** — unreadable/future-version file → re-prompt; gate error → IPC rejection (no data); timeout/Escape/no-window → deny; mid-flight revocation re-checked before returning provider data.
- **Consent key has no collision surface** — exact `(pluginId, capability)` equality inside a per-`envKey` file / per-`userId` bucket; "keeps environments/users apart" tests pass.
- **Isolation intact** — embed WebContentsView with `nodeIntegration:false`, `contextIsolation:true`, `sandbox:true`, `webSecurity:true`, per-recipe partition; embed preload exposes only dumb pipes; consent/resolve channels live in the trusted preload with `assertTrustedSender` + nonce.
- **Store is atomic** (`tmp`+`rename`, 0600, mkdir 0700) with serialized read-modify-write; corruption/future-version files are moved aside and treated as empty.
- **Suites**: broker/capabilities/surface (56), consent/store/audit (37), and the seam suite (52) all pass, `tsc --noEmit` clean. The one exception is the flaky consent-gate test in #3.

One process note: the file count (37) is over the soft ≤25 budget. It's a cohesive single-service feature with its suite, so probably fine — just worth a line in the body on why it isn't split.

zach88:
Did a round-2 pass focused on the fix commit (`45dd9c98`) plus the clean dev merge. Good news first: the three main-process fixes (rate-limiter intra-pair accounting, `gfs.open` dropping `reason`, and the consent-gate reordering) all do what they say, and their regression tests genuinely fail against the previous head. Trust boundary still fails closed, no new security surface. The fix did leave one real regression and one incomplete spot, and there's still one medium untouched from round 1.

Ratified separately: keeping the embed-scoped backdrop without `aria-modal` (the old R1-M3) is fine as an intentional design choice — closed, no action needed.

## Medium

**1. [R2-M1] Mutation error sticks in the permissions banner after a later success**
`ui/src/hooks/domain/usePluginPermissionsController.ts:90-102`. The R1-M4 fix wired `revokeMutation.error` / `clearActivityMutation.error` into `error`, but they're independent observers: one mutation succeeding doesn't clear the other's error, and there's no `.reset()`. So `clearActivity` fails → banner shows the error → then a `revoke` that *succeeds* still shows the stale clear-activity error. In a security surface that reads as "your revoke failed" when it didn't. The new comment also claims "React Query clears each mutation's error on its next run, so a stale message does not linger" — that's only true for the *same* mutation re-running, so the comment is inaccurate as written. Fix: reset the mutation errors on `onSuccess` / when starting any action, or derive the banner from just the last action fired. A test here (mount the controller with a rejecting `revoke`, assert the error banner) also closes the still-open test gap below.

**2. [R1-M8] Untrusted-navigation dispatch still has no test** — *still open since round 1*
`sandboxUiPartitionPolicies.ts:151`. `applySandboxUiNavigationPolicies` routes `gfs_open` / `oauth_authorize` and the `window.open` handler, and nothing asserts that `onGfsOpen`/`onOauthAuthorize` fire, that the legacy positional callback still gets oauth, or that `window.open('gfs://…')` now dispatches. This is the seam that routes *untrusted* navigation, so it's worth a test against a fake `WebContents` (`will-navigate` / `setWindowOpenHandler`) covering both handler shapes. Heads up: this is the only round-1 medium left open — if it slips another round it goes to re-triage.

## Regressions (introduced by `45dd9c98`)

**3. [R2-L1] Consent gate leaks a pending prompt + 120s timer if the surface-hide rejects**
`pluginConsentGate.ts:280-299`. The fix moved `this.pending` + the `setTimeout(120s)` before `await setSurfaceVisible(false)`, inside the `try`, but there's no `catch`. If that await rejects (real case: the `WebContentsView` is destroyed mid-flip so `view.setVisible` throws — the exact race this fix targets), `runPrompt` throws but leaves `this.pending` set and a live timer for up to 120s. **Reproduced**: a throwaway probe passes at `59941b9a` (pre-fix the pending was created *inside* the promise executor, after the await, so a rejecting hide never left a pending) and fails at the current head. Blast radius is low — `hasPendingPrompt()` has no production consumer, the queue serializes so the next prompt overwrites it, and the timer self-heals via its `promptId` guard — which is why it's Low, not Medium. Fix:
```ts
} catch (err) {
  if (this.pending?.promptId === promptId) { clearTimeout(timer); this.pending = null }
  throw err
} finally { await this.deps.setSurfaceVisible(true).catch(() => undefined) }
```

## Grab-bag

**4. [R2-L2] `flushUntil` swallows its timeout** — `src/__tests__/pluginConsentGate.test.ts:31`. The helper returns silently when `maxTicks` runs out instead of throwing, so "cancels a pending prompt when its plugin goes away" (`:174`) would still go green even if `presentPrompt` stopped showing the prompt. Add `if (!pred()) throw new Error('flushUntil timeout')` at the end of the loop.

**5. Test gap for the R1-M4 security fix (T5, ties into the still-open R1-L6)** — the revoke/clear-activity error surfacing — a security action that used to silently look successful — ships with no test. Nothing references `PluginPermissions`, `usePluginPermissionsController`, or `PluginConsentModal`. Worth a test that mounts the controller with a rejecting `revoke` and asserts the error banner shows (assert the visible banner, not `mutation.error`). One test covers this and item 1.

## What checked out clean
- Dev merge (`cf77f1ee`) touches nothing in the plugin SDK — empty diff over these files, no collateral.
- Rate limiter: `take()` contract unchanged for its only caller (`pluginSdkBroker.ts:223`); allow/deny and `retryAfterMs` identical to pre-fix except the double-charge is gone; refund is symmetric and capped by capacity, can't inflate the budget.
- `gfs.open`: the `{ opened }` narrowing is only on the plugin path — the navigation path and the audit line don't consume `reason`, so both are unaffected.
- Consent gate stays fail-closed: `presentPrompt` still runs after the hide; neither the timer nor `resetPlugin` can grant (both resolve to a denial); the pre-hide window isn't exploitable (the renderer doesn't know the `promptId` yet).
- Removing `aria-modal` doesn't break anything: Escape handling and initial focus don't depend on it, and there was no focus-trap tied to it.
- Tests: `pluginConsentGate` (13), `pluginRateLimiter` (5), `pluginSdkCapabilities` (23) all green, 41/41; consent-gate run 3× extra with no flakiness (the R1-M5 flake is gone). T3 confirmed — all three new regression tests fail against `59941b9a`. Fixtures are injected / derived from the real producer, no hand-rolled ones.

zach88:
Round 3 pass. The round-2 fixes hold up well: the consent gate now fails closed on a destroyed WebContentsView (R2-L1), `flushUntil` throws on timeout (R2-L2), and the nav-policy dispatch finally has tests (R1-M8). The dev merge into `App.tsx` is clean too — deep-link and plugin-consent don't step on each other. **0 Blockers/Critical/High.** One catch though: the R2-M1 fix traded one masking bug for its mirror image.

## Regressions

**1. [R3-M1] "Latest action wins" masks a failed revoke** — `desktop-app/ui/src/hooks/domain/usePluginPermissionsController.ts:91-95`

The R2-M1 fix (commit `a3318bd`) switched the banner to show the error of whichever action has the greater `submittedAt`. That closes the reported direction (clear fails → later revoke succeeds), but opens the symmetric one:

- User hits **Revoke**, it **fails** (permission is NOT revoked). Then they hit **Clear activity log**, which succeeds. Since clear's `submittedAt` is now greater, `latestAction` is the clear mutation, `actionError` is null, and the banner **wipes the revoke failure**. The user walks away believing the plugin lost a capability it still has.

This is the same class as R1-M4 (silent failure on the security-relevant action), reintroduced for the revoke-fails→clear-succeeds order. It's sequential, not a concurrency race — react-query holds `revokeMutation.error` until a reset or re-run, and in real use the two `submittedAt` values differ by seconds, so the masking is deterministic. The R2-M1 test only covers one direction, which is why it slipped through green.

Suggested fix: don't let an unrelated action clear another's error. Surface the error of any mutation still in `error` state that hasn't been superseded by a re-run of that *same* action, and add a property test covering **both** masking directions so this can't regress a third time.

## Grab-bag

**2. [R3-L1] Asymmetric disabled guard** — `desktop-app/ui/src/components/PluginPermissions/index.tsx:127,153,204`

The fix added `disabled={controller.clearingActivity || controller.revoking}` to the *Clear activity log* button (204), but *Revoke* (153) and *Revoke all* (127) still only guard on `controller.revoking`. So while a `clearActivity` is in flight you can still fire a concurrent `revoke`. Separate from R3-M1 (which is sequential), but it's the sibling instance the guard change left half-done — worth closing in the same pass.

**3. Dead legacy overload** — `desktop-app/src/sandboxUiPartitionPolicies.ts:159,163-164`

The positional-callback overload of `applySandboxUiNavigationPolicies` (oauth callback as 3rd arg) no longer has a production caller — the only one (`sandboxUiDriver.ts:380`) moved to the object form in this PR. The `typeof handlers === 'function'` branch and its dedicated test now only exercise themselves. Not a bug, just dead once this lands; could drop the overload + test.

## What checked out clean

- **Trust boundary, traced end-to-end:** identity isn't plugin-asserted (`getUserId()` + surface pinning by `senderId`), consent can't be bypassed, and the new try/catch in the consent gate fails closed on every branch — a WebContentsView destroyed mid-flip withdraws the prompt (deny), never approves it. No token leakage, rate limiter / audit log aren't reachable around the broker chokepoint.
- **R2-L1 / R2-M1 regression tests reproduce:** both fail against the pre-fix head (verified).
- **Suites:** `desktop-app` src 1652 passed, renderer suites touched by the fix 22/22 green.

zach88 (approved):

PR #304, opened 2026-08-07
feat: editable display name for agents and contexts

zach88:
## What

Decouples the visible name of Agents and Contexts from their identifier, and hardens the admin resource error path that the old rename flow exposed.

- **Agents**: the identifier (`metadata.name`, RFC1123, referenced across the system) is now read-only in the UI. The visible name is `spec.host` (already free text) and is edited with a single `PUT`. The create+migrate+delete rename dance is removed.
- **Contexts**: new optional `spec.displayName` (additive CRD field) is the editable visible name; the identifier/`contextId` stays read-only. Existing contexts without the field fall back to `metadata.name`.
- **Creation wizards**: the user types a free name, the RFC1123 slug is auto-derived for `metadata.name`, and the free name is stored in the spec. The generated identifier is previewed before creating.
- **Backend hardening**: invalid `metadata.name` on create now returns `422 invalid_name` instead of a `500`; raw Kubernetes 4xx errors are forwarded with their real status instead of collapsing to `500` on every admin route.
- **Propagation**: control-ui tables/detail pages and desktop-app surfaces render the visible name; `external-rest-api` declares `displayName` on its agent type.

## Why

Renaming an agent to a value with spaces or capitals ("Sales Agent") returned a `500`, and contexts could not be renamed at all (the field was disabled). The root cause was that the UI made the user edit the identifier itself, which is immutable, RFC1123, and referenced by channels, users/teams, budgets and shared filesystems. The old agent flow worked around this with a fragile create+migrate+delete orchestration that broke on invalid names; contexts simply blocked the field.

Fixing the identifier once at creation and editing only a separate visible label removes the whole class of problem: no reference migration, no cascade, no orphaned refs. Two robustness defects the old flow surfaced (unvalidated `metadata.name`, and K8s status codes collapsing to `500`) are fixed as defense in depth.

## How

Key decisions:

- **Identifiers stay RFC1123, display fields are free text.** `metadata.name`/`contextId` keep RFC1123 (enforced by the apiserver and, at the application level, by control-api). `spec.host` and `spec.displayName` are free text, length-limited and stripped of control and bidirectional-formatting characters.
- **CRD change is additive only.** Only an optional `spec.displayName` is added to the context CRD. No existing field is hardened with a `pattern`/`maxLength`, so the apiserver never rejects controller `/status` writes on legacy CRs. Those rules live in control-api with a change-only ratchet: a field is validated only when its value changes, so a legacy out-of-norm value never blocks unrelated edits.
- **Minimal, explicit fallback.** Agents read `spec.host` directly (required, always present). Contexts use `spec.displayName ?? metadata.name`, the single fallback justified by the optional field. Sites whose source is the cross-team directory (not the access catalog) keep an explicit `?? name` fallback rather than a blanket defensive one.
- **Latent prerequisite.** `resolveHostRef` in mcp-host used `spec.host` as an identifier fallback; with an editable display that would publish activity under a non-existent ref. It now uses `metadata.name`.
- **Error handler.** The inline global handler is extracted to `clerumErrorHandler`, reads the status from `.status ?? .statusCode ?? .code ?? .response.statusCode ?? .response.status`, and sanitizes the client message (and internal logs) so a Kubernetes `ApiException`, whose message embeds the full `metav1.Status` and response headers, never leaks them.

## Testing

TDD, red-first, per unit. Focused suites are green and typecheck is clean:

- control-api: resource validation + ratchet, error forwarding, error-handler sanitization, external access enrichment. `tsc --noEmit` clean.
- mcp-host: `resolveHostRef` returns the identifier, not the display.
- control-ui: agent/context detail (read-only identifier, single PUT, zero POST/DELETE), table rendering + filter, slug contract.
- desktop-app: display rendering with the three fallback branches (agent without fallback, context with display, context without display), including the cross-team fallback.

Invariants asserted by executable tests in this branch: `metadata.name` and the URL are unchanged on a visible rename (control-ui detail test asserts one PUT with the identifier intact and zero POST/DELETE); a context rename leaves `contextId` and host `contextRef` untouched (no cascade); Kubernetes 4xx are forwarded with their status and the client/log payloads carry no response headers; each hosts/contexts PUT reads the CR exactly once.

Coverage here is **unit/integration only** (unit/integration `.test.ts(x)` across control-api, control-ui, desktop-app and mcp-host). The Playwright end-to-end specs for the four critical flows (agent/context display rename, wizard auto-slug, API defense) are **not included in this PR** and are tracked as a follow-up.

## Agent-create wizard — create-only seam (redesign)

The create wizard no longer materializes resources through a PUT-first "upsert". Each resource (secret, context, channel, host) is now created **create-only via POST**, so a name collision surfaces the apiserver's `409 AlreadyExists` instead of silently overwriting an existing context/channel/host and reporting "created". `submitAll` tracks only the resources it actually created this run and, on any failure before the host is created, best-effort deletes them in reverse order (no orphaned secret/context/channel); a grant failure *after* the host leaves the agent created with partial access and a retryable error, no rollback. All free-text display fields (`spec.host`, context `spec.displayName`) and every derived slug are validated client-side **before the first write** against the exact rule the server enforces — the display-field rule is extracted into a shared `@clerum/display-field` package consumed by both control-api and control-ui, so the two sides can never drift. The standalone context-create page gates on the full RFC1123 identifier rule and the same display-field rule.

## Risks and follow-ups

- **Deploy order matters — CRD → control-api → control-ui.** Apply the additive CRD first and verify it (`kubectl get crd contexts.clerum.io -o yaml | grep displayName`); then control-api (its read-after-write guard only defends `spec.displayName` once the new control-api is live); then the control-ui context/wizard forms. Shipping the UI against an old CRD makes the apiserver silently prune `spec.displayName`, so "save"/create returns 200 while losing data. mcp-host and desktop follow; the rolling restart from the mcp-host image bump is the normal cost of any release. (The durable version lives in `charts/clerum-crds/README.md`.)
- **Editing the visible name is fully hot.** `spec.host`/`spec.displayName` are not read by any reconciler, do not enter a pod hash, and propagate via watch without a restart.
- **Contexts on desktop still show the id** until a `contextId to displayName` source is decided (there is no external context detail endpoint today).
- Known follow-ups tracked separately: a non-production 5xx client response still returns the raw error message; the communication-channel PUT and credentials cascade repeat CR reads; the mcp-server install path does not RFC1123-validate a user-supplied `serverName`; GFS filename validators do not reject bidi characters.

## Note on size

This PR spans 6 services (62 files) because it is a single coordinated change: the display-name contract crosses the CRD, control-api, both UIs, mcp-host and the external API, with a locked deploy order between them. Splitting it would fragment a unit that was implemented and reviewed together and would break the deploy-order guarantees (the frontend forms depend on the CRD field existing). Each phase is independently shippable if a split is later required.


control-ui/components/HostWizard/index.tsx
```diff
@@ -895,7 +902,9 @@ export function HostWizard({
       )
 
       const hostSpec: Record<string, unknown> = {
-        host: normalizedHostName,
+        // Visible name is the free text the operator typed; metadata.name and
+        // every reference (channels.hostRef, access grants) use the derived slug.
+        host: hostName.trim() || normalizedHostName,
```
alfredolopez80: [P1] Prevent free-text slug collisions from becoming destructive upserts

This new raw-display path still passes normalizedHostName to the PUT-first upsert at lines 710-721. Once two display values converge to the same slug, such as My Agent! and my-agent, an existing Host is full-replaced and the wizard reports success instead of surfacing a collision. Please make creation POST-only or perform an explicit existence check before any write.

control-ui/components/HostWizard/index.tsx
```diff
@@ -235,10 +235,17 @@ function isStepValid(
     directoryLoadFailed: boolean
   }
 ): boolean {
-  if (stepIndex === 0) return state.hostName.trim().length > 0
+  // Gate on the DERIVED slug, not the raw text: a name with no ASCII
+  // alphanumerics ("!!!", "日本語", "---", whitespace) trims to a non-empty
+  // string but toKebabCase()s to "", which would send metadata.name: "" — and
+  // because the secret/context are created BEFORE the host in submit(), a
+  // failed host create would orphan those siblings.
+  if (stepIndex === 0) return toKebabCase(state.hostName).length > 0
```
alfredolopez80: [P1] Run the complete display/identifier preflight before the first mutation

This gate only proves that toKebabCase(hostName) is non-empty. It does not enforce the server display control/bidi and 120-character rules or the RFC1123 63-character identifier limit. submitAll writes the Secret, Context, and optional Channel before the Host, so an overlong or control-character display can leave siblings behind after the final 422. Please mirror the server validator before any write, or add verified reverse-order compensation and zero-orphan tests.

control-ui/components/HostWizard/index.tsx
```diff
@@ -235,10 +235,17 @@ function isStepValid(
...
   if (stepIndex === 1) {
     if (state.contextMode === 'existing') return state.selectedExistingContext.trim().length > 0
-    return state.contextName.trim().length > 0
+    // Same slug-derivation gate: the new context upsert runs before the host
+    // create, so an empty derived name here has the same orphan risk.
+    return toKebabCase(state.contextName).length > 0
```
alfredolopez80: [P1] Preserve the raw Context display separately from the derived slug

This new-context gate still leads to state that is normalized with toKebabInput in the render field, while the POST/PUT payload at lines 823-842 omits spec.displayName. A user-entered Sales Context is therefore persisted only as sales-context. Please keep raw display text and contextId/metadata.name as separate values, send spec.displayName, and assert the submitted payload.

control-ui/app/contexts/[name]/page.tsx
```diff
@@ -387,21 +391,28 @@ export default function ContextDetailsPage() {
...
   async function save() {
     if (!canSave) return
+    const trimmedDisplay = displayNameDraft.trim()
+    // Full-replace safety: echo every spec field the form knows about
+    // (identifier + all editables). The visible name lives in spec.displayName;
+    // omit it when empty so an unset display stays absent, not "".
     const payload = {
       spec: {
         contextId: contextNameDraft.trim(),
+        ...(trimmedDisplay ? { displayName: trimmedDisplay } : {}),
```
Comment (alfredolopez80) [inline control-ui/app/contexts/[name]/page.tsx]: [P1] Do not erase spec.gfs during a display-only rename

This full replacement now writes displayName but still omits spec.gfs, although the Context CRD defines GFS mount intents there. Renaming only the display can therefore drop existing mounts. Prefer a field-owned PATCH, or merge a fresh complete spec and add a regression fixture with gfs.mounts.

control-api/src/routes/admin/resources.ts
```diff
@@ -532,6 +669,20 @@ export function createAdminResourcesRouter(gateway: K8sGateway): Router {
             return
           }
         }
+        if (
+          plural === 'contexts' &&
+          body.spec &&
+          contextDisplayNameNotPersisted(body.spec, updated)
+        ) {
+          // Deliberately NO spec restore (unlike the CC update rollback): the
```
alfredolopez80, line 788: [P2] Do not report a clean 409 after a partial Context update

updateResource has already persisted the replacement when this guard fires. If an old CRD prunes only spec.displayName, description and mcpServers can remain committed even though the client receives 409. Please preflight capability or restore and verify the previous object; add a test that asserts persisted state and retry behavior.

desktop-app/ui/src/components/agents/__tests__/AgentWorkspace.displayName.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+    contextsMock.contextDisplayById = {}
+    mcpMock.agentDisplayByName = {}
+    navMock.selectedAgent = 'product-agent'
+    navMock.selectedAgentRoute = 'contexts'
+    vi.clearAllMocks()
+  })
+
+  it('renders the context display when present and the id when absent', () => {
+    contextsMock.contextIds = ['ctx-with', 'ctx-plain']
+    contextsMock.contextDisplayById = { 'ctx-with': 'Prod Ctx' }
```
alfredolopez80, line 99: [P1] This test injects a producer state that production cannot currently create

contextDisplayById is supplied directly by the mocked hook, while the production session/catalog path has no producer for this map. The test proves a renderer branch but not display propagation. Please add the real producer contract and an integration assertion, or defer/remove this test and add the real Playwright journey claimed by the PR body.

control-api/src/routes/admin/resourceFieldValidation.ts
```diff
@@ -0,0 +1,135 @@
...
+//     that visually impersonates another resource in the UI/desktop (spoofing).
+//   - Line/paragraph separators (U+2028/U+2029): line breaks a single-line label
+//     must not contain (JS regex `.`/`\s` and JSON treat them specially too).
+// Deliberately NOT rejected: directional marks (U+200E/U+200F/U+061C) and
+// zero-width/invisible characters (U+200B–U+200D, U+2060, U+FEFF). This set is
+// the canonical Trojan-Source bidi mitigation; zero-width joiners (U+200D) are
+// legitimate in emoji sequences, so blanket-rejecting them would break valid
+// display names. Confusable/invisible-glyph spoofing is out of scope here.
+// eslint-disable-next-line no-control-regex
+const CONTROL_CHAR_RE = /[\x00-\x1f\x7f\u202a-\u202e\u2066-\u2069\u2028\u2029]/
```
alfredolopez80: [P2] Cover the full control-character range promised by the display contract

The current class rejects C0, DEL, bidi formatting, and line separators but omits C1 controls U+0080-U+009F. U+0085 can break the single-line invariant and U+009B can introduce terminal control sequences. Include the C1 range or document why it is intentionally allowed, with a regression test.

control-api/src/routes/admin/resourceFieldValidation.ts
```diff
@@ -0,0 +1,135 @@
...
+  if (typeof value !== 'string') {
+    return { field, message: `${field} must be a string` }
+  }
+  if (CONTROL_CHAR_RE.test(value)) {
+    return {
+      field,
+      message: `${field} must not contain control or bidirectional formatting characters`,
+    }
+  }
+  if (value.trim().length > DISPLAY_FIELD_MAX_LENGTH) {
```
alfredolopez80: [P2] Reject or normalize whitespace-only display values

validateDisplayField currently accepts an empty or whitespace-only string. That value can be persisted and rendered as a blank Context/agent label by some consumers. Please either reject whitespace-only values or persist the trimmed fallback and make every consumer use the same display-or-identifier rule.

alfredolopez80 (dismissed):
## Review decision

**REQUEST CHANGES**

I reviewed the exact PR head 7f100ed4a5acadfa423e557f974749cdc71dfa7b against base a18dfd490b8eed55ec96527b8a09659d6e8a7ae3.

The direction is good: keeping immutable identifiers separate from editable display text, fixing identifier-based channel matching, sanitizing Kubernetes errors, and propagating agent display names are all sound improvements.

The PR is not ready to merge because the current head still contains destructive create/update paths and its claimed Playwright coverage is absent.

### Blocking findings

#### [P1] HostWizard can overwrite an existing Host when a display slug collides

control-ui/components/HostWizard/index.tsx:710-721 and :928-938 still use a PUT-first upsert and fall back to POST on any PUT error. Once the PR accepts free-text display names, inputs such as My Agent! and my-agent converge to the same metadata.name. If that Host already exists, the PUT replaces its spec and the wizard reports a successful creation instead of surfacing a collision.

This mechanism predates the PR, but the new free-text contract materially increases the exposure and contradicts the promise that the identifier is immutable. Please make creation POST-only, or perform an explicit existence check before any write and fail with a clear conflict error.

#### [P1] The embedded context wizard still drops the free display name

control-ui/components/HostWizard/index.tsx:1164-1172 still normalizes the new-context input with toKebabInput, and the POST/PUT bodies at :823-842 omit spec.displayName. A user-entered Sales Context is therefore persisted as sales-context instead of preserving the free display name and deriving only the identifier.

The changed tests only verify whether Next is enabled. They never submit the wizard or assert the persisted payload. Please keep raw display text and the derived context identifier as separate values, send spec.displayName in both bodies, and add a submit-level regression test plus a persisted E2E assertion.

#### [P1] Server-only display validation occurs after side effects

The wizard gate at control-ui/components/HostWizard/index.tsx:238-249 only checks that the derived slug is non-empty. submitAll then creates the Secret, Context, and optional Channel at :799-889 before sending the Host at :904-938. control-api rejects control/bidi formatting and trimmed values over 120 characters at resourceFieldValidation.ts:68-85; metadata.name also has an RFC1123 limit of 63 characters.

A valid slug combined with an overlong, newline, or bidi display can therefore create siblings and fail only at the final Host write. Please add a no-side-effect preflight that mirrors the server rules, or implement verified reverse-order compensation and test zero-orphan behavior for every failure boundary.

#### [P1] The PR body claims Playwright specs that are not in the diff

The exact PR diff contains zero files under tests/e2e. The new UI tests mock APIs/hooks and do not prove the real create, rename, persistence, or no-orphan journey. The existing Playwright suite also contains stale Display ID selectors while the head UI exposes Display name.

Either add real Playwright coverage through the user-visible flow, asserting route/state, visible UI, and a persisted business signal, or correct the PR description so it claims only unit/integration coverage.

#### [P1] Context display rename can erase existing GFS intents

control-ui/app/contexts/[name]/page.tsx:406-419 sends a full replacement spec that includes the new display field but omits spec.gfs, which is defined by charts/clerum-crds/crds/context.yaml:83-120. A display-only rename can therefore erase existing GFS mount intents. The omission is pre-existing, but PR #304 makes this destructive path a primary feature flow.

Please prefer a field-owned PATCH, or merge a fresh complete spec while preserving gfs and controller-owned/unknown fields. Add a fixture containing gfs.mounts and assert that a display-only rename preserves it.

#### [P2] The pruning guard can return 409 after committing other fields

control-api/src/routes/admin/resources.ts:672-684 returns context_crd_outdated after updateResource has already persisted the replacement. If an old CRD prunes only spec.displayName, description and mcpServers can remain committed even though the client receives 409.

Please preflight CRD capability before mutation, or restore and verify the previous object. Add tests for persisted state, failed rollback, and retry behavior.

### Additional findings

- [P2] Deleting an McpServer rebuilds Context specs field-by-field at control-api/src/routes/admin/resources.ts:769-780 and can silently remove spec.displayName. This is pre-existing code, but it is data loss for the feature introduced by this PR. Preserve the complete current spec or use a field-owned patch.
- [P2] The desktop Context display test injects contextDisplayById manually, while the production session path has no producer for that map. Either add the real producer contract or defer/remove the unreachable test.
- [P2] resourceFieldValidation.ts:38 omits the C1 control range U+0080-U+009F. Extend the control-character class if the display contract is intended to be single-line and terminal-safe.
- [P2] Empty or whitespace-only display values are accepted and can render as blank labels in ContextTable and desktop access selectors. Reject or normalize whitespace-only values and use the same fallback in every consumer.
- [P2] The deployment order must be documented as CRD, then control-api, then control-ui. With an old control-api and old CRD, a new UI can receive a successful response while Kubernetes silently prunes displayName.
- [P2] K8sNotFoundError/K8sConflictError expose httpStatus, but errorHandler.ts does not read it; some service-layer 404s still become 500.

### Positive observations

- Agent display propagation through control-api, external REST, AppService, and renderer fallback is coherent.
- HostOverview now matches channels by immutable identifier and has useful cross-agent isolation coverage.
- mcp-host resolves Host references by metadata.name rather than display text.
- Error-handler and field-validation tests exercise real Express/API boundaries where appropriate.
- The CRD change is additive and the identifier-validation intent is clear.

### Verification boundary

- Exact base/head diff reviewed; git diff --check passed.
- GitHub checks reported green for the head.
- No changed E2E files were found.
- Focused local Vitest was not run because Vitest was unavailable in the detached dependency tree; no packages were installed.
- No Docker, Minikube, live-cluster, or GitHub write action was performed.

Please address the P1 findings, add or honestly re-scope the E2E evidence, and then rerun the exact-head review before requesting approval.

alfredolopez80:
## Review — multi-agent pass (verified against real code)

Reviewed the diff with parallel area/security/contract/quality/test agents on an isolated worktree; each finding verified against real code, many by **mutation testing** (break prod → confirm the test goes red). Overall direction is a clear improvement and the core (immutable identity vs editable display, `resolveHostRef` fix, channel matching by identifier, error sanitization) is solid. **Recommendation: hold merge** on 3 blockers below.

### 🔴 Blockers

**B1 · Host "create" is a PUT-first upsert that can silently overwrite an existing agent**
`control-ui/components/HostWizard/index.tsx:710-721` + `:928-938`. Create issues `PUT /admin/hosts/<slug>` and only falls back to POST on failure. If the derived slug collides with an existing agent, the PUT full-replaces that agent's spec (contextRef, secretRef, channels, model policy, display) and the wizard reports "Agent created successfully". Reproduced: `toKebabCase('My Agent!') === toKebabCase('my-agent') === 'my-agent'`. The mechanism is pre-existing (`git blame` → `5d663a49`) but this PR raises the exposure by moving from a visible identifier to free text. **Fix:** check the slug against the loaded host directory in `isStepValid`/`submitAll` and block with "identifier already in use", and/or materialize with a plain POST so a collision fails loud with 409.

**B2 · Deleting an McpServer wipes `spec.displayName` on referencing contexts**
`control-api/src/routes/admin/resources.ts:769-780` (pre-existing DELETE path) + `resourceService.ts:363` (wholesale `spec: body.spec`). The allowlist cleanup rebuilds the context spec by hand with `{ contextId, description, mcpServers }` and omits the new `spec.displayName` → silent data-loss of this PR's own feature, with no test. The anti-pruning guard doesn't fire (this internal update doesn't send displayName). **Fix:** `spec: { ...ctx.spec, mcpServers: filtered }` (or merge-patch) + a red-first regression test.

**B3 · The description claims Playwright E2E specs are included; they are not**
Body says *"The Playwright end-to-end specs (agent/context display rename, wizard auto-slug, API defense) are included but were not run here"*. `git diff --name-only <base>..HEAD | grep -iE 'spec|e2e|playwright'` → empty. The PR adds 10 `.test.ts` + 10 `.test.tsx` (unit/integration) and **zero** `.spec.ts`; the 4 critical flows have no E2E user-path coverage. **Fix:** add the specs, or correct the body to say coverage is unit/integration only.

### 🟠 Should fix before merge

- **H1 — Wizard slug-gate doesn't mirror server validation → orphaned sibling resources.** `HostWizard/index.tsx:721` only checks `toKebabCase(...).length > 0`; a 70-char name passes the gate, secret (POST) → context → channel are created, then the host is rejected `invalid_name` → orphaned secret/context/channel; retry fails on the existing secret. Fix: gate on `RFC1123_RE.test(toKebabCase(name))` + mirror `validateDisplayField`, and truncate/preview at 63.
- **M1 — Empty/whitespace `displayName` renders a blank label (converged across 3 agents).** `validateDisplayField` accepts `""`/`"   "`, CRD has no `minLength`, stored verbatim. `ContextTable.tsx:50` uses `?? name` without trim (vs `HostTable.tsx:96-97` `trim() || name`); the GFS "Manage access" dialog (`GfsAgentAccessSection/index.tsx:86`) shows a selectable, unlabeled agent (reproduced). Only out-of-band writes trigger it. Fix: reject/normalize whitespace-only in `validateDisplayField`; `(x ?? '').trim() || name` in ContextTable and `agentSubjectOptions`.
- **M2 — Deploy-order window with silent data-loss (converged across 2 agents).** The read-after-write guard lives in control-api and only defends when control-api already ships `e984ebb1`. Old control-api + old CRD + new UI → apiserver prunes `spec.displayName`, returns 200, data lost. kubectl/GitOps against the old CRD also prune. Safe order is **CRD → control-api → control-ui** (body only says "CRD before UI"). Fix: document the full order; consider CRDs in a `pre-install/pre-upgrade` hook.
- **M3 — A service-layer 404 still collapses to 500 (the class this PR claims to fix).** `errorHandler.ts:24-40` `extractErrorStatus` (`.status ?? .statusCode ?? .code ?? .response.*`) doesn't read `.httpStatus`, where `K8sNotFoundError`/`K8sConflictError` expose their status (and `http/k8sError.ts:18` already reads it). `GET /admin/hosts/<missing>` → 500 instead of 404; the comment at `resources.ts:582` is wrong and `routes.resources.test.ts:331` pins 500. Reproduced. Fix: add `.httpStatus` to the chain (or map `K8sNotFoundError → 404`) and fix the comment/test.

<details>
<summary>🟡 Lower severity (M4 / Low) & structural debt</summary>

- **M4 — C1 controls (U+0080–U+009F) not rejected** in `resourceFieldValidation.ts:38`; U+0085 (NEL) breaks the "single line" premise, U+009B (CSI) injects ANSI in 8-bit terminals. Reproduced. Fix: extend the class to include `\x7f-\x9f`.
- **L1 — Length limit is post-`trim()` but value is stored verbatim** → `"a"*120 + " "*5000` passes and persists (bounded only by the 150mb JSON body limit). Admin-only.
- **L2 — Rollback logs leak raw K8s `err.message`** (`resources.ts:90,148,214,477`) which embeds `metav1.Status` + Audit-Id headers — exactly what the global handler scrubs. Fix: reuse `safeErrorLogMessage`.
- **L3 — GHAS "Log injection" (`resources.ts:214`) is a false positive** (confirmed by 2 agents): `name` passes RFC1123 validation before the rollback; CodeQL doesn't see the gate. Optional `sanitizeForLog()` to silence the scanner.
- **L4 — Non-integer 4xx status throws `RangeError` inside the handler** (`errorHandler.ts:163,175`) → falls through to Express default. Theoretical. Fix: `Number.isInteger` guard.
- **Coverage gaps:** N1 "PUT reads the CR once" tested only for contexts, not hosts (`routes.resources.test.ts:1473-1543`); the new `TeamsPage.tsx:97-103` display branch has no assertion (mock only for compile).
- **Structural (thermo-nuclear, non-blocking, with a date):** the read-after-write pruning guard is now written **twice** (extract `prunedFieldGuard.ts`); **three** near-identical HTTP-status extractors with different precedence (a canonical `httpStatusOf(err)` also fixes M3 at the root); the "single sanctioned fallback (Decision #6)" exists in ~6 syntactic + 2 semantic variants (one `displayOrSlug(display, slug)` per package). `resources.ts` grows 647→798 at ~+150/PR of per-plural if-blocks — worth arresting before it crosses 1k.

</details>

### 🟢 Verified clean (credit)

- **`resolveHostRef` → `metadata.name`** and **channel matching by identifier (R1-H1)** — both verified **by mutation** (revert → red). The identifier↔display confusion is properly closed.
- Removal of the `create+migrate+delete` rename dance (−162 in `hosts/[name]/page.tsx`) deletes a client-side distributed transaction — the best structural decision here.
- **Security clean:** no `dangerouslySetInnerHTML` (display is escaped JSX, never in URLs); authz unchanged (`requireAuthForControlUI`); RFC1123 enforced server-side before K8s (cap 63); bidi Trojan-Source rejected.
- **Tests are honest:** 169/169 green, **4/4 mutations detected, 0 false positives**. The gaps are coverage (E2E/hosts-N1/TeamsPage), not weak tests.
- `errorHandler.ts` and `resourceFieldValidation.ts` are well-earned abstractions (pure, ratchet as a parameter, rationale in situ, independently testable).

*Gates (reproduced, exit captured without pipe): control-api 109/109 ✓ · control-ui 37/37 ✓ · desktop 21/21 ✓ · mcp-host 2/2 ✓. Typecheck failures are all pre-existing in untouched files.*

alfredolopez80:
## Re-review of the R4 fixes — verification pass (head `38adea6e`)

Follow-up to my earlier review. Re-verified the 9 `R4-*` commits with parallel agents on an isolated worktree, **mutation-verifying** every fix (revert the prod change → confirm the test goes red). Great progress — **8 of 9 prior findings are closed and every fix has a test that dies under mutation** (no cosmetic "fixed"; the API side did real class-level fixes via `{...ctx.spec}` spread). But a fresh-eyes regression pass surfaced a **new Critical** in the same class as B1, and **B3 is still open**. **Hold merge.**

### ✅ Resolved (mutation-verified)

| Prior finding | Commit | Evidence |
|---|---|---|
| **B2** McpServer delete wiped `displayName` | `e9ffaff4` | `resources.ts:775` now spreads `{...ctx.spec, mcpServers: filtered}` — closes the class (future fields too). Mutation → red. |
| **B4 / B5** install/uninstall wiped `displayName` | `fa508ea9` / `d7883d18` | `registry.ts:1294` / `:1655` same spread. Mutations → red, discriminated. |
| **M1** empty/whitespace `displayName` → blank label | `0e5dc572` | server rejects `''`/whitespace (`resourceFieldValidation.ts:82`) + `(x ?? '').trim() || id` in 4 consumers. Mutation → 3 red. |
| **M2** deploy order | `6bd17cc1` | `charts/clerum-crds/README.md:57-75` documents CRD → control-api → control-ui with the why. Actionable. |
| **M3** 404 collapsed to 500 | `86b90db3` | `.httpStatus` read in the right position; `K8sNotFoundError` → 404. Mutation → red. |
| **M4** C1 control chars accepted | `8262b5fe` | `\x7f-\x9f`; verified by execution (28 samples), no collateral damage to accents/CJK/emoji. |
| **M5** secret name limit | `38adea6e` | exact client↔server parity (`isValidDNSSubdomain` ≤253); no off-by-one. |

**Mutation gate: 12/12 fixes with a test die under mutation, 0 false positives.** Suites green: control-api 184 · control-ui 1307 · desktop-ui 773.

### ❌ Still open / newly found

- **N1 · CRITICAL (new) — the collision guard only covers hosts.** Creating a **context** or **channel** whose free name derives to an existing slug still silently overwrites the existing resource. `HostWizard/index.tsx` step-1 (`~:575`) and step-4 (`~:603`) gate only `isValidResourceSlug` (no collision check); `submitAll` PUT-firsts `admin/contexts` / `admin/communication-channels`. A context PUT full-replace drops `displayName`/`mcpServers`; a channel PUT re-points `hostRef` and replaces `access` — silent hijack, reported as "created". Reproduced. Same class as B1, extended to two resource types the R4 fix didn't cover.
- **B1 · PARTIAL — the host collision guard is client-side and fail-open.** `HostWizard/index.tsx:592` `.catch(() => ({ items: [] }))` swallows the hosts fetch error with no signal; if it fails, `existingHostNames=[]` and the guard disappears while the PUT-first upsert can still overwrite. The server still accepts the PUT-upsert.
- **N2 · HIGH (new) — H1 mirrors the slug but not `spec.host` validation.** The gate validates `toKebabCase(name)` but the host sends `host: name.trim()` raw; a name with bidi (U+202E), NEL, or >120 mostly-non-ASCII chars (slug still ≤63) passes the gate and the server rejects the host create (422) *after* secret+context+channel are created → orphaned siblings. Reproduced.
- **B3 · STILL OPEN — no E2E specs, and the description still claims they exist.** `git diff --name-only <base>..HEAD | grep -iE 'spec|e2e'` → empty; the body still says "The Playwright end-to-end specs … are included but were not run here". Add the specs or correct the body.

**Root-cause fix that closes N1 + B1 + N2 at once:** make create a **plain POST** (create-only) for host/context/channel instead of PUT-first upsert — K8s returns `409 AlreadyExists`, which this PR's own `clerumErrorHandler` already forwards cleanly. That turns three best-effort client guards into one real server-side control.

<details>
<summary>Lower-severity new findings (Medium/Low)</summary>

- **N4 (Med)** — Latent twin of M3 left unfixed: `control-api/src/k8s.ts:679 extractHttpStatus` doesn't read `.httpStatus` (consumed by `routes/admin/secrets.ts`). control-api now has 5 duplicated status extractors — worth one canonical `httpStatusOf(err)`.
- **N5 (Med)** — `control-ui/app/contexts/[name]/page.tsx:412-419` rebuilds the PUT spec from the form; a future additive field unknown to the form would be lost (no loss today — `displayName` is echoed). Same class as B2, one layer up. Fix: spread the fetched spec.
- **Low** — dead constraint-message branches (`HostWizard/index.tsx:525,540`); stale comments still documenting `?? id` without trim (`desktop-app/src/types.ts:471`, `useContextsDataController.ts:247`); `routes.resources.test.ts:331` still names the old 500 contract; `AppHeader/index.tsx:409` whitespace fallback has no test; raw C1 chars in `routes.resourceFieldValidation.test.ts:23,30` (prefer ``); mock restore at end of test body instead of `afterEach` (`HostWizard.slug-gate.test.tsx:868,972`).
- Structural debt (non-blocking): helper `patchContextMcpServers` (3 copies), canonical `httpStatusOf()` (5), `displayLabel()` per package (fallback still inline in ~11 sites), shared `@clerum/rfc1123` (client mirrors the server regex by hand).

</details>

*Note: the PR is currently **CONFLICTING** — needs a rebase on `dev` before merge.*

Nice work on this round — 8/9 closed with mutation-proof tests and actionable deploy docs. The remaining gaps are all variants of the same create-collision root that a POST-first would close in one move.

alfredolopez80:
## Closure-verification review — head `aa42121a` (multi-agent, verified against real code)

Re-reviewed the current head with parallel area/security/contract/tests agents on an isolated worktree. This is a **closure pass** against the prior `CHANGES_REQUESTED` (head `7f100ed4`): every prior P1/P2 was re-checked against the **real code** of the current head — several **reproduced**, one **mutation-verified**, and the four suites were actually run.

**Recommendation: mergeable (with nits). 0 Blocker · 0 Critical · 0 High.** All prior blocking findings are resolved. Note: GitHub will not clear the earlier `CHANGES_REQUESTED` on its own — a re-review/approve (or dismiss) is still needed to unblock merge.

### ✅ Prior blockers — closure verdicts (verified in `aa42121a`)

| Prior finding | Verdict | Evidence (real code) |
|---|---|---|
| **B1/N1** PUT-first upsert could silently overwrite an existing host/context/channel | **RESOLVED** | `HostWizard/index.tsx:214` `createOrThrow` = `apiSend('POST', …)` only; no PUT-first path remains; coded 409 → "already exists — choose another name". Free-text field removed from the wizard entirely. |
| **P1 — context rename erased `spec.gfs`** | **RESOLVED** | `contexts/[name]/page.tsx save()` spreads `...loadedSpec` and overwrites only form-owned fields; test `ContextDetail.spec-preserve.test.tsx:112` asserts `spec.gfs` deep-equal preserved (fixture with `gfs.mounts`). |
| **N2** server-only display validation → orphaned siblings | **RESOLVED** | Preflight via shared `@clerum/display-field` before the first write + `compensateCreated` (`:181`) inverse-order, 404-safe, runs only when `!hostCreated`. Zero-orphan traced. |
| **B2** McpServer delete/install/uninstall dropped `spec.displayName` | **RESOLVED** | `registry.ts:1289/1660`, `resources.ts:888` all now `spec: {...ctx.spec, …}` — closes the class. |
| **M3/N4** service-layer 404 collapsed to 500 | **RESOLVED** | `errorHandler.ts:39` and `k8s.ts:693` read `.httpStatus`; `K8sNotFound/Conflict` → 404/409. |
| **M1** empty/whitespace displayName → blank label | **RESOLVED** (reproduced) | `@clerum/display-field` rejects `""`/whitespace; consumers use `(x ?? '').trim() || id`. |
| **M4** C1 controls (U+0080–U+009F) accepted | **RESOLVED** (reproduced) | `CONTROL_CHAR_RE` now includes `\x7f-\x9f` + bidi + LS/PS. |
| **N5** PUT rebuilt spec from form → additive fields lost | **RESOLVED** | context (`...loadedSpec`) and host (`...currentHost.spec`) spread the fetched spec; `resourceVersion` preserved. |
| `resolveHostRef` used `spec.host` as identifier | **RESOLVED** | `mcp-host/src/main.ts:250` returns `host?.name`, never the display; test pins it. |
| **M2** deploy order undocumented | **RESOLVED** | `charts/clerum-crds/README.md` documents CRD → control-api → control-ui + silent-prune risk. |
| **B3** body claimed E2E specs were included | **RESOLVED** | Body now states coverage is unit/integration only and E2E is a tracked follow-up. Honest. |
| **P2** pruning guard returns 409 after persist | **CLOSED (re-scoped)** | Still post-commit but only affects `displayName` (an outdated CRD can't store it anyway); PUT is idempotent and converges; the CREATE path rolls back. The 409 message is accurate, not misleading. |

**Positive security fix (verified):** `hostOverviewService.ts` now matches channels on the RFC1123 identifier (`ch.spec?.hostRef === hostName`) instead of the `[hostName, spec.host]` set. Since `spec.host` is now editable, the old match would have surfaced **another agent's channels and recipients** (telegram userIds, emails, slack userNames) under a colliding display name. This was the real hazard of making the field editable — closed correctly.

### 🟠 Open — Medium (non-blocking)

- **Desktop context-display is inert + a test proves an unreachable branch.** `contextDisplayById` has **no production producer** (`appService.refreshAccessCatalog` at `appService.ts:1764-1779` never sets it; `useContextsDataController.ts:248` always resolves `EMPTY_CONTEXT_DISPLAY_BY_ID`), yet `AgentWorkspace.displayName.test.tsx:97-121` injects it via a mocked hook and asserts the render — green over a dead branch. A context display name edited in Control UI never surfaces on desktop. It **is** a documented deferral (`desktop-app/src/types.ts:466-472`), and the **agent**-name path is real and honestly tested (`appService.agentDisplay.test.ts`). Suggest either wiring the real producer + an integration assertion, or marking the context test `skip`/deferred so it doesn't stay green over an unreachable production branch.
- **Rollback logs leak raw K8s `err.message`.** `resources.ts:215` (`rollbackPrunedContextCreate`, only strips `\r\n`) and `resources.ts:900-903` (mcp-server-delete context cleanup, logs raw) emit the `ApiException` message, which embeds the `metav1.Status` body and apiserver headers (`Audit-Id`, `x-kubernetes-pf-*`). Not secrets/PII (hence Medium-Low), but exactly what the global handler scrubs. Fix: route these through `safeErrorLogMessage` (`errorHandler.ts:109`, would need exporting).

### 🟡 Open — Low / coverage (optional)

- **Detail-page display edits are not client-preflighted** (converged across 2 agents). `contexts/[name]/page.tsx` and `hosts/[name]/page.tsx` don't import `@clerum/display-field`; an invalid display (>120/bidi/control) is caught only by the server 422. Single atomic PUT → no orphan; UX asymmetry vs the create page.
- **Wizard context-name gate weaker than `/new`.** `HostWizard` step-1 uses `contextName.trim().length > 0` (`:251`) instead of `isValidResourceSlug`; a >63-char name reaches `submitAll`. Safe (server 422 → secret compensated), but inconsistent.
- **Oversize whitespace-padded displayName stored verbatim.** Length is measured post-`trim()` but the value isn't mutated; `"x"×120 + " "×1000` persists (bounded only by the body limit). Admin-only. (reproduced)
- **Coverage:** host PUT "reads the CR exactly once" is asserted for contexts only, not hosts; `TeamsPage` display branch tests only the empty/fallback path; the `HostWizard` host POST body isn't asserted to carry `spec.host`.

### ℹ️ Informational (not defects)
- ~5 distinct HTTP-status extractors coexist; each reads `.httpStatus` where it matters now — consolidating into one `httpStatusOf(err)` is cleanup, not a bug.
- Non-integer 4xx → RangeError is moot: Node validates the 100–999 range, not integer-ness; `writeHead(404.5)` doesn't throw.

### 🟢 Verified clean (not assumed)
- **Security:** no SQL/command/path/template injection introduced; authz unchanged, fail-closed (validation before side effects); RFC1123 enforced server-side before K8s; bidi/Trojan-Source rejected; no `dangerouslySetInnerHTML`.
- **Contract:** `@clerum/display-field` is the single source of truth, imported by both control-api and control-ui (parity reproduced byte-for-byte: `MAX=120` post-trim, `RE=[\x00-\x1f\x7f-\x9f‪-‮⁦-⁩  ]`). CRD additive-only (`context.yaml` adds optional `displayName`, not in `required`, no pattern; `host.yaml` description-only). Packaging (`main/types/exports`, `.d.ts↔.cjs`, `file:` lockfile) correct — no `npm ci` break.
- **Concurrency:** `preReadCurrent` snapshot used only on attempt 1; a 409 falls back to a fresh read → last-write-wins unchanged.

### 🧪 Suites (executed — real deps installed, `@clerum/*` built via `tsc -b`, electron reinstalled)
| Package | Result | Tests executed |
|---|---|---|
| control-api | **PASS** (exit 0) | 4003 passed / 61 skipped |
| control-ui | **PASS** (exit 0) | 1502 passed / 1 skipped |
| mcp-host | **PASS** (exit 0) | 3272 passed |
| desktop-app (incl. ui) | **PASS** (exit 0) | 1676 passed |

No suite ran 0 tests (`total > 0` everywhere). **Mutation-verified:** reverting the pruning guard at `resources.ts:185` turned `routes.resources.contextDisplayNamePruning.test.ts` red (2 failed); reverted, tree clean. **E2E:** zero Playwright specs in the PR — the four critical flows remain a tracked follow-up, now honestly declared in the body.

---

*Automated multi-agent analysis, verified against real code on an isolated worktree. Findings above are advisory; merge/approve/dismiss decisions remain the maintainers'.*

alfredolopez80:
the medium issue will be resolve in another PR # 339

alfredolopez80 (approved):
the medium issue will be resolve in another PR # 339

Issue #301, opened 2026-08-07
workflow-recipes: reconciler creates workloads before applying their egress NetworkPolicy (TOCTOU window)


alfredolopez80:
## Summary

The workflow-recipes reconciler creates the workload (Deployment / StatefulSet / DaemonSet / Job / CronJob) **before** applying that workload's specific egress `NetworkPolicy`. This leaves a TOCTOU window in which a workload can start — and issue egress — before its egress allow-list has converged.

## Distinction from #299 (not a duplicate)

#299 covers the *stale* FQDN→`/32` drift: a policy that becomes obsolete over time as the provider rotates IPs. **This issue is a separate, secondary defect: the creation ORDER (workload-before-policy) at reconcile/creation time, not the later drift.**

The live diagnosis on #299 showed the *currently observed* failure is continuous drift (affected pods had been running 15–30h), **not** the startup window — so this ordering defect is intentionally split out as a follow-up rather than folded into the minimal #299 fix.

## Evidence (paths on origin/dev)

- `workflow-recipes/src/reconciler/workflowRecipeReconciler.ts` creates the workload around the ensure-workload step.
- The same reconciler applies the workload-specific egress `NetworkPolicy` in a later step, after the workload already exists.
- Result: between "workload created" and "egress policy applied and confirmed", the workload can be scheduled and attempt outbound connections without its intended egress enforcement being fully in place.

(Exact line ranges should be re-confirmed against the current `origin/dev` head before implementation, since the file evolves.)

## Impact

- **Reliability:** a workload that fetches an external dependency immediately on start may fail transiently because its allow-list is not yet applied.
- **Security posture:** for the brief window, egress is not governed by the workload's intended policy. This does not remove the namespace deny-all, but it weakens the "policy-before-workload" guarantee the platform intends.

## Prior art / related

- HCC already introduced an `ExternalEgressConvergenceCoordinator` and decoupled readiness from initial fleet convergence (PR #205; see #204, #215).
- WRC has **no** equivalent "apply-and-confirm egress policy before creating/resuming the workload" ordering. The HCC semantics are a reuse candidate.

## Proposed direction (no implementation committed)

- Render and apply all WRC-owned egress `NetworkPolicies` for the workload **before** creating or updating it; read them back and confirm ownership/selector/ports/CIDRs.
- For a `Job`, do not create it until the egress policy is applied and confirmed.
- For a `CronJob`, keep it suspended until the egress state is convergent.
[...]

Issue #299, opened 2026-08-07
GKE: exact-host FQDN-to-/32 egress becomes stale and intermittently blocks workloads


alfredolopez80:
# Issue draft — GKE exact-host egress drift blocks external workload access

## Proposed title

`GKE: exact-host FQDN-to-/32 egress becomes stale and intermittently blocks workloads`

## Suggested labels

`area/networking`, `area/workflow-recipes`, `area/gke`, `bug`, `security`, `priority/high`

## Summary

An SDK-enabled analyzer workload installed in the `clerum-dev` GKE cluster
fails intermittently before it can execute its optional SDK prompt path. The
failure is in the analyzer's outbound HTTPS access to GitHub:

```text
TypeError: fetch failed
at ghJson (/app/dist/github.js:56)
at resolveRef (/app/dist/github.js:84)
```

The analyzer has a valid GitHub token, the Plugin Workload SDK grant is
validated, the SDK mcp-host is Ready, and the recipe declares all required
GitHub/OCI/CDN hostnames. The failure is caused by the effective GKE egress
allowlist being a point-in-time set of `/32` IP rules that can diverge from the
current DNS answers for GitHub before the next WRC reconciliation.

This issue is reported from GKE because the failure is reproducible there under
the enforced NetworkPolicy dataplane. The underlying `/32` design is broader
than GKE and also produced transient failures in the branch-owned Minikube
profile. The issue must therefore distinguish the GKE runtime manifestation
from the general platform design that permits the drift.

## User-visible impact

- A real analyzer run can fail with `fetch failed` and persist no findings.
- The operator may see healthy analyzer, API, database, SDK, and grant status
  while the actual scan cannot reach GitHub.
- A manually derived Job can fail immediately, which makes the workload look
  misconfigured even when its credentials and SDK policy are correct.
- The failure is intermittent: rerunning the same operation may pass after DNS
  and NetworkPolicy happen to converge.
- The CronJob is not the direct source of the failing request. It launches the
  operation, but the failed GitHub request is issued by the analyzer Deployment.

## Confirmed environment

```text
Cluster context: gke_eventfire-491421_us-central1-a_clerum-dev
Namespace:       sandbox-recipes
Workload class:  SDK-enabled analyzer Deployment
```

The active analyzer image came from the private registry. Its repository and
tag are intentionally omitted from this general issue; the immutable image
digest and deployment manifest remain available in the incident evidence.

[...]

PR #306, opened 2026-08-08
fix(egress): sliding-window FQDN egress accumulation to fix stale /32 drift (#299)

alfredolopez80:
## Scope note (post-adjudication)

Deliberately scoped to **#299 only** (stale FQDN→/32 egress drift). Two commits: the reviewed sliding-window egress fix + one build/packaging commit (`fix(build): package and publish network-policy-core consumers`) that makes the new `@clerum/network-policy-core` package installable via `npm ci` and copied into the service images — without it the images cannot build and the fix could never reach a cluster.

It does **not** include the separate #301 "policy-before-workload / readiness-gate" lifecycle work: an orthogonal concern (creation-order TOCTOU, not stale drift — the #299 incident pods had been running 15–30h, i.e. continuous drift, not a startup window). Verified code-independent: the egress path (`externalEgressAccumulator`, `fqdnResolver`, `networkPolicyReconciler`) never calls workload creation/ordering/readiness-gate code, and #301 has no implementation to depend on. The #299 fix delivers GitHub-egress survival across DNS rotations on its own; live-cluster proof is the T1/T2 gate below.

## Summary

Fixes **#299** — GKE exact-host FQDN→/32 egress becomes stale and intermittently blocks workloads.

Both controllers (WRC = `workflow-recipes`, HCC = `host-context-controller`) materialized FQDN→/32 egress as a **single-generation DNS snapshot with no requeue**. For a provider that serves **one rotating A record per query** (GitHub's `api.github.com` / `github.com` / `codeload.github.com`, TTL ~15s, rotating over `140.82.112.0/20`), the pinned /32 is wrong on the pod's *next* resolution and the NetworkPolicy silently drops egress. Measured success probability for the pinned-IP approach against that pool ≈ 6%.

The fix is a **platform-wide** sliding-window IP accumulator (not per-plugin, not per-domain): each DNS snapshot is folded into the *union of recently-live IPs* persisted on the policy's annotations, plus a TTL-aware periodic requeue so the window keeps sliding without a CRD event.

## Design (shared pure core: `@clerum/network-policy-core`)

Pure module, injected clock, `node:test` (no Kubernetes, no DNS). Guarantees:

- **§0 monotonically permissive** — the latest OK response is always in the set; never regress vs the single snapshot.
- **H1 fail-static** — a *declared* FQDN whose DNS *transiently* fails is **frozen** (never pruned while failing); a *permanent* no-records answer expires normally; an FQDN with **no observation** (removed from the spec) is **revoked** immediately.
- **H3 eviction-as-alarm** — on overflow, evict least-recently-observed; **never reject** the policy.
- **H4 no-op** — change hash over `(fqdn,ip,port,protocol)` only; a timestamp-only refresh writes nothing.
- **H5 rehydration** — legacy `egress-fqdn-targets` is not blanked; malformed IPs are dropped.
- **renewalDue** — re-persist the window before it lapses (only when a write can actually advance expiry) so a *stable-then-rotated* IP keeps its overlap grace.
- **`classifyDnsError`** — single source of truth for transient-vs-permanent, shared by WRC and HCC.

## Controller changes (WRC + HCC)

- `resolve4({ ttl: true })` → per-fqdn TTL feeds the window.
- Per-fqdn accumulator adapters fold snapshots and persist state on the policy annotations.
- TTL-aware **self-rescheduling** refresh/resync: advance to `≤ TTL/2`, floored; config knobs + startup invariant `floor ≤ interval < overlap`.
- Write gate on a **representation-independent egress signature** (order-insensitive) so a timestamp-only refresh is a no-op (no apiserver/CNI churn) while genuine set changes, out-of-band drift, and window renewals still write.
- Security gates (blocked/private/metadata CIDRs) run on the **fresh** snapshot before any IP enters the accumulator; rehydrated IPs are IPv4-validated — the union can never contain a blocked or malformed IP.

## Scope / architecture

No architecture change. Both `clerum-dev` and `clerum-prod` run **Calico (legacy)** — no Dataplane V2 / FQDN-aware policies — so a DPv2 migration is explicitly out of scope. This fix works within the existing Calico + /32-ipBlock model.

## Quality — unit gates (T0) GREEN

- `@clerum/network-policy-core`: **32/32** (`node --test`)
- `host-context-controller`: **1066/1066** (`vitest`), `tsc --noEmit` clean
- `workflow-recipes`: **2401/2401** (`vitest`), `tsc --noEmit` clean

Hardened across **three rounds of adversarial review** (Claude Opus + Fable): fail-static vs revocation, write-storm anti-churn, rehydration validation, TTL-aware timer safety (no double-timer / post-stop fire / overlap; `unref`; floor holds), and a key-ordering write-gate defect (a raw `JSON.stringify` spec compare is always unequal against the apiserver's deserialization order — replaced with a semantic signature).

## ⚠️ Required before merge (draft)

- [ ] **T1 — minikube (Calico) e2e**: deterministic fixtures per plan §7 (1-A rotating, large-pool rotating, stable, established-flow, controller restart, sustained DNS failure). **Must assert the `unchanged — no-op` log appears on a real cluster** (that live assertion is what would catch the key-ordering class of write-gate bug). Also **measure the status-churn from #307**: WorkflowRecipe status-patch rate + `resourceVersion` growth per egress recipe (expect ~8/min under GitHub's 15s TTL), WRC→apiserver request rate, and `syncFromRecipeExecution` DB write rate per tick.
- [ ] **T2 — real plugin stack**: `evenfire-vulnerability-plugin` in minikube against real GitHub (RED: CronJob never completes → GREEN: Completed across several rotations) + no-regression for `storage.googleapis.com` / Gmail / Anthropic.

## Known deferred: status-patch churn (#307)

This PR's TTL-aware egress refresh loop force-reconciles egress recipes on a fast cadence (~7.5s under GitHub's TTL), which **amplifies** a **pre-existing** unconditional WorkflowRecipe status write (`lastTransitionTime: now`, no dirty-check — identical in `dev`): egress-only recipes go from ~0 to ~8 status writes + watch events/min. This is **churn, not a correctness bug** (no feedback loop; the resulting MODIFIED is status-only and short-circuits). The correct fix is an HCC-style normalized status dirty-check, but that touches the status path shared by **all** WorkflowRecipe reconciles and HCC's own precedent OOM-looped prod when done naively — too risky for this PR. Tracked in **#307**; T1/T2 measure the real rate to gate its priority.

## Notes

- `@clerum/network-policy-core` is linked via `file:../packages/network-policy-core` from both controllers (same pattern as the other `@clerum/*` packages).
- Config env knobs: `HCC_EXTERNAL_EGRESS_{RESYNC_SEC,OVERLAP_SEC,REFRESH_FLOOR_SEC,MAX_ENTRIES}` and the WRC `externalEgress*` equivalents.

Refs #299


packages/network-policy-core/index.cjs
```diff
@@ -0,0 +1,301 @@
...
+  //  - Declared but DNS transiently failing (explicit transient observation):
+  //    FREEZE — never prune while the resolver is failing.
+  //  - No observation AT ALL this round: the FQDN is no longer DECLARED (removed
+  //    from the spec) → REVOKE immediately. Freezing an undeclared FQDN would
+  //    keep a removed host's egress forever — a regression worse than the
+  //    original single-snapshot behavior. Only an explicit transient failure
+  //    freezes; absence of an observation means removal.
+  const frozen = new Set()
+  for (const [k, e] of map) {
+    if (okFqdns.has(e.fqdn) || permanentFqdns.has(e.fqdn)) {
```
alfredolopez80: [High] Revoke removed binding identities, not just FQDNs

`okFqdns`, `permanentFqdns`, and `transientFqdns` are keyed only by FQDN, while the state map is keyed by `(fqdn, ip, port, protocol)`. If `api.example.com:443` changes to `:8443`, the old port is still treated as declared and survives until the previous TTL plus overlap. Because WRC renders `entry.port`, both ports remain allowed. Please track declaration/revocation by the full `(fqdn, port, protocol)` identity and add a regression for removing one port while keeping another on the same FQDN.

packages/network-policy-core/index.cjs
```diff
@@ -0,0 +1,301 @@
...
+      const eq = pair.indexOf('=')
+      if (eq < 0) continue
+      const fqdn = pair.slice(0, eq).trim()
+      const cidr = pair.slice(eq + 1).trim()
+      if (!fqdn || !cidr) continue
+      const ip = cidr.split('/')[0]
+      if (!isValidIpv4(ip)) continue
+      entries.push({
+        ip,
+        port: LEGACY_PORT,
```
alfredolopez80: [High] Do not invent TCP/443 during legacy migration

The legacy `egress-fqdn-targets` annotation contains only `fqdn=ip/32`, but this parser assigns every entry `443/TCP`. A pre-existing policy for `db.example.com:5432` is therefore widened to both `443` and `5432` during the first reconcile. Please derive the port/protocol from the current declaration or the live NetworkPolicy, and reject ambiguous legacy state instead of guessing.

workflow-recipes/src/reconciler/workflowRecipeReconciler.ts
```diff
@@ -627,6 +634,39 @@ export class RuntimeScopeResolutionPendingError extends Error {
...
+ * apiserver/client deserializes rule objects with keys ordered differently than
+ * the builder emits them, so a raw compare is always unequal on a live cluster
+ * and would defeat the no-churn gate. Unlike the earlier tuple-only signature,
+ * this captures the FULL rule — including `namespaceSelector`/`podSelector` `to`
+ * entries for internal/cluster-local egress (audit H1): a tuple signature over
+ * ipBlock only was blind to selector changes, stranding a stale policy. Mirrors
+ * HCC's `egressSignature`.
+ */
+export function egressSignature(policy: k8s.V1NetworkPolicy): string {
+  return JSON.stringify(canonicalize(policy.spec?.egress ?? []))
```
alfredolopez80: [High] Include top-level policy semantics in the no-op signature

This signature hashes only `spec.egress`. A live policy whose `podSelector` or `policyTypes` drifted can therefore be treated as unchanged when its destination rules match, leaving external access applied to the wrong pods. Please compare a canonical projection of the full managed enforcement spec, including `podSelector`, normalized `policyTypes`, and `egress`. The HCC signature should use the same projection.

workflow-recipes/src/reconciler/workflowRecipeReconciler.ts
```diff
@@ -3905,20 +3951,76 @@ export class WorkflowRecipeReconciler {
...
+
+    let effectiveResolved: rb.ResolvedExternalEgressInput[] = resolved
+    let stateAnnotations: Record<string, string> | undefined
+    let existing: k8s.V1NetworkPolicy | null = null
+    let egressRenewalDue = false
     if (externals.length > 0) {
+      // issue #299: fold this DNS snapshot into the accumulated sliding-window
+      // set persisted on the live policy's annotations (rehydrate H5).
+      existing = await this.readNetworkPolicyOrNull(policyName, ns)
+      const acc = accumulateExternalEgress({
```
alfredolopez80, line 3997: [High] Process permanent DNS observations before surfacing failure

This new accumulator path is unreachable for permanent DNS failures because the earlier guard throws before `readNetworkPolicyOrNull()` and `accumulateExternalEgress()` are reached. The existing NetworkPolicy therefore never enters the permanent-expiry path and can retain an old `/32` indefinitely after NXDOMAIN/ENODATA. Please reconcile the persisted state first, apply or delete the resulting policy, and only then report degraded/requeue status. Add an integration test starting from a live accumulated policy.

workflow-recipes/src/reconciler/workflowRecipeReconciler.ts
```diff
@@ -3928,6 +4030,17 @@ export class WorkflowRecipeReconciler {
       return
     }
 
+    // issue #299: NO-OP when the accumulated egress set and rules are already
+    // live — a TTL-only refresh must not churn the apiserver/dataplane. But DO
+    // write when the persisted window is aging (renewalDue, audit M1), even if
+    // the set is unchanged, so a stable-then-rotated IP keeps its overlap grace.
+    if (!this.egressWriteNeeded(existing, policy) && !egressRenewalDue) {
```
alfredolopez80: [High] Persist accumulator identity changes even when egress rules are equal

`accumulateExternalEgress()` returns `changed`, but this write gate uses only the rendered egress signature and `renewalDue`. A rename from `old.example.com` to `new.example.com` with the same IP/port therefore leaves the old FQDN in persisted state because the rendered rule is identical. When the new hostname later rotates, the intended overlap is lost. Please include `acc.changed` in the write decision and add a rename -> no-op -> later-rotation regression.

alfredolopez80 (commented):
Request changes — do not merge yet.

I reviewed PR #306 at base `38be8570239691b9481e6e35ea9648943ce5dcf1` and head `b9abfae22f749fdcd6fd01d9bbd798fe0cf32d44`.

The focused core tests and image-manifest checks pass, but I found five High-severity blockers. Four are introduced by this PR; the permanent-DNS integration gap is pre-existing at the call site but makes the new expiry contract unreachable.

Additional Medium findings:
- `packages/network-policy-core` is missing from the aggregate CI/test/preflight inventories.
- `packages/network-policy-core/**` is unmapped by the incremental GHCR pre-gate.
- HCC accepts timer values that can overflow Node's timer range, and the entry cap is not byte-aware for serialized annotations.
- The process-wide minimum TTL only ratchets down while each refresh forces broad full reconciles, so the PR's required churn measurements are still needed.

The PR's required T1/T2 evidence is also still pending: exact-head Node 24 build/typecheck, Docker/image startup, branch-owned Minikube, real DNS rotation, restart/rehydration, API-server acceptance, no-op/resourceVersion evidence, and Calico enforcement. The focused green checks are not a substitute for those runtime gates.

Please address the inline blockers and add regressions before merge.

packages/network-policy-core/index.cjs
alfredolopez80: **Fixed in 442031dd.** `reconcileEgressState` now classifies revocation/freeze by the full declared identity `(fqdn,port,protocol)` via `idOf()`, not by fqdn alone. A declared-port change (443→8443) now revokes the old-port entry immediately instead of letting it ride the still-declared fqdn's `ok` branch until its old TTL+overlap lapses, and the H1 transient freeze is identity-scoped (a removed port is revoked even while the still-declared port's DNS is failing).

Regressions added in `index.test.cjs`: port-change revoke, identity-scoped revoke-under-transient, and same-identity freeze retained (H1). Mutation-verified — reverting to fqdn-keying fails exactly the two H-A tests while H1/F1 stay green. Both accumulators emit one observation per declared `(fqdn,port)`, so a still-declared identity always has an observation (never wrongly revoked) and a removed one never does (never wrongly retained).

packages/network-policy-core/index.cjs
alfredolopez80: **Fixed in 442031dd.** `parseState` now takes the current declarations and derives the migrated port/protocol from them, dropping legacy pairs whose FQDN is no longer declared, instead of guessing `443/TCP`. A pre-existing `db.example.com=ip/32` legacy target declared on `5432` now migrates to `5432` (not widened to `443`+`5432`). It falls back to the historical `443` grace only when declarations are unavailable (isolated callers); both accumulators (WRC `accumulateExternalEgress`, HCC `accumulateHostExactHostEgress`) pass their declared targets.

Regressions in `index.test.cjs` (declared-port migration, undeclared drop, no-declarations fallback). Mutation-verified — reverting the legacy branch to unconditional `443/TCP` fails exactly the two H-B tests. The new-format `egress-fqdn-state` path is unchanged.

workflow-recipes/src/reconciler/workflowRecipeReconciler.ts
alfredolopez80: **Fixed in 442031dd.** `egressSignature` now hashes `canonicalize({podSelector, policyTypes (sorted), egress})` in both WRC and HCC, not `spec.egress` alone, so a policy whose `podSelector` or `policyTypes` drifted out-of-band with identical destination rules counts as changed and the reconcile re-owns it. `policyTypes` is order-insensitive.

Regressions in `egressSignature.test.ts` (both packages): podSelector drift detected, policyTypes change detected, `Egress,Ingress == Ingress,Egress`. The R2-1 key-order no-op and the H1 selector-in-egress detection are preserved — identical content still yields an identical signature, so the #299 no-churn gate is not weakened. Mutation-verified — reverting to `canonicalize(spec.egress)` fails exactly the two H-C tests per package.

workflow-recipes/src/reconciler/workflowRecipeReconciler.ts
alfredolopez80: **Dismissed — won't fix in #306 (nor as a follow-up).** Two independent Fable 5 passes confirm the mechanic is real (WRC throws on a permanent DNS failure before the accumulator's permanent-expiry branch runs, so the live policy keeps its /32s) but conclude it does not warrant action:

- **Pre-existing, and the PR improved it.** The throw is from the initial commit; #306 narrowed it to permanent-only and added accumulation for transient failures. Not a regression.
- **Fail-loud, not silent.** A permanent DNS failure drives the recipe to the terminal `failed` phase; removing the FQDN from the spec revokes its /32s immediately via F1. There is no silent data-plane state.
- **The suggested fix is unsafe.** "Apply the shrunken policy, then degrade" would silently mutate the policy on a permanent failure — the opposite of the repo's fail-loud contract that the T0/T1 gates just certified.

Retaining a dead host's last-known /32s while the recipe is visibly `failed` is a bounded, operator-recoverable, defense-in-depth item, not a correctness bug — so it is closed rather than carried as debt.

workflow-recipes/src/reconciler/workflowRecipeReconciler.ts
alfredolopez80: **Fixed in 442031dd.** The write gate now includes `acc.changed` (WRC ui + workload sites, HCC), so a rename `old.example.com → new.example.com` onto the same ip/port re-persists the re-attributed state annotation instead of no-op'ing it — preserving the overlap grace when the new name later rotates. A pure timestamp refresh still has `changed=false` (the change hash is over `(fqdn,ip,port,protocol)`, H4), so there is no write storm; a genuine IP rotation already changes `spec.egress`.

Guarded by a mutation-verified reconciler test — `reconciler.test.ts` R.8.24 reconciles with `old`→IP, then with `new`→same IP, and asserts the ui-egress policy is re-written; reverting the gate term makes it fail. Accumulator-level re-attribution covered in `externalEgressAccumulator.test.ts`.

zach88:
Took a deep pass through this one across the core, both controllers, build/packaging, security, and cross-controller contract consistency. Overall this is solid work: the #299 fix is correct, security checks out clean (the rehydration re-gate is present at all four render sites, so nothing that skips the blocklist can reach an `ipBlock`), and the shared annotation format via the core is the right call. No blockers. What I'd want addressed before merge is one review-visibility issue plus a handful of real HCC/WRC divergences and test gaps. IDs below are stable across rounds (from the ledger).

## High

**1. [R1-H1] Raw NUL bytes in `idOf` hide the whole core from diff review** — `packages/network-policy-core/index.cjs:115`. `idOf` builds its key with literal `0x00` bytes (`` `${fqdn}\x00${port}\x00${protocol}` ``). Confirmed with a hexdump: exactly 2 raw NUL bytes, both here. That makes git classify the entire file as binary (`Bin 0 -> 14305`), so all 391 lines of the core (the heart of this PR) never showed up as a reviewable diff on GitHub. It's also fragile: any prettier/editor/copy-paste pass can silently strip the NUL and break the delimiter. Not a runtime bug (the in-memory string is identical), but worth fixing before merge so the core actually gets a diff review. The sibling `keyOf` (line 98) already does the same thing with the textual escape `` — just use that here. NUL is the right delimiter choice; only the raw-byte encoding is the problem.

## Medium

**2. [R1-M2] WRC's no-churn gate drops HCC's `desiredState` guard, so internal-only policies get rewritten every reconcile** — `workflow-recipes/src/reconciler/workflowRecipeReconciler.ts:4613`. HCC's `externalEgressWriteNeeded` (`host-context-controller/src/networkPolicyReconciler.ts:248`) ends with `return Boolean(desiredState) && !existing...[STATE_ANNOTATION]`; WRC's `egressWriteNeeded` just does `return !existing.metadata?.annotations?.[STATE_ANNOTATION]`. For a recipe with `ui.egress.internal[]` but no `external[]`, `externals.length === 0` skips the accumulator, the desired policy carries no `STATE_ANNOTATION`, and `egressWriteNeeded` returns `!undefined === true` forever, so the `if (!writeNeeded && !renewalDue && !changed) return` guard at `:4056`/`:4491` never trips and the policy is rewritten on every reconcile. HCC returns `false` in the same spot. This is amplified by the 300->60s resync bump in this PR. Fix: mirror HCC's `Boolean(desiredState) &&` guard in WRC.

**3. [R1-M1] Byte-aware eviction only budgets the STATE annotation, not STATE+TARGETS against the 256KB apiserver cap** — `packages/network-policy-core/index.cjs:220-244`. `serializeState` also emits `egress-fqdn-targets`, and the 256KB limit is on the total of all annotations. Reproduced with a 253-char FQDN: STATE = 199,708 B + TARGETS = 151,994 B = 351,702 B > 262,144 B, which is exactly the apiserver-422 reconcile loop the comment at lines 20-28 says it prevents (egress goes stale, it doesn't widen). WRC's own ceiling comment (`config.ts:139`, "1300 entries ~= 195KB, a safe ceiling") only counts STATE. Realistic exposure is low with GitHub's short FQDNs, but the mitigation is incomplete for its own stated failure mode. Fix: budget against STATE+TARGETS combined, or cap the TARGETS length, or lower the max-entries ceiling to something safe for long FQDNs.

**4. [R1-M3] The central no-op invariant has no reconciler-level test** — `workflow-recipes/src/reconciler/workflowRecipeReconciler.ts:4049` and `:4491` (and the HCC equivalent). The whole point of the PR is "a TTL-only refresh must not churn the apiserver", but the no-op branch is only exercised in the negative: the rename test asserts a write *does* happen. Nothing asserts that a second reconcile with identical resolver output plus a read of the prior policy does *not* call `createOrReplace`. Related T1 nuance: the rename test feeds the builder's own output back as the "live" policy, but a real apiserver read injects defaults (`protocol: 'TCP'`) and reorders keys, so the equality path against an apiserver-shaped policy (the exact thing the key-ordering fix motivates) is untested. Fix: add a no-op reconcile test with a fixture derived from a normalized read.

**5. [R1-M4] The self-rescheduling resync timer has no test** — `host-context-controller/src/k8sClient.ts:1848-1868`. Only the pure helper `externalEgressResyncDelayMs` is covered. The actual behavioral change of the PR (the timer) has no coverage for the idempotency guard `if (!this.externalEgressResyncTimer)` ("never arm a second chain"), the fire-after-`stop()` path, or the reschedule-after-completion cadence. They read correct by hand, but the invariants the comments promise aren't executable. Fix: a fake-timer test (arms once; a second `start()` doesn't double it; `stop()` prevents a further tick; the next tick is scheduled only after the resync resolves).

**6. [R1-M5] Divergent max-entries ceiling plus a false contract comment** — `host-context-controller/src/config.ts:398-400`. HCC caps `externalEgressMaxEntries` at 4096 with the comment "matches WRC's accumulator contract"; WRC caps at 1300 (`workflow-recipes/src/config.ts:140`). The comment is false (1300 != 4096), and the minimum also differs (HCC 1 vs WRC 8). The core's byte-cap trims anything past ~1300 short entries anyway, so a lot of HCC's advertised 4096 headroom is unreachable. No cross-drift (each controller writes its own policies), but it's a false contract claim and inconsistent knob bounds. Fix: correct the comment and reconcile (or justify) the ceilings.

**7. [R1-M6] Docs say `floor <= interval < overlap`, code enforces the stricter `interval <= overlap/2`** — `host-context-controller/src/config.ts:388`, `workflow-recipes/src/config.ts:90`, the `k8sClient.ts` JSDoc, and two test names (`config.externalEgressResync.test.ts:36`, `config.test.ts:302`). An operator setting `interval=200, overlap=300` (valid per the docs) gets a startup crash. It fails loud so it's safe, but the docs actively misdescribe the contract and the only boundary tested is `interval == overlap`. Fix: align the comments/body/test names to `<= overlap/2` and add a boundary test (`interval = overlap/2` passes, `+1` throws).

## Minor (grab-bag)

- **[R1-L1]** `host-context-controller/src/config.ts:692-695` — the four new knobs use `getEnvInt` (parseInt), which silently coerces `"60.5"->60` / `"60abc"->60`, unlike the sibling loud parsers. The invariant re-checks ranges, so an in-range typo slips through quietly.
- **[R1-L2]** `maxEntries` means per-FQDN in HCC (one policy per FQDN) but per-policy across all FQDNs in WRC (`accumulateExternalEgress` takes `externals[]`). Same knob name, different capacity. Not a bug, but a hidden divergence.
- **[R1-L3]** Disable semantics and bounds differ: HCC allows `interval=0` to disable the resync and uses `overlap` min 1; WRC forbids `interval=0` and uses `overlap` min 30. The ordering invariant itself is identical.
- **[R1-L4]** `packages/network-policy-core/index.d.ts:103` — the `stateHash`/`changed` JSDoc says the hash is over "(ip,port,protocol) only", but the impl (and H4) key over "(fqdn,ip,port,protocol)". WRC's rename detection (`workflowRecipeReconciler.ts:3986`) depends on fqdn being in the hash, so the published JSDoc is the thing that's wrong.
- **[R1-L5]** `workflow-recipes/src/reconciler/externalEgressAccumulator.ts:21` — JSDoc says "H3 evict soonest-to-expire", but the core evicts least-recently-observed (expiry is only a tiebreak). Comment-only.
- **[R1-L6]** Protocol divergence: WRC models egress as `{fqdn,port}` (always TCP via the core default), HCC carries the real protocol. Both serialize through the core, so the wire annotation does carry protocol. Latent: if HCC ever wrote a `UDP` policy and WRC reconciled it, WRC wouldn't emit the matching observation and that entry would fall to revocation. They don't share policies today, but WRC can't express UDP external egress at all.
- **[R1-L7]** `packages/network-policy-core/index.cjs:191-201` — the fail-static freeze has no upper time bound. Someone who can sustain a transient DNS failure (SERVFAIL) for one FQDN pins its already-resolved public IPs indefinitely. It's the intended availability tradeoff and it's bounded by spec-removal revocation, permanent-failure expiry, eviction, and the public-only re-gate, but the freeze itself is unbounded. Consider a max-freeze that reverts to fail-closed.
- **[R1-L8]** `workflow-recipes/src/reconciler/externalEgressAccumulator.ts:112-121` — `buildObservations` indexes failures by `fqdn` only, then emits a transient/permanent observation for every declared `(fqdn,port)` including a port that resolved OK this round. Only reachable when the same FQDN is on multiple ports and two lookups diverge; self-heals next round.
- **[R1-L9]** `egressSignature` + `canonicalize` (and the 18-entry CIDR blocklist + `ipv4ToNumber`/`cidrRange`/`cidrOverlaps`) are duplicated byte-for-byte in HCC (`networkPolicyReconciler.ts:118-151`) and WRC (`workflowRecipeReconciler.ts:643-680`), even though this PR added the shared core to dedupe exactly this. Equivalent today, but a future edit to one and not the other would diverge the no-churn gate silently. Should live in the core.
- **[R1-L10]** `packages/network-policy-core/index.test.cjs` — `reconcileEgressState` is a pure `(state, observations, now) -> state` merge with precedence rules but has only enumerated cases, no `fast-check`. The properties that matter (total ordering of entries, no duplicate composite keys, no declared-OK IP dropped, idempotency, composability across rounds) aren't asserted. Both accumulators depend on this function. Dev-only, no new runtime dep.
- Three quick nits: `index.cjs` uses three different delimiters for near-identical composite keys (`keyOf` -> ``, `idOf` -> raw NUL, `stateHash` -> `|`), worth unifying on ``; there's an orphaned JSDoc block sitting above the wrong getter in `workflowRecipeReconciler.ts` so `egressWriteNeeded` has no doc; and `scripts/minikube/pre-gate-sync.sh:398` mixes spaces into a tab-indented block.

## What checked out clean

- Security is SAFE. Rehydrated IPs are re-gated against the blocklist at all four render sites (HCC `:1039,:1076`, WRC `:4024,:4465`), verified by injecting `10.0.0.5` into both the annotation and spec egress and watching it get dropped from the rendered policy (`networkPolicyReconciler.test.ts:974`). Metadata (`169.254.0.0/16`) and `100.64.0.0/10` are in both blocklists; config is fully clamped and no knob disables a gate.
- `classifyDnsError` is a real single source of truth in the core, imported by both controllers with no divergent local copy.
- The annotation format is shared through the core (`serializeState`/`parseState`), so HCC and WRC are mutually parseable (protocol caveat in R1-L6 aside).
- Build/packaging is clean: real `npm ci` + `npm run build` in HCC and WRC with the `file:` dep, Dockerfiles copy `packages/network-policy-core` before every `npm ci` in builder and prod stages, dockerignores whitelist it, `deploy/images.json` and `test-images-manifest.sh` (21/21) are consistent, and no consumer Dockerfile was left unupdated.
- Suites: core `node --test` 41/41; HCC vitest 181/181 on the scope suites; WRC vitest green (509 in the batch); `test-images-manifest.sh` 21/21. `tsc` wasn't conclusive under the review's linked-`node_modules` layout (an environment artifact, not a diff defect).

One heads-up: the T1/T2 live-cluster gates in the description (minikube Calico e2e + real plugin stack) are still unchecked, and those are the real correctness gate for this in a cluster. Happy to re-review once the above are addressed.

zach88:
Took another pass, this time as round 2 focused on the fix commit that closed out round 1 (`17eb4d5b`). Good news first: all 10 findings that commit set out to close (H1, M1–M6, L1, L4, L5) actually check out against the real code, and their regression tests are red against the prior head, so they cover the bugs for real. Ran a dedicated collateral pass over the fix diff too and it came back clean: no regressions introduced by the fixes. Security is still SAFE. There's one new Medium worth fixing before merge (it's a one-liner) plus a couple of test/doc nits.

## Medium

1. **[R2-M1] Byte-budget measures UTF-16 code units, not UTF-8 bytes** — `packages/network-policy-core/index.cjs:237`. The R1-M1 fix now charges STATE+TARGETS against one combined budget, which is correct for ASCII. But `sizeOf` uses `JSON.stringify(view).length` + `` `${fqdn}=${ip}/32`.length `` — `.length` counts UTF-16 code units while the apiserver counts bytes. So the commit's stated guarantee ("STATE+TARGETS can never exceed 256KB") doesn't hold for a multi-byte FQDN.
   - Repro: fed `reconcileEgressState` + `serializeState` a multi-byte FQDN and measured `Buffer.byteLength(...,'utf8')` → STATE+TARGETS = **584,792 B**, well over 262,144 — exactly the 422/reconcile-loop this budget exists to prevent. The ASCII/#299 (GitHub) path stays correct with headroom (229,976 ≤ 262,144), so real exposure today is low — hence Medium, not High. `o.fqdn` isn't ASCII-normalized on the way in, though.
   - The test at `index.test.cjs:432` asserts with `.length` too, so it replicates the same measurement bug and could never catch this.
   - Fix is small: `Buffer.byteLength(s,'utf8')` in `sizeOf` and in the test assertion.

## Grab-bag (Low)

2. **[R2-L1] Stale wording in one resync test** — `host-context-controller/src/config.externalEgressResync.test.ts:26,28`. The R1-M6 fix realigned the invariant to `interval ≤ overlap/2` and renamed two tests, but this one's name/comment still say `< overlap`. Not misleading (default 60 satisfies both), just a doc-contract leftover. Worth aligning the phrasing.

3. **[R2-L2] Vacuous "protocol defaulted to TCP" assertion** — `workflow-recipes/src/reconciler/reconciler.test.ts:4578`. The no-op reconciler test added for R1-M3 is good on the key-reorder dimension, but the builder already emits `protocol: 'TCP'` on every port (`resourceBuilder.ts:1408,1420`), so `pt.protocol ?? 'TCP'` changes nothing and the test never actually exercises protocol-defaulting its comment claims. If the apiserver ever injected other defaults (`endPort`, `to:[]`) this wouldn't cover it. Either drop `protocol` from the `live` fixture to force the default, or narrow the comment to "covers key reorder only".

Minor, no change needed: hoisting the read in `reconcileUiEgressPolicy` (`workflowRecipeReconciler.ts:3993`) adds one extra GET per reconcile even for UI recipes with no egress — harmless (404→null via `readNetworkPolicyOrNull`), mirrors HCC.

## What checked out clean

- **No regressions.** Reproduced the six closures against the pre-fix head: R1-M1 normal case is byte-identical, overflow strictly better (351,800 → 229,642 B), §0 preserved (the newest IP is never evicted); R1-M2 the hoisted read only feeds `egressWriteNeeded` and cuts apiserver calls rather than adding them; R1-M5/M6/L1 are doc/tightening changes with byte-identical enforcement and no deployed config that breaks.
- **R1-M2 gate now matches HCC** byte-for-byte (`Boolean(desiredState) && !existing…[STATE_ANNOTATION]`); internal-only policies stop rewriting every reconcile — reproduced.
- **R1-L1 loud parser**: unset → default, malformed → throw; the only deployed knob is `"60"`, so nothing crash-loops on current config.
- **Suites** on the PR head: core `node --test` 41/41, HCC 113/113, WRC 351/351. T3 confirmed red on `17eb4d5b~1` for R1-M1 (bytes) and R1-L1 (`'60.5'`).

Round 2 of 3, on track. What still gates merge is the T1/T2 evidence in the description (minikube Calico e2e + real plugin against GitHub) — both still unchecked. #307 status-churn stays a known, correct deferral.

zach88:
Round 3 pass. This one's clean: all three round-2 findings are closed and verified against the real code, nothing new turned up, and the collateral audit came back with zero regressions. Only egress-scope commit since round 2 is `feaa579e`; the `7186d73b` head is just a dev merge (plugin-UI-SDK / ux work) that doesn't touch any egress file, so it's out of scope here.

Nothing left to action from the review side. What's still gating the merge is your own T1/T2 boxes, not this review.

## Closed & verified

1. **[R2-M1] Byte budget now charges UTF-8 bytes, not UTF-16 length** — `packages/network-policy-core/index.cjs:238-243`. `sizeOf` moved from `.length` to `Buffer.byteLength(...,'utf8')`. Confirmed it's the only spot in the core that meters annotation payload against the 256KB cap (grepped `.length` near the budget; the rest are `array.length`, no siblings). Repro checks out both ways: the new test fails red against the parent `8e70cf9a` (`375896 > 262144` with the old meter), and the ASCII #299 path (GitHub, 900 IPs) is byte-identical before/after (609/609 entries, `overCap=true`), so the real-world path doesn't shift — only the multi-byte case gets fixed.

2. **[R2-L1] Resync doc aligned to `<= overlap/2`** — `host-context-controller/src/config.externalEgressResync.test.ts:26-30`. Title + comment reworded from `< overlap`; doc-only, no assertion change.

3. **[R2-L2] Dropped the vacuous protocol default in the R1-M3 test** — `workflow-recipes/src/reconciler/reconciler.test.ts:3949-3984`. `pt.protocol ?? 'TCP'` → `pt.protocol`. The new comment ("egressSignature treats an absent protocol as distinct from 'TCP'") checks out against the real `egressSignature`: it canonicalizes without injecting a default and `JSON.stringify` drops `undefined`, so `{port:X}` and `{port:X,protocol:'TCP'}` sign differently. The builder always emits an explicit `protocol` on every egress port, so removing the default makes the test more faithful, not less.

## Regressions

None. `feaa579e` didn't break anything — the only runtime change is `sizeOf`, and it can't fail open (`Buffer.byteLength(s) >= s.length` for every string, so it never keeps an entry the old meter would have evicted; multi-byte just evicts more, which is what the apiserver's byte cap actually wants). Return shape, `overCap`, and the LRU eviction are identical for ASCII inputs (reproduced).

## What checked out clean

- Security: the blocked/private/metadata CIDR gates still run on the fresh DNS snapshot before anything enters the accumulator, eviction only removes entries (never re-adds), and rehydrated IPs stay IPv4-validated. Change only hardens the byte accounting.
- Suites in the worktree: core `node --test` **42/42** (incl. the new R2-M1 case); WRC `reconciler.test.ts` **290/290** (R1-M3 logs `unchanged — no-op`). HCC `config.externalEgressResync.test.ts` wasn't runnable in my env, but R2-L1 is doc-only so it can't regress behavior (commit reports 14/14).

## Still open (all Low, carried from round 1, none blocking)

R1-L2/L3 (per-FQDN vs per-policy knob semantics, disable/edge divergence), R1-L6/L7/L9/L10 (deferred by the commit body — UDP format, fail-static upper bound, seam dedup, property tests), R1-L8 (per-`(fqdn,port)` failure collapsed to per-`fqdn`, self-heals). No Medium+ open, nothing living past its shelf life.

## Merge gate

Zero blockers from the review. The two `[ ]` in the body are what's left: T1 (minikube Calico e2e, must assert the `unchanged — no-op` log on a real cluster) and T2 (real plugin stack against GitHub). Without that live evidence the fix isn't proven end-to-end. #307 (status-patch churn) stays a known, correct deferral; T1 measures its real rate.

zach88 (approved):

PR #308, opened 2026-08-09
fix(workflow-recipes)!: default recipe isolation to standard, not minimal

andypavia:
## What

Change the default `spec.security.isolationLevel` for WorkflowRecipes from **`minimal`** to **`standard`**. A recipe that omits the field previously ran as UID 0 with a writable root filesystem; it now gets `runAsNonRoot: true` + `readOnlyRootFilesystem: true`, while keeping the existing hard floors (no privilege escalation, drop ALL caps, seccomp `RuntimeDefault`).

The level *definitions* in `securityContext.ts` are unchanged — only the default *selection* moves. `minimal` remains available and must now be requested explicitly.

Addresses **F1** of the small-webapp / untrusted-workload security review: `minimal` (root + writable rootfs) is the wrong default floor for user/untrusted workloads.

## Changed

Every effective default was updated:
- `workflowRecipeReconciler.ts` (3 resolution sites) — the level used at deploy time.
- `policyEnforcer.ts` — the `requiredSecurityLevel` floor check now evaluates a no-security recipe as `standard`, consistent with what actually runs.
- `resourceBuilder.ts` (5 builder param fallbacks) — safety net for direct callers.

Plus fixups (2nd commit): four recipes that omit `isolationLevel` but run root/writable-fs images now declare `minimal` explicitly, preserving their prior (implicit-minimal) behavior — 2 e2e fixtures (mongodb/postgres) and 2 samples (busybox+postgres, nginx-unprivileged).

## ⚠️ Breaking (operational)

Recipes with no `isolationLevel` that rely on running as root or writing to the container root filesystem (e.g. stock mongo/nginx/redis images) will now fail to start. Fix: set `spec.security.isolationLevel: minimal`, or run non-root with a writable volume where the app needs to write.

Note: the capability floor (`drop: ALL`) is applied at *all* levels, so images whose entrypoint needs a capability (e.g. stock nginx needs `CAP_CHOWN`) require `security.addCapabilities` / `prepareVolumeOwnership` regardless of level — unrelated to this change.

## Testing

- **Unit:** full `workflow-recipes` suite green — **2355 passed, 29 skipped, 0 failures**. Updated the one test that asserted the old default.
- **Live before/after** on a minikube cluster with the rebuilt WRC, two identical busybox recipes:
  - no `isolationLevel` → applied `standard` securityContext → pod `CreateContainerConfigError` ("container has runAsNonRoot and image will run as root") → recipe `degraded`.
  - `isolationLevel: minimal` → applied `minimal` securityContext → pod **Running** → recipe **active**.
  - Confirmed no collateral damage: the pre-existing recipe on that cluster already declares `minimal` and stayed `active`.

### DIFF
--- a/workflow-recipes/src/reconciler/policyEnforcer.test.ts
+++ b/workflow-recipes/src/reconciler/policyEnforcer.test.ts
@@ -130,11 +130,20 @@ describe('policyEnforcer.enforcePolicy', () => {
     expect(result[0].message).toContain('strict')
   })
 
-  it('defaults to minimal when recipe has no security config', () => {
-    const policy = makePolicy({ governance: { requiredSecurityLevel: 'standard' } })
-    const result = enforcePolicy(makeRecipe(), [policy])
-    expect(result).toHaveLength(1)
-    expect(result[0].rule).toBe('requiredSecurityLevel')
+  it('defaults to standard when recipe has no security config', () => {
+    // A recipe that omits spec.security now defaults to 'standard' (non-root +
+    // read-only rootfs), so it satisfies a 'standard' floor but not 'strict'.
+    const meetsStandard = enforcePolicy(makeRecipe(), [
+      makePolicy({ governance: { requiredSecurityLevel: 'standard' } }),
+    ])
+    expect(meetsStandard).toEqual([])
+
+    const belowStrict = enforcePolicy(makeRecipe(), [
+      makePolicy({ governance: { requiredSecurityLevel: 'strict' } }),
+    ])
+    expect(belowStrict).toHaveLength(1)
+    expect(belowStrict[0].rule).toBe('requiredSecurityLevel')
+    expect(belowStrict[0].message).toContain('standard')
   })
 
   // ── imageDenylist ────────────────────────────────────────────
--- a/workflow-recipes/src/reconciler/policyEnforcer.ts
+++ b/workflow-recipes/src/reconciler/policyEnforcer.ts
@@ -138,7 +138,7 @@ export function enforcePolicy(
     }
 
     if (gov?.requiredSecurityLevel) {
-      const recipeLevel = recipe.spec.security?.isolationLevel ?? 'minimal'
+      const recipeLevel = recipe.spec.security?.isolationLevel ?? 'standard'
       if (SECURITY_LEVELS[recipeLevel] < SECURITY_LEVELS[gov.requiredSecurityLevel]) {
         violations.push({
           policy: policyName,
--- a/workflow-recipes/src/reconciler/resourceBuilder.ts
+++ b/workflow-recipes/src/reconciler/resourceBuilder.ts
@@ -948,7 +948,7 @@ export interface WorkloadBuildOptions {
 export function buildDeployment(
   workload: DeploymentDef,
   recipe: WorkflowRecipeCRD,
-  isolationLevel: SecurityIsolationLevel = 'minimal',
+  isolationLevel: SecurityIsolationLevel = 'standard',
   secretKeys?: SecretKeysByName,
   buildOptions?: WorkloadBuildOptions
 ): k8s.V1Deployment {
@@ -1019,7 +1019,7 @@ export function buildDeployment(
 export function buildStatefulSet(
   workload: StatefulSetDef,
   recipe: WorkflowRecipeCRD,
-  isolationLevel: SecurityIsolationLevel = 'minimal',
+  isolationLevel: SecurityIsolationLevel = 'standard',
   secretKeys?: SecretKeysByName,
   buildOptions?: WorkloadBuildOptions
 ): { statefulSet: k8s.V1StatefulSet; headlessService: k8s.V1Service } {
@@ -1121,7 +1121,7 @@ export function buildStatefulSet(
 export function buildCronJob(
   workload: CronJobDef,
   recipe: WorkflowRecipeCRD,
-  isolationLevel: SecurityIsolationLevel = 'minimal',
+  isolationLevel: SecurityIsolationLevel = 'standard',
   secretKeys?: SecretKeysByName,
   buildOptions?: WorkloadBuildOptions
 ): k8s.V1CronJob {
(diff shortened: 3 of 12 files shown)


alfredolopez80 (dismissed):
Request changes

The security direction is correct, but the default flip needs a migration and contract update before merge.

Blocking items:

1. Existing recipes that omit `spec.security.isolationLevel` are changed on their next reconcile. Please inventory the omitted consumers, classify the deployable ones, and explicitly pin only confirmed root/writable workloads to `minimal`. Update the source of truth so those patches are not overwritten by GitOps or registry re-application.
2. The CRD description and documentation still state that omission falls back to `minimal`. Update the installed CRD prose, `docs/crds/workflowrecipe.md`, operator guides, examples, and migration notes.
3. Add direct manifest/reconciler coverage for omitted `standard` across Deployment, StatefulSet, CronJob, Job, and DaemonSet. Keep an explicit-`minimal` compatibility case and validate a real pod readiness/business signal, not only CR creation or table visibility.
4. The required `mcp-host` check is red in CI. The failure is outside this PR's diff, but it must be rerun/root-caused or separately dispositioned before merge.

The exact base/head focused workflow-recipes tests pass, and no new privilege-escalation, capability, service-account-token, or network-policy regression was found. The remaining risk is the incomplete migration and the mismatch between runtime behavior and the repository's published contract.

workflow-recipes/src/reconciler/resourceBuilder.ts
```diff
@@ -948,7 +948,7 @@ export interface WorkloadBuildOptions {
 export function buildDeployment(
   workload: DeploymentDef,
   recipe: WorkflowRecipeCRD,
-  isolationLevel: SecurityIsolationLevel = 'minimal',
+  isolationLevel: SecurityIsolationLevel = 'standard',
```
alfredolopez80, line 951: [P2] Keep one source of truth for isolation resolution

The standard default is correct, but omission semantics are duplicated across the five builders and the reconciler. A future change in one path could make policy and rendered workloads disagree. Prefer consuming one normalized isolation level, or protect this invariant with table-driven coverage across every builder.

workflow-recipes/src/reconciler/policyEnforcer.ts
```diff
@@ -138,7 +138,7 @@ export function enforcePolicy(
     }
 
     if (gov?.requiredSecurityLevel) {
-      const recipeLevel = recipe.spec.security?.isolationLevel ?? 'minimal'
+      const recipeLevel = recipe.spec.security?.isolationLevel ?? 'standard'
```
alfredolopez80, line 141: [P2] Enforce the same security contract in policy evaluation

Policy evaluation must use the same normalized level as reconciliation and workload rendering: omission means standard, while minimal is allowed only when explicitly declared. The unknown-value fail-open behavior should be tracked separately as pre-existing hardening, rather than attributed as a regression in this PR.

workflow-recipes/src/reconciler/policyEnforcer.test.ts
```diff
@@ -130,11 +130,20 @@ describe('policyEnforcer.enforcePolicy', () => {
...
 
-  it('defaults to minimal when recipe has no security config', () => {
-    const policy = makePolicy({ governance: { requiredSecurityLevel: 'standard' } })
-    const result = enforcePolicy(makeRecipe(), [policy])
-    expect(result).toHaveLength(1)
-    expect(result[0].rule).toBe('requiredSecurityLevel')
+  it('defaults to standard when recipe has no security config', () => {
+    // A recipe that omits spec.security now defaults to 'standard' (non-root +
+    // read-only rootfs), so it satisfies a 'standard' floor but not 'strict'.
+    const meetsStandard = enforcePolicy(makeRecipe(), [
```
alfredolopez80, line 136: [P2] Prove the rendered security posture

This assertion proves only the policy-floor interpretation of omission. It does not prove that the reconciler and all five builders render runAsNonRoot: true and readOnlyRootFilesystem: true. Add table-driven manifest assertions for Deployment, StatefulSet, CronJob, Job, and DaemonSet, plus an explicit-minimal regression case.

alfredolopez80 (changes requested):
## Review verdict

**CHANGES_REQUESTED — security contract and evidence gaps remain.**

This PR implements the intended security posture change:

- `standard` is the correct default isolation level.
- An omitted `spec.security.isolationLevel` resolves to `standard`.
- `minimal` remains available only as an explicit CRD exception.
- `minimal` must not become an implicit fallback or an automatic compatibility mode.

This is a security-boundary change. `standard` requires non-root execution and a read-only root filesystem, while `minimal` allows root and a writable root filesystem. Therefore, `minimal` should only be used when the workload has a documented compatibility requirement and the exception is visible in the CRD.

The runtime changes correctly update the fallback across policy evaluation, reconciliation, and workload builders while preserving explicit `minimal`.

### Remaining findings

#### P2 — CRD and documentation contradict the new security default

The runtime now treats an omitted isolation level as `standard`, but the CRD description and public documentation still describe `minimal` as the fallback.

The contract should consistently state:

```text
omitted isolationLevel → standard
explicit isolationLevel: minimal → minimal exception
```

The security-level table should also identify `standard` as the default.

#### P2 — Missing direct manifest-level coverage

The updated policy test proves that an omitted recipe satisfies a `standard` policy floor, but it does not prove that the rendered workload receives the expected security context.

Add direct coverage for omitted recipes across the workload builders and reconciliation paths, asserting `runAsNonRoot: true`, `readOnlyRootFilesystem: true`, `allowPrivilegeEscalation: false`, dropped `ALL` capabilities, and `RuntimeDefault` seccomp. The explicit `minimal` path should remain covered as an intentional exception.

#### P2/P3 — Duplicated isolation-level resolution

The default decision is repeated across policy evaluation, reconciliation, and resource builders. The current values are aligned, but future changes could make policy and rendered workloads disagree. A single normalized isolation-level result should ideally be shared by all paths, or the invariant should be protected by table-driven tests.

### Post-merge operational note

After merge, operators should review existing `WorkflowRecipe` CRDs and explicitly set `isolationLevel: minimal` only where a workload cannot meet the `standard` posture. Where possible, workloads should remain on `standard` and their images, writable mounts, or runtime configuration should be adjusted to satisfy the security requirements.

This is an operational follow-up and is not a blocker for the security-default change itself.

### Conditional metadata requirement

If `spec.security.isolationLevel: minimal` is the required exception metadata, the PR follows the intended model. If the product additionally requires a justification, owner, approval, or expiration for every `minimal` exception, that enforcement is not present in this PR and would be a separate finding.

## Final assessment

The security direction is correct:

```text
standard = baseline posture
minimal = explicit exception
```

The remaining issues concern contract consistency and direct security-context coverage. The current PR checks are green, including `workflow-recipes` and `mcp-host`.

PR #310, opened 2026-08-10
feat(gfs): add self-hosted Desktop operator parity

alfredolopez80:
## Summary

This PR adds self-hosted Desktop GFS operator parity for the first Control Admin and Desktop user created by the supported self-hosted setup flow.

Desktop operator authority exists only when an authenticated external session resolves one active, explicit `initial_setup` link to one active Control Admin. No Control Plane credential or admin password is sent to Desktop.

Self-hosted Desktop GFS operator parity architecture

### Included

- Create the exact Desktop-user ↔ Control-Admin link during initial setup, after both identities exist, in one transaction.
- Resolve GFS operator authority server-side on every request and fail closed for missing, stale, retired, malformed, or conflicting state.
- Provide Control UI revoke/reactivation without deleting either identity.
- Preserve immutable link generations, tombstones, typed retirement actors, request correlation, idempotent outcomes, and governed audit events.
- Fence stale sessions/tokens when user, admin, link, or lifecycle generations no longer match.
- After operator revoke, keep the authenticated Desktop user in ordinary `Shared with me` scope; remove operator root/ACL affordances without logging the user out.
- Keep external families behind independent source-IP, session, and operation gates. The recommended defaults are `60 operation <= 120 session < 1,200 client-IP`, but each positive-integer environment override is independent and preserved exactly; crossed scopes emit an advisory diagnostic instead of crashing startup or silently clamping an operator value.

### Explicit non-goals / separate follow-ups

- No invitation-based provisioning, email matching, admin-to-member bridge, automatic backfill, repair path, or inferred relationship.
- No general Control UI admin access from Desktop.
- IPv6 durable-key normalization and cross-layer 429 header coherence remain separate global issues ([#352](https://github.com/evenfire-ai/evenfire/issues/352), [#353](https://github.com/evenfire-ai/evenfire/issues/353)).

### Follow-up ownership

The pre-existing GFS large-upload transport follow-up is intentionally owned by PR #311 and is not part of this PR's merge scope.

PR #311 carries the related upload contract:

- 200 MiB binary file ceiling;
- 8 MiB preferred parts with a 16 MiB hard per-request limit;
- deterministic rejection of invalid upload geometry;
- preservation of `413 payload_too_large` from GFSC through Control API and External REST without collapsing it into a generic 500.

The configurable GFSC issuer remains a separate follow-up and is not claimed as fixed by PR #310 or by the current scope of PR #311.

## Authority and lifecycle contract

Operator authority requires all checks below:

1. Authenticated external session.
2. Active Desktop user with current authentication generation.
3. Active Control Admin with current session generation.
4. Current immutable link generation active and matching both exact IDs.

A bare user UUID never implies operator authority. Revoke records actor, reason, request ID, idempotency outcome, and a retained tombstone. Reactivation creates a new successor generation; revoked generations are never mutated back to active.

For Desktop error handling: bare 401/session-invalid revokes the browser authority state; resource-scoped 403 stays local; an operator-link lifecycle denial triggers one server-backed downgrade/refetch, then fails closed on a second denial.

## Rate-limit boundaries

| Boundary | Budget | Keying | Purpose |
| --- | ---: | --- | --- |
| External REST GFS token IP | 600/min | trusted source IP | Token traffic backstop |
| External REST GFS client IP | 1,200/min | attested client IP | GFS client backstop |
| External REST GFS aggregate | 1,800/min | process aggregate | Final GFS edge backstop |
| Control API GFS token mint | 10/min | Desktop user and source IP | Short-lived GFS token issuance |
| Control API GFS reads | 120/min | session, trusted IP, resolved actor | Resource/proxy/grants/shares reads |
| Control API GFS mutations | 30/min | session, trusted IP, actor/operation class | Resource, grants, and shares mutations |

GFS keeps its token-IP → client-IP → aggregate order. Non-GFS external families use independent client-IP → session gates before session authentication; authenticated operation buckets remain narrower where defined. Invitation routes expose both edge guards directly on each route so static analyzers and runtime share the same boundary.

### Rate-limit override policy

`APPROVAL_RL_EXTERNAL_PER_MIN`, `APPROVAL_RL_EXTERNAL_EDGE_PER_MIN`, and
`APPROVAL_RL_EXTERNAL_CLIENT_IP_PER_MIN` are independent scopes rather than a
single global budget. Blank/unset values use the documented defaults; any
non-empty value must be a positive safe integer. A partial override never
prevents Control API startup, and the service does not clamp the configured
value. If the resulting tuple crosses the recommended `operation <= session <
client-IP` topology, one boot advisory records the resolved values and their
`default`/`environment` sources so the operator can review redundant or
dominating buckets.

### Breaking rollout and compatibility notes

- Generation-aware external sessions use a hard cut-over. Session JWTs minted
  before this release do not contain `authGeneration` and receive `401`; users
  must sign in again once after rollout. No legacy-token grace path is enabled.
- `DELETE /api/v1/members/:userId` requires JSON `{ "reason": "<non-empty>" }`
  and an `Idempotency-Key` header. Missing values return `400`. Bundled Profile
  UI callers are updated; direct API clients must adopt this contract before
  upgrading.

## Database migrations

| Migration | Responsibility |
| --- | --- |
| `0091_gfs_desktop_operator_links` | Explicit `initial_setup` Desktop/Admin relationship. |
| `0092_gfs_audit_actor_correlation` | Desktop actor, effective admin, source, and request correlation. |
| `0093_gfs_desktop_operator_link_generations` | Immutable generations, lineage, tombstones, CAS, and restrictive parent FKs. |
| `0094_desktop_user_retirement_lifecycle` | Active/retired Desktop lifecycle, typed retirement actors, and idempotent outcomes. |
| `0095_gfs_lifecycle_authority_projection` | GFSC lifecycle authority projection, readiness, and least-privilege reads. |
| `0096_control_admin_session_version_default` | Backfill/default/NOT NULL/positive generation contract for existing Control Admin rows. |

The pre-gate applies migrations in order, reconciles runtime roles/ACLs, verifies readiness, and only then exposes consumers. Rollback disables new issuance/mutations; it does not down-migrate or delete lifecycle evidence.

## Validation

- Final synchronized commit: `3131d334e7f9d595498de69510543f1bef00e320`.
- Branch: `codex/gfs-desktop-operator-parity`, based directly on remote `origin/dev@7d56b10849458a67b15f7d9a0991fa62604913c`; remote branch and local HEAD match, with `0` behind and `64` ahead relative to `origin/dev`.
- Local full suites on the synchronized HEAD: Control API `4,175 passed / 68 skipped`, External REST `146 passed`, and GFSC `518 passed / 8 skipped`; focused Desktop GFS suites `90 passed`.
- Builds on this HEAD: Control API, External REST API, GFSC, Desktop App, and Control UI pass. `git diff --check` passes.
- Shell contracts on this HEAD: runtime access, GFS deploy, full-setup, and pre-gate state contracts pass. The broad GCP script contract still reports unrelated missing deployment artifacts outside this PR.
- CI for synchronized HEAD `3131d334` is green, including CodeQL and Real PostgreSQL 16: [workflow run](https://github.com/evenfire-ai/evenfire/actions/runs/31840830577), [CodeQL](https://github.com/evenfire-ai/evenfire/runs/94898502098).
- Browser T2: the six-journey certificate remains valid for the unchanged product/UI surface from the prior product HEAD; the synchronized T2 pre-gate is currently blocked before deployment by an `origin/dev` Dockerfile defect: `control-api/package.json` declares local `@clerum/display-field`, but `control-api/Dockerfile` does not copy `packages/display-field` before `npm ci`. The reused profile/context remains `clerum-gfs-operator-c8b14240`; no new profile or PVC reset was used.
- E2E Guardian static review found no fixed waits, route mocks, browser storage mutation, forced clicks, direct link-table writes, or skipped/only/fixme scenarios.

Formal review approval and merge remain outside this PR.

## Architecture documentation

- [Detailed architecture](docs/architecture/gfs-desktop-operator-parity.md)
- [Overview diagram](docs/architecture/diagrams/gfs-desktop-operator-parity-overview.svg)
- [Detailed diagram](docs/architecture/diagrams/gfs-desktop-operator-parity.svg)


zach88:
Took a deep pass through this one across all five services (control-api, gfs-controller, desktop-app, external-rest-api, control-ui) plus the deploy/scripts changes. The security posture is solid — authority resolution fails closed, subjects stay disjoint, setup is atomic, and no admin credential crosses to Desktop. But there's a broken test that blocks CI, and two High issues where a stated invariant leaks: one lets a retired user re-mint GFS authority, and one turns an expected permission verdict into a full session lockout. Details below, grouped by severity with ledger IDs so we can track them across rounds.

## Blockers

1. **[R1-B1] Broken assertion fails the gcp-scripts test** — `scripts/tests/test-gcp-scripts.sh:1485`. The new assertion greps for `expected.access_profile IN ('legacy_dml', 'link_lifecycle')`, but that contiguous string never appears in `deploy/scripts/run-control-api-db-migration.sh` (it uses `IN ('legacy_dml','upsert','append','link_lifecycle')` for INSERT, `IN ('legacy_dml','upsert','link_lifecycle')` for UPDATE, `IN ('legacy_dml')` for DELETE). Reproduced: isolating `assert_control_api_runtime_access_contract_is_exact()` yields `FAIL: control-api runtime access contract is incomplete or not exact` → exit 1. The sibling new test `test-control-api-runtime-access-contract.sh` omits this line and passes, which points to a leftover/typo. Fix: drop line 1485 — the INSERT/UPDATE/DELETE assertions already cover the contract.

## High

2. **[R1-H1] Retired-user guard is bypassed on `POST /external/gfs/token`** — `control-api/src/routes/external/gfs.ts:358-382`. The retirement denial (`isDesktopUserActive` → 403 `desktop_user_retired`) lives inside `resolveExternalGfsAuthority`, which only runs via `attachExternalGfsAuthority`, mounted at line 386 — *after* the token route is registered at 358. `requireValidExternalSessionToken` is a pure JWT check with no lifecycle lookup, and the soft-retire path (`src/services/directory/users.ts:563-589`) sets `lifecycle_state='retired'` and revokes the operator link but does not remove the user's `gfs_grants`/`gfs_shares` or bump a session version. So a soft-retired desktop user (the operator-link-history branch, the only one that soft-retires rather than hard-deletes) whose session JWT is still within TTL can call `POST /external/gfs/token`, get a `sub=users.id` token, and use it against gfsc, which authorizes against the still-present grants — exactly the "retired user falls back to ordinary GFS authority" the PR promises to prevent. Reproduced the route ordering and the missing lifecycle check via `test/routes.externalGfs.test.ts:276-292` (mint succeeds with only a session token). Window is bounded by session TTL, hence High rather than Blocker. Fix: hoist a lifecycle guard (403 `desktop_user_retired`) ahead of the `/token` handler, independent of `attachExternalGfsAuthority`.

3. **[R1-H2] An expected 403 verdict nukes the whole file browser session** — `desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts:115-130` + `327-338`. `describeGfsBrowserFailure` classifies *any* message containing `403`/`forbidden` as `unauthorized`; `queryAuthorizationError` aggregates `grantsQuery`/`sharesQuery` errors and calls `revokeAccess()`, which sets `accessState='revoked'`, clears crumbs, and wipes the GFS cache — then `FilesPage.tsx:238-248` closes the dialog and drops to the "not authorized" empty state. But `manage_acl_required` is an *expected* 403 (a user who can view a read-only resource but not manage its ACL), and `gfsGrantErrors.ts` maps it as `severity:'quiet'` for an informational banner. Reproduced by flow tracing: an ordinary user in "Shared with me" opens "Manage" on a read-only resource → `listGrants` returns `403 manage_acl_required` → `revokeAccess()` → the whole Files page locks, and the quiet banner never shows. "Retry file access" re-runs discovery but reopening Manage locks again (persistent trap). Same pattern lives at ~7 sites — `useGfsBrowserController.ts:335,487` and `FilesPage.tsx:305-309` (`failClosedOnAuthorizationError` for `escalation_rejected`/`foreign_agent_forbidden`), `:324-333`, `:348-357`. Fix: distinguish the session/operator-authority 403 (`operator_link_inactive`/`operator_link_invalid`, 401/403 on `listAccessible`/`listChildren`) from the per-operation verdict 403 (`manage_acl_required`, `escalation_rejected`, `foreign_agent_forbidden`, …) and only let the former reach `revokeAccess`. Worth a small design note before this grows into a code table rather than expanding the set by trial and error.

## Medium

4. **[R1-M1] `desktop_user_retired` isn't mapped → 500 instead of 409/404** — `control-api/src/routes/admin/controlAdmins.ts:495-520`. `reactivate()` throws `desktop_user_retired` (`src/services/gfsDesktopOperatorLinkService.ts:715-719`) when someone reactivates a link to a retired desktop user, but the handler maps `invalid_input`/`link_conflict`/`control_admin_inactive`/`malformed_link`/`desktop_user_not_found` and lets this one fall through to a 500. It still fails closed, but returns a server-error status (and alerting noise) for a deliberate business refusal. No route test covers it. Fix: map to 409/404 alongside the others.

5. **[R1-M2] `413 payload_too_large` is dropped at the edge → Desktop sees an opaque 500** — `external-rest-api/src/routes/gfs.ts:27`. gfsc emits 413 (`gfs-controller/src/api/serve.ts:735`, `readJsonBody`), control-api forwards it verbatim, but the edge `PROPAGATED` set doesn't include 413, so `forwardControlApiError` collapses it to 500. Uploading a body over `GFS_MAX_WRITE_BODY_BYTES` gives the user a generic error instead of "file too large". It's the only status gfsc can emit that's missing from the allowlist, and the file's own header comment states the intent to avoid exactly this. Fix: add 413 to `PROPAGATED`.

6. **[R1-M3] The `aggregate-ip` edge bucket is dead under every recommended config** — `external-rest-api/src/routes/gfs.ts:156-157`. Both limiters (`aggregate-ip` at 1800, `authenticated-ip` at 1200) register on the same path with the default per-IP key generator, and since `aggregate = 2×authenticated` always, `authenticated-ip` trips first and the aggregate bucket is never reached. So `EXTERNAL_REST_API_GFS_EDGE_AGGREGATE_RL_PER_MIN` is a no-op, and the "coarse per-process" backstop described in the scope is actually per-IP (a distributed flood from many IPs has no edge cap). The test "enforces the aggregate edge backstop before auth" passes with an inverted config (`aggregate:2 < authenticated:10`) that never happens in prod — a green that hides the dead bucket. Fix: give the aggregate bucket a constant key generator for a real per-process cap, or drop the redundant bucket and adjust the test/env doc.

7. **[R1-M4] Breaking contract change on two DELETE endpoints — confirm consumers are updated** — `control-api/src/routes/admin/users.ts:249-276` and `src/routes/external/members.ts:187-215` now require a `reason` body field and an `Idempotency-Key` header (400 `reason_required` / `idempotency_key_required` if absent); previously both deleted with no body/header. This is the governed-retirement design, but the callers live outside control-api (control-ui / profile-ui / desktop). Any caller not updated hits a hard 400 on user/member deletion. Please confirm all consumers now send both.

8. **[R1-M5] `retireDesktopUser` has no real-Postgres integration coverage (T1)** — `control-api/src/services/directory/users.ts:344-632`. The multi-statement retirement SQL (claim-INSERT / `ON CONFLICT DO NOTHING` / fingerprint-replay / optimistic version — the part most likely to hide a column-name or WHERE bug) is only exercised against hand-written `txQuery` mocks. The real-Postgres test (`test/db.realPostgresMigration.integration.test.ts:418+`) runs manual INSERT/UPDATE against the schema, not `retireDesktopUser` itself. Suggest an integration test that drives `retireDesktopUser` end-to-end (both `deleted` and `retired` outcomes plus a replay) against the temp Postgres the harness already stands up.

9. **[R1-M6] The "expected 403 must not revoke" invariant has no test (T5)** — `desktop-app/ui/src/hooks/domain/__tests__/useGfsBrowserController.test.tsx:428`. The only revocation test uses `403 operator_link_inactive` on `listAccessible`; nothing verifies that `manage_acl_required` (from grants/shares) or `escalation_rejected` (from a mutation) does *not* trigger `revokeAccess`. That's why R1-H2 slipped through green. Closes with R1-H2's regression test, which should fail against the prior head.

## Smaller stuff

10. **[R1-L1] Case-sensitivity mismatch on `controlAdminId`** — `gfs-controller/src/auth/verify.ts:63-70` vs `src/authz/subjectResolver.ts:65-70`. `parseBrokeredAuthority` lowercases `controlAdminId` but returns `sub` verbatim; `subjectResolver` compares them strictly. A linked-admin token with an uppercase-hex UUID (accepted by `UUID_RE`'s `/i` flag) passes verify but gets rejected at resolution as "linked admin is not active" even with an active admin. Reproduced end-to-end. Fails closed and masked today (Postgres `uuid` renders lowercase), but worth normalizing both sides.

11. **[R1-L2] New hard coupling on `desktop_user_id $13::uuid`** — `gfs-controller/src/authz/audit.ts:103` + `src/authz/subjectResolver.ts:122`. The user-session path sets `desktopUserId: sub` raw; any user-path principal whose `sub` isn't a syntactic UUID would flip a previously-served request into a fail-closed 503 at audit-insert time. Safe today (user-path subs are always UUIDs), but the invariant is new and undocumented — consider validating/normalizing on the user path or documenting it next to the cast.

12. **[R1-L3] `isUninitializedDriveError` matches on overly broad substrings** — `desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts:93-104`. Free-substring matches on `'not initialized'`/`'unseeded'`, evaluated before the unauthorized check, so a `403 ... drive not initialized` message would classify as `uninitialized` and mask the 403. Low likelihood; same brittle-substring class as R1-H2.

13. **[R1-L4] Deploy-order constraint: control-api before gfsc** — `gfs-controller/src/authz/subjectResolver.ts:104-108`. A UUID `sub` without a `principalType` marker now resolves as user (deliberate fail-closed). control-api now stamps `principalType:'control-admin'` on the human/proxy mint, so if gfsc ships before control-api starts stamping, in-flight operator tokens resolve as user and operators lose authority during the window. Not a bug — just worth calling out in the release notes.

14. **[R1-L5] `gfsOperatorLinkStatus:'revoked'` for admins that were never linked** — `control-api/src/services/adminAuthService.ts:731-770`. The `gfs_operator_user_id ? … : 'revoked'` fallback reports "revoked" for never-linked admins (`gfsOperatorLink: null`), and Control UI renders "Revoked" (`control-ui/components/ControlAdminsPanel/index.tsx:426`). "Never linked" and "revoked" are semantically different — a `'none'` sentinel / "—" display reads better.

A few nits not worth their own numbers: `resources.ts:373-381` open-codes audit fields instead of reusing `callerAuditFields(caller)`; the `unlink`/`reactivate` handlers in `controlAdmins.ts` don't forward `requestId`/`correlationId` even though the DELETE-user paths do; and the four migration unit tests assert `sql.toContain('<literal>')` against a mocked query, which is near-tautological (real coverage rests on the Postgres integration suite).

## What checked out clean

- **Security**: no credential/password crosses to Desktop; authority resolution is server-side and fails closed on inactive/malformed/conflicting links; no privilege escalation via subject collision (this PR actually removes the old ambiguity of probing any bare UUID against `control_admin_users`); one-shot setup is atomic under a row lock (409 on retry); rate-limit env vars are clamped (`boundedPositiveIntegerFromEnv` rejects <1 and >1e6; control-api budgets are constants); queries are parameterized; no NetworkPolicy/K8s regression; the `link_lifecycle` runtime profile grants SELECT/INSERT/UPDATE and denies DELETE.
- **Contracts**: token aud/kid/issuer-default, the user-vs-control-admin sub scheme, the `{ok,data}` envelope, the `gfs:*` IPC channels (preload↔ipc↔renderer.d.ts), the 4-way revoke/reactivate shape (control-ui↔controlAdmins↔adminAuthService), and 429/`Retry-After`/`X-RateLimit-*` header propagation all line up.
- **Migrations 0091-0094**: additive/idempotent, generation lineage + lifecycle constraints consistent, FK→RESTRICT matched by the two guarded deletes, backfill to `lifecycle_state='active'` with no lockout.
- **Desktop e2e**: no fixed waits, route mocks, storage mutation, Playwright force, or direct link-table DML; the operator link is established through the real setup path.

Suites at head `f107093`: control-api 185 (+108 DB/directory) green, `tsc` clean; gfs-controller 121 green, `tsc` clean; external-rest-api 124 green — but `test-gcp-scripts.sh` fails on item 1. The real-Postgres integration suites were skipped in the review sandbox (no docker), so the SQL there was read, not executed.

One process note: this is 118 files across 5 services (+12637/−522), well over a reviewable budget, and the body doesn't say why it can't be split. Every round re-reviews the whole thing — worth carving into per-service PRs if feasible.

zach88:
Did a full round-2 pass over the rebased head (`8eb7088b`). Good news first: all 14 round-1 findings are resolved. 13 are closed (most with a regression test that I confirmed goes red against the pre-fix head), and one (the 413 propagation) turned out to be pre-existing in the base rather than something this PR introduced, so it's out of scope here. Security came back clean: the retirement bypass is properly closed and fail-closed, subject canonicalization doesn't open any user/admin confusion, and the distributed rate limits are intact.

No blockers or highs this round. What's left is two Medium items, both in the desktop error-classification logic that the R1-H2 fix introduced, plus a handful of low/test-quality nits. Heads up on process: `control-api/src/routes/external/gfs.ts` is now on its 4th `fix:` commit in this PR, and `externalGfsRateLimit.ts` / `external-rest-api/src/routes/members.ts` are on their 3rd, so the rate-limit plane is getting patched a lot.

## Medium

1. **[R2-M1] A bare `401` during a mutation no longer revokes the session (regression).** `desktop-app/ui/src/pages/FilesPage.tsx:254,307,328,350`. The R1-H2 fix moved the mutation handlers from `describeGfsBrowserFailure` (revoked on any `401/403/unauthorized/forbidden`) to `isGfsSessionAuthorityFailure(raw, 'operation')`, which on the `operation` surface only returns true for the four session-authority codes (`desktop_user_retired`, `operator_link_inactive/invalid`, `gfs_operator_link_invalid`). A plain `401` (expired/revoked gfs session) no longer matches. Repro: evaluating the predicate on `'401 not authenticated'` on the `operation` surface returns false, so with an expired session a Delete/rename now just toasts the raw message and leaves `accessState='active'` with stale content and retryable-but-failing mutations, where it used to send the user to re-auth. Splitting `403 manage_acl_required` out was the right call, but a `401` is always a session failure, never a per-resource verdict, so it shouldn't have gone in the same bucket. Suggest treating `401`/`not authenticated` as session-authority on the `operation` surface too, keeping the 403 verdict-vs-code split intact. The new test only covers 403-with-code, so it doesn't catch this.

2. **[R2-M2] A per-resource read `403` on `childrenQuery` revokes the whole session.** `desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts:369-374` and `:135-140`. `childrenQuery.error` is classified on the `discovery` surface, where any `403/forbidden` revokes. That's correct for root discovery (`accessibleQuery`), but subfolder navigation (`listChildren`) is bundled under the same surface. Scenario: in "Shared with me" with 3 shared folders, an operator revokes the share on folder A from another session; the user opens A, `listChildren(A)` returns a per-resource `403`, and the entire session gets revoked, dropping all three folders with a "sign in again". Same class as R1-H2 ("this subfolder failed" is a per-resource verdict, not "root failed" = session dead), and the whole revoke mechanism is PR code, so it's the sibling instance the R1-H2 generalization should have covered. Suggest classifying `childrenQuery` folder-read 403s on the `operation` surface (only the session codes revoke) while `accessibleQuery` stays on `discovery`.

## Low / grab-bag

3. **[R2-L1] The aggregate edge bucket went from per-IP to a single global key.** `external-rest-api/src/routes/gfs.ts:158-167`. The R1-M3 fix ("per-process backstop") was applied literally: `keyGenerator: () => 'gfs-edge-aggregate'` is now one global bucket of `aggregatePerMin` (default `expectedUsers*120`) shared across all IPs. It does correctly cap a distributed flood (the bucket was dead before), but during a legitimate concurrent spike, users that collectively exceed the aggregate all get 429 at once, and the draft-7 `RateLimit-*` headers now decrement from other IPs' traffic. The per-IP `authenticated-ip` bucket below preserves fairness and exhausting the global pool needs 2+ IPs, so it's low, but the sizing is worth a sanity check (see the question at the end).

4. **[R2-L2] The edge 429 body omits `retryAfterSeconds` that the desktop client parses.** `external-rest-api/src/routes/gfs.ts:41-47`. The new `edgeLimiter` responds `{error, rateLimitLayer, rateLimitBucket}` with draft-7 headers only, but `desktop-app/src/gfs/uriHandler.ts:649-668` reads the retry hint from the JSON `retryAfterSeconds` field, not the header. The control-api-originated 429 does carry it, so a bulk request that trips the edge bucket loses the hint the UI shows for the control-api bucket. Cosmetic (it fails loud with the original error), but easy parity fix: add `retryAfterSeconds` to the edge body, or fall back to the `Retry-After` header on the client.

5. **[R2-L3] Unreachable fixture in the header-allowlist test.** `external-rest-api/test/routes.gfs.test.ts:444`. The test hand-builds a `ControlApiError` carrying `set-cookie`/`server`/`x-internal-debug`, but the real producer (`external-rest-api/src/controlApiClient.ts:31-38` `forwardableResponseHeaders`) never emits those. The double filter is legit defense-in-depth, but the test isn't exercising a reachable path, so it only documents intent. Derive it from the real producer or drop it to a comment.

6. **[R2-L4] The `row_hash` tamper-evidence invariant has no test.** `gfs-controller/src/authz/audit.ts:93-94`. `dbAuditSink.test.ts` asserts the INSERT value array, not that the hash changes when `desktopUserId`/`authoritySource` change, so a refactor that dropped them from the hash would stay green. Suggest a case that writes two events differing only in `desktopUserId` (or `authoritySource`) and asserts `row_hash` differs.

7. **[R2-L5] The default delete path isn't covered client-side.** `profile-ui/test/memberApi.test.ts`, `control-ui/lib/__tests__/adminUsersApi.test.ts`. Both tests pass explicit `reason`/`idempotencyKey` (they do satisfy T3), but the path the profile-ui components actually use is the defaults one (`deleteManagedUser(id)` with no request). If someone regressed the default generation, the server would 400 every profile-ui delete and no test would go red. Suggest a case that calls delete with no request and asserts a non-empty `Idempotency-Key` + `reason` still go out, plus uniqueness across two calls.

Minor nits, no need to gate on these: `adminDeleteUser` in `control-api/src/services/directory/users.ts:631` is now dead (both callers moved to `retireDesktopUser`, so the `gfs_operator_link_history_retained` guard is unreachable, drop or `@deprecated`); `externalGfsRateLimit.ts:249` emits `X-RateLimit-*` from the last allowed bucket instead of the most restrictive; the "sub recorded verbatim" comment in `gfs-controller/src/authz/permissionClient.ts:34` is stale now that UUID subs are lowercased; and the UUID regex is copied in three files (`verify.ts:33`, `subjectResolver.ts:41`, `audit.ts:71`).

## Two decisions for you

- **Aggregate edge sizing (R2-L1):** now that the aggregate bucket is a single global pool, do you want to re-check `aggregatePerMin` against realistic concurrent-user peaks so legit spikes don't all 429 together, or are you happy taking it as a deliberate per-process backstop tradeoff and just documenting it?
- **Two pre-existing Mediums (not introduced by this PR, non-blocking):** while tracing the seams, two base-level issues showed up that reproduce without this diff. (a) `413 payload_too_large` isn't in the `PROPAGATED` set in `external-rest-api/src/routes/gfs.ts:27`, so gfsc's real 413 on oversized writes collapses to a generic 500 at the edge (one-line add to the set). (b) `gfs-controller/src/index.ts:158` calls `verifyGfsToken` without passing `issuer`, so it falls back to a fixed `"control-api"` while control-api signs with the configurable `CONTROL_API_RPC_JWT_ISSUER`; a non-default issuer deployment would reject every gfs token. Neither is caused here. Do you want these split out as their own follow-up items, or left as-is for now?

## What checked out clean

- Security: retirement bypass closed and fail-closed (`attachExternalGfsUserLifecycle` hoisted before the mint), subject canonicalization with no user/admin confusion or bare-UUID elevation, distributed limits intact (10/min mint, 30/min mutation; only reads widen to 120/min), member boundary is post-auth with no leak to unauthenticated callers, `reason`/`idempotencyKey`/`correlationId` bounded and injection-safe, no secret crosses to Desktop.
- Contract seams all consistent end-to-end: DELETE `reason`+`Idempotency-Key` (all four UI callers updated, `Idempotency-Key` is per-op `randomUUID`, not constant), gfs-token `principalType`/brokered/case-insensitive `controlAdminId` across sign/verify/resolve, `gfsOperatorLinkStatus` union with `none` on both sides (and fixed at the source, not just the label), and the `gfs:*` IPC channels.
- The R1-H2/M6 fix generalized across every operation handler, and its regression test is genuinely red against the pre-fix head. The rate-bucket read/mutation split is applied consistently across all three planes via a single helper, with the mutation bucket identity preserved (no distributed-counter reset).

Test status from the touched files: control-api 147, external-rest-api 34, gfs-controller 49, desktop-app 28 + 40, control-ui 19, profile-ui 68, all green. The real-Postgres retirement integration test is env-gated and was reviewed by reading.

One structural note worth flagging: this is 137 files across 5 services, which makes every round re-review the whole thing. If there's a natural seam to split it, it'd cut the review cost going forward.

zach88:
Took another deep pass through this one (round 3). Security looks solid: the generation fencing fails closed and I couldn't find a bypass. But this round adds 1 blocker and 2 highs, and all three of them come from the fixes themselves rather than fresh design. Two pieces keep generating the findings round after round: the desktop error->revoke classifier and the external rate-limit plane. I'd genuinely suggest pausing the patching on those two and writing a short design note for each before another fix cycle, because we're on the third round of the same classes reappearing.

## Blockers

1. **[R3-B1] Invited control admins get locked out of all GFS authority.** `control_admin_users.session_version` is `INTEGER NOT NULL DEFAULT 0` (`control-api/src/db.ts:212,1089`). The 0095 migration only normalizes existing rows (`db.ts:2936`), it never changes the column default, and `completeControlAdminInvitation` inserts without setting it (`control-api/src/services/adminAuthService.ts:944`), so every admin invited through the new ControlAdminsPanel lands on `session_version = 0`. Nothing bumps it on login (`registerAdminSuccessfulLogin` leaves it), only a password change does. The new generation guard then rejects `< 1`: `control-api/src/routes/gfs/token.ts:51-56` and `routes/gfs/proxy.ts:38-42` return 401, and the brokered operator path denies too (`gfs-controller/src/auth/verify.ts` throws on non-positive authGeneration, `subjectResolver.ts` `positiveGeneration(0)` -> null). Repro (by reading the guards + the default/insert): deploy, invite a new control admin, they accept + log in + open GFS, they get 401 and can't act as a GFS operator. The bootstrap first admin escapes only because its row is seeded before the 0095 normalize. This is the sibling of the `users` change you did correctly (`users.lifecycle_version DEFAULT 1 NOT NULL`); `control_admin_users` was left asymmetric. Fix: set the column `DEFAULT 1` and/or stamp `session_version = 1` on admin creation.

## High

1. **[R3-H1] `openUri` revokes the whole session on a per-resource 403.** In `desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts:537` the `openUri` catch runs `if (isGfsSessionAuthorityFailure(message, 'discovery')) revokeAccess()`, and the `discovery` surface treats any `401/403/forbidden` as a session-authority failure. Opening a `gfs://` link to a resource the user just lacks read on (the main use of the Open-link dialog for ordinary shared users) now clears all crumbs and shows "File access is not authorized". Repro (throwaway test): `openUri` returning `Error('403 Forbidden')` -> `accessState = revoked`, `current = none`. At base this path did `setOpenError(...)` and left the session intact. This is the same class as R2-M2 that got fixed for `childrenQuery`; `openUri` should use the `'operation'` surface the same way, so only lifecycle codes revoke.

2. **[R3-H2] New CodeQL edge rate limiters key by `req.ip`, which collapses behind the proxy.** The limiters added to `control-api/src/routes/external/{directory,teams,members,oauthGrants,sharedFilesystems,users,invitations,decisions}.ts` use the default key generator. With `app.set('trust proxy', 1)` and the external-rest-api -> control-api chain, `req.ip` resolves to the shared infra-proxy address, so all external clients fall into one 60/min pre-auth bucket per route family. For `directory/teams/members/oauthGrants/sharedFilesystems` there's no per-user inner limiter at all, so those routes went from unlimited to a shared 60/min, and one busy Desktop client 429s everyone else before auth. This is inconsistent with the GFS sibling you fixed in this same PR (`external/gfs.ts:95-96,264-269` keys by validated XFF with a wide 1800/min backstop). Fix: key these by a validated client IP the same way, and make the pre-auth backstop wider than the per-user quota (or lean on the existing per-user `rateLimitMiddleware`).

## Medium

1. **[R3-M1] PR body rate-limit claim is stale.** The body says "30/min for session/IP/actor operations", but the code splits reads from mutations this round: reads are 120/min and mutations 30/min (`control-api/src/config.ts`, `middleware/externalGfsRateLimit.ts:160-161,288-306,336`); the global IP backstop is 1200/min and token-IP 600/min. Token mint = 10/min is correct. Worth syncing the body since it's part of what gets reviewed.

2. **[R2-M1] A bare 401 during a mutation still doesn't revoke (still open since round 2).** `FilesPage.tsx:254` + `useGfsBrowserController.ts:129`: a plain `401` (expired/revoked session token, inherently session-wide) doesn't match `isGfsSessionAuthorityFailure(raw,'operation')` (only the 4 lifecycle codes do), so the user gets the raw error as a toast and the operator write controls stay live over a dead session. Repro (throwaway test): `createFolder` throwing `Error('401 Unauthorized')` -> `revokeAccess` called 0x, no friendly unauthorized banner, create-folder still present. The existing test masks it because it throws `` `${status}: gfs_operator_link_invalid` ``, passing via the embedded lifecycle code rather than a bare 401/403. This has been open two rounds now, so it should either go up in severity or get closed with a written call. Cleanest path: fold it into the same decision-table fix as R3-H1 (treat `401`/`not authenticated` as session-wide in the `operation` surface, keep `403`/`forbidden` local).

## Regressions / collateral from this round's fixes

All three serious findings above trace back to fix commits rather than the original design, which is the actionable signal here:
- **R3-B1**: the generation guard (`82e46b0b`, `2652dfd2`) meeting the pre-existing `session_version DEFAULT 0`, and the `users`-vs-`control_admin_users` asymmetry.
- **R3-H2**: the CodeQL hardening pass (`9174afa8`, `f5fa81b8`).
- **R3-H1**: the error-classifier fix that closed R2-M2 for `childrenQuery` but didn't generalize to `openUri`.

## Smaller stuff

- **[R3-L1]** `AuthClaims.authGeneration` is typed optional (`control-api/src/profileTypes.ts`) but required at runtime (`externalSessionAuthToken.ts:16-23` throws on sign, returns null on verify). Fail-closed, but the type doesn't express the invariant.
- **[R3-L2]** The admin generation fence is untested. `gfs-controller/src/authz/subjectResolver.test.ts:44-49` hard-codes `session_version: 1` for every admin, so neither the brokered nor direct-control-admin `generationMatches(admin.session_version, claims.authGeneration)` has a mismatch case. If that check regressed, the suite stays green, on the very thing this round is about. Add a `session_version=2` / `authGeneration=1` case (and mirror it in the real-postgres test).
- **[R3-L3]** `control-api/test/middleware.externalSessionAuth.lifecycle.test.ts` hand-builds a fixture with `lifecycle_version` as a number, but the real producer (`users.lifecycle_version` is int8, node-pg returns a string) emits a string. No live bug (`Number()` coerces), but the fixture doesn't represent the real layer.
- **[R3-L4]** `control-api/src/routes/external/invitations.ts:44` comment says "without double-counting", but the edge limiter does stack on top of the existing per-route `rateLimitMiddleware`. Only hardens, but the comment is misleading.
- **[R3-L5]** `control-api/src/middleware/externalSessionAuth.ts` now makes every `/external/*` request depend on a `pool` query (`isCurrentExternalSession`); if the DB is down it throws -> 500 where a valid token used to pass without touching the DB. Correct fail-closed, just a new failure mode worth knowing.
- **[R2-L2]** (still open) the edge 429 body omits `retryAfterSeconds` that the desktop parses (`external-rest-api/src/routes/gfs.ts:41`, `control-api/.../external/gfs.ts:264`).
- **[R2-L3]** (still open) `external-rest-api/test/routes.gfs.test.ts:444` uses an unreachable hand-built `ControlApiError` fixture (T1).
- **[R2-L4]** (still open) the `row_hash` tamper-evidence invariant has no test that asserts the hash changes when `desktopUserId`/`authoritySource` vary (`gfs-controller/src/authz/audit.ts:93`).
- **[R2-L5]** (still open) the default delete route (no explicit request) is uncovered in `control-ui/lib/__tests__/adminUsersApi.test.ts` and `profile-ui/test/memberApi.test.ts`.

## What checked out clean

- Security is SAFE. Generation fencing fails closed on all three branches (user/control-admin/brokered): `positiveGeneration` rejects 0/negatives/decimals/leading-zero/whitespace/non-safe-integer, `generationMatches` needs both sides non-null and strict `===`, so a stale-lower generation is denied and default-0/rollover can't bypass. The retirement bypass class (R1-H1) stays closed (`assertDesktopUserActive` runs before the flag and the link read). No new IDOR/enumeration, no secret leakage (header allowlist in `controlApiClient.ts`), brokered tokens keep `sub === controlAdminId` with a strict key allow-list. No K8s posture change.
- The bigint-vs-number trap is handled: `users.lifecycle_version` (int8 -> string), `control_admin_users.session_version` (int4 -> number) and the link generation are all normalized before comparison on both sign and edit paths. Covered by `subjectResolver.test.ts:192`.
- **R2-M2 is fixed**: per-resource subfolder 403 on `childrenQuery` no longer collapses the session (`useGfsBrowserController.ts:369-385` passes `'operation'`).
- **R2-L1 is addressed**: the edge aggregate backstop is now configurable and right-sized via `EXTERNAL_REST_API_GFS_EDGE_*` (1800/1200/600), wired through config -> middleware -> deploy. The fairness tradeoff is now a tunable, documented decision (though the same class reappears untuned on the non-GFS routes, see R3-H2).
- Collateral trace of the round is clean: every changed signature has its callers updated in the same range (`requireExternalWorkflowCaller` sync->async with `await` at all call sites, `verifyExternalSessionToken`->null handled as 401 everywhere, invitation/login acceptors returning `{error:'user_retired'}` handled gracefully).

One rollout note that's by-design but worth flagging in the deploy: the fencing invalidates every in-flight external session token (they lack `authGeneration`), so all `/external/*` return 401 until the TTL window clears and users re-login, and control-api needs to deploy together-with or ahead-of gfs-controller.

Suites/typecheck at this head: control-api `tsc` clean + 4056 unit passed (realPostgres skipped, no live DB), gfs-controller 138 passed, desktop-app 69 passed, control-ui 19, profile-ui memberApi 1. The two desktop findings were reproduced with throwaway tests and reverted.

zach88:
Took another deep pass through this one for round 4. Good news first: the stuff that was blocking merge is actually resolved this time. The invited-admin `session_version=0` blocker (R3-B1), the `openUri` session-revoke bug (R3-H1), and the shared per-IP edge bucket (R3-H2) all check out as closed and verified. This round found no new blockers and no new highs, which is the first time the serious count came down instead of up.

What's left is a small set of concrete fixes plus a couple of policy calls. Two of the new findings are regressions introduced by fix commits in this same round, so those are the most actionable.

## Regressions (introduced by a fix commit this round)

1. **[R4-M2] Per-client edge fairness fix dropped the per-IP bound** (`control-api/src/middleware/externalClientIdentity.ts:22-27`, from `0aff79b5`). The new key is `ip:session:hash(token)`, but the session token is attacker-controlled *before* JWT verification, so rotating `x-user-session-token` mints a fresh bucket every request and the 120/min pre-auth cap on the non-GFS families (`directory`, `teams`, `members`, `users`, `oauthGrants`, `sharedFilesystems`, `invitations`, `decisions`, `notifications`, `workflow-approval-mediums`) never trips. Reproduced. Two things soften it: it's not a cross-tenant poison (you can't forge someone else's token), and external-rest-api's client-IP bucket (1200/min) still caps absolute volume at the edge, so this is a defense-in-depth weakening rather than an unbounded hole. It's also a consistency gap: the GFS surface does this right with two separate buckets (per-IP *and* per-session) in `externalGfsRateLimit.ts:301-307`. Suggest mirroring that here (one limiter keyed by attested IP, one by IP+session) instead of the single combined key.

2. **[R4-L1] A bare 401 on one folder's children now revokes the whole session** (`desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts:394`, from `da22e463`). The R2-M1 fix flipped `isGfsSessionAuthorityFailure` so a generic 401 on the `operation` surface revokes. That's correct for `openUri` (the intended fix) and for the FilesPage mutation callsites, but `childrenQuery` at :394 inherits the same helper and it contradicts the contract documented right above it at :378-382 ("a generic 401/403 there can mean one shared folder was revoked while the rest of the session remains valid. Only explicit lifecycle codes may revoke from that surface"). Reproduced at unit level: `isGfsSessionAuthorityFailure('401 Unauthorized', 'operation')` is now `true` where it was `false`; the 403 path is unchanged. Low because no known producer emits a bare per-resource 401 on childrenQuery, but there's no test pinning it either, so the divergence slipped through silently. Suggest giving childrenQuery its own classification, or excluding bare 401 from revoke on that callsite.

## Medium

3. **[R4-M1] `deleteControlAdmin` lost delete idempotency** (`control-api/src/services/adminAuthService.ts:551-560`). Base did `DELETE ... RETURNING`, so a second delete of the same admin returned rowCount 0 and a 404. The diff switched to `UPDATE ... SET status='disabled', session_version = session_version + 1 WHERE id=$1` with no status guard. So DELETE the same admin twice: the second call still matches (rowCount 1), returns `200 {deleted:true}` again, bumps `session_version` a second time, and writes a duplicate `control_admin_deletion_audit` row plus event for an already-retired admin. That's a broken idempotency contract (200 instead of 404), unbounded generation inflation, and audit pollution. Suggest `AND status <> 'disabled'` in the WHERE and returning `not_found` on rowCount 0.

4. **[R2-L5] Profile-ui default delete path still has no test, now on its third round** (`profile-ui/test/memberApi.test.ts` + callers `profile-ui/app/members/page.tsx:76`, `[userId]/page.tsx:134`). Both real callers invoke `deleteManagedUser(member.id)` with no request, so they take the defaults branch (`reason='profile_ui_user_retirement'`, generated Idempotency-Key). The only test exercises the explicit-args branch that no caller uses. Since the endpoint now returns 400 without `reason`/`Idempotency-Key`, a break in the default reason or key generation ships green and fails in prod. This has been open since round 2, so per our own aging rule it either gets the test (my vote) or a written close, not a fourth round floating.

5. **[R4-M4] Every in-flight Desktop session hard-invalidates on deploy** (`control-api/src/utils/auth/externalSessionAuthToken.ts:38-45`, `middleware/externalSessionAuth.ts:16-36`). Tokens minted before this deploy carry no `authGeneration`, so on rollout every authenticated Desktop user gets 401 until they re-login, with no grace window. It's an intended consequence of binding sessions to a generation epoch, but it isn't called out anywhere. Worth confirming the forced re-login is acceptable and noting it in the body/runbook (or staging it).

## Smaller stuff

6. **[R4-L2] Readiness probe doesn't gate on migration 0096** (`gfs-controller/src/authz/storeProbe.ts:121-125`). `AUTHORITY_PROJECTION_SQL` checks the `session_version` column exists (0095) but not that its DEFAULT is 1 / the backfill ran (0096). If 0095 applies but 0096 lags, the probe goes green while every legacy admin at `session_version=0` gets denied GFS, which is exactly the R3-B1 symptom the probe exists to make loud. Note the asymmetry: `users.lifecycle_version` is checked `attnotnull`, `session_version` is checked for neither. Suggest a `pg_attrdef`/`atthasdef` check keyed to "0096 not applied".

7. **[R4-L3] HEAD stream abandons the upstream body without cancel** (`external-rest-api/src/routes/contextSharedFilesystems.ts:113`). The HEAD fix does `res.end(); return` without `cancel()`, so if upstream sends a body with content-length on a HEAD, the undici connection is retained until GC. Trivial: `void upstreamRes.body?.cancel()` before `res.end()`.

8. **[R4-L4] Missing the negative test for "per-resource 403 does not revoke" on the mutation surface** (`desktop-app/ui/src/pages/__tests__/FilesPage.test.tsx:236`). The `it.each` only covers cases that *do* revoke. Behavior is correct today, but this is the exact class that came back three times (R1-H2, R2-M2, R3-H1); if someone reintroduced `'discovery'` at :254, no FilesPage test would catch it.

9. **[R4-L5] `DELETE /members/:userId` is now a breaking change for external clients** (`external-rest-api/src/routes/members.ts:90-102`). At the base the handler required nothing; now it 400s without `reason` and an `Idempotency-Key` header. First-party is updated (`profile-ui/lib/api.ts:329`), and nothing in desktop-app calls it, but a third-party integrator doing `DELETE .../members/<uuid>` with no body/header breaks on deploy. Suggest calling this out as an API version bump in the body and release notes.

There's also a handful of test-quality nits (SQL-string asserts on the 0096 migration test instead of the observed backfill, a hand-written link-service fixture missing NOT NULL columns, the row_hash tamper test only exercising 3 of 14 hashed fields, and `deriveExternalGfsEdgeLimits` being exported and tested but never wired into production). Happy to expand if useful, none of them block.

## What checked out clean

- Security came back clean: server-side GFS authority resolution fails closed on every branch (retired user, malformed/conflicting link, DB failure to 503), no user/admin subject confusion, SQL is parameterized, no admin credential or secret reaches Desktop.
- Generation fencing is fail-closed for 0/NULL/legacy on the data-plane, validated against real Postgres for the int4/int8 coercion.
- The brokered authority contract matches field-for-field emitter to validator (6/6 keys), and `retryAfterSeconds` is now consistent end to end (closes R2-L2).
- No tests mask a regression: the new round-4 tests (401-revoke on operation, per-client fairness) fail against the previous head, so they cover the change rather than pinning it.
- Focused suites green: control-api 122, gfs-controller 143 plus `tsc --noEmit` clean, desktop-app 72, external-rest-api 51, control-ui 20, profile-ui 1. Real-Postgres and Playwright suites weren't run here (gated/CI).

One process note since this is round 4: the two pieces that looked like they needed a redesign (the desktop error classifier and the rate-limit plane) actually converged, so my read is this doesn't need another full patching round. Closing the three concrete items above (R4-M1, R4-L1, R4-M2), making the three policy calls (R4-M3, R4-M4, R4-L5), and filing the rest as follow-ups gets this to merge.

zach88:
Took another pass, this time scoped tight. Since the last review the branch rebased again, but a range-diff shows the previous 58 commits are byte-identical (the only difference is a package version bump), so the real delta is just 4 commits (`0ed02047..8ce298ff`, 59 files). That set is the convergence tail we agreed to land: it closes every open finding from the last round with real tests, typecheck is clean, and every touched suite is green. Only one new regression fell out of it, and it's a latent Low. The remaining blockers to merge aren't code, they're two policy calls that need an explicit decision.

## Regressions introduced this round

1. **[R5-L1] Rate-limit boot invariant can crash startup on manual overrides** (`control-api/src/config.ts:363-370`, introduced by `2afdb91f`). The new `externalRateLimitConfig` IIFE enforces `operation <= session < clientIp` at load time and throws if it doesn't hold. Before this commit the three values were independent `Number(env || default)` reads with no cross-constraint. Reproduced with node: setting `APPROVAL_RL_EXTERNAL_EDGE_PER_MIN=1200` without also raising the client-IP limit gives `session=1200` and `clientIp=1200` (default), so `session < clientIp` is false and `config.ts` throws on load, the process never comes up. Same for bumping only `APPROVAL_RL_EXTERNAL_PER_MIN` past the edge default, and `0`/non-integer values now throw via `positiveIntegerFromEnv` where `Number()` used to tolerate them. It's Low because no `deploy/` manifest sets these envs (grep is empty), so production runs on the sane `60/120/1200` defaults and never trips it. It only bites a deploy that hand-overrides these limits. Fix options: validate only when all three are explicitly overridden, or downgrade the throw to a warn-and-clamp.

## Decisions needed before merge (carried, still open)

These two aren't defects, they're intentional behavior that needs a sign-off and a doc note. Both have been open since the last round.

1. **[R4-M4] Hard cut-over of legacy sessions on deploy** (`control-api/src/utils/auth/externalSessionAuthToken.ts:16`, `middleware/externalSessionAuth.ts:24`). Confirmed live at head: `signExternalSessionToken` throws `auth_generation_required` without `authGeneration >= 1`, `verifyExternalSessionToken` rejects tokens missing the claim (401), and the middleware does a per-request DB lookup requiring `lifecycle_version === authGeneration`. Net effect: every session token minted before this deploy is invalid on rollout, so every user gets bounced to re-login, with no grace window and nothing about it in the PR body. Either accept and document it in the body/runbook, or add a grace period that treats a missing `authGeneration` as legacy for N days.
2. **[R4-L5] Breaking change on `DELETE .../members/:userId`** (`external-rest-api/src/routes/members.ts:85-110`). The route now requires a `reason` body field and an `Idempotency-Key` header, returning 400 otherwise. First-party callers are updated, but any external consumer deleting without those breaks. Either document it as a versioned breaking change in the body plus release notes, or make the fields optional with a default.

## Closed this round

All verified against real code, R4-M1 with a regression test confirmed red on the prior head:

1. **[R4-M1] `deleteControlAdmin` idempotency** restored (`services/adminAuthService.ts:569-591`): `SELECT ... FOR UPDATE` plus a `status !== 'active' -> not_found` guard before any audit/retire/`session_version` bump, and `AND status = 'active'` on the UPDATE. Replay no longer re-audits or inflates the generation.
2. **[R4-M2] Per-IP rate-limit bound** back in place (`middleware/externalClientIdentity.ts:53-73`): two gates, a source-IP bucket (1200/min) and a session bucket (IP + token, 120/min). Rotating the pre-auth token no longer evades the cap, and this didn't re-introduce the 60/min throttle from the earlier round.
3. **[R4-M3] Default member-delete route now covered** (`profile-ui/test/memberApi.test.ts`): the test drives `deleteManagedUser(USER_ID)`, the exact signature the real callers use. This was the Medium in re-triage since round 2, now resolved.
4. **[R4-L1] `childrenQuery` 401 handling** enters as `surface:'operation'` (`useGfsBrowserController.ts:430`), so a 403 stays local and a 401 revokes, matching the `openAt` contract at `:378-382`. One thing to confirm: the 401-revokes behavior already existed on the prior head, so if the original intent was literally "401 must not revoke," that contradicts the documented contract. Worth a quick confirm with whoever filed it.
5. **[R4-L2] Readiness now validates migration 0096** (`gfs-controller/src/authz/storeProbe.ts:125-137`): checks the default via `pg_attrdef` + `pg_get_expr(...) IN ('1','1::integer')`, `attnotnull`, no rows `< 1`, and the positive constraint validated.
6. **[R4-L3] Upstream stream is cancelled** before `res.end()` (`external-rest-api/src/routes/contextSharedFilesystems.ts:114`).
7. **[R4-L4] Negative test added** for "resource-scoped 403 does not revoke" on the mutation surface (`FilesPage.test.tsx`), asserting the observable result and using the real `isGfsSessionAuthorityFailure`.
8. **Invitation rate-limit guards** are now inline on the boundary of all six `invitations.ts` endpoints, ordered consistently with the other families.

## What checked out clean

- `tsc --noEmit` on control-api is clean.
- Suites run green: control-api (config + adminAuthDelete + externalClientIdentity, 15/15), gfs-controller (storeProbe 41/41), external-rest-api (contextSharedFilesystems 6/6), profile-ui (memberApi 2/2), desktop-app (useGfsBrowserController + FilesPage 76/76), control-ui (Toast + ControlAdminsPanel 6/6).
- Rate-limit keying across all families uses distinct buckets with the correct gate order, no shared failure domain, no premature 429 for pre-auth anonymous traffic.
- The `retireParentInTransaction`, `deleteControlAdmin`, and `membership` (`lifecycle_state='active'`) contract changes are all propagated to their consumers in the same delta.
- The preview refetch path (`GfsImagePreview`/`GfsMarkdownPreview`) has stable memoized identity, no re-download loop.

One CI note before merge: please confirm `db.realPostgresMigration.integration.test.ts` is green. It's the one place a `pg_get_expr` mismatch from migration 0096 would block gfs-controller startup, since `storeProbe.test.ts` uses a hand-built fixture.

Process nit, not blocking: the operator-downgrade state machine in `useGfsBrowserController.ts` keeps growing inside `fix:` commits to close Low findings. It'd be worth half a page of decision-table spec before it grows further.

zach88:
Did a focused pass on the delta since the last review round (head `0040daae`). This one's clean, and from my side the code has converged. I'm good with merging this; the remaining items are all policy calls, not code defects, and can go out as follow-ups.

Since the branch got rebased again, the only genuinely new work since the previous reviewed head (`c884a43f`) is 3 convergence commits, so I scoped the pass to `git diff c884a43f..0040daae` with a single question: did these fixes break anything? Answer: no regressions, no new findings.

## What the delta did

1. **[R5-L1] Rate-limit boot no longer crashes on a crossed topology** (`c15ca381`). `control-api/src/config.ts` used to `throw` at boot when `operation <= session < clientIp` didn't hold; it now emits a `console.warn` advisory and preserves the exact configured values. Verified nothing depended on the crash: the config only feeds three scalars, and the 13 `routes/external/*.ts` call sites build independent source-IP and session limiters that don't assume any hierarchy, so a crossed topology just makes one bucket stricter than another. No deploy manifest sets `APPROVAL_RL_EXTERNAL*`, and the docs were updated in the same delta. This closes R5-L1 (the latent boot-crash regression flagged last round). Reproduced: the 6 config tests fail against `c884a43f` and pass on head, so the test covers the fix rather than freezing the bug.

2. **Image builders now include `display-field`** (`0040daae`). `control-api/Dockerfile` and `control-ui/Dockerfile` were missing the `COPY packages/display-field` that the `file:` dependency needs, which was blocking the build gate. Checked it out: the package is data-only with the expected shape, the `COPY` lands before `npm ci`, and the control-ui `cp -R` sits in the right spot in the `RUN` without clobbering the other `@clerum/*` packages.

3. **Legacy session token test** (`3131d334`). Hardens the fixture using a real RS256-signed token with no `authGeneration`, matching the legacy signer. It's a characterization test for already-intended behavior (rejecting tokens without `authGeneration`), not a repro pinned to a fix, so it doesn't shield anything.

## Follow-ups (non-blocking, fine to land after merge)

These are decisions rather than bugs, carried over unchanged:

1. **[R4-M4] Hard session cut-over.** Session JWTs minted before this release lack `authGeneration` and get a `401`, so everyone signs in once after rollout. Intentional and now covered by the legacy-token test; just wants a line in the release notes.
2. **[R4-L5] Breaking `DELETE /members/:userId`.** It now requires `reason` + `Idempotency-Key`. First-party callers are updated; external API clients need the version bump documented (or the fields made optional with defaults).
3. **[R5-L1] Warn-vs-clamp choice.** Resolved here via warn + preserve-exact-values. If you'd rather clamp crossed values than just warn, that's a quick swap, but warn is the safer default.

## What checked out clean

- Rate-limit builder composes independent limiters with no hierarchy assertion.
- `.dockerignore` doesn't exclude `packages/display-field`; the package has the shape both Dockerfiles expect.
- No dangling references to the old hierarchy contract; docs (`gfs-desktop-operator-parity.md`, `production.md`, `external-rest-api/README.md`) are in sync.
- Delta tests green on head (19/19); the config suite is red against the prior head, confirming coverage.

One thing to confirm on your side before merge: `db.realPostgresMigration.integration.test.ts` green in CI (carried over from last round).

alfredolopez80:
Validation update — synchronized, tested, and pushed

The branch is now rebased onto the current origin/dev at d9d79d7f and force-pushed at commit c59707dd8ec41aa6151e6cef1c073349187104dc.

The three carried decisions from this review are now explicit:

- R4-M4 — Hard session cut-over is intentional. Legacy RS256 session JWTs without authGeneration are rejected with 401 before the lifecycle database lookup. Users must sign in once after rollout. The behavior is covered by the real legacy-token test and documented in the rollout/production notes.
- R4-L5 — DELETE /members/:userId intentionally requires a bounded reason and Idempotency-Key. First-party callers are updated and the breaking external-client contract is documented. The fields were not made optional and no silent defaults are synthesized.
- R5-L1 — Rate-limit values are parsed independently as positive safe integers. Crossed topologies no longer crash Control API startup; configured values are preserved exactly and one advisory warning is emitted. Focused coverage includes defaults, partial overrides, equality/reversed tuples, blanks, and invalid scalars.

The display-field image-builder fix from PR #357 is present exactly once in both Dockerfiles, with manifest/change-detection coverage for both affected images.

Validation for c59707dd:
- Control API full suite: 4,175 passed, 68 skipped.
- Control UI full suite: 1,510 passed, 1 skipped.
- Focused R5/config/legacy/client-identity suite: 27 passed.
- Control API and Control UI builds: passed.
- Image manifest contract: passed, including display-field-only rebuild selection.
- Exact-head local T2 on profile/context clerum-gfs-operator-c8b14240: local Control API and Control UI images built, migrations completed, GFS reader/writer wiring verified, and relevant workloads Ready.

This comment records the validation state; it is not a merge approval.

zach88:
Took another pass, this time scoped to what actually changed since the last round. The branch got rebased again, and the only genuinely new work is a small 6-commit delta (about 480 lines) that packages the local `@clerum/display-field` into the control-api and control-ui Docker images so `npm ci` can resolve its `file:` dependency. That's the exact Dockerfile gap the PR body called out as blocking the T2 pre-gate, plus one docs-only comment tweak. None of it touches the GFS authority/session/rate-limit core, which has been converged and clean for a couple rounds now.

### 🟡 Approve with nits
- **Optional / non-blocking: 2**

Nothing new this round. The deploy delta checked out clean and introduced no regressions.

## Optional (non-blocking)

These two are policy calls that have been open since round 4, not code defects. The PR can merge once you make the call and document it; they don't gate on code quality. Flagging them as re-triage: since they've survived multiple rounds, they should either get an explicit decision now or be closed in writing rather than floating another round.

1. **[R4-M4] Hard session cut-over on deploy.** Session JWTs minted before this release don't carry `authGeneration`, so `signExternalSessionToken` / verify reject them and users get a `401` until they sign in again once (`control-api/src/utils/auth/externalSessionAuthToken.ts:16`, `control-api/src/middleware/externalSessionAuth.ts:24`). This is intentional with no legacy grace path. Just needs a conscious "yes this is acceptable" plus a line in the body/release notes so nobody's surprised by the forced re-login.

2. **[R4-L5] Breaking change on `DELETE /members/:userId`.** The route now requires a JSON `{ "reason": "<non-empty>" }` and an `Idempotency-Key` header, returning `400` when missing (`external-rest-api/src/routes/members.ts:85-110`). First-party callers are updated, but direct external clients will break. Either document it as an API version bump in the body + release notes, or make the fields optional with a default.

## What checked out clean

Verified against the real code this round (reproduced except where noted):

- **Dockerfiles net to exactly one `COPY packages/display-field`, before `npm ci`.** `control-api/Dockerfile:13` and `control-ui/Dockerfile:11` each land the copy ahead of the `npm ci` that needs it. The add-then-dedup pair of commits netted to a single copy with no duplicate and no accidental zero-copy.
- **Change-detection is coherent across all three layers.** `packages/display-field/**` is consistently declared in `.github/workflows/build-publish.yml:125,129`, `deploy/images.json`, and `scripts/minikube/pre-gate-incremental.sh:154`. No image depends on the path in its Dockerfile without listing it for rebuild, and vice versa. Ran `scripts/tests/test-images-manifest.sh`, 22/22 pass.
- **The new manifest test is a real guard.** Reproduced T3: it fails at the commit that added it (`filters for packages/display-field/** are <none>, expected control-api,control-ui`) and passes at head. It asserts the observable set of selected images, not an intermediate effect.
- **The `config.ts` comment tweak matches runtime.** The rate-limit topology comment (`control-api/src/config.ts:792-795`) accurately describes the code at `config.ts:386-401`: independent per-env-var scopes, crossed values preserved after a boot advisory, no clamping and no crash.
- **Lockfiles stay consistent.** control-api and control-ui `package.json` and `package-lock.json` agree on version and on the `@clerum/display-field: file:../packages/display-field` entry, so `npm ci` stays deterministic.

One CI check worth confirming before merge: `db.realPostgresMigration.integration.test.ts` green, and the T2 pre-gate now that display-field is in the images.

zach88 (approved):

Issue #300, opened 2026-08-07
Desktop GFS upload cap is a hardcoded compile-time constant (make it runtime-configurable)


alfredolopez80:
### Symptom (observed live in `clerum-dev`)
After PR #280 (which raised the cap to 16MB via #284) merged and deployed, an **installed desktop build still rejected GFS uploads above ~10MB** — the pre-#284 value. The rejection happens client-side, before any request leaves the app: no server log is produced for the over-cap attempt. Meanwhile the server side already allows 24MiB end to end (funnel `client_max_body_size 25165824`, gfsc `GFS_MAX_WRITE_BODY_BYTES 25165824`, external-rest-api/control-api `express.json` 150mb).

### Root cause
The desktop upload cap is a **hardcoded, compiled-in constant** with no runtime/server override:

- `desktop-app/ui/src/constants/gfsFileUpload.ts:3-4`
  ```ts
  export const GFS_FILE_UPLOAD_MAX_BYTES = 16 * 1024 * 1024
  export const GFS_FILE_UPLOAD_MAX_MEGABYTES = 16
  ```
- `desktop-app/ui/src/lib/gfsFileUpload.ts:5` — `assertGfsFileUploadSize` throws `"GFS uploads are limited to ${GFS_FILE_UPLOAD_MAX_MEGABYTES} MB per file."` **before** the file is sent.

Because it is baked into the renderer bundle at build time:
1. **Installed binaries silently diverge** from the server cap. A build made before #284 caps at 10MB even though the server (and the current repo) allow 16MB. Users don't get the new cap until they install a new desktop release.
2. **No server-side toggle.** Every other GFS size/timeout limit in this stack is runtime-configurable (funnel `client_max_body_size`, gfsc `GFS_MAX_WRITE_BODY_BYTES`, `CONTROL_API_GFSC_PROXY_TIMEOUT_MS`, control-ui `GFS_UPLOAD_TIMEOUT_MS`), but the **desktop client cap is the one leg that requires a binary rebuild** to change.
3. The client cap can never track a future server bump without shipping a new app.

### Proposed fix
Make the desktop GFS upload cap **runtime-configurable**, fetched from the server so it tracks `GFS_MAX_WRITE_BODY_BYTES` (÷ ~4/3 for base64 inflation) without a rebuild:

[...]

Issue #286, opened 2026-08-06
GFS Phase 2: replace base64-in-JSON uploads with multipart/streaming (enables 50MB+)


alfredolopez80:
Follow-up from #280 (deferred Phase 2; design in `.ralph/plans/2026-08-06-gfs-upload-integral-solution.md`).

base64-in-JSON inflates the wire +33% and forces O(body) buffering (~5 copies at control-api, ~4.3 at gfsc), which is why the 16MB cap is memory-bound (control-api 768Mi). Multipart/streaming makes memory **flat** regardless of file size and removes the amplification, unlocking 50MB+ cleanly.

Scope (4 surfaces): gfsc `application/octet-stream` endpoints for POST children / PUT content (metadata via headers/query); control-api raw passthrough `fetch(req, { duplex: 'half' })` bypassing express.json for those routes; control-ui streaming route handler + File-body client (drop `fileToEncodedData`). Keep the JSON+contentBase64 surface for hosts/agents.

Acceptance: a 50MB upload succeeds end-to-end at bounded (flat) memory, byte-match verified.

Issue #285, opened 2026-08-06
GFS upload: close test-coverage gaps (RC4 expect-strip, timeout signal, 16MB e2e)


alfredolopez80:
Follow-up from #280 (audit coverage notes).

The fix shipped with these gaps:
- **RC4** `expect`/`proxy-connection` header strip in `control-ui/app/control-api/[...path]/route.ts` has **no unit test**.
- control-api proxy timeout tests assert error-mapping only, never that `signal:` is actually attached to the gfsc fetch (deleting it stays green).
- `parseTimeoutMs` in `control-api/src/config.ts` is untested.
- **No e2e test uploads a large file** — the exact gap that let this ship (dev 'passed' because nobody uploaded >3.2MB). 

Acceptance: add the unit tests above, and a **large-file (near-cap) upload** to the GFS e2e gate that asserts 201 + sha256 byte-match (regression guard).

Issue #282, opened 2026-08-06
GFS: route-scoped 24MiB body cap on /gfs/proxy ahead of the 150mb global express.json


alfredolopez80:
Follow-up from #280 (audit finding **H2**).

`control-api/src/config.ts` `jsonBodyLimit` = **150mb** (`CONTROL_API_JSON_BODY_LIMIT`) applied app-wide in `app.ts`. An authenticated operator can POST ~150MB to `/api/v1/gfs/proxy/*`; base64-in-JSON buffers ~5 copies (~750MB) and can OOMKill control-api even at 768Mi, and gfsc's 413 (16/24MiB) fires far too late.

Acceptance: mount a route-scoped `express.json({ limit: '24mb' })` on the GFS proxy routes (ahead of the global parser) so oversized bodies are rejected before buffering, without lowering the global limit other routes rely on.

alfredolopez80:
## Adversarial-audit sharpening (from PR #280 meta-review)

An adversarial audit of PR #280 sharpens this issue's scope: the uncapped body path is reachable **pre-authentication**, not just for authenticated admins.

- `control-ui/app/control-api/[...path]/route.ts` does `await req.arrayBuffer()` with **no size limit, before any auth** (auth lives in control-api, downstream).
- control-api applies `express.json({ limit: config.jsonBodyLimit })` **globally, before the auth gate** (`app.ts:78-90`), default `'150mb'` (`config.ts:487`).

**Chain:** an unauthenticated POST through the public operator tunnel → control-ui (512Mi) buffers the full body → whatever passes → control-api (768Mi) buffers + parses up to 150MB **pre-auth**. With base64's ~5× body-copies-per-hop, a single 150MB body is ~450MB+ transient in control-api → OOMKill — which invalidates the 768Mi sizing PR #280 derived for ≤22.4MB bodies. PR #280 homologated the funnel (24MiB) and gfsc (24MiB) caps but left the operator front-door uncapped.

**Recommendation (unchanged, with one addition):** route-scoped ≤24MiB body cap ahead of the global 150mb `express.json` — and apply it **before** the auth gate so the DoS is bounded for anonymous callers too, not only authenticated admins.

**Provenance:** both lines pre-exist PR #280 (`git blame` → commit `5d663a49`, Jose Ramirez, 2026-07-14). Mechanism verified by reading code; OOM not reproduced live (Medium-High severity; drops if the public tunnel caps body size below ~24MiB upstream — worth confirming).

PR #311, opened 2026-08-10
feat(gfs): resumable uploads up to 200 MiB

alfredolopez80:
## Summary

Implements the approved GFS Upload v2 contract for the packaged Desktop app and Control UI with a hard 200 MiB binary file ceiling.

- Product maximum: `209,715,200` bytes (200 MiB binary); `209,715,201` is rejected before session allocation.
- Indexed resumable binary parts: 8 MiB preferred, 16 MiB hard per-request maximum, four concurrent parts with a bounded 2-part fallback after repeated retryable failures.
- Server-owned sessions validate geometry, idempotency fingerprints, checksums, pause/resume/cancel, response-loss-safe completion, transactional publication, receipts, expiry, and writer reconciliation.
- Legacy JSON/base64 compatibility remains separate and capped at its existing 16 MiB limit.
- Control UI uses bounded Blob slices and XHR progress; Desktop main-process streaming/hash work keeps whole files out of IPC and renderer memory.

## Security and transport closure

- Desktop persisted upload state is versioned and scoped to owner, team/tenant, normalized environment/base URL, drive, and auth epoch; legacy or invalid unscoped records are quarantined.
- Logout, team switch, and environment changes fence GFS dispatch, abort and await in-flight work, clear credentials, and persist `suspended_auth`; only an explicit same-scope resume is allowed.
- Every Desktop v2 lifecycle request carries a canonical drive. The external relay forwards the selector opaquely; Control API owns missing/whitespace/multivalue/body-query validation, ownership, authorization, and writer-token minting.
- Upload admission is replica-safe through PostgreSQL atomic request/weighted-byte buckets and advisory-lock concurrency leases. Admission fails closed when PostgreSQL is unavailable and returns stable `429` plus `Retry-After`.
- The legacy fallback opens one `O_NOFOLLOW` descriptor, validates it with `fstat`, and reads from that descriptor.
- The relay preserves GFS transport status/body and safe headers for `408`, `413`, `425`, `429`, `507`, and other `5xx` responses so clients distinguish retryable failures from terminal storage exhaustion.
- GFSC now carries signed `authGeneration`, `principalType`, and brokered authority metadata through the upload service and finalizer reauthorization; this prevents a valid token from being downgraded to a generation-less second check.

## Exact provenance

- PR branch: `codex/gfs-upload-v2-200mib`
- Exact head: `cf2509b5964e55e3c2c31379a62aeff1a7bfcb90`
- Base and merge-base (`github/dev`): `d3905ffd1feb51c6d77ddd2f9d42583f09da2fcc`
- The branch and worktree were clean at validation time; `git diff --check` passed.
- The v2 capability remains disabled by default; this PR does not claim production enablement.

## T0 local validation

- GFSC focused route/session/finalizer tests: `64/64` passed.
- Control API GFS relay/admission tests: `90/90` passed.
- E2E Guardian static audit: Control UI and packaged Desktop specs passed.
- GFSC, Control API, External REST, Desktop (main + renderer), and Control UI builds/typechecks passed.
- Deployment, Docker local-package, Minikube pre-gate, and full-setup contract suites passed; `git diff --check` passed.

## T1 real PostgreSQL validation

The branch-owned profile `clerum-codex-gfs-upload-v2-200mib-9c072c78` executed the real PostgreSQL lane with the anti-skip guard enabled:

- Control API: 10 suites, `66` tests, `0` skipped.
- GFSC: 2 suites, `8` tests, `0` skipped.
- Coverage includes session/part concurrency, idempotent replay, response-loss reconciliation, finalizer publication, stale leases, expiry, missing-file recovery, and rate-limit admission.

## T2 runtime validation

Against the same branch-owned profile and exact head, using the real Control UI/Desktop user journeys, PostgreSQL-backed sessions, visible progress/business signals, and real download byte/hash verification:

- Control UI Playwright: `8/8` passed — 199 MiB and 200 MiB create/replace, drag/drop pause plus persisted reload/resume, visible 200 MiB + 1 byte rejection, v2-disabled legacy fallback, and writer restart recovery.
- Packaged Desktop Playwright: `8/8` passed — 199 MiB and 200 MiB create/replace, visible oversize rejection, v2-disabled legacy fallback, writer restart recovery, and grant revocation during upload.
- Profile marker after the run: `gitHead=cf2509b5`, `imageSource=local`, writer flag restored to `true`.

## CI and security

- CI for the exact head: [31880965074](https://github.com/evenfire-ai/evenfire/actions/runs/31880965074) — successful.
- CodeQL for the exact head: [31880965081](https://github.com/evenfire-ai/evenfire/actions/runs/31880965081) — successful; open alerts are `0` on the branch and PR merge ref.
- The GitHub Advanced Security unused-variable finding in `control-api/test/routes.externalGfs.test.ts` was removed and the resulting CodeQL alert is fixed.
- GitHub reports the PR mergeable; any remaining repository approval requirement is governance, not a failing check.

## Architecture diagram

GFS Upload v2 — PR #311 architecture

- [SVG export](https://github.com/evenfire-ai/evenfire/blob/cf2509b5964e55e3c2c31379a62aeff1a7bfcb90/docs/architecture/diagrams/gfs-upload-v2-pr311.svg)
- [Editable JSON source](https://github.com/evenfire-ai/evenfire/blob/cf2509b5964e55e3c2c31379a62aeff1a7bfcb90/docs/architecture/diagrams/gfs-upload-v2-pr311.json)

## Scope boundary

This PR changes only GFS Upload v2, its clients, relays, persistence/admission, deployment contracts, and their tests. It does not modify PR #310, operator-link lifecycle, configurable GFSC issuer behavior, or unrelated rate-limit paths.

References: #282, #285, #286, #300.


zach88:
Took a deep pass through this one across all six services (control-api, gfs-controller, desktop-app, control-ui, external-rest-api, plus the deploy/docs/scripts). The core is genuinely solid: the field/type/geometry contract is tight end-to-end, the security boundaries hold up (writer-token minting can't escalate, no IDOR, path/symlink writes are locked down, admission fails closed on Postgres down), and the desktop scoping/quarantine work is careful. Where it wobbles is at the edges of the resumable promise: the finish-line `complete` call and the two headline safety paths the PR is named after aren't actually exercised. Nothing here is a data-loss Blocker, but three High items are worth closing before merge.

Two process notes up front: this PR is 112 files across 6 services (the soft budget is ~25 files / 1 service) — worth a line in the description on why it can't be split, or splitting off docs/deploy/scripts at least. And the body's T0/T1 + CodeQL evidence is pinned to `3e9d6e01`, but head is `083a869b` with 4 commits on top — including the `reconcile ambiguous part responses safely` fix, which is exactly the path that has no test coverage (see #3). The evidence predates the riskiest change.

## High

**1. [R1-H1] 429/Retry-After is only retried on part PUTs, not on lifecycle calls**
`control-ui/lib/gfsFileUpload.ts:1001`, `desktop-app/src/gfs/upload.ts:1111`. The server emits 429 + `Retry-After` on every upload route (admission, the edge limiter's `public_edge_requests`, and gfsc's `quota_exceeded`). Both clients classify 429 as retryable but only apply the retry inside the part loop (`uploadPartWithRetries`/`putPart`). The lifecycle helpers — create/complete/pause/resume and the resume's HEAD+first status page — have no retry loop, and neither client reads `Retry-After` (fixed exponential backoff instead). So if two users share a source IP and trip `source_ip_requests`, a 200 MiB file that's fully uploaded gets a 429 on `POST …/complete` and the job goes to `failed` at the finish line, even though the server explicitly asked it to retry. Fix: route lifecycle 429s through a retry that honors `Retry-After` in the shared request helper of both clients.

**2. [R1-H2] A transient team hop tears down in-flight uploads, against the file's own documented invariant**
`desktop-app/src/appService.ts:977`. This adds `await this.suspendDesktopGfsUploadsForAuthBoundary()` inside `switchSessionToTeam`, which is shared by the deliberate `switchTeam` and the transient per-operation hops (`runWithTeamContext`/`issueRpcTokenForHostRefs`, lines 1045/1056). The comment at 2727-2729 spells out the invariant: only the user-initiated switch should close streams, the transient hops must not. As written, an agent tool-call that mints a cross-team RPC token now aborts the user's in-flight upload, bumps `gfsAuthEpoch`, persists `suspended_auth`, and clears `gfsUploadJobs` — and the restoring hop doesn't resume it. Committed bytes survive (it's resumable, so not data loss), but it's a silent interruption of the user's work triggered by a background op they didn't start. Fix: move the suspend to the deliberate boundaries (`switchTeam`/`installAuthenticatedLogin`/`applyRuntimeEnvironmentChange`/`logout`), out of `switchSessionToTeam`. No test covers the "transient hop leaves uploads intact" invariant either.

**3. [R1-H3] The two headline invariants have no executable test**
`gfs-controller/src/upload/uploadSession.test.ts`. The PR is named for "response-loss-safe completion" and "reconcile ambiguous part responses safely," but neither is tested. The response-loss-safe branch of `complete()` (`uploadSession.ts:1214-1234`, the `CommitOutcomeUnknownError`/`RollbackOutcomeUnknownError` → idempotent re-read path) never runs — `grep CommitOutcomeUnknown *.test.ts` is 0. `reconcileCommittedPart()` (1276-1307) and the ambiguous catch in `putPart` (819-854) don't run in any test. And all three `complete()` tests use `sizeBytes: 0` with a `finalize` that always throws/expires, so the successful-completion path and its idempotent retry are never exercised. Combined with #1 (client dies on `complete` under 429), the most critical path in the PR is unverified on both sides. Worth: a test that completes a non-zero session with real parts and retries `complete()` asserting idempotency, plus one that forces `CommitOutcomeUnknownError` and checks the re-read resolution — each should fail against an implementation that breaks idempotency.

## Medium

**4. [R1-M1] The state machine is only tested against a hand-written MemoryDb fake**
`gfs-controller/src/upload/uploadSession.test.ts:13-184`. The fake reimplements Postgres semantics by string-match: `contiguous_bytes` (a JS loop at 126-132 vs the real `generate_series … MIN(missing)` subquery at 798-810), the reserved-bytes formula, the counters. `tx.transaction` is `fn => fn(db)` with no isolation, so `pg_advisory_xact_lock`, `FOR UPDATE [SKIP LOCKED]`, the `session_epoch` fencing, and all concurrency go unverified. The repo already has the right pattern and its literal justification in `db/renamePublication.realPostgres.integration.test.ts` (gated on `CONTROL_API_REAL_PG_ADMIN_URL`) — worth extending to create/putPart/complete/cancel/reconcile, including a two-concurrent-parts race.

**5. [R1-M2] Progress bar jumps backwards on concurrent multi-part uploads** (reproduced)
`control-ui/lib/gfsFileUpload.ts:970` + `components/GfsBrowser.tsx:286-289`. Two producers call `onProgress`: the accumulator `progress()` (941-953, monotonic, sum of completed parts) and `onByteProgress` (970 → `putPart` 464-469) which forwards the in-flight part's absolute position in the file. With `concurrency=4` parts upload out of order, so part 3 emits ~24-32 MiB while parts 0-2 are still going; when part 0 completes the accumulator emits 8 MiB, and `GfsBrowser:288` assigns `uploadedBytes` with no clamp — so the bar jumps 32→8 MiB (on a 200 MiB file it can swing from ~95% to ~40%). Fix: route byte-progress through the accumulator (track in-flight `loaded` per part) or clamp to a monotonic max. The only progress test uses a 4-byte / single-part file, which never exercises concurrency.

**6. [R1-M3] The `reconcile()` test proves nothing**
`gfs-controller/src/upload/uploadSession.test.ts:372-405`. The fake returns `{rows:[]}` for the stale-lease/expiry/committed-missing queries, so `expect(reconcile()).toEqual({staleParts:0, expiredSessions:0})` is true by construction of the fake, not by the logic. The stale-lease cleanup, the expiry sweep (1000-1018), and the `corrupt_part_missing` marking (1029-1041) are all uncovered. Best folded into the real-PG harness from #4.

## Smaller stuff (grab-bag)

- **[R1-L7] Dead code**: `control-ui/lib/gfsFileUpload.ts:985-996` — the `if (result.paused || this.pauseRequested)` block is duplicated; the first copy (973) returns unconditionally, so 985-996 is unreachable. (flagged by two passes)
- **[R1-L4] D4 duplication**: the 200 MiB ceiling and part geometry live as separate literals in `gfs-controller/upload/protocol.ts:10`, both clients' constants files, and hardcoded again in `desktop-app/src/gfs/upload.ts:6`. Consistent today (server clamps to 209,715,200 exactly), but a divergence-on-edit risk — worth a shared constant.
- **[R1-L1] A2**: the operator proxy (`control-api/src/routes/gfs/proxy.ts:14`) forwards only content-type/-disposition/-length, so the web resume size-drift guard reading `upload-length` (`gfsFileUpload.ts:323-328`) is dead over that path. Backstop exists (the follow-up `GET status` compares `expectedBytes`), so no corruption, just a silently-disabled check that desktop still has.
- **[R1-L2] A3**: web `UploadCapabilities` (`gfsFileUpload.ts:110-114`) declares `preferredPartBytes`/`maxPartBytes` that the writer never emits (it sends `*ChunkBytes`), so they're always undefined dead fields; and the web client never reads `maxFileBytes`, so it can't honor a server-advertised ceiling below its local constant. Desktop uses the right names.
- **[R1-L3] A5**: the web client doesn't verify the response `drive` (leaves `UploadReceipt.drive` optional, never compares), the body-promised check that desktop does at `upload.ts:1024,1129`.
- **[R1-L5] N1**: `gfs-controller/src/api/serve.ts:900-905` — the hardened `optionalNumber` (rejects negatives/non-safe-int) is consumed by pre-existing resource create/replace routes (715/751), not just upload. Only rejects already-invalid input, but it's a silent contract change in a shared helper — worth a test or a comment.
- **[R1-L9] Rate-limiter test confidence**: `control-api/test/services.rateLimiter.test.ts:116,139` — the "no race"/"replica-safe advisory slots" tests run against a synchronous in-memory Map/Set mock that can't exhibit the race or cross-session contention the names claim. The underlying SQL is genuinely atomic, so it's false confidence rather than a bug — a real-PG test (gated like the session one) would actually exercise it.
- **[R1-L13] Source-IP header trust**: `control-api/src/middleware/gfsUploadAdmission.ts:67-72` trusts `x-gfs-upload-source-ip` when it parses as an IP. Mitigated — the relay sets it and drops the client's copy, so spoofing needs direct in-cluster access plus a valid token; net effect if reached is weaker IP metering only. Worth binding the header to the relay's identity.
- **[R1-L14] Per-part file reopen** (integrity, not security): `desktop-app/src/gfs/upload.ts:246,497` re-opens the source by path per part instead of a single validated descriptor (the way the legacy fallback does with `O_NOFOLLOW`+`fstat`). A local file swap between parts yields an inconsistent object, but only on the uploading user's own machine, no privilege boundary — acceptable for a restart-surviving design, just noting it.
- A few more low-value notes in the ledger: geometry pure functions with no property-based tests (`protocol.ts:91-137`), `get()` not throwing on a soft-expired session (`uploadSession.ts:564-576`), the `uploadId` regex format coupling across two routers with no contract test, and a no-op `inc({result:'rejected'},0)` metric.

## What checked out clean

- **Security: SAFE, nothing at Medium+.** Writer-token minting is scoped to the caller's subject with drive validated before mint and re-checked at gfsc; every session op is scoped by `upload_id AND drive AND owner_subject` (no IDOR); blob paths derive only from UUID + int part number with `O_CREAT|O_EXCL|O_NOFOLLOW` 0600 and containment checks; all SQL is parameterized; part size (16 MiB, three ways) and whole-file (200 MiB + partCount ≤ 1024 before insert) are clamped; admission fails closed on Postgres down (503, never admit); auth runs before the edge rate-limit so there's no existence leak; the desktop persistence envelope carries no token; no NetworkPolicy lanes were opened.
- **control-api**: canonical-drive validation is applied on all nine v2 routes; the weighted byte-cost is charged against observed bytes (a false Content-Length can't under-charge); the `checkAndIncrement` change is additive and backward-compatible.
- **desktop-app**: the 5-component scope (`ownerId+teamId+environmentKey+baseUrl+drive+authEpoch`) makes cross-scope resume impossible, baseURL normalization is robust, legacy records are quarantined without rehydration, the file never crosses IPC (renderer passes only `filePath`), hashing is streamed, and in-flight work is awaited before credentials are cleared.
- **contract**: field names, receipt shape, part headers, checksum format, and the geometry formula are identical across all three tiers; idempotency is computed server-side only (no duplicated client fingerprint to diverge); the legacy 16 MiB base64 path stays separate and functional; cancel-during-finalize reconciles via `GET status` on both clients.
- **Suites** (run during review): control-api 88 ✓, gfs-controller 84 ✓ + `tsc --noEmit` ✓, desktop-app 31 ✓, control-ui 16 ✓. CI is green on head including CodeQL and the real-Postgres integration job.

zach88:
Took another deep pass through this one for round 2. Good news up front: all six of the serious round-1 items (the three Highs and three Mediums) are genuinely closed, several backed by a real Postgres harness and T3-verified regression tests. The remaining stuff clusters in two places: the desktop team-fence state machine in `appService.ts`, and the progress/retry orchestration on the web client. One thing that still blocks, a handful worth deciding on, and some follow-ups.

Heads up on two process things before the findings: this branch got rebased between rounds (old base `a6a4e570` to `21d9a7d5`), so the collateral audit ran against the rebased equivalent of the round-1 head. And the PR body's T0/T1 + "CodeQL green" evidence is pinned to `3e9d6e01`, which isn't an ancestor of the current head anymore, so that evidence predates ~3k lines of round-2 fixes.

## Blockers

**1. [R2-H1] Control/read ops on in-flight uploads still throw during a transient team hop.** `desktop-app/src/appService.ts:2168,2187,2204,2234,2264`. The R1-H2 fix added `allowTransientTeamHop: true` only to the in-flight job scope check. But `getGfsUploadSnapshot`, `pauseGfsUpload`, `resumeGfsUpload` (existing-`entry` branch), `cancelGfsUpload` (existing-`entry` branch) and `listGfsUploadSessions` still call `currentDesktopGfsUploadScope(drive)` without the flag, so they run `assertGfsDispatchAllowed()` and throw `GFS upload dispatch is unavailable during a transient team hop` while `gfsTransientTeamHopDepth > 0`. The code's own comment at `:2238` says "Only new user-facing dispatches are gated" which these five paths violate; snapshot and list are read-only and shouldn't gate at all.
Repro: with an existing job in `gfsUploadJobs`, calling `getGfsUploadSnapshot('up-1','main')` / `pauseGfsUpload('up-1','main')` inside `runWithTeamContext('team-b', ...)` throws and `job.pause` is never invoked.
Fix: pass `{ allowTransientTeamHop: true }` on the five routes that operate on an existing entry or are read-only; keep the gate only where a brand-new job mints a fresh token (`startDesktopGfsUpload`, `createGfsFileFromPath`, the no-entry branch of `resumeGfsUpload`). This is the sibling instance the R1-H2 fix should have generalized to.

## High

None beyond R2-H1.

## Medium

**2. [R2-M1] `getUploadSnapshot` emits `state: 'active'`, which is outside the declared IPC union, so a restarted upload loses its pause/resume controls.** `desktop-app/src/appService.ts:2179` (persisted-record branch) returns `{ state: record.status }`, and `record.status` can be `'active'` (set at `:2088`). The union in `desktop-app/src/renderer.d.ts` and its twin in `desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts:70-79` don't include `'active'`. After an app restart, an interrupted upload is persisted with `status: 'active'` and no in-memory job; the snapshot takes the persisted branch and returns `'active'`, and `desktop-app/ui/src/pages/FilesPage.tsx:615-663` has no branch for it, so you get no pause/resume button (only cancel) and a default label. `listUploadSessions` already includes `'active'` in its type; only the snapshot branch drifts.
Fix: map `record.status` to a contract state in that branch (e.g. `'active' -> 'uploading'`), or add `'active'` to both unions.

**3. [R2-M4] The production finalize seam in the writer has no executable test.** `gfs-controller/src/index.ts` (~26-120, ~268-370). `orderedUploadStream`, the permission-epoch revocation recheck ("authorization changed during upload finalization" / "upload authorization was revoked before publication"), the transactional `onPublished` write, and the streaming `writeService.create/replace` path are all wired here, but every test injects a mock `finalize` and the real-PG suite never touches this closure. That "revoked before publication" invariant lives only in prose plus code right now. Worth an integration test that wires the real `GfsWriteService` and asserts (a) happy-path publish produces the resource with the streamed sha, and (b) a revoked epoch aborts publication and marks the session failed.

**4. [R2-M3] `switchTeam` now forces a logout if GFS fence persistence fails, which flips the failure contract.** `desktop-app/src/appService.ts:2846` (from the team-fence reorder + fail-closed change). Before, the fence ran before the token swap with no try/catch, so a state-write failure aborted the switch and left you authenticated on the old team. Now that same transient write failure (full disk / bad perms on the GFS state file) runs `clearAuthenticatedSessionState` + `clearSessionToken`, forcing a re-login. It's tested (`appService.rpc.test.ts:328`) and it's a deliberate fail-closed after the new token is installed, which is defensible on security grounds, so I'm flagging it for a call rather than as a hard block. It only affects the deliberate `switchTeam`; transient hops stay best-effort, so it doesn't block tool-calls. This is a behavior regression introduced by the round-2 team-fence commits.

**5. [R2-M2] The "monotonic progress bar" invariant breaks on a part failure+retry, and nothing tests it.** `control-ui/lib/gfsFileUpload.ts:1270-1273` (`onPartFailure`) drops the failed part's in-flight bytes and recomputes `uploadedBytes` lower, and `GfsBrowser.tsx`'s `mergeUploadProgress` (unlike `mergeUploadSnapshot`) applies no monotonic max-guard, so the retry's next byte-progress visibly regresses the bar. It's deliberate (comment: a failed retry shouldn't leave the aggregate permanently overstated) and arguably correct, but it contradicts the "monotonic bar" claim and the only monotonicity test covers reorder, not failure. Pick one: document the invariant as "monotonic absent failures" and add a failure-case test, or floor `mergeUploadProgress` to a monotonic minimum. Also note the two-merge split (guarded snapshot vs unguarded progress) is the kind of duplicated seam logic that drifts.

**6. [R2-M5] The real-PG advisory-slot test doesn't exercise the production function.** `control-api/test/services.rateLimiter.realPostgres.integration.test.ts:72-99` hand-writes `pg_try_advisory_lock(hashtext($1), 0)` with two clients instead of calling `acquireRateLimitConcurrencyLease`, so it never covers bounded slot selection, the `heldConcurrencySlots` reentrancy guard, ordered multi-requirement acquisition, or partial rollback, and it uses `hashtext` rather than `advisoryKey()` (sha256 to int32). The only test of the real function is still the synchronous Set mock. So R1-L9's advisory half is still false confidence even though it reads as closed. The production SQL is atomic, so this is a test-trust gap, but if wiring the real function is awkward because of the module-global `concurrencyClient`, that awkwardness is the finding.

## Grab-bag (low)

- **[R2-L1] Fail-closed catch in `switchTeam` skips `stopAllStreams()`.** `desktop-app/src/appService.ts:2848-2862`. The happy path a few lines down calls it; the catch doesn't, so a fence failure leaves the old team's progress/activity sockets open after logout. Minor leak on a rare error path. Also a round-2 regression.
- **[R2-L2] `canceledUploads` set leaks on repeated cleanup failure.** `gfs-controller/src/upload/uploadSession.ts:1253/1283/1295` vs `1343`. The set is added to for every terminal/corrupt session inside the reconcile transaction but only deleted in the success branch; if `removeUploadDirectory` keeps failing, the session stays in the set forever. Bounded, slow growth, fencing stays correct meanwhile.
- **[R1-L4] Geometry constants duplicated across 4 modules, and the docstring is wrong.** `gfs-controller/src/upload/protocol.ts:4-6` claims "the writer and both clients import this module" but neither client imports it, they each keep a numeric copy. They haven't diverged, but the docstring is false. Still open from round 1.
- **[R1-L8] `uploadId` shape hardcoded as `[0-9a-f-]{36}` in three routing spots with no contract test.** `control-ui/app/control-api/[...path]/route.ts:137`, `external-rest-api/src/app.ts:23`, `external-rest-api/src/routes/gfs.ts:459`. Over-matches vs the strict UUID validators so nothing breaks today, but the coupling to what the writer actually mints is untested. Still open from round 1.
- A few R1 lows are still open but benign: `optionalNumber` hardening (R1-L5), `get()` not throwing on soft-expired (R1-L6), `advisoryKey()` int32 collision (R1-L10, fail-safe), the `inc(...,0)` no-op metric (R1-L11), geometry without property-based fuzz (R1-L12), and the `x-gfs-upload-source-ip` header trust (R1-L13, mitigated by the relay overwriting it).

## What checked out clean

- Security verdict is SAFE, nothing >= Medium, even though the round-2 fixes touched a lot of sensitive code (drive delegation to control-api, team-hop fencing, the DB recovery/fence machinery, admission). IDOR stays impossible (ops scoped by `upload_id AND drive AND owner_subject`), path/symlink closed (`O_CREAT|O_EXCL|O_NOFOLLOW 0600`, drive never enters a filesystem path), drive validated before minting the writer token, recovery fence fails closed, desktop persistence carries no token with legacy quarantine.
- The collateral audit ruled out the four scariest regression hypotheses: create/complete/pause/cancel idempotency is intact (reconcile-before-retry, Retry-After capped at 5s over 3 attempts), `complete()`/`reconcile()` signatures unchanged, and progress does reach 100% on completion.
- Contract tracing came back consistent end to end: `Retry-After` is seconds everywhere, `parseRetryAfter` is byte-identical in both clients, the drive contract is validated only in control-api with both clients verifying the response, the legacy/v2 seam is decided by a single consistent signal, and the file never crosses IPC (renderer passes only `filePath`).
- Suites green on the touched unit tests: gfs-controller 89 + clean tsc, control-api 90, desktop 29, control-ui 55. The real-PG suites are gated so they didn't run locally, but they read as genuine harnesses, not empty skips.

Two size notes worth a look: this is 115 files across 6 services, and four code files are past a healthy fix-commit budget (`control-ui/lib/gfsFileUpload.ts` at 6, `desktop-app/src/gfs/upload.ts` and `external-rest-api/src/routes/gfs.ts` at 5, `desktop-app/src/appService.ts` at 4). Given that R2-H1 + the two regressions all land in the same `appService.ts` fence machine, a short decision-table spec for that seam (deliberate switch vs transient hop vs logout vs control/read op on an in-flight job, and what happens to uploads in each) would probably close R2-H1, R2-M3 and R2-L1 together instead of three more patches.

zach88:
Took a full pass over round 3. Good news first: the round-2 merge gate (R2-H1) is closed and properly generalized, security came back clean, and the collateral pass found **zero regressions this round** (0/9, down from 25% in round 2). The finalizer seam, the monotonic-progress invariant, and the advisory-slot test all got real coverage. So the code isn't what's blocking here.

What's left is a handful of new Medium/Should-fix items (mostly functional gaps and test quality), plus a couple of things that are genuinely **your call**, collected at the bottom. Nothing new is a Blocker/High.

## Medium

1. **[R3-M1] Boot-time upload reconcile never runs** (`gfs-controller/src/index.ts:209,273`). The `void runCleanup?.()` this round added is a guaranteed no-op: `void cleanup()` at :209 runs synchronously up to its first `await`, sets `cleanupRunning=true`, and checks `if (uploadServiceForCleanup)` while it's still `undefined` (it gets assigned at :248), so it skips the upload reconcile; then :273 sees `cleanupRunning===true` and returns immediately. Net effect: after a crash, stuck `finalizing/uploading` sessions aren't recovered at boot, only on the first interval tick (`blobCleanupIntervalMs` later), which is exactly what that line was meant to fix. **Repro:** deterministic synchronous sequence (no `await` between :209 and :273). Fix: assign `uploadServiceForCleanup` before :209, or drop the `void cleanup()` at :209 and keep only `void runCleanup?.()` after the handler is built.

2. **[R3-M5] `listUploadSessions` is an IPC producer with no renderer consumer, so R2-M1's actual goal is still un-wired** (`desktop-app/src/appService.ts:2311`, consumer expected in `desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts:302,338`). The IPC union fix from this round is correct (`active` maps to `uploading`, types line up), but the renderer only ever learns an `uploadId` from an upload created in the live session and never calls `listUploadSessions` or rehydrates `uploadIdRef` from persisted state. So after you close and reopen the app, a persisted `active/paused/suspended_auth` upload is resumable by `resumeGfsUpload`, but the Files page never discovers its `uploadId` and never renders the row with Pause/Resume. The union fix was necessary but not sufficient for "resume after restart." Fix: a mount effect in `useGfsBrowserController` that calls `listUploadSessions(DRIVE)`, and if there's a non-terminal session, sets `uploadIdRef` and starts `waitForUpload`.

3. **[R3-M2] The "active" concurrency lease is applied to every upload route, not just part PUTs** (`control-api/src/middleware/gfsUploadAdmission.ts:212-234`). Every lifecycle request (status, pause, resume, complete, delete, HEAD) grabs and releases a `gfs-upload:active:{global|user|ip}` slot, while a part PUT holds its slot for the whole upload (up to a 10-min timeout). So a user with 8 concurrent parts in flight hits `gfsUploadMaxActivePerSubject=8`, and the 9th request, which might be the `pause`/`complete`/`delete`/`status` they need to drain or observe those uploads, gets `429 active_requests`. The client can wedge itself on its own uploads. Fix: scope the concurrency lease to part PUTs (or exempt the light lifecycle ops), leaving request/byte buckets as the gate for the rest.

4. **[R3-M4] The streaming upload proxy path only validates the `Upload-Chunk-Length` header, never the actual stream byte count** (`control-ui/app/control-api/[...path]/route.ts:190-234`). The buffered branch (:200-211) bounds the real size via `readBodyCapped(req.body, cap)`; the new `rawUpload` branch reads `declaredChunk` from the header, does the 413 fast-path, then forwards `req.body` with `duplex:'half'` without counting bytes. A client can declare `Upload-Chunk-Length: 1048576` and stream 500 MiB, and the proxy forwards it with no cap of its own, and per the file's own comment this handler runs before auth. Not a Blocker (gfsc downstream is authoritative on size and the streaming avoids the proxy's own OOM), but the proxy's 16 MiB guarantee is now a trusted header. Fix: interpose a byte-counting `TransformStream` that aborts past `16*1024*1024`, or document that the proxy delegates the cap downstream and the header is only a fast-path.

5. **[R3-M3] Switching auth/me/invitations to `createRateLimiter` changes their 429 semantics and IP keying with no dedicated coverage** (`external-rest-api/src/middleware/rateLimit.ts:8-52`). GFS v2 uses `express-rate-limit` (a different module), so this change only touches auth/me/invitations: it adds `X-RateLimit-*` to all responses, `Retry-After`+`retryAfterSeconds` to their 429s, and switches `getClientIp` from `X-Forwarded-For[0]` (spoofable) to `req.ip`. The `getClientIp` change is a legit security improvement (with `trust proxy 1`), but the bundle changes three unrelated features' 429 contract and only `routes.invitations.test.ts` was touched. **Repro:** suites green, no regression, but the new headers on auth/me are unasserted. Fix: split the `getClientIp` fix into its own commit with a test, or justify the blast radius in the body and assert the new headers.

6. **[R3-M6] "An explicit resume must not fall through to the legacy fallback" has no test (T5)** (`desktop-app/src/appService.ts:2154,2192`). The guard `if (!(error instanceof DesktopUploadCapabilityError) || resumeUploadId) throw error` stops a resume that hits a non-v2 writer from creating a duplicate non-resumable resource. The path is reachable in prod (`ipc.ts:465-473` / `preload.ts:96-103` propagate `resumeUploadId`), but nothing exercises it: drop `|| resumeUploadId` and the suite stays green while a resume against a legacy writer creates a second resource. **Repro:** code read plus grep-confirmed absence of a test. Fix: a test that forces `DesktopUploadCapabilityError` with `resumeUploadId` present and asserts `createResource`/`replaceFile` was not called.

7. **[R3-M7] The R2-M1 test injects a fixture the real producer can't emit on that branch (T1)** (`desktop-app/src/__tests__/appService.gfs.test.ts:366-387`). It mocks `readDesktopGfsUploadState` to return `status:'active'`, but the real producer reconciles any orphaned `active` to `suspended_auth` (`appService.ts:1978-1980`). The production path where `record.status==='active'` reaches the mapping (:2221) is the pending-job branch (`hasLiveDesktopGfsUpload` true, `entry` absent), not what the test simulates. The mapping is correct and reachable; the test just cuts the reachability with a hand-built fixture. Fix: derive the `active` state from the real producer by mounting a pending job.

## Smaller stuff

8. **[R3-L1] An orphaned `finalPath` part can block retry after an inconclusive reconciliation** (`gfs-controller/src/upload/uploadSession.ts:1005-1016`). If the transactional `commit` throws and `reconcileCommittedPart` also throws (readback I/O), in the sub-case where the COMMIT didn't actually land you get an orphaned `<n>.part` with the part marked `failed`: neither the `stale` loop (WHERE reserved) nor the `committed` loop cleans it, and a PUT retry hits `stat(finalPath)` at :938 with `conflict: committed part file already exists`. Double-fault, rare, degrades to a stuck upload until expiry rather than corruption.

9. **[R3-L2] `advisoryKey` is 32-bit with an XOR-folded slot (`low ^ slot`), new logic this PR** (`control-api/src/services/rateLimiterService.ts:184-293`). Two buckets that collide on the high 32 bits of the sha256 and have overlapping slot sets share advisory-lock identity. The direction is fail-closed (it could deny a bucket that had capacity, never over-admit) and needs a 32-bit collision, so it's Low, but worth either documenting the trade-off or using full 64 bits.

Grab-bag nits (only if you're already in the file): dead `errorCode` option in `external-rest-api/src/middleware/rateLimit.ts:15,17,44`; redundant `setHeader` in `control-api/src/routes/gfs/proxy.ts:156-163`; stale "any 5xx" comment in `desktop-app/src/gfs/upload.ts:614-616` (the set was narrowed to `{408,500,502,503,504}`); `createOversizedDiskUploadFixture` duplicates `createDiskUploadFixture` in `tests/e2e/gfsUploadV2Fixtures.ts:53-70`; the hop test doesn't call `listGfsUploadSessions` with `hopDepth>0` so R2-H1's read-only sibling has no executable regression; `retryAfterOf` isn't byte-identical across the two clients (desktop has a JSON-body fallback, benign since the header is always set).

## Still open from earlier rounds

- **[R2-L1] The fail-closed catch in `switchTeam` doesn't call `stopAllStreams()`** (`desktop-app/src/appService.ts:2902-2913`, twin in `switchSessionToTeam:1038-1048`). Happy path calls it at :2916; the catch (fence-persist fails, forced logout) leaves the old team's sockets alive. Still open since round 2, Low.
- **R4, body evidence is anchored to a head that isn't in this branch.** All the T0/T1 and "CodeQL green" evidence points at `3e9d6e01` and remediation commits `ba859fb5`/`dacd47e6`/`4bc1484d`, and none of them are ancestors of the current head `60e51561` (the round-3 rebase rewrote them again, with ~15 commits on top that touch that very evidence). **Repro:** `git merge-base --is-ancestor 3e9d6e01 60e51561` returns non-zero. This has been open since round 2. Please re-anchor the evidence to `60e51561`, and sync the "monotonic bar" line with the doc, which is already reconciled to "may move downward" on a failed retry.

## Decisions for you

These aren't things I can just fix, they need your call:

1. **[R2-M3] Deliberate `switchTeam` now forces logout if the GFS fence write fails** (`desktop-app/src/appService.ts:2895-2914`). A transient failure (disk full, permissions on the GFS state file) after the replacement token is already installed forces `clearAuthenticatedSessionState`+`clearSessionToken`, so the user re-logs-in. It's intentional fail-closed and well tested (T4, `appService.rpc.test.ts:498`), and it only applies to the deliberate switch, not transient hops. **This has been open two rounds, so it can't just float into a third:** either it escalates to Blocker or it gets closed on the record. My recommendation is to close it on the record (you accept the fail-closed contract). Your signature either way.

2. **Process, F4 + D3: `7d13aa43` is labeled `test(gfs):` but carries production logic across 15+ files** (the `appService.ts` security fencing, the new `uploadFinalizer.ts` with an added epoch revalidation, `rateLimiterService.ts`, `protocol.ts`, `index.ts`). It's already merged into the PR, so reverting it isn't practical this late. `appService.ts` is now on its 5th patch touch (4 `fix:` plus this disguised `test:`), and that team-fence machine is the piece that keeps generating findings round after round (R1-H2 to R2-H1 to R2-M3/L1). My recommendation: note in the body that this commit contains production, and write the retroactive mini-spec for the `appService.ts` team-fence (a decision table: for each op, gated vs read-only vs dispatch, and what happens on transient hop vs deliberate switch vs logout vs env change), with `uploadFinalizer.ts` covered by it. That closes the pattern instead of patching it a 6th time.

3. **Whether to do a fix round at all vs merge with follow-ups.** Round 4 trips the stop rule, so if you want these seven Mediums closed, my recommendation is a single tight fix round (one commit per ledger ID, with Class-check + Repro-test), not another full review round, or move the less urgent ones to issues. If you'd rather merge and follow up, R3-M5 (restart-resume still doesn't reach the user despite being marked resolved) and R3-M1 (boot recovery is a no-op) are the two I'd least want to ship open.

## What checked out clean (verified, not assumed)

- R2-H1 resolved and generalized: the 5 control/read routes no longer gate on transient hops, new-job creation still does, security fence intact per the collateral pass.
- Finalize seam (R2-M4) actually runs now, with a fail-closed epoch revalidation and real-PG coverage of the producer.
- State machine idempotency (complete/pause/cancel) intact; the new ambiguous-part reconcile doesn't break it.
- All four end-to-end contracts line up: canonical drive, IPC state union, 429/`public_edge_requests` retry, geometry constants.
- The body's `express-rate-limit@^8.5.1` fast-reject claim, which was false in round 2, is now true and wired to the 9 GFS routes.
- Suites the reviewers ran: gfs-controller 97, control-api 92+51, external-rest 117, desktop 97, control-ui 62, host-context-controller 27, geometry contract PASS. Real-PG suites are gated and were read as real harnesses, not run.

zach88:
Took a deep pass through the exact head (`0db14195`) across all five services. Good news first: this round closed all seven round-3 mediums cleanly, with zero regressions from the fixes, security came back clean, and the cross-service contracts line up end to end (the IPC-union `active` leak from R2/R3 is verified fixed). The catch is that patching the upload client one more time surfaced a genuine new correctness bug in the part-reconciliation path, plus a couple of test-coverage gaps that are hiding real behavior. Details below.

Heads up on process: this is round 4, and `gfsFileUpload.ts` is now at 6 `fix:` commits (double the budget). The right move for the top finding isn't another patch on that file, it's a small design note first, because the reconciliation logic never had a decision table to review.

## High

**1. [R4-H1] Ambiguous part failure fails the whole upload under concurrency** — `control-ui/lib/gfsFileUpload.ts:587-599`, `:699-705`

With the default 4-part concurrency, an ambiguous but ordinarily-retryable part failure (502/503/504/timeout) escalates to `upload_part_outcome_unknown` and rejects the entire upload instead of just re-uploading that part. The only path back to "part is definitely missing, replay it" is `if (activePartCount === 0) return 'missing'` at `:595`, but `activePartCount` is per-session, not per-part (the writer bumps `active_part_count` per reserved part, `gfs-controller/src/upload/uploadSession.ts:842`). So while any sibling part is still in flight, `activePartCount > 0`, reconciliation never reaches `'missing'`, it retries ~3 times (~750ms) and throws a non-retryable error. It also runs before the normal exponential backoff, so the per-part retry budget is bypassed for ambiguous states, and the 4→2 fallback can be skipped entirely because the unknown-outcome aborts first.

Reproduced: 8-byte file / 4 parts, statuses `[204,204,204,502,204]`, `/status` returning `activePartCount:3, parts:[]`, and `start()` rejects with "One or more upload parts could not be committed." instead of resolving `completed`. For real large files (16 MiB parts taking seconds), siblings are reliably active during the reconcile window, and this fires exactly when the backend is unstable, which is the scenario the whole 4→2 fallback exists for. It fails safe (persisted resume recovers, no corruption), but it defeats the concurrent resilience that's the point of this feature.

Suggested direction: this needs a design decision before code, not another patch on this file. The core question is whether the writer should expose active state per-part instead of a session-wide counter. Its test twin is finding 4 below.

**2. [R4-H2] The reconcile/contiguous SQL is only tested against a hand-rolled in-memory DB** — `gfs-controller/src/upload/uploadSession.test.ts:19-319` vs `uploadSession.ts:962-988`, `:1216-1327`

The heart of this PR (the contiguous-bytes subquery with `generate_series` + `NOT EXISTS`, and the whole `reconcile()` family) is verified only against a `MemoryDb` that reimplements the same logic in JS. The assertion `contiguousBytes === 2097152` (`:507`) certifies the MemoryDb, not the SQL. The reconcile queries return `{rows:[]}` in the fake, so those tests only assert `{staleParts:0, expiredSessions:0}` and never exercise the logic. The fake also diverges from the real contract (it ignores the `lease_epoch` filter the real SQL applies at `:1040`). Net effect: a parser or semantics bug in the real SQL sails through green.

The real-Postgres harness already exists and even documents this exact risk in its header comment (`gfs-controller/src/db/renamePublication.realPostgres.integration.test.ts`), so a `uploadSession.realPostgres.integration.test.ts` following that pattern is the fix, and it's not awkward to build.

## Medium

**3. [R4-M1] Orphaned `finalizing` session has no recovery path** — `gfs-controller/src/upload/uploadSession.ts:1262-1268`, `:1387-1408`

If the writer (a singleton) crashes during the finalize window after `complete()` durably marks the session `finalizing` but before `onPublished` commits, restart leaves the session stuck `finalizing` with `inFlightFinalizers` empty. The client's retry of `complete` gets `upload_finalizing`, and `cancel` throws `upload_finalizing`, so the session is wedged until the 24h TTL. The expiry query only looks at `expires_at <= now()`, and nothing in `reconcile()` reclaims an orphaned `finalizing`. This partially misses the "writer reconciliation" invariant the description promises. Suggested fix: have `reconcile()` return a `finalizing` session with no in-flight finalizer and parts still present back to `uploading`/`paused` with `session_epoch+1` (audit the collateral so a legitimately-running finalizer in the same process isn't reactivated, which the epoch bump covers).

**4. [R4-M2] The R3-M6 fix covers the create branch but not the replace branch** — `desktop-app/src/appService.ts:2188`

The "explicit resume never falls through to the legacy fallback" fix landed on both `startGfsFileUpload` and `startGfsFileReplace`, but only the create branch has a test. Reproduced by mutation: dropping `|| resumeUploadId` on `startGfsFileReplace` keeps `appService.gfs.test.ts` green (20/20), while the same mutation on create fails. So `startGfsFileReplace(resourceId, path, drive, ifMatch, 'resume-id')` with v2 capability down would create a second non-resumable legacy resource with no test catching it. Add the twin case asserting `rejects DesktopUploadCapabilityError` and `replaceFile` not called.

**5. [R4-M3] The reconcile tests hand-write `activePartCount:0`, masking finding 1** — `control-ui/lib/__tests__/gfsFileUpload.test.ts:547`, `:600`, `:693`, `:754`, `:867-965`

Every reconcile/fallback test writes the `/status` fixture by hand with `activePartCount:0` (and mostly `partCount:1`), so the "missing → replay" path is only ever exercised with no active siblings, which is exactly the condition that hides R4-H1. The real producer emits `activePartCount > 0` under concurrency. This is the test face of finding 1; they close together, with a fixture derived from the real producer (siblings active) and, ideally, a property-based test for the pure progress reconciliation.

## Smaller stuff

- **[R4-L1] Writer 429 doesn't match the edge/admission 429 shape** — `gfs-controller/src/upload/uploadSession.ts:796`, `:798`. The edge and admission limiters emit `{error:'gfs_upload_rate_limited', limit, retryAfterSeconds}` plus a `Retry-After` header; the writer's own concurrency 429 emits the ok-envelope `quota_exceeded` with no `Retry-After` and no `limit`. The client treats all 429s as retryable so nothing breaks, but the "stable, distinguishable public 429" claim is only partly realized.
- **[R4-L2] `/me/gfs/capabilities` now requires `?drive=`** — `control-api/src/routes/external/gfs.ts:280-287`. Capabilities are drive-independent, but the route is now behind `requireCanonicalUploadDrive`, so a caller omitting `?drive=` gets `400 drive_required`. Both shipping clients always send one, but any future non-drive-scoped probe now 400s on a global-geometry endpoint.
- **[R4-L3] Drag-drop upload can't resume** — `control-ui/components/GfsBrowser.tsx:316-386` vs `:416-506`. Two near-duplicate upload entry points; the drag-drop path doesn't compute `resumeUploadId`, so a dragged upload can't resume a persisted session while the modal path can. Worth extracting a shared helper.
- Carryover lows still open since round 1, all benign: dead `preferredChunkBytes`/`maxChunkBytes` capability fields (`gfsFileUpload.ts:133-134`), `get()`/HEAD not calling `assertNotExpired` (`uploadSession.ts:654-665`), non-UUID uploadId falling through to the JSON parser for a different error code (`external-rest-api/src/app.ts:23`), and geometry constants still duplicated across 5 modules (values agree today).

## What checked out clean

- No regressions from any round-4 fix.
- Security is clean: 200 MiB ceiling rejected before session allocation (hardcoded, config can't raise it), sum-of-parts closed at `complete`, drive delegation with no bypass (the writer takes drive from the token claim, not body/query; cross-drive access 404s), admission fails closed when Postgres is down, source-IP header only trusted from the internal relay, no tokens persisted to disk, TOCTOU closed via `O_NOFOLLOW`+`fstat`+`assertSourceIdentity`, and least-privilege DB grants.
- All eight upload-v2 capabilities are contract-consistent across both the Desktop and Control UI planes, and the IPC-union `active` regression from R2/R3 is verified closed.
- All seven round-3 mediums (R3-M1 through M7) are closed and verified, several by mutation testing. R2-M3 stays closed-by-write (intentional fail-closed, T4 test still solid).
- Suites green in the touched areas: gfs-controller 518/518, control-api subset 92/95 (3 real-PG skips), external-rest 118/118, desktop 99/99, control-ui 39/39. Typecheck clean where run.

One process note worth surfacing: this is round 4 with 128 files across 6 services, and control-api/external-rest and gfs-controller have both converged (no correctness findings in control-api/external-rest this round). Those are good candidates to split out and merge on their own, leaving the desktop/control-ui reconciliation piece, where R4-H1 lives, as the part to redesign.

alfredolopez80:
Thanks for the detailed round-4 review.

Validated on **4141cb20**.

### Resolved

- **R4-H1 — fixed.** Ambiguous part reconciliation now uses the exact active part-number lease set, not only the session-wide counter. A missing part can be replayed while sibling parts remain active; the normal retry/fallback path is preserved.
- **R4-H2 — fixed.** Added gated real-PostgreSQL upload-session coverage for concurrent reservations, out-of-order and non-zero completion, idempotent completion, ambiguous part/complete commit outcomes, active lease snapshots, and stale-lease cleanup.
- **R4-M1 — fixed.** reconcile() reclaims an orphaned finalizing session only when no in-process finalizer owns it, increments session_epoch, and safely reopens the session.
- **R4-M2 — fixed.** Added the explicit-resume regression test for startGfsFileReplace; it rejects the capability mismatch and never falls back to a legacy replace.
- **R4-M3 — fixed.** The UI reconciliation fixtures now model active sibling leases and assert the four-way concurrency path.
- **R4-L1 — fixed.** Writer quota responses now carry the same limit plus retryAfterSeconds envelope and Retry-After header contract as the edge/admission paths; relay tests preserve both.
- **R4-L2 — clarified/fixed.** Capabilities are a global probe and accept an omitted drive; upload mutations still require the canonical drive.
- **R4-L3 — fixed.** Drag-and-drop now reuses the matching persisted resumable session, with regression coverage.

### Follow-ups

The remaining carryover lows — dead capability aliases, expiry-check and error-taxonomy cleanup, and duplicated constants — are non-blocking follow-ups and are intentionally outside this upload correctness change.

### Validation

CodeQL, the full CI matrix, and Real PostgreSQL integration are green.

zach88:
Took a focused pass over the round-5 delta on this one (isolated the 19 new commits with `range-diff` so the rebase noise stays out). Short version: this round actually broke the loop instead of feeding it. Every open finding from round 4, including both Highs, is now closed and verified against the real code, with tests derived from the real producer. Zero blockers left. I found one new Low regression and a couple of nits, nothing that has to hold up the merge.

The R4-H1 fix is the highlight: it's a real model-level correction (reconcile decides per-part off `activePartNumbers`, not off the session-wide `activePartCount`), not another patch on the symptom. Verified by mutation: reverting to the old model turns the concurrent tests red.

## Low

1. **[R5-L1] `finalizing` sessions with a null start timestamp can get stuck forever**
   `gfs-controller/src/upload/uploadSession.ts:1362-1368` (the `expired` reconcile query) and `:1333-1341` (`staleFinalizers`). Commit `a6915e78` dropped `'finalizing'` from the `expired` query and now leans entirely on `staleFinalizers`, which filters `finalizing_started_at IS NOT NULL`. So a session in `finalizing` with a null `finalizing_started_at` is touched by neither path: not `expired` (excluded) and not `staleFinalizers` (requires the column), and it hangs permanently with orphaned parts and staging dir. Before this round it got expired once it passed `expires_at`.
   Reachable in the rolling-deploy window: an old replica transitions to `finalizing` without setting that column, and migration `0093` backfills it as null on existing rows. In steady state the new `finalize()` always sets it, so it doesn't recur, hence Low. Confirmed by reading the head; not reproduced against Postgres (needs a version-skew setup).
   Fix is tiny: put `'finalizing'` back in the `expired` query, or handle the `IS NULL` case in `staleFinalizers`.

## Grab-bag (nits)

2. **[R5-N1] Button-path resume matching is still inline**
   `control-ui/components/GfsBrowser.tsx:451-461`. Drag-drop and replace now share the `matchingPendingResumeUploadId` helper, but the button path (`uploadFile`) keeps its own inline matching. Harmless today since that path is always `create`, but it can drift from the helper. Worth folding into the same helper.

3. **[R5-N2] `X-RateLimit-Limit` carries two different shapes**
   The writer's 429 puts a string label in `X-RateLimit-Limit` (e.g. `active_sessions_subject`), while the relay's own edge 429 (`external-rest-api/src/routes/gfs.ts:99`) puts a number. Same header, inconsistent semantics. Nothing parses it numerically today so it doesn't break anyone, but it's a trap for a future consumer.

## One process note

`a6915e78 "test(gfs): close reclaim and resume evidence gaps"` isn't only tests: it changes production semantics in `uploadSession.ts` (the `expired`-query change above, plus the `ownsFailureCleanup` fence and the epoch-guarded cleanup). The last two are legit and check out clean, but burying behavior changes under a `test(...)` subject makes the collateral pass almost skip them. Worth re-labeling or splitting next time.

## What checked out clean

- **R4-H1** (concurrent upload failing on a single ambiguous part): closed at the model level, reproduced by mutation. Concurrent upload no longer dies on a 502/500 for one part while siblings are active.
- **R4-H2** (reconcile SQL only tested against a hand-rolled `MemoryDb`): now exercised against real Postgres (`control-api/test/gfsUploadSession.realPostgres.integration.test.ts`), with CI enforcing the lane via `CONTROL_API_REAL_PG_REQUIRED=1` plus an anti-skip guard.
- **R4-M1** (`finalizing` orphan with no reconcile path): closed via the `staleFinalizers` block (`session_epoch+1` reopen). Note the recovery latency is `finalizeTimeoutMs` (default 24h), which is a design tradeoff, not a bug. See R5-L1 for the null-timestamp gap that the same commit opened.
- **R4-M2** (replace branch untested): now covered, verified by mutation (mutating only the replace guard reds exactly the replace test).
- **R4-M3** (hand-written `activePartCount:0` fixtures masking R4-H1): fixtures now emit `activePartNumbers` in the real producer's shape and fail against the old model.
- **R4-L1** (non-uniform 429): writer now attaches `retryAfterSeconds`/`limit`, `serve.ts` sets the headers, and relay/control-api forward them; `x-internal-secret` confirmed not crossing the edge.
- **R4-L2** (`/external/gfs/capabilities` requiring `?drive=`): drive is optional again, global probe works with no query params.
- **R4-L3** (drag-drop not resuming): shared helper wired into replace and folder-drop, covered by integration tests.
- Suites run green locally: gfs-controller 57/57 (`uploadSession`+`serve`), external-rest-api 24/24 (`routes.gfs`+`app`+`middleware.rateLimit`), control-ui 25/25 (`gfsFileUpload`) + 30/30 (`GfsBrowser`), desktop-app 22/22 (`appService.gfs`). Real-PG suites validated by reading + the CI wiring (they need Postgres).

Nothing carried over: all eight R4 findings are closed, none survived three rounds. `R5-L1` is the only thing still open. My take: close it as a one-commit fix (or a follow-up issue since it's upgrade-window only) and ship, no need for a round 6.

zach88 (approved):

zach88:
Took a focused pass over what actually changed since the last review round. After the rebase onto dev, the only genuinely new work here is two commits (`44c4dc7a` recover expired finalizing uploads, `d0ba9532` build wiring for the shared display-field package); everything else in the range is the unrelated display-name feature that already landed on dev. Ran a collateral audit scoped to that real delta with a single question: what did these commits break? Answer: nothing. This is good to merge from my side, the remaining items below are all non-blocking follow-ups.

## Closed this round

1. **[R5-L1] Orphaned `finalizing` session hung forever** (`gfs-controller/src/upload/uploadSession.ts:1367-1394`). The regression from `a6915e78` is fixed: the `expired` reconcile query now re-includes `state='finalizing' AND finalizing_started_at IS NULL`, so a stale row (dead replica during a rolling deploy, or a NULL row from migration 0093) gets picked up and expired. The new `if (inFlightFinalizers.has(uploadId)) continue` guard does not reintroduce the hang and does not self-cancel a live finalizer: the local finalizer writes `finalizing_started_at = now()` (NOT NULL) inside the `start` transaction before it registers in the map, so any row the guard skips already fails the `IS NULL` predicate. Reproduced: the new assertions in `uploadSession.test.ts:779` and the real-Postgres integration at `:839` fail against the previous head and pass now.

2. **[R5-N2] `X-RateLimit-Limit` mismatched string vs number** (`gfs-controller/src/api/serve.ts`, `errors.ts`). gfs-controller now emits a numeric limit and moves the scope label to a new `X-GFS-RateLimit-Scope` header, matching what control-api and external-rest-api already sent. `quotaExceeded` grew a required numeric third arg and all 5 call sites pass it, so a miss would fail typecheck. No consumer parses the header, so nothing downstream breaks.

3. **[R5-N1] Duplicated inline resume matching** (`control-ui/components/GfsBrowser.tsx:452`). The button path now reuses `matchingPendingResumeUploadId` like drag-drop already did. Behavior is unchanged on the create path.

## Follow-ups (non-blocking, for later PRs not this one)

These are structural, not defects in the diff, and none of them should hold the merge:

1. **PR size / scope (D1).** 142 files across 6+ services in one PR means every review round re-scanned the whole surface. Worth splitting future work on this piece (client vs server vs deploy) into separate PRs.
2. **Fix churn on hot files (F3).** A few production files (`gfsFileUpload.ts`, `uploadSession.ts`, the external `gfs.ts`) accumulated more than three `fix:` commits over the review history. Not a bug today, but a signal for the retro that a couple of these areas deserve a small design note before the next change rather than another patch.

## What checked out clean

- The `quotaExceeded` signature change, the `toResponse` return shape, and the `X-RateLimit-Limit` contract, all callers migrated, no consumer parsing the header.
- No resume UI path lost (button, drag-drop, replace).
- Build commit is safe: only control-ui and control-api depend on `@clerum/display-field`, both Dockerfiles copy it before `npm ci`, the package is prebuilt and dependency-free, and `.dockerignore` whitelists it additively. It actually closes a build gap dev had opened.
- Collateral audit of the two new commits: 0 regressions.

No open findings left on my end. Good to merge whenever the required approval is in.

Issue #312, opened 2026-08-11
Updating a CommunicationChannel can add a provider without its credentials


jozer-rami:
## Summary

Creating a `CommunicationChannel` enforces that a provider (telegram / slack / teams / email) comes with the credentials it needs. Updating one does not. You can add a second provider to an existing channel and the update is accepted, because the channel already carries a `credentialsSecretRef` for the *first* provider and nothing checks that the referenced Secret holds the new provider's keys.

The result is a channel that advertises a provider it cannot serve. Nothing fails at save time. The failure surfaces much later, at the provider's first inbound request.

## Impact

Found while wiring a Slack app to a live deployment. A Telegram channel had a Slack App Name added to it in the Control UI. The update saved cleanly, and the channel then:

1. Started appearing as a **Slack approval target** in the Profile UI, labelled identically to the real Slack channel on the same host, with no way for a user to tell them apart.
2. Started rendering a **Slack Request URL** on its edit page, which is easy to copy by mistake.
3. Returned `409 slack_signing_secret_missing` when Slack called that URL, because the referenced Secret only contained `telegram-bot-token`.

Slack reports this to the operator as "Your request URL responded with an HTTP error / Your URL didn't respond with the value of the challenge parameter", which gives no hint that the real cause is a credential-less provider on the wrong channel. Diagnosing it needs cluster log access.

## Reproduction

1. Create a Telegram `CommunicationChannel` with a valid `telegram-bot-token`. It gets `credentialsSecretRef: cc-<name>-credentials`.
2. Edit the channel in the Control UI and set **Slack App Name** to any value. Do not supply Slack credentials.
3. Save. The update is accepted (HTTP 200).
4. Observe `spec.slackSettings.botHandle` on the resource, with a Secret that has no `slack-bot-token` / `slack-signing-secret`.
5. Point a Slack app's Event Subscriptions at that channel's Request URL. Slack's `url_verification` gets a 409.

## Root cause

`control-api/src/routes/admin/resources.ts`

- **POST** enforces the coupling:
  - `:323` rejects a spec with any provider unless a `credentials` envelope or a `credentialsSecretRef` is present.
  - `:340` calls `missingCreateCredentialKey`, which requires the provider-specific key (`slack-bot-token` for a Slack provider, and so on).
- **PUT** (`:450`) runs only `validateCommunicationChannelSpec`, a shape check. There is no `ccSpecHasProvider` guard and no `missingCreateCredentialKey` call.

[...]

PR #313, opened 2026-08-11
docs: add a how-to for connecting Slack

jozer-rami:
## What

Adds `docs/how-to/connect-slack.md`, the missing counterpart to `connect-telegram.md`, plus index entries and a cross-link from the approvals guide.

Slack was documented only as a set of CRD fields and a name in the approvals table. Nothing explained how to get a Slack app talking to an agent, and the Telegram guide is not a usable template because the two providers work differently.

## Why it is not a copy of the Telegram guide

Telegram is polled and needs nothing inbound. Slack is pushed:

- `channel-reader/src/main.ts:1321-1328` explicitly skips Slack polling. `fetchMessages` is only called for Telegram (`:1299`) and email (`:1316`).
- Events arrive via the Slack Events API on `webhook-proxy`, go to `workflow-approval-request-reader`, and reach `channel-reader` over `/internal/slack/handoff`.

So the Request URL is required for **chat and for the setup code**, not just for approval buttons. That inverts the setup order relative to Telegram, and it means a laptop-only minikube needs a tunnel before anything works at all.

The env-var path from the Telegram guide also does not apply: `CLERUM_SLACK_BOT_TOKEN` is read only by the dev-mode resolver and ignored in-cluster.

## Verified live

Every claim was checked against the code, and the flow was run end to end on a real deployment: app creation, channel creation, Request URL, `verify` linking, chat, and tool approvals including `Always approve`.

## Things the guide calls out because they cost real time

- The Request URL is **per channel** and its tail base64-decodes to the channel name. Copying from the wrong channel's page yields `409 slack_signing_secret_missing`, which names no channel. The guide includes a decode one-liner.
- Slack's **Retry re-sends to the stored URL**, so retrying without editing the field loops forever on the old value.
- Chat is gated on a linked identity. Before verification, messages are dropped with `Ignoring unauthorized slack message`, which presents as a dead bot.
- Tool approvals post **three** buttons (Approve / Always approve / Deny), not two, and require Interactivity enabled with the same Request URL. If it is unset Slack blocks the click client-side, so every cluster log stays silent.
- The approval token is in-memory with a 10-minute TTL, and saving a channel edit in the Control UI restarts `channel-reader` and invalidates it.
- An unanswered approval suspends that conversation and queues later messages behind it. Scope is one session, not the host.

Troubleshooting leads with the two `kubectl` commands that expose all of this, then keys rows on the actual status codes.

## Follow-ups, not in this PR

- #312, provider-without-credentials accepted on update. The guide warns readers away from the trap in the meantime.
- `resolveSlackTarget` returns 409 rather than 404 for a channel with no Slack provider, and the Control UI renders a Slack Request URL for such channels. That combination is what made the wrong-URL mistake so easy.
- `slack[]` items are missing `confirmedByUserId` / `confirmedAt` in the CRD schema, so control-api writes them and Kubernetes prunes them. `telegram[]` has both.
- `channel-reader/test/asyncProgressApprovalFlow.test.ts:396` only asserts the Slack buttons render, while the Teams case at `:474` round-trips a decision. The seam that broke is the untested one.

## Notes for review

- Based on `dev`.
- Docs only, no code paths touched.
- `docs/` is outside the repo's prettier targets, so formatting-only churn was deliberately kept out of `configure-approvals.md`; that file gets a two-line cross-link and nothing else.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -48,6 +48,7 @@ Use this index for long-form docs.
 | Doc                                                                           | Description                                                |
 | ----------------------------------------------------------------------------- | ---------------------------------------------------------- |
 | [Connect Telegram](how-to/connect-telegram.md)                                | Minikube env or CRD channel path                           |
+| [Connect Slack](how-to/connect-slack.md)                                      | Slack app scopes, request URL, conversation verification   |
 | [Configure approvals](how-to/configure-approvals.md)                          | Human-in-the-loop tool gates                               |
 | [Add an MCP server](how-to/add-mcp-server.md)                                 | Dev-mode JSON or CRD allowlist path                        |
 | [LLM providers overview](llm-providers/README.md)                             | Which providers are supported and how they fit together    |
--- a/docs/how-to/configure-approvals.md
+++ b/docs/how-to/configure-approvals.md
@@ -104,7 +104,8 @@ stay approval-required by default regardless of what is set here. See
 | Control plane APIs | Automation and integration tests |
 
 Channel callbacks are signature-verified; unauthorized actors cannot mint
-approvals by replaying chat.
+approvals by replaying chat. Wiring a chat surface up:
+[Connect Telegram](connect-telegram.md), [Connect Slack](connect-slack.md).
 
 ## Verify
 
--- a/docs/how-to/connect-slack.md
+++ b/docs/how-to/connect-slack.md
@@ -10,7 +10,12 @@ as an Events API callback on `webhook-proxy`, which hands it to
 able to reach your cluster before anything works, including the setup code you
 send to link your account.
 
-## Create the Slack app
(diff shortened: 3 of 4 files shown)


jozer-rami:
Superseded by #343. Both commits from this branch (`a07c0dff`, `52754ece`) are merged into `fix/slack-manifest-panel-size` and verified as ancestors of its head — nothing here is lost.

Folded in deliberately rather than landed separately: this how-to was written on 2026-08-11, before #320, #321, #323 and #331 shipped, and it had gone stale describing a flow the product no longer has — build the Slack app first, fetch the Request URL after. Its corrections are only *correct* alongside the code changes in #343 (manifest on the create page, the `app_home` DM fix, the Settings/Features credential paths), so reviewing them apart is what let this drift in the first place.

PR #314, opened 2026-08-11
fix(control-ui): sync package-lock for sharp @emnapi deps

miguelo-999:
Regenerate lockfile so npm ci succeeds in the Docker build; the previous lockfile was missing @emnapi/{core,runtime,wasi-threads} entries required by sharp@0.35.3.


miguelo-999:
@claude bro check and provide greenlight if everything is solid

miguelo-999:
@josue-tech-94 done

josue-tech-94 (approved):

PR #317, opened 2026-08-12
feat(control-api): mcp-server OAuth broker base — adapters, PKCE, grant-presence gate, shared identity

zach88:
## What

Lands the **control-api base** for OAuth-authorized mcp-servers (v1), sequenced as four units on one service:

- **U1** — data model for both identity flavors (`oauth_grants` gains `owner_kind`, `context_id`, `bootstrapped_by_user_id`; `grant_kind` gains `shared`; partial unique index + CHECKs), the `spec.oauth` block on the `McpServer` CRD (immutable `grantScope`/`id`/`provider` via CEL), and the broker endpoint `POST /api/v1/mcp-oauth/user-token` (gated by the mcp-host control-JWT + `oauth:user-token` scope).
- **U2** — three known-shape provider adapters: **monday**, **clickup**, **vercel**. PKCE (S256) for monday+vercel; clickup is non-PKCE, non-expiring, no refresh.
- **U3 (control-api)** — opens the rpc-proxy direct-rail resolver (`filterInvocable`) to admit `oauth`, plus a per-flavor grant-presence gate as defense-in-depth for that rail.
- **U6 (control-api)** — exercises the shared-identity (`grantScope='context'`) branch end-to-end across store/endpoint/gate.

## Why

mcp-servers that require OAuth (monday, clickup, vercel, google-workspace, office365, ...) need a per-user (and optionally per-team shared) identity brokered by control-api, so the user token is never mounted in the mcp-server pod and token exchange/refresh stays on the control plane. This PR is the control-api half of that design; the mcp-host runtime (U4) and desktop consent (U5) follow on their own branches.

## How (key decisions)

- **PKCE without server-side state.** The `code_verifier` is neither stored nor randomly generated: it is derived deterministically as `base64url(HMAC-SHA256(stateSecret, "pkce-verifier:" + state))`, recomputed at authorize (for the challenge) and at callback (for the token exchange). This preserves the existing stateless OAuth flow and keeps the five legacy recipe adapters byte-identical. Defeating it requires leaking both `stateSecret` and `client_secret` -- the same bar as a random verifier in a session store. `code_challenge_method=S256` is always explicit.
- **Isolation lives in JIT, not in the catalog.** Opening the auth-type allowlist admits OAuth servers on both rails by design. The rpc-proxy direct rail gets a grant-presence gate (fail-closed, keyed per flavor: `user`->`(server, callerUserId)`, `context`->`(server, server.spec.contextRef)` decoupled from the caller). The LLM-catalog rail is intentionally left ungated -- its isolation is the JIT token resolution shipping in U4.
- **Shared identity keyed by the server's authoritative Context.** The `context` flavor always keys the grant by `server.spec.contextRef`, never a request/body value, closing the cross-context token-theft class. Bootstrap is first-wins (`INSERT ... ON CONFLICT DO NOTHING`); refresh is an `UPDATE` that never rewrites the tokens' bootstrapper or `bootstrapped_by_user_id`. A CHECK enforces `grant_kind='shared' => bootstrapped_by_user_id NOT NULL`.
- **F2** -- the three duplicated `KNOWN_PROVIDERS` sets are collapsed into one exported set derived from `Object.keys(ADAPTERS)`, eliminating the enum-drift class.

## Testing

- `control-api` build (tsc + artifact check) and full suite green: **3960 tests passed** at head.
- PKCE: round-trip test from the real `signOAuthState` (challenge at authorize == SHA256(verifier re-derived at callback)); byte-identical golden for the five legacy adapters; RFC 7636 B.1 vector + property tests.
- Grant gate: `fast-check` property test of the pure `filterInvocable` (no-grant not invocable; allowlist never admits bearer/basic/apiKey; context servable IFF shared grant, decoupled from caller).
- Shared identity: endpoint `context` branch happy-path + `no_grant` + `server_missing_context`; store first-wins/refresh-preserves-bootstrapper; audit-integrity CHECK.
- **Note:** the real-Postgres integration tests (store race/CHECK and the grant-gate context branch against a live DB) are gated on `CONTROL_API_REAL_PG_ADMIN_URL` and are skipped in default CI; they run when that env is set.

## Risks / follow-ups

- **Deployment ordering:** after this PR the LLM catalog admits OAuth mcp-server *names*, but the runtime isolation (JIT fail-closed token resolution) ships in U4 (mcp-host). **Do not enable OAuth mcp-servers in production until U4 lands.** This is now enforced code-level: the broker endpoint is gated behind `MCP_OAUTH_BROKER_ENABLED` (default **off**) and returns `404` before auth when disabled (see Review round 1 below). The rpc-proxy gate here returns names only (no token/URL) and protects that rail alone.
- **Follow-ups on other branches:** U4/U6 mcp-host runtime; U5 desktop consent + the shared-grant revocation endpoint; the control-ui OAuth registration form (the `McpServer` create form has no OAuth UI yet).
- **Recipe scope:** the collapsed provider set means the recipe OAuth path would accept the three new providers too; the `WorkflowRecipe` CRD enum stays at five so recipe admission remains fail-closed. Defense-in-depth only.
- **D1:** 44 files, but one code service (control-api) -- the units are sequenced parts of one spec, coupled on U1's data model/endpoint, so they ship together rather than as stacked PRs against an unmerged base.

## Review round 1 (post-review fixes)

Five findings from the multi-agent review were closed on this branch (each its own commit):

- **H1 (High) — broker kill-switch.** `POST /api/v1/mcp-oauth/user-token` is now gated behind `MCP_OAUTH_BROKER_ENABLED` (default **off**), returning `404` before auth when disabled. Per spec 06 §U1 the `userId` is deliberately *asserted* over the host-bound control-JWT (the same trust model as `internal/oauth.ts`<->rpc-proxy), and per-user principal binding + scope provisioning are U4's explicit responsibility -- so the flag makes the "not until U4" invariant code-level rather than building U4's binding early. **The per-user/per-server principal binding, and provisioning `oauth:user-token` to oauth-fronting runtimes, land with U4** (branch `feat/mcp-server-oauth-host`); until then no producer emits the scope and the endpoint is disabled.
- **M1 (Medium)** -- widened the HCC `McpServerOAuth.provider` union from 5 to 8 to match the CRD enum, and rewrote the `providers.ts` "adding a provider" checklist to list all four sites (it had omitted HCC, which is what let the union drift).
- **M2 (Medium)** -- documented why `oauth_grants_service_unique` is intentionally owner-blind (service grants are recipe-only; mcp-servers never create them).
- **L1 (Low)** -- closed a two-step `spec.oauth` immutability bypass in the mcpserver CEL by forbidding removal of `spec.oauth` once set. **Structural CRD change -- apply to the dev cluster before merge.**
- **L2 (Low)** -- documented that `OAuthProvider` is the control-api adapter set, not the recipe-admissible set (the recipe CRD enum is the authoritative gate).

Collateral audit of the fix round: clean, no regressions.


alfredolopez80 (commented):
Two inline findings on the OAuth broker authorization boundary.

control-api/src/utils/auth/mcpHostJwtToken.ts
```diff
@@ -148,6 +148,11 @@ export const ALL_MCP_HOST_CONTROL_SCOPES = [
   // (status.pluginWorkloadSdk) declares the SDK; SDK routes reject access
   // tokens without it via scope_denied.
   'plugin-workload-sdk',
+  // OAuth mcp-server broker (U1). Lets mcp-host exchange its control JWT for a
+  // per-connection provider access token at POST /api/v1/mcp-oauth/user-token.
+  // Provisioners request it only for hosts that front OAuth mcp-servers; the
+  // route rejects control tokens lacking it with 403 insufficient_scope.
+  'oauth:user-token',
```
alfredolopez80, line 180: **Must fix — the scope is accepted but never provisioned.**

At this head, neither trusted producer can request `oauth:user-token`: `HostWorkflowControlScope` and the Host CRD enum exclude it, and WRC's `WorkflowControlScope` / `deriveWorkflowControlScopes()` exclude it as well. Consequently, the real HCC/WRC → `/auth/mcp-host/issue` path signs control JWTs without this scope, so production broker requests reach `403 insufficient_scope`. The route tests mint a scoped JWT directly and therefore miss the broken wiring.

Please derive this scope in the trusted provisioner from authoritative Host/recipe MCP-server configuration, only for runtimes that front an enabled `auth.type: oauth` server; include it in token rotation/drift checks; and add contract tests that obtain the token through the real HCC and WRC issuance payloads. The mcp-host must never choose or widen the scope itself.

control-api/src/routes/mcpOauth.ts
```diff
@@ -0,0 +1,333 @@
...
+        return res.status(403).json({ error: 'insufficient_scope' })
+      }
+
+      try {
+        const body = (req.body ?? {}) as {
+          mcpServerName?: unknown
+          userId?: unknown
+          contextId?: unknown
+        }
+        const mcpServerName = body.mcpServerName
```
alfredolopez80, line 395: **Security boundary — bind the broker request to the signed caller, not only the coarse scope.**

Once issuance is wired, a runtime legitimately carrying `oauth:user-token` can choose `mcpServerName` and, for user grants, `userId` from the body. The verified claims (`sub`, `hostRefs`, recipe binding) are not used to prove that the caller may front that server or act for that end-user/session. A compromised authorized host that knows another valid user ID could therefore request that user's provider access token. Encryption and refresh-token custody remain intact, but the returned access token is usable.

Please resolve the allowed server set and user/session principal from platform-signed or authoritative context, or require a short-lived signed binding and verify it here. Add negative cross-host, cross-recipe, cross-server, and cross-user tests. The context branch's authoritative `contextRef` check is the right shape; user grants need equivalent principal binding.

control-api/src/utils/auth/mcpHostJwtToken.ts
zach88: You're right that at this head neither trusted producer can request `oauth:user-token`: HCC's `HostWorkflowControlScope` (5 values) and WRC's `WorkflowControlScope` (6) both exclude it, so `resolveWorkflowControlScopes` / `deriveWorkflowControlScopes` never emit it, and a real HCC/WRC-issued control JWT would hit `403 insufficient_scope`. And you're right that the route tests mint the scoped JWT directly, so they don't catch that.

The reason it's wired this way is deliberate sequencing, not an oversight. Per spec 06, U1's only remit for this scope is adding it to `ALL_MCP_HOST_CONTROL_SCOPES` (§U1) — the *provisioning* of the scope to runtimes that front an enabled `auth.type: oauth` server is U4 work, on `feat/mcp-server-oauth-host` (the same branch that lands the runtime `spec.oauth` projection in HCC and the JIT token resolution that actually calls this broker). Until U4, no producer emits the scope by design.

To make that non-exploitable rather than merely un-provisioned, this round added a fail-closed kill-switch: the endpoint is now gated behind `MCP_OAUTH_BROKER_ENABLED` (default **off**) and returns `404` before auth, so it can't serve in production until U4 lands and the flag is flipped.

Your provisioning requirements are exactly what U4 will implement — derive the scope from authoritative Host/recipe config only for oauth-fronting runtimes, include it in token rotation/drift, and (the key testing point) obtain the token through the **real** HCC/WRC issuance payloads rather than a hand-minted JWT. We're tracking those as U4 preconditions so the contract test derives the JWT from the real producer, not a fixture. The mcp-host never chooses or widens its own scope.

control-api/src/routes/mcpOauth.ts
zach88: Agreed on the shape of the gap — this is the same object-level authz concern our security review raised, and you're right that the `context` branch's authoritative `contextRef` check is the correct model. Two things on how we're closing it:

1. **The `userId` binding can't live in control-api by design.** The mcp-host control JWT carries no end-user identity — `sub` is `${recipeNamespace}/${recipeName}` (the host/recipe), and the broker is inherently multi-user (a per-request `userId`). Per spec 06 §U1 the `userId` is *asserted* by the mcp-host over the host-bound, mutually-authenticated control-JWT channel — the same trust model as the existing `internal/oauth.ts` ↔ rpc-proxy seam, which also takes `userId` from the body. The "userId always from the authenticated session, never a body param" invariant is explicitly scoped to **end-user surfaces** (§U1 / framing), not this control-plane seam. The per-user principal binding you're describing is threaded from the authenticated task sender in **U4** (`this.task.sourceMessage.sender`, `userId` from `auth.sub`), which is where per-user partitioning and JIT fail-closed resolution live.

2. **Server/context binding for the `user` flavor is likewise U4's connect-policy.** The isolation of which servers a host may front is the JIT resolution shipping on `feat/mcp-server-oauth-host`, not U1 base.

Rather than build U4's binding early (which would contradict U1's "zero owner-minting infra" decision), this round makes the risk non-exploitable until U4: the endpoint is now gated behind `MCP_OAUTH_BROKER_ENABLED` (default **off**), returning `404` before auth. So no host — compromised or not — can reach the broker in production until U4 lands both the principal binding and the scope provisioning. Tracking the cross-host/cross-recipe/cross-server/cross-user negative tests as U4 acceptance criteria.

control-api/src/utils/auth/mcpHostJwtToken.ts
zach88: Follow-up + correction. I earlier said this was already handled in U4 — that was wrong, and you were right that it was a genuine end-to-end gap: I verified across the full stack (#317/#319/#322) that **no producer requested `oauth:user-token`** (HCC's `HostWorkflowControlScope` and WRC's `WorkflowControlScope` both excluded it, and the Host CRD `workflowControl` enum didn't allow it either), so a real HCC-minted runtime JWT never carried it → the broker would 403. The route/broker tests masked it by minting the scoped JWT directly, exactly the contract-test point you raised.

Now implemented in **#319** (`feat/mcp-server-oauth-host`), commit `464e3cc1`: HCC derives `oauth:user-token` **only for a host whose Context fronts an enabled `auth.type: oauth` mcp-server**, resolved once per reconcile and threaded into both the mint scopes and the drift-hash (so the token and the scope-hash annotation can't diverge — no re-issue loop), failing closed to no-scope on an uncertain read. It's a derive-only scope (not user-declarable), so the Host CRD enum is unchanged. WRC is untouched — recipes can't front oauth McpServer CRDs in v1 (recipe OAuth is a separate broker).

On your contract-test point: the new coverage runs a host that fronts an enabled oauth mcp-server through the **real** derive → issuance payload (asserting `oauth:user-token` in the scopes passed to the issuer), and a non-oauth host does not get it — no hand-minted JWT. Full HCC suite green.

Note this lands in #319, not #317: #317's broker still trusts the host-bound control JWT (userId asserted over the channel, per the spec's control-plane seam), and the endpoint stays gated behind `MCP_OAUTH_BROKER_ENABLED` (default off) until the chain ships.

alfredolopez80:
**Multi-agent review (4 parallel agents + Fable 5 adversarial pass) — cc @zach88**

Context: reviewed **only #317** as the base of the stack #317 → #319 (U4) → #322 (U5); #319/#322 used as contract-stability context. Verified against the real code in a worktree of the head; suite run locally.

## Recommendation: nothing blocks the merge ✅

The fail-closed posture is solid and verified three ways (direct verification + security agent + adversarial pass): broker gated by `MCP_OAUTH_BROKER_ENABLED` (default **OFF**, `failClosedBooleanFromEnv` throws on invalid values), `404` **before** auth, single route with no duplicate router, direct-rail serves no tokens (names + internal cluster DNS only). Suite: `tsc` 0 errors; **4462 passed / 111 skipped / 0 failed** (non-vacuous).

## Verdict on the open comments

| # | Location | Verdict |
|---|---|---|
| **C1** (Must fix) | `mcpHostJwtToken.ts:155` | **Legitimate, correctly deferred to U4.** #317 only defines the scope in `ALL_MCP_HOST_CONTROL_SCOPES`; no producer emits it. The kill-switch makes it inert. U4/#319 precondition: derive the scope from authoritative config + contract tests using the **real** HCC/WRC issuance. |
| **C2** | `mcpOauth.ts:215` | **Legitimate and CONFIRMED by the adversarial pass.** The `user` branch (`mcpOauth.ts:269-278`) takes `userId` from the body without binding it to `claims.sub`; with the flag ON → cross-user token theft if the victim has a grant. The `context` branch is properly hardened (authoritative `contextRef`). Correctly deferred to U4; **"flag-on = low risk" is false** — this is the hard precondition before enabling the flag. |
| **C3** | `pkce.ts:38` (CodeQL) | **False positive.** Not password hashing: a deterministic KDF/MAC over an already HMAC-signed, high-entropy `state`. Security rests on the entropy of `stateSecret` (≥32 enforced, prod-required), not on hash cost. The verifier is never stored. |
| **C4/C5** | `mcpOauth.ts:193,343` (CodeQL) | **False positive (covered at HEAD).** The file has a single route and it does carry `rateLimitMiddleware` (60/min). The flagged lines are from a pre-round-1 commit; CodeQL doesn't follow Express's middleware chain. |

## Follow-up (do not block #317)

**HIGH**
- **H1** — close the `user`-branch principal binding in **#319** (bind `userId` to `claims.sub` / task sender) with negative cross-user/host/recipe/server tests **before** enabling the flag in prod. This is the materialization of C1+C2.

**MEDIUM**
- **M1** — `spec.contextRef` is not CEL-immutable in `mcpserver.yaml`, unlike `grantScope`/`id`/`provider`. An actor with `update` RBAC on the CR can repoint `contextRef` ctxA→ctxB and redirect the shared token. Two independent agents concur. Suggestion: add `!has(oldSelf.contextRef) || self.contextRef == oldSelf.contextRef`.
- **M2** — `OAuthProvider` duplicated across 3 TS sites (`control-api providers.ts:25`, `HCC types.ts:36`, `WRC types.ts:441`) + 2 CRD enums, with no shared `@clerum/*` package (its sibling `LlmProviderId` has one). Enum-drift risk on the 9th provider.
- **M3** — real DB/CEL behavior is only mocked or skip-gated (silently) in CI: the `*.realPostgres.integration.test.ts` skip without `CONTROL_API_REAL_PG_ADMIN_URL`, so `ON CONFLICT DO NOTHING`, the CHECKs, and the uniqueness superset never execute in CI. Add a **loud** skip + a Postgres lane.
- **M4** — JWT scope *provenance* is untested (confirms C1): `routes.mcpOauth.test.ts` injects the scope at mint time; nothing exercises real issuance. Close with a contract test when U4 lands.

**LOW / optional**
- `redact.ts` misses snake_case `access_token`/`refresh_token` (no live leak; tokens encrypted at rest) · rate-limit runs after `requireControlCaller` (minor unauth CPU-DoS, gated by the flag) · the "adding a provider" checklist omits the 2nd hand-maintained list `isOAuthProvider` in `WRC types.ts:451` · `utils.log.redact.test.ts` asserts membership, not behavior · CRD immutability tests are string-assert only (no `kubectl --dry-run=server`) · label rot "0091" vs the real migration "0100" in `db.oauthGrantsOwnerGeneralization.test.ts`.

## What came out clean (verified)
Fail-closed kill-switch · fail-closed grant gate (never admits bearer/basic/apiKey; `context` servable IFF shared grant, decoupled from the caller; keying always by `server.spec.*`) · store race/integrity (first-wins, refresh never rewrites the bootstrapper, CHECK rejects `shared` without a bootstrapper, parameterized queries) · PKCE (explicit S256, random nonce per authorize, signature+TTL verified) · providers with no SSRF (per-adapter hardcoded URLs) · provider union aligned (8 in the full set, 5 in the recipe CRD = intentional fail-closed) · R1-L1 immutability-bypass closure present at `mcpserver.yaml:518`.

## Deployment note
The structural `spec.oauth` immutability change in the CRD must be applied to the dev cluster before merge (present and correct). "Do not enable OAuth mcp-servers in prod until U4" is enforced at code level by the kill-switch; **H1/M4/C1/C2 are the concrete preconditions of that "until U4".**

---
<sub>Automated analytical review. Findings were verified against the real code; none block the merge.</sub>

control-api/src/routes/mcpOauth.ts
```diff
@@ -0,0 +1,347 @@
...
+            ownerKind: 'mcpserver',
+            recipeNamespace: config.mcpServersNamespace,
+            recipeName: mcpServerName,
+            contextId: authoritativeContextId,
+            oauthClientId: resolved.oauthClientId,
+          }
+        } else {
+          if (typeof userId !== 'string' || userId.length === 0) {
+            return res.status(400).json({ error: 'invalid_request' })
+          }
```
alfredolopez80, line 447: **[HIGH · follow-up for U4/#319]** The `user` branch takes `userId` from the request body without binding it to the signed caller (`claims.sub`). Once the broker is enabled (`MCP_OAUTH_BROKER_ENABLED=true`), a runtime legitimately carrying `oauth:user-token` can request `{mcpServerName, userId: <victim>}` and receive that user's provider access token if a grant exists. The `context` branch is correctly hardened via the authoritative `contextRef`; the `user` flavor needs equivalent principal binding (thread `userId` from the authenticated task sender / `auth.sub`). Confirmed by the adversarial pass. Not exploitable at this head (kill-switch OFF by default) — this is the hard precondition to close in #319 with negative cross-user/host/recipe/server tests before the flag is ever enabled in prod.

charts/clerum-crds/crds/mcpserver.yaml
```diff
@@ -388,6 +489,34 @@ spec:
...
+                # OAuth servers must not also carry a static credential.
+                - rule: "!has(self.oauth) || !has(self.auth) || (!has(self.auth.secretRef) && !has(self.auth.secretKey))"
+                  message: "OAuth mcp-servers must not declare a static auth.secretRef/secretKey"
+                # Immutability via has()-guard (patrón `managed`): these are
+                # transition rules (they reference oldSelf), so they are SKIPPED on
+                # CREATE and only enforced on UPDATE. Guarding on has(oldSelf.oauth)
+                # means create-time grantScope:'context' is allowed, and a
+                # none→oauth UPDATE is allowed, while any change to a grantScope /
+                # id / provider that ALREADY existed is rejected.
+                - rule: "!has(oldSelf.oauth) || !has(self.oauth) || oldSelf.oauth.grantScope == self.oauth.grantScope"
```
alfredolopez80, line 505: **[MEDIUM]** `grantScope`/`id`/`provider` are immutable here, but `spec.contextRef` is not — yet for a `grantScope: context` server it is the authoritative coordinate that selects which Context's shared identity is brokered (`mcpOauth.ts` keys the grant by `resolved.contextRef`). An actor with `update` RBAC on the McpServer CR can repoint `contextRef` ctxA→ctxB in place and redirect the shared token, defeating the very immutability guarantee the sibling fields enforce. Attenuated by triple gating (CR-write RBAC + flag ON + control JWT), hence Medium not High. Suggest adding a transition rule mirroring the others: `!has(oldSelf.contextRef) || self.contextRef == oldSelf.contextRef`.

control-api/src/oauth/providers.ts
```diff
@@ -1,10 +1,33 @@
...
+ * source of truth) — no per-call-site provider lists.
+ *
+ * U2 (spec 06 §U2) adds monday, clickup, vercel. monday + vercel require PKCE
+ * (see `usesPkce` / pkce.ts).
  */
-export type OAuthProvider = 'salesforce' | 'slack' | 'notion' | 'microsoft-graph' | 'google'
+export type OAuthProvider =
+  | 'salesforce'
+  | 'slack'
+  | 'notion'
```
alfredolopez80, line 25: **[MEDIUM]** This `OAuthProvider` union is hand-maintained across 3 TS sites (here, `host-context-controller/src/types.ts:36`, `workflow-recipes/src/types.ts:441`) plus the 2 CRD enums — 5 places kept in sync only by a checklist comment. The sibling `LlmProviderId` already lives in a shared package (`@clerum/llm-providers`); OAuth doesn't follow that pattern. Adding a 9th provider requires touching 5 synchronized sites; a miss in HCC or WRC compiles fine and only fails at runtime/admission. Consider a `@clerum/oauth-providers-core` (full set + recipe-admissible subset). Non-blocking.

alfredolopez80 (commented):
Inline anchors for the follow-up findings from the multi-agent review (cc @zach88). None block the merge; H1/M1/M2 are tracked follow-ups.

control-api/src/oauth/providers.ts
zach88: Agreed, and thanks — this is the right structural direction. Two notes on where it stands:

- The immediate drift risk you'd expect from the round-1 collapse is already closed: R1-M1 (commit on this PR) widened the HCC union that had silently fallen behind (`host-context-controller/src/types.ts` was at 5 while everything else was at 8) and **rewrote the `providers.ts` header checklist** to enumerate every synchronized site, so "compiles fine, fails at runtime/admission" now has an explicit lockstep list. You're right that the checklist still doesn't call out the second hand-maintained list — `isOAuthProvider` in `workflow-recipes/src/types.ts` — I'll fold that into the checklist.
- A shared `@clerum/oauth-providers-core` (full adapter set + recipe-admissible subset, mirroring how `LlmProviderId` lives in `@clerum/llm-providers`) is the real fix that removes the checklist entirely. It's a cross-package refactor touching 5 sites + the two CRD enums, so I'll track it as a follow-up issue rather than fold it into this stack — non-blocking, as you noted. Adding the 9th provider stays a checklist-guarded change until then.

control-api/src/routes/mcpOauth.ts
zach88: Confirmed and closed on the mcp-host side (#319, commit `a3e0adc1`). You're right that control-api's broker trusts the body `userId` — by design that's the trusted-channel seam (the control-JWT `sub` is the host/recipe, not a user, same shape as `internal/oauth.ts` ↔ rpc-proxy). The enforcement lives at the authenticated producer, and I traced that the mcp-host can only ever send the caller's own identity:

- broker POST body `userId` = the provider instance's baked-in subject (`brokerTokenProvider.ts`), a per-instance closure — no cross-user reuse
- subject ← `principal.userId` from the manager (`main.ts` `createMcpTokenProviderFactory`) — never a body value
- `manager.callTool` reads **only** `options.userId`, fails closed on missing/`anonymous` before any token resolves, and never reads tool args
- `options.userId` = `this.userId` on `McpToolRegistryAdapter` — a constructor field, separate from `params`
- adapter `userId` = `task.sourceMessage.sender`, which `handleMessageRoute` binds to the rpc-proxy edge `caller.userId` (`auth.sub`), **overwriting** any body-supplied sender and 401-ing if the edge context is absent; a direct `Authorization` is rejected 401 at the edge

So an LLM / tool-arg cannot choose the identity — a `userId` field inside the tool `params` is inert. I locked this with negative tests that bite by mutation: reading `params.userId` fails the adapter tests, and crossing the broker `userId` fails a real-`McpManager` + real-`createBrokerTokenProvider` E2E (only `fetch` stubbed) asserting the POST body `userId` per caller (alice's identity never serves bob). The residual — a fully-compromised mcp-host process forging the body — is the trusted-channel assumption inherent to this seam, covered by the kill-switch + control-plane trust model, not by control-api re-deriving an identity it does not hold.

charts/clerum-crds/crds/mcpserver.yaml
zach88: Fixed (#317, commit `527ec08b`) — with one deliberate deviation from the literal suggestion, after auditing its collateral.

Global `contextRef` immutability would regress non-oauth servers: `contextRef` is a required field on *every* McpServer, and WRC's full-spec REPLACE (`mcpDelegation.ts:500`) re-points a stable-named McpServer's `contextRef` whenever an operator edits the owning recipe's (mutable) `contextRef`. A global `oldSelf` rule would fail that at admission and wedge the reconcile loop — for servers that carry no shared grant and no redirection threat. The threat is oauth-specific (`mcpOauth.ts` keys the shared grant by `resolved.contextRef` only when `auth==oauth && grantScope==context`), so I guarded the rule on `oldSelf.oauth`:

```
!has(oldSelf.oauth) || !has(self.oauth) || self.contextRef == oldSelf.contextRef
```

This pins `contextRef` once oauth exists (closing the ctxA→ctxB shared-token redirection at admission — for the admin route, kubectl, and GitOps alike), while leaving the non-oauth WRC re-parent flow intact. No two-step drop-oauth bypass — the sibling rule already forbids removing `spec.oauth` once set. Structural CRD change, so it needs `kubectl apply -f charts/clerum-crds/crds/mcpserver.yaml` on dev before merge.

Follow-up worth its own issue: `WorkflowRecipe.spec.contextRef` is itself mutable (`workflowrecipe.yaml`, no `oldSelf` rule) — the upstream knob that drives that WRC re-point. Out of scope for this finding; flagging it since if recipe-level re-parenting should also be delete+recreate, that's the place to close it.

alfredolopez80:
@claude Please perform a **complete independent re-review** of the FULL current PR. Review only — do not make code changes.

https://github.com/evenfire-ai/evenfire/pull/317

Current expected state:
- base: `4862d9c2ff306e7727f35539ae696bc2c465b24a` (`origin/dev`)
- head: `50f1608fc57d7a4400aa0d4f95f6a420d4c63727`
- branch: `feat/mcp-server-oauth`
- this is the **control-api base** of the mcp-server OAuth stack: #317 (this PR, U1/U2/U3/U6-api) → #319 (U4 + U6-host, OPEN) → #322 (U5, MERGED)
- the last human review comment from `@alfredolopez80` was at `771e23fc` (2026-08-19). Many commits have landed since: `527ec08b` (M1 `contextRef` pin), minikube overlay enabling the broker, several `origin/dev` syncs, migration renumber `0100` → `0101` + `legacyVersions`, and the T2 public-boundary comment rewrite at this head

If the live head/base have moved, review the **actual current state** and report the live SHAs rather than relying on the values above.

This is **not** a delta-only pass and **not** a recap of prior reviews. Re-audit the entire current diff against live `origin/dev`. No prior `@claude` review job was found on #317 in the conversation/reviews/timeline search — still search the full PR (including Actions / Claude Code jobs) and treat any Claude output you find as in-scope.

---

## Required deliverable 1 — historical-findings closure ledger

Build an exhaustive ledger of **every** finding, residual, inline thread, review observation, CodeQL alert, and claimed closure that appears anywhere on this PR: conversation comments, review bodies, inline review threads, author replies, the PR description (“Review round 1”), commit messages that claim to close a finding, and any `@claude` job output.

For **each** item report:

| Field | Required content |
|---|---|
| ID / short title | Stable ID if one exists (`R1-H1`, `C1`, `M1`, …) or a precise coined title |
| Origin reviewer | Who raised it |
| Original severity | Blocker / Critical / High / Medium / Low / note / CodeQL |
| Current disposition | `CLOSED` / `OPEN` / `PARTIAL` / `DEFERRED_TO_319` / `PRE_EXISTING` / `NOT_FOUND_IN_PR` / `REOPENED` |
| Live-head evidence | File:line, test name, CRD rule, env flag, migration, or explicit absence |
| Claimed vs verified | Independently confirm or refute the author’s claimed closure |
| Residual | What remains, if anything |

Rules for this ledger:

- Do **not** trust author replies, “resolved” GitHub threads, the PR body’s “closed” list, commit-message claims, or prior human conclusions. Re-verify against the live head.
- A resolved thread is **not** evidence of a fix.
- If a reviewer left no findings on #317 after a full search, say so explicitly as `NOT_FOUND_IN_PR`. Do not skip the search.
- Include optional/non-blocking items, not only must-fix items.
- If a previously `CLOSED` item has reopened after later commits (especially the minikube overlay that sets `MCP_OAUTH_BROKER_ENABLED=true`, the `dev` syncs, or the migration renumbers), mark it `REOPENED` and explain why.
- Stack context is allowed only to verify a claimed “closed on #319 / #322” disposition. Do **not** expand this review into a full audit of #319 or #322 unless needed to confirm or refute a #317 closure claim.

### Reviewers whose findings must be inventoried

1. **@claude** — search conversation comments, review bodies, inline threads, and Claude Code / Actions jobs on #317. If none exist, report `NOT_FOUND_IN_PR` after the search. Still perform deliverable 2 in full.
2. **@zach88** — PR author (Diego Zacarias). Inventory every author disposition, claimed closure, deferred-to-U4/#319 statement, and inline reply (including the PR-body “Review round 1” list and replies on C1/C2/H1/M1/M2). Independently confirm or refute each claimed closure.
3. **@alfredolopez80** — every round on this PR:
   - 2026-08-14 inline threads C1 / C2 (`mcpHostJwtToken.ts`, `mcpOauth.ts`)
   - 2026-08-19 multi-agent review comment (C1–C5 verdicts; H1; M1–M4; LOW/optional list)
   - 2026-08-19 inline anchors (H1 user-branch binding, M1 `contextRef`, M2 `OAuthProvider` drift)
4. **@jozer-rami** — a review was **requested** and is still pending. Search the full PR conversation, reviews, inline threads, and mentions. If this reviewer left no findings on #317, report `NOT_FOUND_IN_PR` after the search.
5. **Josué Palmera / @josue-tech-94** — search the full PR conversation, reviews, inline threads, and mentions. If this reviewer left no findings on #317, report `NOT_FOUND_IN_PR` after the search.

### Known IDs that must appear in the ledger

This list is a minimum checklist, not the complete set. Add any other finding you discover on the PR.

**Round 1 (PR body + fix commits, claimed closed on this branch):**
- R1-H1 — broker kill-switch `MCP_OAUTH_BROKER_ENABLED` (default off, `404` before auth). Re-evaluate after commit `fdf7d53b` enabled the flag in the **minikube overlay** (`deploy/...` control-api config). Confirm base/managed stay fail-closed. Confirm `failClosedBooleanFromEnv` still throws on invalid values. Confirm there is still a single route and no duplicate router.
- R1-M1 — HCC `McpServerOAuth.provider` union widened 5 → 8; `providers.ts` “adding a provider” checklist rewritten
- R1-M2 — `oauth_grants_service_unique` left owner-blind; documented rather than migrated
- R1-L1 — two-step `spec.oauth` immutability bypass closed (forbid removal once set)
- R1-L2 — `OAuthProvider` documented as adapter set, not recipe-admissible set (recipe CRD enum stays at 5)

**Alfredo C1–C5 (open-comment verdicts):**
- C1 (Must fix / deferred) — `oauth:user-token` is accepted in `ALL_MCP_HOST_CONTROL_SCOPES` but no trusted producer (HCC `HostWorkflowControlScope`, Host CRD enum, WRC `WorkflowControlScope` / `deriveWorkflowControlScopes`) emits it. Route tests mint a scoped JWT directly.
- C2 — `user` branch takes `userId` from the body without binding it to `claims.sub`. Adversarial confirmation: flag ON → cross-user token theft if the victim has a grant. `context` branch is claimed hardened via authoritative `contextRef`.
- C3 (CodeQL) — `pkce.ts` “password hash with insufficient computational effort”. Claimed false positive (deterministic KDF/MAC over HMAC-signed high-entropy state).
- C4 / C5 (CodeQL) — missing rate limiting on `mcpOauth.ts`. Claimed false positive (single route already has `rateLimitMiddleware` 60/min; CodeQL does not follow Express middleware).

**Alfredo follow-ups (do not treat “do not block #317” as “verified closed”):**
- H1 — close `user`-branch principal binding in #319 (bind `userId` to task sender / `auth.sub`) with negative cross-user/host/recipe/server tests **before** enabling the flag in prod. Author later claimed this closed on #319 commit `a3e0adc1`. Verify whether that claim is still true **and** whether #317 itself still has the unbound body `userId`. Separately re-evaluate residual risk now that minikube overlay has the broker **ON**.
- M1 — `spec.contextRef` was not CEL-immutable. Author claimed closed in `527ec08b` with an oauth-guarded rule (`!has(oldSelf.oauth) || !has(self.oauth) || self.contextRef == oldSelf.contextRef`) instead of a global pin, to avoid regressing WRC full-spec REPLACE on non-oauth servers. Verify the CEL, the two-step drop-oauth interaction with R1-L1, and the test at `control-api/test/crd.mcpServerOauth.test.ts`.
- M2 — `OAuthProvider` hand-maintained across 3 TS sites + 2 CRD enums; no `@clerum/oauth-providers-core`. Author agreed, promised to fold `isOAuthProvider` into the checklist, and deferred the shared package to a follow-up issue. Verify whether the checklist update landed on this head.
- M3 — `*.realPostgres.integration.test.ts` skip silently without `CONTROL_API_REAL_PG_ADMIN_URL`; CHECKs / `ON CONFLICT DO NOTHING` / uniqueness never execute in default CI
- M4 — JWT scope provenance untested (confirms C1); close with a contract test when U4 lands

**Alfredo LOW / optional (all must appear):**
- `redact.ts` misses snake_case `access_token` / `refresh_token`
- rate-limit runs after `requireControlCaller` (unauth CPU-DoS, claimed gated by the flag)
- “adding a provider” checklist omits the second hand-maintained list `isOAuthProvider` in WRC `types.ts`
- `utils.log.redact.test.ts` asserts membership, not behavior
- CRD immutability tests are string-assert only (no `kubectl --dry-run=server`)
- label rot `"0091"` vs later migration numbers (`0100`, then `0101` after the minimax seed collision). Verify the current test labels, ordering assertions, and `legacyVersions` (`0091_...`, `0100_...` → `0101_oauth_grants_owner_generalization`)

**Author residuals / follow-ups that must be dispositioned:**
- claimed “do not enable OAuth mcp-servers in production until U4” — is this still a comment-only promise, or still enforced in code on every non-minikube overlay?
- U4/#319 still OPEN: scope provisioning, producer issuance, JIT isolation
- U5/#322 MERGED: confirm #317 does not assume U5 surfaces that are missing here (desktop consent, shared-grant revocation endpoint)
- `WorkflowRecipe.spec.contextRef` remains mutable (author-flagged follow-up after M1)
- LLM-catalog rail intentionally ungated; isolation claimed to be U4 JIT
- recipe CRD enum stays at 5 while adapter set is 8

**CodeQL / github-advanced-security (threads currently marked resolved):**
- `control-api/src/oauth/pkce.ts` insufficient computational effort
- `control-api/src/routes/mcpOauth.ts` missing rate limiting (two alerts)

---

## Required deliverable 2 — full independent audit of the current PR

Review the complete current PR diff, not only historical items. Cover at least:

- correctness and regressions vs live `origin/dev`
- security, authentication, and authorization boundaries of `POST /api/v1/mcp-oauth/user-token`
- kill-switch correctness: default, invalid-value fail-closed, `404` before auth, overlay vs base/managed
- principal binding for `user` vs `context` grant flavors; whether body `userId` / `contextId` can steal tokens
- grant-presence gate (`filterInvocable` / `computeGrantPresence`): allowlist never admits bearer/basic/apiKey; `context` servable IFF shared grant, decoupled from caller; fail-closed per server
- PKCE: deterministic verifier from `stateSecret`, S256, no stored verifier, domain separation from state signature, refresh path carries no PKCE
- provider adapters (monday / clickup / vercel): hardcoded URLs (no SSRF), PKCE vs non-PKCE, refresh behavior, token-body shape
- store invariants: first-wins bootstrap (`ON CONFLICT DO NOTHING`), refresh never rewrites bootstrapper / `bootstrapped_by_user_id`, CHECKs, uniqueness (including the owner-blind service index)
- CRD / CEL: `spec.oauth` coupling, secretRef forbidden alongside oauth, immutability of `id` / `provider` / `grantScope`, oauth-removal forbid, oauth-guarded `contextRef` pin
- migration 0101 + `legacyVersions` (no false-skip against `0100_seed_minimax_allowed_model`; no re-exec on clusters that applied 0091/0100)
- logging / redaction of tokens, verifiers, DSNs, callback URLs
- test quality: minted-JWT vs real HCC/WRC issuance; silent real-PG skips; string-assert CEL; property tests that actually bite; T2 public-boundary comment-only rewrites
- rollout / deployment: structural CRD must be applied before merge; minikube overlay enables the broker — state whether that changes the residual-risk story
- stack contract: what #317 still leaves to #319 (OPEN) and what it may incorrectly assume from #322 (MERGED)

Report:

- new findings with severity (Blocker / Critical / High / Medium / Low) and exact evidence
- whether any previously `CLOSED` item has reopened on the current head
- remaining **#317 must-fix** vs **acceptable deferred to #319** vs **pre-existing / out of diff scope**
- whether the current exact head is still a sound base for #319, and whether anything now blocks merge of #317 itself
- an explicit reminder that a stacked “until U4” deferral is only acceptable if the kill-switch (or equivalent) still makes the deferred hole non-exploitable on every environment that is not a deliberate local overlay

Be as detailed as possible. Trace dispositions to code and tests, not to comments. Prefer observable-behavior or mutation-style evidence over narrative. Read tests for what they actually assert.

Review only — no code changes. This is analysis, not a formal GitHub approval.

alfredolopez80:
## Independent review — exact head `50f1608f`

Read-only review of the **full** current PR against live `origin/dev` `4862d9c2`. The prior independent review at https://github.com/evenfire-ai/evenfire/pull/317#issuecomment-5457248072 was used as a **checklist only**. Every disposition below was re-found in head (and compared to base when claiming `introduced_by_pr`).

cc @zach88

This is **analysis, not a GitHub Approve**. Local Vitest was not re-executed (disposable worktrees had no `node_modules`). CI on this head is green, including control-api build/test, CodeQL, leak guards, and **Real Postgres integration** (the job that sets `CONTROL_API_REAL_PG_REQUIRED=1` and fails if any suite is skipped).

**Legend:** CLOSED · PARTIAL · OPEN · DEFERRED_319 · NOT_FOUND_IN_PR · independently verified

Live state:
- head `50f1608fc57d7a4400aa0d4f95f6a420d4c63727` (`feat/mcp-server-oauth`)
- base `4862d9c2ff306e7727f35539ae696bc2c465b24a` (`dev`)
- 43 files, +3610 / −174
- stack: #317 (this PR) → #319 (U4, OPEN) → #322 (U5, MERGED)

---

## Review verdict

**YES** — no Blocker / Critical / High product defect blocks merge of #317 itself.

This head is a **sound base for #319**. It is **not** permission to enable OAuth mcp-servers in production.

Conditions that remain operator / follow-up work, not merge refusals:

1. Apply the structural `mcpserver.yaml` CEL to the target cluster before relying on immutability (`kubectl apply` / `kubectl --dry-run=server`). Unit tests only string-assert CEL.
2. A multi-replica `control-api` rolling update has a write-window break on the **existing recipe** OAuth upsert (**RP-317-001**). Sequence the deploy or accept a short Connect / refresh outage.
3. Keep `MCP_OAUTH_BROKER_ENABLED` **OFF** on base/managed until #319 closes C1 / C2 / H1 / M4 (scope provenance + `userId` binding + negative tests).
4. Shared-grant refresh TOCTOU (**RP-317-002**) should land before or with U5, which is the first production writer of shared grants.

---

## PR-introduced issue count

| Severity | Count | IDs |
|---|---|---|
| Blocker / Critical / High | 0 | — |
| Medium | 5 | RP-317-001, RP-317-002, RP-317-003, RP-317-004 (deferred), RP-317-005 (accepted residual) |
| Low | 8 | RP-317-006 … RP-317-014 |
| Pre-existing material | 0 | recipe OAuth path is unchanged in steady state |
| Uncertain | 1 | local suite execution |
| Root-cause clusters | 4 | C-SCHEMA-CUTOVER, C-SHARED-REFRESH, C-U4-BINDING, C-KILL-SWITCH-SCOPE |

No previously CLOSED ledger item reopened. Two Medium findings are **new relative to the prior independent review**: RP-317-001 and RP-317-002.

---

## Deliverable 1 — historical-findings closure ledger

Reviewer search on this PR: findings exist from **alfredolopez80**, **zach88** (author), and **github-advanced-security**. **jozer-rami** (review requested) and **josue-tech-94** left no findings (`NOT_FOUND_IN_PR`). There was no prior independent job on #317 before comment 5457248072.

| ID | Origin | Sev | Prior review | This review | Live-head evidence |
|---|---|---|---|---|---|
| **R1-H1** broker kill-switch | zach88 | High | CLOSED | **CLOSED** | `config.ts` `failClosedBooleanFromEnv('MCP_OAUTH_BROKER_ENABLED')` throws on garbage; `mcpOauth.ts:182-187` returns `404 not_found` **before** auth; single `api.use(createMcpOauthRouter)` in `app.ts`. Overlay ON only in `deploy/overlays/minikube/configmaps/control-api-config.yaml`. |
| **R1-M1** HCC union 5→8 | zach88 | Med | CLOSED | **CLOSED** | `host-context-controller/src/types.ts:36-44` = 8 providers. |
| **R1-M2** service unique owner-blind | zach88 | Med | CLOSED | **CLOSED** (documented) | `db.ts:6110-6121`. Residual type hole = RP-317-012. |
| **R1-L1** oauth-removal forbid | zach88 | Low | CLOSED | **CLOSED** | `mcpserver.yaml:518` `!has(oldSelf.oauth) \|\| has(self.oauth)`. Test does **not** search this rule (RP-317-009). |
| **R1-L2** adapter set ≠ recipe set | zach88 | Low | CLOSED | **CLOSED** | Recipe CRD enum stays at 5; comment added in `workflowrecipe.yaml`. |
| **C1** scope accepted, never provisioned | alfredo | Must-fix | DEFERRED_319 | **DEFERRED_319** | Scope present in `ALL_MCP_HOST_CONTROL_SCOPES` (`mcpHostJwtToken.ts:180`). HCC `HostWorkflowControlScope` and WRC `WorkflowControlScope` omit it. Issue route **will** mint it if asked (RP-317-003). |
| **C2 / H1** unbound body `userId` | alfredo | High | DEFERRED_319 | **DEFERRED_319** (still present) | `mcpOauth.ts:269-280` takes `userId` from the body. Context branch **is** hardened via authoritative `spec.contextRef` (`:246-268`). |
| **C3** PKCE CodeQL “weak hash” | CodeQL | note | CLOSED (FP) | **CLOSED (FP)** | `pkce.ts` is HMAC-SHA256 over an already HMAC-signed high-entropy `state`. Not password hashing. |
| **C4 / C5** missing rate-limit CodeQL | CodeQL | note | CLOSED (FP) | **CLOSED (FP)** | `mcpOauth.ts:194` already has `rateLimitMiddleware` (60/min). Order caveat = RP-317-014. |
| **M1** `spec.contextRef` CEL pin | alfredo | Med | CLOSED | **CLOSED** | `mcpserver.yaml:538` oauth-guarded pin. Global pin would wedge WRC non-oauth REPLACE. |
| **M2** `OAuthProvider` shared package | alfredo | Med | PARTIAL | **PARTIAL** | Unions aligned at 8. Checklist still omits WRC `isOAuthProvider` (RP-317-010). |
| **M3** real-PG silent skip | alfredo | Med | OPEN | **PARTIAL** | Default `npm test` still `describe.skip`. The dedicated CI job sets `CONTROL_API_REAL_PG_ADMIN_URL` + `CONTROL_API_REAL_PG_REQUIRED=1`, runs `--run realPostgres`, and **fails if any test is skipped**. That job is green. |
| **M4** JWT scope provenance untested | alfredo | Med | DEFERRED_319 | **DEFERRED_319** | `routes.mcpOauth.test.ts` injects `scopes: ['oauth:user-token']` into the real minter. |
| LOW redact snake_case | alfredo | Low | OPEN | **OPEN** = RP-317-008 | `redact.ts` has camelCase tokens + PKCE; no `access_token` / `refresh_token`. |
| LOW rate-limit after JWT | alfredo | Low | OPEN | **OPEN** = RP-317-014 | `mcpOauth.ts:190-201`. |
| LOW checklist omits `isOAuthProvider` | alfredo | Low | OPEN | **OPEN** = RP-317-010 | Author agreed; not landed. |
| LOW redact membership-only test | alfredo | Low | OPEN | **OPEN** | `utils.log.redact.test.ts` uses `toContain`, not a Pino round-trip. |
| LOW CEL string-assert | alfredo | Low | OPEN | **OPEN** = RP-317-009 | `crd.mcpServerOauth.test.ts` admits there is no CEL evaluator. |
| LOW label rot `0091` | alfredo | Low | CLOSED | **CLOSED** | Migration is `0101_...` with `legacyVersions` `0100_oauth_grants_owner_generalization` / `0091_...`, distinct from `0100_seed_minimax_allowed_model`. |

---

## Deliverable 2 — findings on this head

### Medium (introduced)

| ID | Sev | Location | Defect | Gate | Cluster |
|---|---|---|---|---|---|
| **RP-317-001** | Med | `control-api/src/db.ts:6092` + `store.ts:136` | One-step DROP+ADD of `oauth_grants_unique` plus rewritten `ON CONFLICT` (4 columns → 5). After the first new pod runs `initDb`, remaining **old** replicas still name `ON CONFLICT (recipe_namespace, recipe_name, user_id, oauth_client_id)` and Postgres rejects: no matching unique. Recipe Embed Connect and token refresh fail until the last old replica dies. Steady-state after rollout is fine. | Rolling update window | C-SCHEMA-CUTOVER |
| **RP-317-002** | Med | `store.ts:166` + `tokenHelper.ts:139-174` | Shared refresh exchanges at the provider first, then a plain UPDATE whose `rowCount` is ignored, then returns `{ kind: 'ok' }`. Concurrent delete between `getOAuthGrant` and the UPDATE burns the rotated refresh token. User/service branches use `ON CONFLICT DO UPDATE` and do not have this hole. | Latent until U5 writes shared grants | C-SHARED-REFRESH |
| **RP-317-003** | Med | `issue.routes.ts:30-48,164-166` | Adding `oauth:user-token` to `ALL_MCP_HOST_CONTROL_SCOPES` makes the existing issue endpoint a producer if any provisioner puts the scope in the body. HCC/WRC TypeScript unions still omit it, so a normal reconcile cannot request it. | Broker OFF on base/managed; overlay ON + hand-crafted issue call | C-U4-BINDING |
| **RP-317-004** | Med | `mcpOauth.ts:269-280` | User flavor keys the grant by body `userId` with no bind to a signed end-user. Same trusted-channel seam as `recipeOauth` / `internal/oauth.ts`. Context flavor is correctly hardened. | Deferred to #319; non-exploitable while no producer emits the scope and the flag stays OFF in prod | C-U4-BINDING |
| **RP-317-005** | Med | `mcpInvocable.ts:157-186` | Kill-switch hides the **broker**, not catalog admission. Base excluded any `authType !== 'none'`; head admits `oauth` on both rails. Catalog rail (`resolveMcpServersForAgents`) passes no `grantPresence`. Existence leak + broken connect, not token theft. Documented as “isolation is U4 JIT”. | Acceptable residual unless the promise is literal “servers must look absent” | C-KILL-SWITCH-SCOPE |

### Low (introduced)

| ID | Location | Defect |
|---|---|---|
| RP-317-006 | `providers.ts:401-416,521` | monday/vercel omit PKCE params when challenge/verifier is missing instead of throwing. Production helpers always pass the fields. |
| RP-317-007 | `mcpInvocable.ts:258-262` | New `console.error` bypasses control-api’s redacting Pino logger (logging standard). |
| RP-317-008 | `redact.ts:2-22` | Missing snake_case `access_token` / `refresh_token`. No live leak today. |
| RP-317-009 | `crd.mcpServerOauth.test.ts` | CEL tests are substring searches; R1-L1 oauth-removal rule has no guardian. Product CEL at `mcpserver.yaml:494-538` is consistent when read directly. |
| RP-317-010 | `providers.ts:10-16` | “Adding a provider” checklist omits WRC `isOAuthProvider`. |
| RP-317-011 | `mcpInvocable.ts:219-264` | Serial per-server `oauthGrantExists` (N+1) on the rpc-proxy access path. Fail-closed per server is correct. |
| RP-317-012 | `store.ts:49-56` | Service grant types `ownerKind?: OAuthOwnerKind`, so `'mcpserver'` compiles against the owner-blind service unique. Narrow to `'recipe'`. |
| RP-317-013 | `mcpOauth.ts` / `mcpserver.yaml` / HCC `types.ts` | Shipped comments contain Spanish words (`híbrido`, `patrón`, `carril`). |
| RP-317-014 | `mcpOauth.ts:190-201` | Rate limit runs after RS256 JWT verify (unauth CPU-DoS). Flag-gated. |

---

## Verified clean

Fail-closed kill-switch (default OFF, invalid env throws, `404` before auth, single route) · overlay-only flag ON · context flavor keys strictly by `server.spec.contextRef` and rejects `context_mismatch` before any grant query · grant allowlist never admits `bearer` / `basic` / `apiKey` · rpc-proxy rail is grant-gated · PKCE HMAC-derived S256, never stored, re-derived after state verify · adapter URLs hardcoded (no SSRF) · shared bootstrap first-wins (`ON CONFLICT DO NOTHING`) · refresh does not rewrite `bootstrapped_by_user_id` · CHECK `shared ⇒ context_id + bootstrapped_by_user_id` · `legacyVersions` cannot false-skip `0100_seed_minimax_allowed_model` · `hookFetcher.ts` in the file list is whitespace-only.

---

## Must-fix on #317 vs follow-up

- **#317 must-fix at head:** none.
- **Acceptable deferred to #319:** C1 / C2 / H1 / M4 / RP-317-003 / RP-317-004 — bind `userId` to the task sender, provision `oauth:user-token` from real HCC/WRC issuance, and add negative cross-user / host / recipe / server tests **before** the flag is enabled in prod.
- **Fix before or with U5:** RP-317-002 (shared refresh persist-or-fail).
- **Deploy sequencing on this PR:** RP-317-001 (expand-then-contract the unique, or a documented write-stop / single-replica cutover).
- **Product decision:** RP-317-005 — accept oauth **names** in the catalog while the broker is off, or gate the allowlist on the same flag.

Reminder: the “until U4” deferral holds **only** while the kill-switch stays default-OFF on base/managed and no trusted producer emits `oauth:user-token`. Both are true at this head for a normal HCC/WRC reconcile. The moment #319 provisions the scope, C2 / H1 / RP-317-004 must land in the same change or the flag must stay OFF in prod.

This is analysis only — not a formal GitHub approval.

alfredolopez80 (approved):
Approved with follows

PR #320, opened 2026-08-12
feat(slack): reply to unverified senders instead of dropping their messages

jozer-rami:
## What

When an unverified Slack user messages an evenfire agent, they now get one ephemeral reply telling them how to link their account. Previously the message was dropped by policy and **nothing was said to anyone** — the bot simply looked dead, and diagnosing it required cluster logs.

This is the first of three PRs from a Slack onboarding review. It fixes the failure with 100% incidence on every new install: the one that makes a working setup look broken at the exact moment a new operator is testing it.

## Why the classification is by HTTP status

The obvious implementation is wrong and would have shipped this as a no-op.

control-api already distinguishes the cases we care about: **404 `medium_account_not_found`** for an unlinked identity, **403 `communication_channel_access_denied`** for revoked access. But `WorkflowBrokerClient.request` throws on any non-OK response, and `resolveProviderWorkflowCallerContext` does not catch, so a real unverified sender arrives at mcp-host's catch block — not as a null caller context. Classifying on "null context" would have meant the reply never fires, with tests green because they mocked the wrong boundary.

So the broker now throws a typed `WorkflowBrokerRequestError` carrying `status` and `code`, and only a recognized 404/403 classifies as `'unresolved'`.

## The safety properties, and why each exists

| Property | Why |
| --- | --- |
| Only 404/403 with a recognized code classify `'unresolved'`. 5xx, timeouts, network failures, 401, 429, unknown codes and a null context are all `'error'` | `'error'` means silence. If a control-api outage classified as `'unresolved'`, we would tell **fully-linked users** they are not linked |
| An absent `reason` means silence | An older mcp-host paired with this channel-reader behaves exactly as today, never a wrong message |
| Identical copy for unlinked and revoked users | Differing copy would leak which state an account is in |
| One notice per user per conversation per 24h, key recorded on **attempt** | If Slack keeps rejecting the post, recording on success would produce one outbound API call per inbound message forever |
| The limiter has its own 24h TTL, pruned by the existing `cleanupStaleApprovals` sweep rather than a new loop | That sweep is not generic; it iterates specific maps against hardcoded constants, so the limiter needed its own TTL added to it. Reusing the 10-minute dedupe TTL, or skipping the prune, would permanently re-silence the exact users this feature exists to help |

The notice is ephemeral (`chat.postEphemeral`), so it is visible only to the person who typed and adds no noise to a shared channel. No new Slack scope is needed: `chat.postEphemeral` uses `chat:write`, already required.

## Reviewer notes

- **A user linked to a *different* CommunicationChannel gets the same notice.** The 403 `communication_channel_access_denied` fires on an `accountRef` mismatch. This is correct given per-channel linking, and the identical-copy rule requires it, but it will look surprising.
- **`CLERUM_PROFILE_UI_URL` is optional and ships commented out in the base ConfigMap.** When unset, the notice sends without a link rather than with a broken one. The minikube overlay sets `http://127.0.0.1:3001`, matching the existing `CONTROL_API_DESKTOP_PROFILE_UI_BASE_URL`. Without this wiring the link feature would have shipped inert in every environment.
- **No production behavior changes outside the unauthorized-Slack-sender path.** Telegram, email, and Teams are untouched; `sendEphemeral` is optional on the adapter interface.

## Testing

`mcp-host` 3270 passing / 293 files (baseline 3253 / 290). `channel-reader` 354 passing / 36 files (baseline 331 / 34). `tsc --noEmit` clean in both.

`kubectl kustomize deploy/overlays/minikube` exits 1 on a clean checkout, because `patches/k8s-api-ip.yaml` is gitignored and generated at deploy time — that is pre-existing and not caused by this PR. Once the generator has run it renders clean and `CLERUM_PROFILE_UI_URL: http://127.0.0.1:3001` is present in ConfigMap `clerum-channel-reader-config`, which HCC's per-Host channel-reader Deployment consumes via `envFrom`.

Three of the eleven commits are pure test commits with no production change. Each closes a case where correct code was shipping unprotected: handler wiring that could be deleted, a whitelist that could become a passthrough, and an argument order that could be swapped into a silent 24-hour outage. Every guard in this PR was verified by breaking the implementation, watching the test fail, and reverting.

## Known follow-ups, deliberately not in this PR

- The ephemeral notice is not threaded (`thread_ts` is not passed) though normal replies are, so a notice triggered inside a thread appears at the channel root.
- The limiter key is recorded before adapter resolution, so a message that resolves to no adapter costs that conversation 24h of silence with nothing sent.
- No test asserts the ConfigMap declares `CLERUM_PROFILE_UI_URL`; a future manifest edit could drop it silently.
- The `.trim()` on the profile URL is unpinned by any test.

## Not yet proven in a real deployment

Every proof here is a unit test with a stubbed Slack client. The behavior only manifests when a genuinely unverified Slack user messages a real agent, and the reply is visible to that user alone, so it will not appear in a channel transcript.


jozer-rami:

jozer-rami:
All four addressed in `b93c20fa`. Each behavioral fix was verified by breaking it and confirming the intended test fails.

**1. Unused `vi` import** — removed.

One correction to the mechanism, since it changes who catches this next time: it does **not** trip lint. `mcp-host/eslint.config.mjs` is a bespoke architecture-invariant config declaring only `no-restricted-syntax`, with no recommended ruleset, so no `no-unused-vars` rule exists; `tsconfig.json` has no `noUnusedLocals` either. Verified by re-adding `vi` and linting the file: exit 0. The real gate is CodeQL (`codeql.yml:65` runs `security-and-quality`, which includes the unused-import query). The finding was right; only its stated mechanism was wrong. Also worth knowing: `channel-reader` has no runnable lint at all (`eslint: command not found`, no config), and no workflow invokes `npm run lint` for either service.

**2. Uncaught `await sendEphemeral`** — wrapped at the call site, with record-on-attempt ordering preserved so a conversation Slack keeps rejecting still cannot produce one outbound call per inbound message.

Your evidence was the decisive part, and better than the reasoning I had accepted earlier. A previous review argued the path was unreachable because `pollCycle` skips Slack and `handoffServer` catches. You pointed at the test having to `.catch(() => undefined)`, which proves the throw escapes `handleMessages`. That `.catch` is now gone and the test proves containment instead. I also added a two-message batch test, since containment within one message does not prove the rest of the batch survives.

**3. Code-to-status binding** — now a `Map<code, status>` requiring both to match. Two new tests pin the cross pairs (`403 medium_account_not_found` → `error`, `404 communication_channel_access_denied` → `error`). Reverting to the `Set` fails exactly those two and nothing else, which confirms the pre-existing suite never covered it.

**4. Limiter key scope** — now `providerChannelId || channelId`, matching `pendingApprovalChannelScope`. Only the key changed, not the send target: `replyChannelId` returns `msg.channelId` for Slack (its provider fallback is Teams-only), so sending on transport while scoping on identity matches the rest of the class. Noted in a comment.

On the deferred items: agreed, and the record-key-before-adapter-resolution one is the one I would also rank most user-visible, since it reproduces the exact silence this PR exists to kill on a narrower path. It is recorded in the PR body as a follow-up.

mcp-host 3270 passed / 293 files, channel-reader 354 passed / 36 files, `tsc --noEmit` clean in both.

alfredolopez80:
## Multi-agent review — `feat(slack): reply to unverified senders instead of dropping their messages`

Reviewed at head `b93c20fa` with 5 parallel agents (mcp-host/contract, channel-reader, security, tests, deploy/config), each verifying every finding against the real code. **Analysis only — this does not change the PR.**

### Recommendation: nothing blocks merge ✅

No **Blocker / Critical / High** findings. Well-engineered PR with disciplined tests and a sound fail-closed security design.

### Verified for real (not assumed)

- **Suites actually run** (exit code captured, no masking pipe): mcp-host **3270 passed / 293 files**, channel-reader **354 passed / 36 files**, `tsc --noEmit` clean in both — an **exact** match to the PR's claimed counts. No empty runs (`total > 0` everywhere).
- **Mutation-verified**: reverting the `Map<code,status>` back to a `Set` breaks *exactly* the two cross-pair tests (`providerMessageAuthorization.reason.test.ts:59` and `:64`) and nothing else. The guard tests catch the regression they claim to catch.
- **Fail-closed toward silence, confirmed in two layers** (`classifyAuthorizationFailure` in mcp-host + `reason === 'unresolved'` in channel-reader): 5xx / timeout / network / 401 / 429 / unknown code / null context all classify as `'error'` = silence. A control-api outage never tells a linked user they are unlinked.
- No account-state leak (single `UNRESOLVED_NOTICE_COPY` for 404 and 403), no injection (ids passed as API params, broker code sanitized by regex), no new Slack scope (`chat.postEphemeral` rides existing `chat:write`), no secrets/PII in logs.
- Contract end-to-end: `reason` travels intact and type-safe from the mcp-host handler through `routes.ts` to the `rpcClient` whitelist; every `authorizeProviderMessage` consumer migrated from `boolean` to `{ authorized, reason? }`.

### Previous `@claude` review — all 4 findings resolved (reproduced)

1. `Map<code,status>` now requires **both** to match (was `Set`) — `mcp-host/src/main.ts:1797-1817`; guard `expectedStatus !== undefined` blocks the unknown-code false positive.
2. `await sendEphemeral` wrapped at the call site, record-on-attempt preserved, batch not aborted — `channel-reader/src/main.ts:1576-1587`, test `unresolvedSenderNotice.test.ts:314`.
3. `providerWorkflowCallerContextClient.ts` does **not** catch — the error propagates to the handler.
4. Backward-compat: absent `reason` ⇒ silence — `rpcClient.ts:255-259` strict whitelist.

The 4 deferred follow-ups are confirmed out of scope and correctly triaged.

### New findings (none block merge)

| # | Sev | Finding | Note |
|---|-----|---------|------|
| **M1** | Medium | The `'unresolved'` path depends on **control-api** returning a thrown 404/403 with the code in JSON `.error`. If it returned `200 {}` for an unlinked sender, the feature is a silent no-op. `providerWorkflowCallerContextClient.ts:81-90`, `workflowBrokerClient.ts:146-157` | Outside this diff; **validate in a real deployment / T2**. The PR already admits "Not yet proven in a real deployment". |
| **M2** | Medium | `kubectl kustomize deploy/overlays/minikube` exits **1 on a clean checkout** (missing `patches/k8s-api-ip.yaml`, gitignored & generated at deploy-time). After the generator runs, **exit 0 and `CLERUM_PROFILE_UI_URL: http://127.0.0.1:3001` appears in ConfigMap `clerum-channel-reader-config`** (consumed by HCC via `envFrom`, `hostReconciler.ts:1860`). | Not a regression; only the body claim is misleading as written. |
| **L2** | Low | Dedup key is recorded **before** the adapter is resolved (`main.ts:1567-1575`). If no `sendEphemeral` resolves, the conversation is silenced 24h with nothing sent — the exact silence this PR kills, on a narrower path. | Already-listed follow-up; the most user-visible. Worth a tracking issue. |
| L1 | Low | `channel-reader/README.md:45` says *"verification code"* (a Telegram concept); this Slack notice is **account-linking** (`main.ts:102-104`). | Align wording to the base-YAML comment. |
| L3 | Low | The body's "its own prune loop" actually reuses `cleanupStaleApprovals` (`main.ts:2205-2211`) — with its own 24h TTL, so functionally correct, but not a separate loop. | No functional impact. |
| L4 | Low | Notice delivery failures are invisible except a `console.warn` (double swallow: `slack.ts:766-772` + `main.ts` try/catch). Intentional best-effort. | Consider a metric/counter. |
| L5 | Low | Identical 404/403 copy depends on control-api mapping **both** to `unresolved`; if it ever differentiated "revoked", that user gets silence instead of the copy. | channel-reader side is correct. |
| L6 | Low | Bounded growth of the `unresolvedNoticesSent` map (`main.ts:289`, `:1567`); pruned by the TTL sweep, each notice reaches only its own sender. | Informational. |

### Decisions for the maintainer

- **M1** is the only material open question: the logic is correct, but *that the feature actually fires* is unproven against the real control-api. If certainty is wanted pre-merge, verify against a deployed stack.
- **L2** — leave as documented follow-up, or open a tracking issue?

<sub>Multi-agent review · findings verified against code at `b93c20fa` · report is analysis-only.</sub>

PR #321, opened 2026-08-12
fix(slack): make the wrong Request URL impossible to paste (#312)

jozer-rami:
> **Stacked on #320.** Base is `feat/slack-unresolved-sender-notice`, not `dev`. Retarget to `dev` once #320 merges. Zero file overlap between the two, so they can land in either order.

## What

Connecting a Slack app had three traps that compound into one bad afternoon. This closes all three.

1. **The Control UI rendered a copyable Slack Request URL for channels with no Slack provider.** Two channels' URLs differ only in a base64 tail encoding the channel name, and both our field and Slack's input truncate right around where they diverge. Pasting the wrong one yields `409 slack_signing_secret_missing` — an error that names no channel and is visible only in cluster logs. Slack itself reports "your request URL responded with an HTTP error."
2. **control-api accepted adding a provider to a channel whose Secret lacked that provider's credentials** (#312). That is how such a channel comes to exist in the first place.
3. **Slack needs the same URL in two config places**, Event Subscriptions and Interactivity & Shortcuts, each with its own Save button. Setting only the first leaves approval buttons silently dead: Slack blocks the click client-side, so nothing reaches the cluster and every server log stays quiet.

## Why the obvious fix for #312 is a no-op

Mirroring POST's existing guard onto PUT closes nothing. `resources.ts:323` fires only when there is **neither** a `credentials` envelope **nor** a `credentialsSecretRef`, and the channel that motivated this already had a ref for Telegram. `missingCreateCredentialKey` runs only when an envelope is present, and typing an app name sends none. Both checks miss the case on POST as well as PUT.

So validation keys on the **absent → present transition**: enabling a provider requires that provider's keys, in the envelope or already in the referenced Secret. Presence of a ref proves nothing. Channels already broken in a cluster stay editable, so this cannot fail an unrelated edit.

## Design decisions worth reviewing

**Two error directions, deliberately opposite.** The write validator fails **closed** — an unreadable Secret rejects the write. The Profile-UI target listing fails **open** — an unreadable Secret drops that one target. Refusing one write is safe; blanking a user's entire verification picker because one unrelated channel has a broken Secret is not.

**Secret reads are ordered after authorization, and that is a security property.** `listSlackApprovalTargets` lists channels across all namespaces and is gated only by a session token. Reading keys inside `projectTarget` would let any authenticated profile user trigger a read of every channel's credentials Secret cluster-wide. Reads now run only for channels that passed `userCanAccessChannel`, memoized by `namespace/secretName`, so they are O(caller's channels).

**No secret value crosses any new surface.** `secretKeyNames` returns key names only, and both consumers use it as a set-membership test.

**The URL guard lives in the target-id builder**, not the page, so every derived surface — including the manifest — inherits it.

## Reviewer notes

- **`ApiException` sets `.code`, not `.statusCode`.** The first version of `secretKeyNames` branched on `.statusCode` and therefore never detected a missing Secret; it rethrew exactly like an RBAC denial. Fixed by reusing the shared `extractK8sStatus`, and pinned by a test that constructs a **real** `ApiException` rather than a fabricated object.
- **The validator is broader than Slack.** Enabling telegram, teams or email without keys is now refused too, per the plan's key table. Three pre-existing tests needed their gateway doubles to model Secret contents; each edit is mock-setup only, with assertions byte-identical.
- **`hasSlackConfig` also reads a `clerum.io/slack-bot-label` annotation**, so a channel carrying a stale annotation shows a Slack URL. Reachable but not a reopening of the trap: nothing in this repo writes that annotation, the App Name field visibly populates from it, `projectTarget` uses a spec-only predicate so it is never offered in the picker, and the write validator still refuses a credential-less save. Worth pinning with a test.
- **A non-string credential used to 500.** The whitespace fix introduced it, because `providerTransitionError` runs before credential validation and therefore sees the raw body. A present-but-wrong-type value now falls through to the validator that owns the "must be a string" 400.

## Testing

control-api 3910 passed / 61 skipped / 324 files (baseline 3889 / 61 / 321). control-ui 1373 passed / 139 files (baseline 1351 / 137). `tsc --noEmit` clean in both.

Every guard in this PR was verified by breaking the implementation, watching the intended test fail, and reverting. That discipline caught three separate **substring-shaped assertions** that looked rigorous and could not fail: an "appears exactly twice" check defeated by a suffix; a scopes list where `app_mentions:read` satisfied `app_mention`; and `mpim:read` satisfying `im:read`. All three now assert on exact emitted lines.

## Known follow-ups, deliberately not in this PR

- `DELETE /credentials/:key` has no counterpart guard, so deleting `slack-signing-secret` from a Slack-enabled channel restores the broken state. Needs a product decision (refuse, or also disable the provider).
- A present-but-empty key value passes both new gates: `secretKeyNames` reports names, so `data['slack-signing-secret'] = ''` still yields the 409 this PR targets. Reachable via kubectl, not via `PUT /credentials`.
- The single-target resolvers still accept a keyless channel, so a stale target id can still reach a 409.
- The manifest is only reachable for a **new** Slack app via the existing-app paste path; the create page has no manifest, and creating a Slack channel requires credentials that only exist after the app is installed.
- Three now-unused item-based Slack URL exports in `control-ui/lib/communicationChannels.ts` should be deleted.
- The Slack required-key list is duplicated in four places.

## Not proven in a real deployment

Every proof here is a unit or page test. Nothing has been exercised against a real Slack workspace or a live cluster.

### DIFF
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.438",
+  "version": "0.1.439",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.438",
+      "version": "0.1.439",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.433",
+  "version": "0.1.434",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.433",
+      "version": "0.1.434",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.434",
+  "version": "0.1.435",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.434",
+      "version": "0.1.435",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.435",
+  "version": "0.1.436",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.435",
+      "version": "0.1.436",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.436",
+  "version": "0.1.437",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.436",
+      "version": "0.1.437",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.437",
+  "version": "0.1.438",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.437",
+      "version": "0.1.438",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.439",
+  "version": "0.1.440",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.439",
+      "version": "0.1.440",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
         "@clerum/llm-providers": "file:../packages/llm-providers",
--- a/control-api/package.json
+++ b/control-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.436",
+  "version": "0.1.437",
   "description": "Internal Clerum control plane REST API + UI for managing CRDs and secrets",
   "main": "dist/main.js",
   "engines": {
--- a/control-api/package.json
+++ b/control-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.433",
+  "version": "0.1.434",
   "description": "Internal Clerum control plane REST API + UI for managing CRDs and secrets",
   "main": "dist/main.js",
   "engines": {
--- a/control-api/package.json
+++ b/control-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.435",
+  "version": "0.1.436",
   "description": "Internal Clerum control plane REST API + UI for managing CRDs and secrets",
   "main": "dist/main.js",
   "engines": {
--- a/control-api/package.json
+++ b/control-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.439",
+  "version": "0.1.440",
   "description": "Internal Clerum control plane REST API + UI for managing CRDs and secrets",
   "main": "dist/main.js",
   "engines": {
--- a/control-api/package.json
+++ b/control-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.438",
+  "version": "0.1.439",
   "description": "Internal Clerum control plane REST API + UI for managing CRDs and secrets",
   "main": "dist/main.js",
   "engines": {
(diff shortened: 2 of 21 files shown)


jozer-rami:

jozer-rami:
**Correction to this PR's reviewer notes, plus a fix.**

The body originally said the `clerum.io/slack-bot-label` annotation case was "reachable but not a reopening of the trap" and implied it was pre-existing behavior. That characterization was wrong on provenance, and a peer reviewer was right to push back on it.

**What actually happened, in three states:**

1. **Base branch** — no gate at all. `slackWebhookUrlForChannel(item)` rendered a Slack Request URL for every channel, which is the bug this PR exists to fix. The annotation was irrelevant because everything leaked.
2. **After the URL-guard commit (`74837891`)** — spec-only gate. Annotation-**immune**.
3. **After the draft-gate fix (`edaf0ad1`)** — the gate became `hasSlackConfig(draft)`, and `createCommunicationChannelDraft` seeds `slackBotHandle` from `spec.slackSettings?.botHandle || annotationValue(item, ['clerum.io/slack-bot-label'])` (`communicationChannelEdit.ts:86`). Annotation-**sensitive**.

So the annotation-sensitivity was introduced between steps 2 and 3, inside this PR. The net outcome was still better than base — a Telegram channel used to always show a URL and afterwards only did so carrying a stale annotation — but calling it pre-existing would have sent whoever read that later looking for it in the base branch, where it is not.

There was an irony too: the test asserting `slackWebhookUrlForChannel(telegramOnly)` is null pinned the annotation-**immune** function, which the follow-up list marks for deletion as unused. The safe version was the one being deleted; the unsafe one was the one the page called.

**Fixed in `6a5ca491`.** `CommunicationChannelDraftState` gains an optional `slackBotHandleFromAnnotation`, set only when the spec had no `slackSettings.botHandle` and the annotation supplied one. A new `hasSlackConfigForRequestUrl` delegates to `hasSlackConfig` with an inherited handle blanked, and gates the URL and manifest only. Typing in the App Name field clears the flag, so an operator still gets the URL and manifest before saving — which was the whole reason the draft gate replaced the spec gate.

`hasSlackConfig` itself is untouched, so the initial tab, the `buildCommunicationChannelSpec` payload, and the annotation-populated App Name field are all unchanged.

Rejected: a predicate that discounts a handle equal to the annotation value. It touches less but cannot tell "seeded from the annotation" from "operator typed exactly that" — on a channel labelled `Evenfire`, typing `Evenfire` would yield no URL under a hint promising one, which is the same lying-hint dead end this PR fixes.

Three mutations verified: removing the exclusion fails the annotated-Telegram test; gating on the persisted spec fails the type-before-save test; gating on `true` fails all four no-Slack-provider tests.

**Also merged the updated base.** #320 gained four review fixes (`b93c20fa`), and this branch now includes them. Post-merge: control-ui 1377 passed / 139 files, channel-reader 354 passed / 36 files.

alfredolopez80:
## Multi-agent review — PR #321

Reviewed against the real head `0dd6f13a` (base `feat/slack-unresolved-sender-notice` @ `b93c20fa`), 21 files, +1458/−40. Four agents in parallel (control-api, control-ui, security + contract, tests + typecheck), each verifying findings by **mutation** against the real code rather than reading optimistically.

### Recommendation: **APPROVE / mergeable**

Nothing blocks the merge. The three Slack traps are real, the fixes are correctly layered, and the suite is honest (exact-line assertions, not substring-shaped). The author's three security properties all hold, verified independently. Per the author's own caveat, nothing here is proven against a live Slack workspace or cluster — that remains for human validation before relying on the manifest end-to-end.

### Findings by severity

| Severity | Count |
|---|---|
| Blocker / Critical / High | **0** |
| Medium | **1** (non-deterministic test flake) |
| Low | **4** (3 already deferred by author; 1 new, reproduced) |

**[Medium] Low-frequency flake in control-ui — not a defect in the implementation.**
In 1 of ~26 cold-start runs, the 3 tests of the security fix itself (`hasSlackConfigForRequestUrl`) failed (`communicationChannelEdit.test.ts:89`, `EditCommunicationChannelPage.test.tsx:258,274`). In isolation the result is always correct and the functions are pure → this is cross-file contamination under vitest's parallel pool, not a logic bug (the implementation is proven correct by mutation). Suggested follow-up: **test isolation**; note the pre-existing global `retry:2` in `vitest.config.ts:14` can mask chronic flakiness. Non-blocking.

**[Low, new, reproduced] Misleading error on `PUT` to a nonexistent channel that declares a provider** — `control-api/src/routes/admin/resources.ts:606-622`.
`providerTransitionError` runs before `updateResource`; when the snapshot load 404s, `previousSpec` is `null`, so every provider counts as newly added and the guard returns `400 credentials["…"] is required` instead of a 404 (`updateResource` never called). Not a regression (it was an unhelpful 500 before) and not unsafe. Follow-up: short-circuit (or word the 400 generically) when the snapshot itself 404s.

**[Low, informational] The listing fail-open `.catch(() => [])` conflates all read failures** — `workflowApprovalMediumSlackVerificationService.ts:241-247`. A systemic API-server outage or a 403 is caught by the same handler as the expected missing-Secret 404, silently emptying the picker with only a `console.warn`. Within the accepted read-only design; observability follow-up, not this PR.

**[Low, cosmetic] Required-key order differs** between `SLACK_REQUIRED_SECRET_KEYS` (service) and `PROVIDER_REQUIRED_KEYS.slack` (resources). No effect (set-membership). Subsumed by the already-listed "required-key list duplicated in four places" follow-up.

### Verdicts on the prior review + author's fix

- **Finding #2 (stale `clerum.io/slack-bot-label` → copyable manifest): CLOSED by `6a5ca491`.** Verified by mutation by **three independent agents** — removing the inherited-handle blanking turns exactly the 3 fix tests red; reverting → green.
- Finding #1 (URL guard location / 3 dead item-based exports): **confirmed dead code** (grep: zero prod callers), correctly deferred.
- Finding #3 (key order): cosmetic, subsumed by dedup follow-up.
- Finding #4 (`Promise.all` fan-out): note, not a defect — bounded to the caller's channels, memoized per Secret.

### Author's security properties — all CONFIRMED
1. **No secret leak:** `secretKeyNames` returns `Object.keys(secret.data)` only — never values; no new surface (manifest, URL, logs, HTTP response) serializes credentials; `yamlScalar` escapes `" \ \r \n` (no YAML injection).
2. **Read-after-authorization:** `listSlackApprovalTargets` filters by `userCanAccessChannel` before any Secret read, memoized by `namespace/secretName` — no cluster-wide enumeration via a session token.
3. **Opposite error directions:** write validator fails **closed** (unreadable Secret → 400); listing fails **open** (dropped target + `console.warn`). Each correct for its context.

**#312 closed:** the absent→present transition validation only fires when `added.length > 0`; an already-broken channel stays editable with no Secret read. Reproduced with a disposable route test (400 + `updateResource` not called) and by `routes.providerTransition.test.ts:188`.

### Verified clean (not assumed)
`tsc --noEmit` clean in both packages (no errors in PR files; the initial `TS2307`s were an unbuilt workspace package, not the diff). Slack key contract **identical across all 5 sites**. PR suites green with `total > 0` (no zero-test false green). Exact-line assertions (three previously substring-shaped assertions confirmed fixed). No placeholders, unrequested fallbacks, or production code shaped to pass a test. The `console.log in resources.ts` warning is a **false positive** — all 5 are pre-existing; the PR adds none.

### Deferred follow-ups (all correctly out of scope)
`DELETE /credentials/:key` has no guard; an empty key value still 409s (reachable via kubectl only); single-target resolvers still accept a keyless channel; 3 unused item-based exports to delete; required-key list duplicated in 4 places; manifest only reachable via the existing-app paste path.

---
<sub>🤖 Automated multi-agent review via Claude Code (`/review-pr`). Analysis only — no approval/label/merge performed.</sub>

PR #322, opened 2026-08-12
feat: reactive OAuth consent for mcp-servers (U5), desktop connect prompt + durable resume

zach88:
## What

U5 of the mcp-server OAuth work: the **reactive consent flow**, end to end. When a conversation calls a tool on an OAuth mcp-server for which the user has no grant (or an expired/revoked one), the tool-call 401s, mcp-host attaches a typed `connect_required` marker and durably suspends the task, the desktop shows a "Connect &lt;server&gt;" prompt, the user authorizes in their browser, and the return deep-link resumes the task so mcp-host re-executes the tool with a just-in-time token.

It spans the four services the flow crosses:
- **mcp-host**: reactive 401 to `connect_required` detection, durable suspension (migration 013), resume re-exec, and the session-rejoin projection carrying the connect fields.
- **control-api**: OAuth state subject binding (`subjectKind:'mcp'`), the internal authorize-url mint endpoint (rpc-proxy-gated, membership-checked), the callback that persists the grant with a membership gate, and the return deep-link carrying `source=mcp&mcpServerName`.
- **rpc-proxy**: forwards the authorize-url mint, deriving `userId` from `auth.sub`.
- **desktop-app**: the "Connect" prompt, the IPC/authorize flow, deep-link routing and correlation, and cold-start queueing.

## Why

Per-user and per-context (shared) OAuth identities for mcp-servers need an end-user consent path. v1 is **reactive-only** by design (the proactive "connect before use" surface is an explicit non-goal): the mcp-server serves `tools/list` without auth (self-hosted image with a static catalog, which is every v1 provider), so the LLM can attempt the tool and trigger the 401 that starts consent. This change closes that loop.

## How (key decisions)

- The reactive trigger is the **live tool-call 401** surfaced as `McpAuthError(401)` in `manager.callTool`. `403` (insufficient scope) stays terminal. There is no machine loop: `McpAuthError` fires only after the single forced-refresh retry, and resume is always human-gated.
- **Durable suspension** reuses the approval-gate machinery (`reason=connect_required`, additive so an absent reason stays a byte-identical generic approval), persisted via migration 013 and projected on the REST rejoin path so a cold restart does not degrade a connect suspension into a generic approval.
- **Defense-in-depth membership**: both the shared-identity callback bootstrap and the authorize-url mint verify the consenting user is a member of the server's Context (`getUserContexts`), which is the single membership rule shared by both, not reimplemented.
- **Authorization symmetry**: the mint gates every `grantScope` by Context membership, closing a gap where a `grantScope:'user'` server could otherwise be connected across Contexts.
- **Correlation on return**: the deep-link carries `mcpServerName`, and the desktop correlates it against the FSM snapshot (robust for concurrent suspensions, survives an app restart) and resumes through the existing central decider, adding no new resume machinery.
- The recipe/sandbox OAuth paths stay **byte-identical**: the state subject discriminator and the deep-link `source` branch are additive and fail-closed.

## Testing

- **control-api**: build and 4002 tests pass. Covers the recipe wire byte-identical golden, the mcp subject round-trip and tamper rejection, the internal endpoint rejecting non-rpc-proxy callers and ignoring a body-supplied `oauthClientId`, a 403 with no authorizeUrl for a non-member at both callback and mint, and the deep-link `mcpServerName` guardian that keeps the recipe path unchanged.
- **mcp-host**: build, lint, and 3307 tests pass, PR #232 fencing intact. Covers 401 oauth to `connect_required`, 403 terminal, static 401 producing no marker, the durable round-trip on real SQLite, resume re-executing without a double approval, and the rejoin projection carrying the connect fields (repro-test fails against the pre-change head).
- **rpc-proxy**: build and 474 tests pass. Covers auth required, `userId=auth.sub` with the body ignored, and 403/404 propagation.
- **desktop-app**: build (tsc and vite) and 1539 tests pass. Covers `source=mcp` routing to resume with the sandbox path byte-identical, concurrent-suspension correlation, `connect_required` rendering "Connect", and the cold-start queue delivering a completion that arrives before the renderer is ready.

## Risks and follow-ups

- **PR size**: 74 files across 6 services, over the soft budget. U5 is one atomic feature whose flow inherently spans mcp-host to control-api to rpc-proxy to desktop, so no single service delivers testable value alone, and it was reviewed per phase (three rounds, zero Blocker/Critical/High, security SAFE, collateral clean). The earlier stacked PRs were the major-unit split; this is the U5 layer.
- **Stacking**: this targets `feat/mcp-server-oauth-host` and re-targets to `dev` as the parent PRs merge.
- **Deployment invariant**: OAuth mcp-servers must not be enabled in production until this chain ships, since the reactive isolation is the just-in-time token resolution.
- **Deferred, not in this PR**: the control-ui admin registration form (its own UI branch) and the proactive connect surface (a v1 non-goal that does not affect the reactive flow).
- **Known follow-up (pre-existing)**: the shared-grant token-resolution path in control-api has no membership re-check at call time; it is gated upstream by the same access model and is a candidate for a separate issue.


alfredolopez80 (commented):
One inline finding on stacked OAuth journey coverage.

mcp-host/src/mcp/__tests__/managerConnectRequired.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    auth: { type: 'none' },
+    enabled: true,
+    status: { deployed: true, ready: true },
+  }
+}
+
+/** Always hands out a token, so the connection has a Bearer and the 401 is real. */
+function tokenFactory(): McpTokenProviderFactory {
+  return (_server: McpServerInfo, principal: McpPrincipal) => ({
+    resolve: async () => (principal.kind === 'user' ? `tok-${principal.userId}` : undefined),
```
alfredolopez80, line 84: **Coverage gap — this test bypasses the broker and platform-issued scope.**

`tokenFactory()` always returns a usable token, so the green U5 tests cannot detect that the stacked HCC/WRC issuance path never grants `oauth:user-token`. Please add a stacked integration journey that starts with the real provisioner-issued runtime token, exercises the broker, and then reaches the consent classification. It should prove that `no_grant` leads to the intended connect flow while `insufficient_scope` remains a distinct platform misconfiguration. Do not repair this by treating every 403 as consent; repair the scope producer.

PR #323, opened 2026-08-12
fix(control-ui): stop claiming credentials exist when they do not

jozer-rami:
> **Third and last of the Slack onboarding series.** Base is `feat/slack-unresolved-sender-notice` (#320), which already absorbed #321. Merging this collapses the whole series into #320, which then goes to `dev`.

## What

The credentials panel claimed credentials existed that did not.

`ChannelCredentialsPanel` received one boolean for the whole channel and applied it **inside `fields.map`**, so every credential field rendered `**********` whenever the channel had any Secret at all. On a Telegram-only channel whose Secret held only `telegram-bot-token`, the Slack Signing Secret and Slack Bot Token fields both displayed as populated.

That is what made a half-configured channel look configured. An operator read it as "Slack is set up here" and spent an hour looking for the problem somewhere else.

## Why it took a new endpoint

`control-api/src/routes/admin/communicationChannelCredentials.ts` had only `PUT` and `DELETE`. With no `GET`, the UI had no data source for per-key presence and fell back to `!!credentialsSecretRef?.name` — the only signal available to it.

So this adds `GET /api/v1/admin/communication-channels/:name/credentials` returning `{ name, secretName, namespace, keys: string[] }`. Names only, never values.

## Design decisions worth reviewing

**An absent Secret returns 200 with `keys: []`, not 404.** "This channel has no credentials yet" is a normal state the panel must render, and a 404 would force every caller to handle two shapes for one meaning.

**Secret-name resolution mirrors the PUT** — an operator-supplied `credentialsSecretRef.name` wins, else the conventional `cc-<name>-credentials`. Without that, a channel pointing at a custom Secret would report the keys of a Secret that does not exist, reintroducing the exact lie this endpoint kills.

**Pending is a distinct state from empty.** While the GET is in flight the fields render a pending affordance with `aria-busy`. Defaulting to "empty" would flash "no credentials" on every page load before correcting itself — a worse lie than the current one, because it appears and vanishes and nobody can reproduce it.

**The reset effect keys on a sorted joined signature, not array identity.** Keying on the array would reset the panel mid-typing on any re-render and silently discard what the operator entered.

**`savedKeys`**, added beyond the plan: without it, a key the panel had just saved rendered empty, because the parent's fetched list predates the save. The same lie in the opposite direction.

## The assertion shape is deliberate, and the reason is in this file

The existing `PUT` returns `result: merged` — the full patched Secret with base64 `.data` — and its tests assert status and call arguments but never the body. That leak shipped through a green suite and is still there; #231 owns fixing it, and this PR deliberately does not touch it.

So the GET's test deep-equals the **entire** body (never `objectContaining`) and plants a sentinel in realistic wire form, asserting neither `hunter2` nor `aHVudGVyMg==` appears anywhere. A raw-only check stays green while every value leaks base64-intact.

Every guard here was verified by breaking the implementation, watching the intended test fail, and reverting. Mutations run include: reintroducing the boolean masking, keying the reset effect on `[ccName]` only, rendering empty while pending, spreading the Secret into the response, returning 404 on an empty key list, and changing one word of each copy string.

## Copy

Three sentences, each checked against the runtime rather than assumed:

- **Under the Slack credential inputs** — you do not need the `xapp-` App-Level Token or the Client Secret, and the Signing Secret is not in Slack's "app is ready" dialog at all.
- **On an empty conversation list** — no conversations confirmed yet; each user links their own from their profile page. Verified against `main.ts:693-694`: only conversations in `spec.slack` are polled, so an empty list genuinely means the channel cannot chat.
- **Beside "Answer only when the app is mentioned"** — replies inside a thread still require a mention. Verified against `slack.ts:379-387`: the mention gate has no thread exemption.

## Reviewer notes

- A failed credentials read leaves the panel in "unknown" and disables Edit and Delete, so a read-permission gap becomes a write-path outage. Deliberate and tested; flag if that trade should instead keep Edit available.
- The empty-state copy says "the verify command from their profile page" rather than a literal `verify <code>`, because that string is the shared default label and Telegram's command is `/verify <code>` while Slack's is `verify <code>`. profile-ui renders the exact per-provider string.
- A GET on a nonexistent CommunicationChannel returns 500, not 404, because `ApiException` sets `.code` while the error middleware forwards `.status`. Pre-existing and shared with the sibling DELETE; it deserves its own issue, since every k8s 403/404/409 in this codebase collapses to an opaque 500.
- The mention hint is on the edit page only. The create wizard renders the same checkbox through `ProviderPanel`, and adding it there means changing that component's API.

## Testing

control-api 3915+ passing, control-ui 1395 passing across 139 files, `tsc --noEmit` clean in both.

## Not proven in a real deployment

Every proof is a unit or page test. Nothing in this series has been exercised against a real Slack workspace or a live cluster.


jozer-rami:
Whole-branch review found three Important issues. All fixed in `e73279b0`, each verified by breaking it and watching the intended test fail.

**I1 — the panel lied in a new place, and this is the one worth reading.**

A failed credentials read left the panel showing "Checking stored credentials…" **forever**, with every input, Edit and Delete disabled, no banner, no retry, and the error object discarded. Only recovery was a page reload, which repeats the failure if the cause is persistent. Reachable whenever control-ui is newer than control-api — a window on every rolling update, permanent on mismatched self-hoster tags.

The reviewer's framing is the accurate one: *Task 1 made the state fallible, Task 2 made it blocking, Task 2 step 4 made it silent. Each is defensible alone.* A panel whose entire purpose is to stop lying about credentials should not claim a request is in flight that isn't. Same class of lie, new location.

Fixed by surfacing the failure: the page now distinguishes failed from pending and passes that down, the panel renders the existing `RetryBanner` with the underlying cause, **Edit stays enabled** (rotation is a blind PUT that doesn't need current state), **Delete stays disabled** (deleting a key you can't see isn't safe), and `aria-busy` is dropped. The implementer also routed a 200-with-unusable-body down the same path, since it produced the identical permanent brick.

One test had pinned the bug **as intent** — it asserted `aria-busy="true"` and the pending placeholder on a *failed* read. It was rewritten rather than deleted: the half worth keeping (never render as "nothing stored") survives, and the new half is that it must not render as "still loading" either.

**I2 — nothing tested what the GET asks Kubernetes for.** Three separate mutations left 18/18 green: replacing the `credentialsSecretRef` lookup with the conventional name, deleting the `getResource` call entirely, and passing the wrong namespace. Every fixture used a secret name byte-identical to the fallback, and nothing asserted call arguments — so the branch the route's own comment calls load-bearing was untested. A custom `credentialsSecretRef` would have reported another Secret's keys, reinstating exactly the lie this endpoint kills. Now covered with `getSecret` argument assertions.

**I3 — deleting the `!pending` guard bricked the create wizard with every suite green.** All three call sites in `new/page.tsx` pass `pending` without `storedKeys`, so the mutation disables every credential input during channel creation. The test that appeared to cover it passed `storedKeys`, which made the guard irrelevant to its assertion.

**M5** — copy reworded; "sending the verify command from their profile page in the conversation…" parsed as sending it *from* the profile page.

control-api 3917 passed / 325 files, control-ui 1403 passed / 139 files, `tsc --noEmit` clean in both.

Two things the reviewer verified rather than assumed, worth recording: no secret value escapes any new path (mutating the handler to spread the Secret fails both the whole-body deep-equal and the base64 sentinel), and the `savedKeys` mechanism cannot mask a key that does not exist, because it is set only after a 2xx and the PUT rejects unknown keys and drops empty values.

Remaining follow-ups, none blocking: retry discards a value typed during the failure (pre-existing reset behavior, now reachable by an explicit click); an empty-valued Secret key still renders as configured, reachable only through an operator-managed custom Secret; and the 404→500 collapse, now tracked in #324.

PR #332, opened 2026-08-12
feat(desktop): improve chat scroll behavior

josue-tech-94:
## Summary

Improves the desktop chat reading experience for long conversations.

- Shows a centered, accessible control that returns the reader to the latest messages after scrolling away from the bottom.
- Keeps the control scoped to the active conversation, with a 250 ms delay that prevents entry flicker and cancels stale timers on chat switches.
- Opens cached and remotely hydrated conversations at their latest rendered message.
- Preserves the reader position when new messages append while they are viewing older history.
- Adds provider-backed workspace coverage for control visibility, cancellation, click-to-bottom, chat switching, and transcript hydration.

## Testing

- [x] `cd desktop-app && npm test` — 161 files, 1,658 tests
- [x] `cd desktop-app && npm run build`
- [x] `npm run style-rules -- --strict`
- [x] `git diff --check`

## Checklist

- [x] This is not a new third-party MCP server / WorkflowRecipe.

### DIFF
--- a/desktop-app/ui/src/components/agents/AgentWorkspace.tsx
+++ b/desktop-app/ui/src/components/agents/AgentWorkspace.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useMemo, useRef, useState } from 'react'
 import type { RefObject } from 'react'
+import { useAgentChatActionsContext } from '@contexts/AgentChatActionsContext'
 import { useChatListContext } from '@contexts/ChatListContext'
 import { useMcpRuntimeContext } from '@contexts/McpRuntimeContext'
 import { useNavigationContext } from '@contexts/NavigationContext'
@@ -76,8 +77,10 @@ export function AgentWorkspace({ mode = 'agents', scrollContainerRef }: AgentWor
   const { sessionStateByChatId, activeChatId } = useChatListContext()
   const activeSessionState = activeChatId ? sessionStateByChatId[activeChatId] : undefined
   const { hostRuntimeStatus } = useMcpRuntimeContext()
+  const { scrollChatToBottom } = useAgentChatActionsContext()
 
   const [chatScrollNavVisible, setChatScrollNavVisible] = useState(true)
+  const [showScrollToBottom, setShowScrollToBottom] = useState(false)
   const [chatAgentRouteMenuOpen, setChatAgentRouteMenuOpen] = useState(false)
   const chatAgentRouteMenuRef = useRef<HTMLSpanElement | null>(null)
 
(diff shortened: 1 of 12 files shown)


josue-tech-94:
@claude please perform one final, independent review of this entire PR at the current head (`23bcc272cb15ed0b8490481e06e79819c3b2aef6`).

Review the full diff for correctness, regressions, race/timing issues, React state/effect lifecycle problems, accessibility, maintainability, and test quality. Please verify the implementation itself rather than relying on the PR description or existing review conclusions.

Pay particular attention to the chat scroll behavior and the recent fixes around:

* keeping delayed “scroll to latest” visibility/timers scoped to the active conversation;
* cancellation and cleanup during rapid chat switches;
* cached and remotely hydrated conversations opening at the latest rendered message;
* preserving reader position when messages append while viewing older history;
* click-to-latest behavior;
* whether the new workspace tests genuinely assert observable user behavior and would catch regressions rather than only validating implementation details.

Also look for collateral regressions introduced by the fixes, including rapid conversation switching, stale timers/effects, and interactions with existing chat behavior.

If you find anything that should block merge, please report it with the exact file/line, concrete failure scenario, impact, and recommended fix/test. If nothing material remains, please say clearly that you found no blocking issues.

zach88:
Took a deep pass through this one across the components, the scroll hooks, and a security sweep. It's in good shape — nothing blocks the merge. The visibility state machine is correctly scoped to `activeChatId`, the 250 ms timer is cleaned up on every path, and the tests that justify the feature actually fail against the base (verified the "opens a remote chat at the bottom after its empty cache hydrates" case goes from `scrollTop=0` to `800`). Only a few nits, all optional.

## Nits

1. **[R1-L2] Orphan JSDoc in useChatScroll** — `desktop-app/ui/src/hooks/domain/useChatScroll.ts:5-9`. The `/** Distance (px) from the bottom... */` block used to document the local `NEAR_BOTTOM_THRESHOLD_PX`, which this diff moved into `@constants/agents`. It's now stranded right above `interface UseChatScrollParams`, describing a threshold the file no longer defines. Worth moving to the constant's new home in `constants/agents.ts:2` or dropping it. This is the one I'd actually apply.

2. **[R1-L1] Redundant guard in the render gate** — `desktop-app/ui/src/components/agents/AgentWorkspace.tsx:74`. The `{activeChatId && showDelayedScrollToBottom && (...)}` repeats an `activeChatId &&` that's already baked into `showDelayedScrollToBottom`. Harmless, but it reads as load-bearing when it isn't. Heads up though: the `Boolean(activeChatId)` *inside* `showScrollToBottom`/`showDelayedScrollToBottom` is NOT redundant and must stay — without it, `null === null` would show the control with no active chat.

3. **[R1-L3] Hand-built cross-layer fixtures in tests** — `AgentWorkspace.chatScroll.test.tsx:276-283` builds `groupedMessages` by hand instead of from the real `ChatThreadStateContext` grouping producer, and `useAgentChatController.test.tsx:620-628` hand-writes the `turns` fixture for the `loadSessionMessages` mock. Both check out as non-shielding: the load-bearing scroll fixtures (`chatEndRef`, `scrollChatToBottom`) do come from the real `useChatScroll` hook, and the `turns` shape matches the `SessionMessagesResult.turns` wire type in `desktop-app/src/types.ts:878-886`. Cosmetic — flagging for the record, not asking for a change.

## What checked out clean

- Traced A→B chat switch, scroll-away→scroll-back, click-to-scroll, and new-message-while-reading. Every path gates the control on `=== activeChatId`, so chat A's visibility never bleeds into B, and the timer clears on both `activeChatId` and `showScrollToBottom` transitions. No stale closures.
- The two entry-scroll effects are ordered correctly: `switchToChat` batches the `activeChatId` set with the message reset, so there's never a render with the new chat key and stale messages from the old one. Non-empty caches always open at the bottom, and the pending key clears after the first scroll, so later appends fall through the near-bottom guard instead of yanking a reader off older history.
- Scroll listener re-registers only on real content change (`groupedWithKeys` is memoized), cleanup removes from the captured node — no churn or leaks.
- Constant rename `NEAR_BOTTOM_THRESHOLD_PX` → `CHAT_NEAR_BOTTOM_THRESHOLD_PX` keeps the same value (120) and now shares one source between `ChatThread` and `useChatScroll`.
- CSS vars all resolve; `pointer-events:none` on the overlay + `auto` on the button keeps it clickable without blocking scroll; button has a label and the svg is `aria-hidden`.
- Security: no new surface. Static SVG with no interpolation, doesn't touch the message-content render path, timers and listeners all have cleanup.

Suites run against the PR head: `AgentWorkspace.chatScroll.test.tsx` + `chatThread.toolSteps.test.tsx` (14 passed) and `useAgentChatController.test.tsx` (37 passed). Didn't re-run the full 1,658-test suite the PR body reports.

zach88 (approved):

PR #339, opened 2026-08-12
feat(control-api): LLM model lifecycle control plane (tolerance, impact gate, scheduled sync, attention, stale quarantine)

zach88:
## What
The control-plane half of the LLM model lifecycle work, in five cohesive pieces:
1. Asymmetric no-worsening tolerance for the allowlist write-gate, so an operator can still
   edit a Host/grant whose model fell out of the allowlist without worsening the incoherence.
2. A live model-impact module plus a ?force gate on llm-models PUT/DELETE, so disabling or
   deleting a referenced model returns 409 with the impact instead of silently breaking
   Hosts/grants. The gate is serialized with the reference writes (Host create/update, grant
   upsert) by a per-model-name advisory lock, so the 409 guarantee holds under concurrent
   writers — not just against a stale snapshot.
3. A scheduled non-destructive catalog sync cron with a deterministic sanity guard.
4. GET /admin/attention, a feed of stale-but-still-referenced models.
5. A soft, non-blocking warning when an operator assigns a stale model to something new.

## Why
When a model becomes unavailable, the system must detect it, not mask it, alert whoever can
fix it, and break nothing on the way. The cron refreshes the catalog and marks disappeared
models stale without disabling them; the attention feed and the impact-gated disable are what
keep that safe. The tolerance and the soft warning keep the operator able to act without
being trapped by a hard gate on a flappy signal.

## How
- Tolerance is role-scoped: a disallowed pair is tolerated only if it is never promoted into
  the active primary slot (a subset-only offer may still become a routable fallback), coverage
  is a superset, and the incoherence pre-existed. Each toleration emits a structured audit log,
  flushed only after the write persists.
- The impact module is the single source of references (Host spec.model / allowedModels /
  fallbacks from K8s, grant allowed_models from Postgres), reused by both the ?force gate and
  the attention feed. It is fail-loud: a K8s LIST failure fails the request closed rather than
  under-reporting.
- Availability-reduction (disable/delete/rename) and the writers that create a reference (Host
  create/update, grant upsert) share a per-model-name Postgres advisory lock, held across the
  write — for Host writes, across the K8s call itself, bounded by
  idle_in_transaction_session_timeout. Whichever side commits first wins; the loser re-reads
  under the lock and is rejected (409 with the impact, or the allowlist error), so no dangling
  reference persists. ?force still skips the enumeration and mutates without the lock.
- The cron is default-off, dedup'd across replicas by a session advisory lock, and its guard
  skips stale-marking when the live catalog looks implausible (per-provider zero floor and an
  absolute global floor), never disabling anything.
- The stale quarantine is an additive, non-blocking warning (never 422/409), orthogonal to
  the tolerance gate. Best-effort lookups log through the structured (Pino) logger, so a
  request-derived model name cannot forge or split a log record.

Packaging: one service, five pieces, because they share the same files (hostSpecValidation.ts,
pluginWorkloadSdk.ts, the llm-models routes) and the deploy builds the service as a unit;
separate PRs would repeatedly churn the same files across review rounds.

## Testing
Full control-api suite green (build, typecheck, 4126 tests). Property-based tests for the
tolerance predicate; real-Postgres integration tests for the grant impact query and for the
concurrency regressions (host↔model and reductor↔grant races, driven by real transactions with
a test-controlled commit order — skipped without a DB URL); fail-loud, persist-ordering and
log-record-integrity covered by observable HTTP tests.

## Risks / Follow-ups
- The cron ships default-off and must only be enabled after the mcp-host error-classification
  change AND the attention feed are deployed, because the remediation the alert induces
  (disable -> PUT) relies on the classification to avoid silent failover.
- MIN_PLAUSIBLE_LIVE_TOTAL is calibrated to 100 from the real snapshot; a legitimately tiny
  catalog would not mark stale (fail-safe direction), env-overridable.
- The per-model lock is held across the Host K8s write; at admin QPS this is bounded by
  idle_in_transaction_session_timeout, but it does consume a pool connection for the duration
  of the K8s call.
- Follow-up: a grant allowed_models entry with no matching promptTarget still has no
  enabled-ness gate (pre-existing, not a concurrency issue) — closing it needs a new gate, not
  serialization. Paired with a dangling-reference attention kind and a durable reconciler.
- Pre-existing: resourceService.listResource('*') swallows per-namespace LIST errors; the
  impact module avoids it, but other callers remain.


josue-tech-94:
@zach88 

## Decision

**Decision:** Blocked

**Why:** The one-service, five-part lifecycle implementation is broadly coherent and its focused
coverage is strong, but the core impact gate is not serialized with writes that create references.
It can therefore silently strand a Host or grant in the precise concurrent case the PR says it
prevents. A new plain-text warning also lets a control-plane request split or forge log records.

**Author next step:** Establish one shared, cross-replica per-model mutation protocol that covers
the destructive impact check and all Host/grant reference writes, then replace the unsafe warning
with structured logging and add deterministic concurrency/log-record regressions.

## Review summary

| Area | Result | Notes |
| --- | --- | --- |
| Correctness | Issues found | `R1-H3` defeats the destructive-mutation safety guarantee under a normal concurrent write. |
| Security | Issues found | `R1-M5` confirms the current CodeQL alert; auth and SQL scopes are otherwise sound. |
| Contracts | Pass | The attention and impact surfaces use the same source and preserve their response shapes. |
| Tests | Gaps found | Strong focused/property coverage; no deterministic race or log-record test proves the two findings. |
| PR scope/body | Needs update | 41 files are a D1 exception that is explained, but the 409 invariant currently overstates reality. |

## Action items

| Priority | ID | Severity | Action | Owner decision |
| --- | --- | --- | --- | --- |
| Must fix | `R1-H3` | High | Serialize/recheck impact versus reference persistence. | Lock/protocol design needed. |
| Must fix | `R1-M5` | Medium | Emit the lookup failure through Pino or sanitize all dynamic fields. | none |

## Must fix before merge

### R1-H3 — Impact gate races Host and grant reference creation

- **Severity:** High
- **Evidence:** `control-api/src/routes/admin/llmModels.ts:269`
- **What is wrong:** `blockedByImpact()` lists Hosts and grants before `updateAllowedModel()` or
  `deleteAllowedModel()`, but neither this path nor the Host/grant write paths use a common lock or
  recheck. The result is only a stale snapshot, not the promised safety gate.
- **User/operator impact:** A model can be disabled, deleted, or renamed without `?force=true`
  while a Host or grant ends up referencing the old pair; runtime then loses its allowed model.
- **Why this blocks merge:** It contradicts the body’s core promise that a referenced model returns
  409 instead of being silently broken.
- **Required fix:** Define a cross-replica per-pair protocol shared by model reduction and Host/grant
  writes. Hold it across impact calculation and the destructive mutation, and revalidate an
  in-flight reference immediately before it persists. The design must cover K8s and Postgres writes.
- **Required test:** Deterministically pause a Host create/update and a promptBridge grant upsert
  after their allowlist read, run a non-forced disable/rename, then release each write. Prove either
  the reduction gets 409 or the reference write is rejected; neither may persist a dangling pair.
- **Confidence:** code-verified

### R1-M5 — Plain console warning permits log-record injection

- **Severity:** Medium
- **Evidence:** `control-api/src/routes/admin/hostSpecValidation.ts:404`
- **What is wrong:** The new `console.warn` interpolates request-derived `provider`/`model` and the
  thrown error text into a line-oriented sink. Host `spec.model.name` only requires a trimmed,
  non-empty string, while allowed-model creation also permits control characters in model names.
- **User/operator impact:** A signed-in control-ui operator can store/send a model containing CR/LF;
  if the best-effort lookup fails, log aggregation can receive forged/split entries and corrupt the
  operational audit trail.
- **Why this blocks merge:** This is a new reachable log-integrity exposure, not a scanner-only
  concern. The existing Pino `rootLogger` produces escaped JSON records, but this code bypasses it.
- **Required fix:** Use the structured Pino logger with fields for the safe identifiers and error, or
  strictly reject/escape CR, LF, and other controls before any line-oriented logging.
- **Required test:** Force the lookup rejection with a model/error containing CR/LF and assert exactly
  one parseable structured log event, with no forged second record.
- **Confidence:** code-verified

## What looks good

- **Allowlist tolerance:** Role promotion is strict for Host primary/grant default roles; demotion,
  coverage-superset handling, and create rejection are centralized in the pure predicate and have
  property coverage.
- **Impact/attention authority:** `computeModelImpact` is the single reference implementation for
  both the force gate and attention feed. It covers primary, `allowedModels`, fallbacks, and the
  parameterized `allowed_models @> to_jsonb($1::text)` grant lookup.
- **Failure handling:** Per-namespace Host LIST and grant-query failures propagate instead of
  returning partial safety data. `?force` bypasses only impact enumeration, after route auth, id,
  and schema validation.
- **Catalog sync:** The default-off config, dedicated session advisory lock, release-on-unlock-error,
  transactional reconciliation, live/vendor and floor semantics, source guard, and no-disable
  behavior are correctly represented and property tested.
- **Write ordering:** Toleration audit events and stale warnings are queued until Host K8s/grant DB
  writes succeed. Stale assignment is advisory-only; it does not introduce a 409/422 gate.
- **Security:** Changed admin routes remain behind the established `/admin` control-ui auth middleware;
  grant impact SQL is parameterized; no secret or Kubernetes-policy change was introduced.

## Detailed findings

### High

#### R1-H3 — Impact gate races Host and grant reference creation

- **Priority:** Must fix
- **Severity:** High
- **Evidence:** `control-api/src/routes/admin/llmModels.ts:269`
- **Confidence:** code-verified
- **Problem:** The read-only impact computation and availability-reduction write are separated from
  the allowlist read and persistence in `resources.ts` and `pluginWorkloadSdk.ts`.
- **Scenario:** A Host create sees model M enabled, then a non-forced disable lists no references and
  disables M, then the Host create persists M. The analogous sequence works for a grant upsert.
- **Impact:** A valid-looking non-forced operation bypasses the advertised 409 protection and leaves
  an unusable/dangling reference.
- **Fix:** Add a shared distributed serialization/revalidation design, not a route-local retry.
- **Test needed:** Controlled interleaving tests for Host create/update, grant upsert, disable, delete,
  and enabled-pair rename.
- **Notes:** No relevant advisory lock/mutex appears in either the model route or reference writers.

### Medium

#### R1-M5 — Plain console warning permits log-record injection

- **Priority:** Must fix
- **Severity:** Medium
- **Evidence:** `control-api/src/routes/admin/hostSpecValidation.ts:404`
- **Confidence:** code-verified
- **Problem:** The CodeQL source is reachable: request `body.spec.model` flows to the interpolated
  message and the schema does not reject newline/control characters. The normal structured logger is
  Pino JSON, but this line is Node `console.warn`, so it gets none of that escaping.
- **Scenario:** An authenticated operator stores/selects `"model\nforged-entry"`; a transient state
  lookup failure reaches the catch block and emits multiple line-oriented records.
- **Impact:** Log/audit consumers can misattribute events or conceal the true failure context.
- **Fix:** Log through Pino structured fields (preferred) or normalize every dynamic value before the
  console sink.
- **Test needed:** A control-character regression proving one valid record only.
- **Notes:** GitHub CodeQL’s current inline comment at lines 404–406 is a real, in-scope finding, not
  a false positive or a pre-existing issue; those lines do not exist at the merge base.

## Testing and validation

### Verified

| Check | Result | Notes |
| --- | --- | --- |
| PR metadata, diff, comments/reviews, and files | passed | Reviewed remote head `db803c63e6c72b3d23629926f20509fe0c431b8d`; base merge point `673c640d…`. |
| Exact-head CI / CodeQL | passed | control-api build/test/typecheck and Postgres 16 job succeeded; CodeQL analysis succeeded but retains the inline log-injection comment. |
| Focused control-api tests | passed | 7 files, 69 passed, 3 real-Postgres tests skipped locally because `CONTROL_API_REAL_PG_ADMIN_URL` is unset. |
| `npm run build` | passed | Reviewed control-api checkout. |
| `npx tsc --noEmit` | passed | Reviewed control-api checkout. |
| `git diff --check` | passed | No whitespace errors. |

### Gaps

- Local real-Postgres coverage legitimately skipped without an admin DB URL; exact-head CI’s
  `Real Postgres integration (Postgres 16)` job passed, including its migration suite.
- The full local `npm test` was started twice from the isolated checkout; its detached runner did
  not return a terminal summary to the review harness. Treat exact-head CI as the full-suite proof.
- T3 attestations appear in the feature/fix commit messages, but no prior local review artifact was
  retained for PR #339; the missing ledger/history prevents independently attributing a prior round.

## Contract compatibility

- **Admin model mutation → Host/grant runtime contract:** issue found (`R1-H3`). The 409 contract is
  correct for a stable snapshot but not for concurrent writers.
- **Impact → attention:** compatible. Both use `computeModelImpact`; attention memoizes only Host
  list promises and preserves fail-loud behavior.
- **Stale warning response:** compatible. `warnings` is additive on successful Host/grant writes.
- **Catalog reconciliation → ConfigMap:** compatible. Enabled rows’ serialized fields are not
  changed by discovery sync; stale is intentionally non-materialized.

## Security review

- **Issues found:** yes (`R1-M5`)
- **Reviewed:** auth/authz, namespace ownership, force semantics, SQL, K8s list/error handling,
  stale provider data/URLs, logs, secrets/redaction, and dependency delta.
- **Notes:** The only new dependency is `fast-check` as a dev dependency. No introduced SQL injection,
  authz bypass, secret exposure, or unsafe forced behavior was found.

## PR description / updates

- **Current state:** needs update after `R1-H3` is fixed.
- **Required update:** Replace the absolute claim that disabling/deleting a referenced model returns
  409 with final-state wording that states the new serialization/revalidation guarantee. Retain the
  D1 five-piece explanation; current scope supports it.
- **Copy-ready text:** `The impact gate and Host/grant reference writes share a per-model concurrency
  protocol, so a non-forced disable, delete, or rename cannot race a newly persisted reference; one
  side is rejected with the actionable impact or allowlist error.`

## Follow-ups and owner decisions

### Follow-ups

- Reconcile the missing prior local review history if its committed `R1-*` references need auditable
  attribution; it was not used to invent a round or prior reviewed head.

### Owner decisions

- Choose the shared locking/validation protocol for `R1-H3`; a route-local patch cannot provide the
  required K8s/Postgres concurrency guarantee.

## Annex: pre-existing findings (out of scope)

No pre-existing findings identified. Count: Blocker 0, Critical 0, High 0, Medium 0, Low 0.

josue-tech-94:
@zach88 Remaining finding
Loq, PR body:30, confidence: reproduced.
The body promises that no dangling grant reference persists, but later acknowledges that an allowed_models entry without a matching promptTarget has no enabledness gate.
Concrete scenario: model reduction commits first, then an allowed_models-only grant upsert succeeds. This behavior predates the PR, but contradicts the body’s absolute guarantee. Change: restrict the guarantee to Host references and promptTarget references, and update the stale “4,126 tests” claim.
Required test: none for the narrowed wording; exact-head real-Postgres tests already prove both winner orders for Host and promptTarget serialization.

josue-tech-94 (approved):
Just check last comment

PR #342, opened 2026-08-12
fix(mcp-host): stabilize dual-store summary parity test against clock jitter

zach88:
## What

Makes the `DualConversationStore` test **"records matching parity for summaries written through both stores"** deterministic. It flaked ~2% of runs (seen as a red `mcp-host` job on unrelated PRs, e.g. #339).

## Root cause

Not the persist race the earlier "restore/stabilize dual store parity" fixes assumed. It's a real, benign timestamp divergence between the two stores:

- **memory** derives a summary's `lastActivityAt` from `conversation.updated_at` — stamped when the turn completes;
- **SQLite** stamps its row from `Date.now()` when the persist **worker** writes it — a few ms later.

Two independent clock reads. ~2% of runs straddle a millisecond boundary, so the inline parity probe compares e.g. `...986Z` vs `...987Z`, records `match:false`, and it never converges no matter how long the test waits. Captured by instrumenting `pagesMatch`:

```
MEM lastActivityAt = 2026-08-12T23:50:11.986Z
SQL lastActivityAt = 2026-08-12T23:50:11.987Z   (all other fields identical)
```

## Fix (test-only)

- Freeze the clock (`vi.useFakeTimers({ toFake: ['Date'] })` — **Date only**, so the persist queue's real `setImmediate`/`setInterval` timing and `vi.waitFor` keep running) on a **whole-second** boundary, which keeps SQLite's seconds↔ms round-trip exact, so both stores stamp the identical instant.
- `drainPrefix` the write chain before probing, then restore real timers and poll for the fire-and-forget `getSessionMessagesByKey` parity record.

No production code changes. Verified: 60/60 isolated runs, 15/15 full-file runs, whole persistence suite 96/96, `tsc --noEmit` clean.

## Not fixed here (reported separately)

The underlying product observation stands: in production, memory and SQLite disagree on `lastActivityAt` by write latency, so the parity **metric** emits ~2% false negatives. Making SQLite copy `updated_at` instead of stamping `Date.now()` would touch cursor ordering and cold-start TTL, so it's out of scope for a flake fix and left for a separate decision.

## Notes

- Pre-existing flake on `dev` (introduced by `0159b09` via #173), unrelated to the in-flight LLM-lifecycle PRs — it just happens to redden their `mcp-host` CI job.
- Repro is probabilistic by nature (~2%, observed 1/20–1/50 locally); no deterministic single-run repro exists.

### DIFF
--- a/mcp-host/package-lock.json
+++ b/mcp-host/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-mcp-host",
-  "version": "0.1.237",
+  "version": "0.1.238",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-mcp-host",
-      "version": "0.1.237",
+      "version": "0.1.238",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.32.0",
--- a/mcp-host/package.json
+++ b/mcp-host/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-mcp-host",
-  "version": "0.1.237",
+  "version": "0.1.238",
   "description": "MCP Host - reads Host CRD and provides LLM access via OpenAI/Claude",
   "main": "dist/main.js",
   "engines": {
--- a/mcp-host/src/core/conversation/persistence/__tests__/dualConversationStore.test.ts
+++ b/mcp-host/src/core/conversation/persistence/__tests__/dualConversationStore.test.ts
@@ -119,11 +119,9 @@ describe('DualConversationStore — dual-write parity', () => {
       await sqlite.persistQueue.drainSessionKey('u-cold:rpc:a:chat-1')
 
       const dual = new DualConversationStore(new InMemoryConversationStore(), sqlite.store)
-      const page = await dual.getSessionMessagesByKey(
-        'u-cold:rpc:a:chat-1',
-        'u-cold:rpc:',
-        { limit: 1 }
-      )
+      const page = await dual.getSessionMessagesByKey('u-cold:rpc:a:chat-1', 'u-cold:rpc:', {
+        limit: 1,
+      })
 
       expect(page?.turns.map(turn => turn.user_input)).toEqual(['cold question'])
     } finally {
@@ -151,6 +149,16 @@ describe('DualConversationStore — dual-write parity', () => {
 
   it('records matching parity for summaries written through both stores', async () => {
     const sqlite = makeSqliteStore({ cacheSize: 4 })
+    // Memory derives lastActivityAt from `conversation.updated_at` (stamped when
+    // the turn completes) while SQLite stamps its row from `Date.now()` when the
+    // persist worker writes it — two independent clock reads. They usually land
+    // on the same millisecond, but ~2% of runs straddle a boundary and the parity
+    // probe reports a spurious mismatch (the source of this test's flake). Freeze
+    // the clock so both reads are identical; a whole-second boundary keeps the
+    // SQLite seconds↔ms round-trip exact. Only Date is faked, so the persist
+    // queue's real setImmediate/setInterval timing (and vi.waitFor) still runs.
+    vi.useFakeTimers({ toFake: ['Date'] })
+    vi.setSystemTime(new Date('2026-01-01T00:00:00.000Z'))
     try {
       const memory = new InMemoryConversationStore()
       const parity: Array<{ op: string; match: boolean }> = []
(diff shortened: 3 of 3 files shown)


zach88:
@claude take a look at this

alfredolopez80:
## Independent review — findings & verified points

Every PR claim was traced to the real code in a worktree of the PR head (not assumed).

### Verified against the code

- **Parity metric compares timestamps at millisecond precision** — `dualConversationStore.ts:53` `normalizeParityValue` does `value.toISOString()` (no rounding); `pagesMatch` (`:67-68`) compares via `JSON.stringify`. `...986Z` vs `...987Z` → `match:false`. Root cause is accurate.
- **Memory `lastActivityAt` = `conversation.updated_at`** — confirmed at `conversationStore.ts:411`.
- **SQLite `lastActivityAt` = `new Date(row.last_activity_at * 1000)`** — confirmed at `sqliteConversationStore.ts:406`; the stamp originates from `timestamp: Date.now() / 1000` on the enqueue (main) thread.
- **`getSessionMessagesByKey` records parity fire-and-forget** (`queueMicrotask`, `:215-218`), while `listSessionSummariesByPrefix` records it synchronously (`:197`) — so `vi.waitFor` is only needed for the former. Correct.
- **`persistQueue.drainPrefix`** awaits every in-flight async write under the prefix (`persistQueue.ts:167`, `Promise.allSettled` over `writeChain`). Correct.
- **Why the fix is legitimate, not failure-masking**: `vi.useFakeTimers` only patches the current realm's `Date`. In production the persist worker is a **separate thread** → two genuinely independent clock reads (the real ~2% flake). In tests, `testHelpers.ts` uses an **in-process worker** (same thread), so a single frozen clock covers both stamps. Freezing the clock hides **no detectable regression** — the test validates *data* parity; timestamp identity was a harness artifact and was never a valid detector of the production divergence.
- **Whole-second boundary** (`2026-01-01T00:00:00.000Z`) is necessary: `Date.now()/1000` and read-back `*1000` round-trip exactly only on a whole second; a fractional value can lose exactness through the float64 seconds↔ms round-trip.
- **Assertions stay fail-loud**: `vi.waitFor(() => expect(parity).toContainEqual({..., match: true }))` for **both** ops throws on timeout — no soft-assert, no swallowed catch, no skip. Not weakened vs the original (same two `match:true` requirements; only the wait mechanism changed from a single `setImmediate` race to a bounded poll).
- **No fake-timer leakage**: `vi.useRealTimers()` in both `try` and `finally`; only `Date` was faked, so `vi.waitFor`'s polling timers were never faked.
- **`package.json` / `package-lock.json`**: pure `0.1.237 → 0.1.238` version bump — no dependency add/change/remove. Consistent with "No production code changes"; no supply-chain risk.
- **CI is green**, including `Build / test / typecheck (mcp-host)` — the exact job this flake reddened. (Suite not run locally: no `node_modules` in the review checkout; the mcp-host CI job is the authoritative gate here.)

### Findings

| # | Severity | Finding |
|---|----------|---------|
| 1 | Low (out of PR scope / follow-up) | In **production**, memory and SQLite stamp `lastActivityAt` from two clocks separated by write latency, so the dual-store parity **metric** emits ~2% false negatives. This PR correctly leaves it out (fixing it — SQLite copying `updated_at` instead of `Date.now()` — would touch cursor ordering and cold-start TTL) and flags it rather than silently masking it. **Suggestion:** make sure a tracked issue exists for that product decision. Non-blocking. |

No Blocker/Critical/High/Medium findings.

**Verdict: approve to merge.** Minimal, correct, test-only; root cause and fix mechanics hold up under independent verification.

alfredolopez80 (commented):
## Review summary

I reviewed PR #342 against the exact range:

- Base: `e0f73568d4aef3973e19db8dde0410808486f272`
- Head: `af4431d0cc97e55b3ea049d2c4afe347e30094cb`

This formal review adds the lifecycle/contract qualification that is not
covered by the earlier conversational LGTM note. It is a `COMMENT`, not an
approval or a request to merge.

The test-only change correctly stabilizes the observed clock-jitter flake:

- Base focused test: 56/60 isolated runs passed; 4 reproduced `match:false`.
- Head focused test: 60/60 passed.
- Persistence-focused suites, typecheck, async-leak checks, lint, and diff
  hygiene passed.
- Required remote checks, CodeQL, mcp-host build/test/typecheck, Postgres
  integration, and repository hygiene are green.
- No production code or dependency graph change is present; only the package
  version changed from `0.1.237` to `0.1.238`.

### Medium finding — narrowed parity contract

The updated test freezes `Date` for the complete-turn happy path and then
asserts exact parity. This removes scheduling jitter, but it also forces two
independently owned production timestamps to be equal.

The test therefore does not prove the broader `lastActivityAt` contract. On
both the exact base and exact head, deterministic probes reproduced `match:false`
for:

- `failTurn`
- `suspendForApproval`
- `recordToolCall`

The underlying divergence is pre-existing: InMemory projects
`conversation.updated_at`, while SQLite derives `last_activity_at` from
message timestamps and independent `Date.now()` calls. State-only transitions
and tool-call persistence do not currently have identical timestamp semantics.

This is a test-contract/coverage finding introduced by the PR, not a new
production regression.

### Recommendation

This PR can merge as a scoped test-only stabilization if:

1. The PR description and test scope explicitly state that it fixes
   deterministic write-plumbing coverage for the complete-turn path.
2. It does not claim that the broader `lastActivityAt` semantic contract is
   resolved.
3. A tracked follow-up with an owner defines the event matrix and verifies
   InMemory, hot SQLite, cold/restarted SQLite, dual parity, and cursor
   pagination.

The dual-store migration/canary should not be considered parity-certified
until that follow-up passes. A frozen-clock `60/60` result is not sufficient
evidence for general lifecycle parity.

I did not publish an approval, request changes, merge, or any code change.

alfredolopez80 (approved):
looks good

PR #344, opened 2026-08-13
fix(profile-ui): show the platform release, not the portal commit counter

jozer-rami:
## What the profile portal was showing

The settings page read `Version 0.1.119`, and the sidebar brand carried the same string as a `title` tooltip. Both came from `profile-ui/package.json`.

That number is a **per-commit counter**: `scripts/precommit/bump-staged-package-versions.mjs` bumps it on every commit touching `profile-ui/`. It is **not** a release coordinate (`scripts/release/release-coordinates.mjs`), so it can never be matched to a tag, a release, or a docs page. A user reading it learns nothing, and neither does support.

## What they show now

Both read `releaseId` from external-rest-api's release manifest via `GET /api/v1/desktop/release`, which is the same source the desktop app's Settings → Information tab already uses. On this HEAD that renders `Release v0.7.0`.

- Settings header: `Release v0.7.0`, with `Release ...` while the read is in flight and `Release unavailable` if the first read fails.
- Sidebar brand: `title="Release v0.7.0"`, and **no** title at all rather than naming a release it could not read.

Deliberate choices, each covered by a test:

- The read is cached for the page session, so the always-mounted sidebar and the settings page share one request.
- A failed **first** read is **not** cached, so one transient error does not pin the label to `unavailable` until the next full page load.
- A failed **refresh** keeps the last good value. Clearing the cache before the new read resolved used to publish `null` to every mounted instance, so a 502 on Refresh turned a correct `Release v0.7.0` into `Release unavailable` (and stripped the sidebar `title`) until the next successful refresh or a full reload.
- Refresh invalidates an in-flight read with a generation token, so the settings button is not a no-op while the mount request is still outstanding. An abandoned promise cannot clobber a newer result.
- The call passes `silentUnauthorized`, so a 401 cannot run the global auth handler and bounce the user to login. The page's own `getMe()` owns that decision.

No release-toolchain change: profile-ui already proxies `/external-rest-api/*` to the service (`profile-ui/next.config.js`), and `requireAuth` accepts the profile session cookie (`extractAuthToken` falls back to `PROFILE_SESSION_COOKIE`).

## Known trade-off

`releaseId` names the last **cut** release, so every `dev` build of the portal reads `Release v0.7.0` until the next release is cut. That is correct for what the label claims, but it no longer identifies which portal build is deployed. If we want that back, the line can carry both.

## Scope

This is the profile-ui half of a wider audit. `control-ui` shows the same class of number in two places (`control-ui/components/ControlSettingsPanel/index.tsx:485` and `control-ui/components/Sidebar/index.tsx:38`) and is **not** fixed here: its browser only reaches `/control-api`, control-api exposes no release endpoint, and the manifest lives inside external-rest-api's build context. That needs a shared manifest module plus a new endpoint, tracked separately.

The dead `DESKTOP_APP_VERSION` constant (`desktop-app/ui/src/constants/appMetadata.ts`, imported nowhere) is also left alone on purpose: it should ride along with the next real desktop-app change. `scripts/release/validate-release-version-bumps.mjs` does not mention `desktop-app` — its `PROJECTS` list is `external-rest-api` and `rpc-proxy` only — and `scripts/precommit/bump-staged-package-versions.mjs` already excludes it from `packageRoots`. Leaving the constant out of this PR is hygiene, not a toolchain constraint.

## Verification

`profile-ui`: `npm test` (75 node + 23 vitest, all green), `npx tsc --noEmit`, `prettier --check` on the PR files.

The refresh-that-fails, refresh-while-in-flight, and throwing-subscriber tests were confirmed to fail against the previous cache/inFlight split, then pass against this change.


jozer-rami:
@claude review this PR

jozer-rami:
Review addressed in 617bc302. Three applied, one declined with reasoning, one deliberately unchanged.

**Applied**

1. **`String()` coercion (`releaseIdentity.ts:22`)** — real defect, fixed. It was worse than "harmless today": an object `releaseId` would have rendered as `Release [object Object]`. Now requires `typeof raw === 'string'`, with a test covering object, array, number, and boolean payloads.

2. **The rendered instance did not self-heal** — this was the finding with actual user-visible value, so it got a real fix rather than a note. Mounted labels now subscribe to the release identity, so one resolved read lands on every instance, and the settings Refresh button routes through a new `refreshReleaseIdentity()` that drops the cached value and re-asks. Refresh only, never the mount path: re-reading on mount would defeat the per-session cache the sidebar and settings page share. A mutation that removes the subscription makes the new `heals a mounted label in place` test fail, so the behavior is actually gated.

3. **Duplicated `Release ` prefix** — `formatReleaseTitle(releaseId)` now owns it and returns `undefined` when the release is unknown, which is exactly what the sidebar brand needs (absent `title` rather than the word "unavailable"). `formatReleaseLabel` composes it, and a test asserts the two agree.

**Declined: moving `useReleaseIdentity` into `lib/`**

Checked before deciding, and the codebase argues against it. All 12 files under `profile-ui/lib` are React-free, and `tsconfig.test.json` compiles `lib/**/*.ts` into the node test run, so a hook there would drag React into that compile. A hook exported from its component's `index.tsx` is the established shape in this app: `useAuth`, `useToast`, `useProfileAccess`, `useConfirmDialog` all do it. Left a comment on the hook recording both reasons.

**Unchanged: the successful-but-empty read**

Agreed with your read that this is the correct retry behavior, so it stays. Worth noting the bound is small: only two labels mount, and the in-flight de-dup collapses simultaneous mounts into one request, so an always-empty manifest costs at most one request per label mount, not per render.

**Verification:** 70 node + 22 vitest tests green (7 new), `tsc --noEmit`, `next build`, `prettier --check` on every touched file.

jozer-rami:
@claude review again

alfredolopez80:
@jozer-rami — consolidated review of `d9764e32`, run as a 5-agent fan-out on the exact PR HEAD in an isolated worktree (correctness/React concurrency · end-to-end contract · security · tests+mutation · architecture & comment veracity). Findings marked **REPRODUCED** were executed, not read: the real module compiled with `tsc`, `renderToString`, a live `next dev` against a fake upstream, and 6 mutations applied and reverted.

**Nothing blocks the merge.** 0 Blocker · 0 Critical · **1 High** · 8 Medium · 8 Low. The High and the top Medium are the *same* design defect and fix together. Three Mediums are corrections to the PR body, not the code.

---

## H1 · A failed Refresh destroys a correct label on every mounted instance — REPRODUCED

`profile-ui/lib/releaseIdentity.ts:87-89` (catch) + `:103`

`refreshReleaseIdentity` sets `cachedIdentity = null` *before* knowing whether the new read works. When it fails, the `catch` runs `publishReleaseIdentity(null)` and **every** subscriber drops to `null`.

Scenario: user is on Settings with `Release v0.7.0` showing and the sidebar carrying `title="Release v0.7.0"`. They hit Refresh, `/api/v1/desktop/release` returns a transient 502. Settings falls to `Release unavailable` and the sidebar **loses its `title` attribute entirely**. It does not self-heal — only another successful Refresh or a full reload restores it.

```
REPRO1 after good mount     sidebarTitle= "Release v0.7.0"  settingsLabel= "Release v0.7.0"
REPRO1 after FAILED refresh sidebarTitle= undefined         settingsLabel= "Release unavailable"
```

This contradicts the file's own comment at `:20-22`:

> *"A failed read is deliberately NOT cached: the label is decoration, and pinning it to `unavailable` until the next full page load would be the wrong trade for one transient error."*

Not caching the failure avoids **persisting** it; publishing it **propagates** it. Those are two independent decisions and the comment treats one as implying the other. No test covers refresh-that-fails in either lane.

Minimal fix: don't publish `null` from the catch when a good value was already held, or don't clear the cache until the new read resolves successfully.

## M1 · `refreshReleaseIdentity` never invalidates `inFlight`, so Refresh is a no-op — REPRODUCED, found independently by 4 of 5 agents

`profile-ui/lib/releaseIdentity.ts:101-106` + `:76-77`

It clears `cachedIdentity` but not `inFlight`, and `loadReleaseIdentity:77` does `if (inFlight) return inFlight`. With a read in flight, Refresh **issues no request at all**, returns the stale value, and re-caches it.

Scenario: direct entry to `/settings/profile`. Mount fires the read against a slow backend (`next.config.js` tolerates up to 30 s of `proxyTimeout`). Two seconds in, the user hits Refresh → zero new fetches. Thirty seconds later the original read fails and the label sits at `Release unavailable`.

```
A3 fetch calls = 1
A3 text after Refresh = Release unavailable
REPRO2 mount   -> {"releaseId":"v0.7.0-OLD"}
REPRO2 refresh -> {"releaseId":"v0.7.0-OLD"}   (expected v0.8.0-NEW)
```

Refresh is inoperative in exactly the state it exists to repair. The earlier bot review called this *"benign — and arguably the safer choice"*; that read was reasoned, not measured, and the measurement disagrees.

The existing test at `releaseIdentity.test.ts:147` `await`s both loads before refreshing, so it only exercises the already-resolved case.

Fix: invalidate `inFlight` alongside `cachedIdentity` — which is what `resetReleaseIdentityCache()` already does. Note it orphans the old promise, which will still publish on resolve, so a **generation token** is needed so the abandoned read cannot clobber the newer one. H1 and M1 share one root: `cachedIdentity` and `inFlight` are a single piece of state invalidated separately.

---

## Corrections to the PR body

**M2 — the stated reason for leaving `DESKTOP_APP_VERSION` alone describes a mechanism that does not exist.** The body says touching `desktop-app/ui/src/**` would make `validate-release-version-bumps.mjs` demand a `desktop-app/package.json` bump. `scripts/release/validate-release-version-bumps.mjs:7-21` defines `PROJECTS` with exactly two entries — `external-rest-api` and `rpc-proxy` — and the file contains **zero** mentions of `desktop-app`, so `codeChanged()` (`:49-55`) cannot match `desktop-app/ui/src/**`. The sibling script states it outright at `scripts/precommit/bump-staged-package-versions.mjs:23`: *"desktop-app is deliberately absent: it is not in packageRoots (this hook never bumps it), and desktopVersion is synced by a separate desktop release flow."* Leaving the dead constant out of scope is still defensible on PR-hygiene grounds — but the written reason isn't, and a false reason outlives the PR: the next person who wants to clean up `DESKTOP_APP_VERSION` will read this and not try.

**M3 — the body is stale: `v0.6.0` → `v0.7.0`.** At this HEAD, `external-rest-api/src/releaseManifest.ts:10` is `releaseId: 'v0.7.0'` and `:13` is `desktopVersion: '0.7.0'`. Dev renders `Release v0.7.0`. The body was written on Aug 13 and wasn't refreshed after `309ba2cc` and `d9764e32`; the *Known trade-off* section carries the same stale value. (Test fixtures also say `v0.6.0`, harmless since they mock, but it confirms the drift.)

**M4 — the verification numbers don't match.** The body says *"70 node + 22 vitest tests green (7 new)"*. Actual: **72 node + 22 vitest**, and **25** new tests, not 7 (16 in `releaseIdentity.test.ts`, 7 in `ReleaseLabel.test.tsx`, 2 in `Sidebar.test.tsx`; baseline on `origin/dev` is 56 + 13). The "7" describes one file.

**L5 —** the `control-ui` paths are `control-ui/components/…`, not `control-ui/app/components/…` (that directory doesn't exist), and the settings panel line is `:485`, not `:483`. Both references do exist and the diagnosis is right.

---

## Other findings

**M5 · The `/desktop/release` payload is declared in three places nothing cross-checks** — `external-rest-api/src/routes/desktop.ts:23-31` builds the response as an untyped object literal, so the handler literal, `desktop-app/src/types.ts:67-75` and `profile-ui/app/types/api.ts:45-53` (added here) never validate against each other. Renaming `releaseId` would compile clean in all three, and `routes.desktop.test.ts:86` would stay green because it compares against the manifest itself (`releaseId: releaseManifest.releaseId` — the test moves with the code). Desktop-app would silently lose its minimum-version check and the portal would show `Release unavailable` forever. Pre-existing; this PR adds the third vertex. Worth its own issue.

**M6 · Weak assert: `Sidebar.test.tsx:99-107` passes even with the failure path fully broken.** It anchors on `waitFor(() => expect(api.getDesktopRelease).toHaveBeenCalled())` and then asserts `not.toHaveAttribute('title')` — but the title is already `undefined` on first render, so it passes before the rejection propagates. Demonstrated: a probe that breaks failure handling in `ReleaseLabel/index.tsx:30` fails 3 `ReleaseLabel` tests while `Sidebar.test.tsx` stays green 4/4. Its only real signal is against the literal-template mutation.

**M7 · The pub/sub is a second convention.** The established shape in profile-ui is a `lib/` cache plus a Context that distributes it (`lib/profileAccess.ts:13` with reset at `:75`, distributed by `ProfileAccessContext/index.tsx:32/34/158`). The cache here is consistent; the `Set<listener>` isn't — there's no other `subscribe*` in `profile-ui/lib`. `useSyncExternalStore` exists for this and the repo already uses it (`desktop-app/ui/hooks/useComposerDraft.ts:20`, `hooks/domain/useAgentChatController.ts:373`) on React `^18.3.1`. It would also fix `ReleaseLabel/index.tsx:22` reading mutable module state in the render body. Your call — no runtime impact.

**Low:** `publishReleaseIdentity` sits *inside* the `try` (`:83`), so a throwing listener is caught by the read's `catch`, which republishes `null` over the good value, throws again uncaught, and **rejects the promise** — measured: listener 1 gets `[{v0.6.0}, null]`, listener 2 gets `[]`, and `cachedIdentity` ends up inconsistent with what was published · `loading` never returns to `true`, so a Refresh doesn't return the label to its placeholder (`ReleaseLabel/index.tsx:26-43`) · the bare `catch {}` (`:85`) makes a 401, a 503, a CORS failure and malformed JSON indistinguishable from the browser — an observability gap, explicitly *not* a banned silent fallback, since the failure is surfaced on screen and never falls back to `packageJson.version` · the comment at `:17-19` ("the value cannot change while the tab is open") is true per-pod but false per-Service: during a rolling update the Service routes to a new pod and the session cache keeps serving the old id · `DesktopReleaseResponse` declares 7 fields where 1 is consumed, and `getDesktopRelease` `as`-casts over `apiGet`'s `unknown`, so the type validates nothing — the real defense is `normalizeReleaseIdentity` at runtime · `resetReleaseIdentityCache` has zero non-test callers and is exported unmarked · desktop-app declares `externalRestApiVersion?`/`rpcProxyVersion?` optional where profile-ui declares them required (M5 already drifting).

---

## What came back clean — verified, not assumed

**Concurrency & rendering.** StrictMode (`next.config.js:3`, active) yields **one** request, not two — the second effect run finds `inFlight` and reuses the promise. SSR is clean: `renderToString(<ReleaseLabel/>)` makes **0 fetches** and leaves `cachedIdentity` at `null`, because all three cache writes are unreachable during server render and the `useState` initializer only reads — so there is **no cross-request state contamination** and no hydration mismatch. No subscription leak; `active` and `unsubscribe` are not redundant, since the `.then(apply)` continuation is something no unsubscribe reaches. The copy-on-iterate at `:41` guards a real, non-obvious case.

**Contract & network.** 7/7 fields match server↔client and none can be `undefined` at runtime. The proxy claim was confirmed by *running* it — a real `next dev` against a fake upstream forwarded `url = "/api/v1/desktop/release"` and `cookie = "profile_session=…"` intact. Auth confirmed: `router.get('/desktop/release', requireAuth, …)` (`routes/desktop.ts:19`) — unlike its public `/desktop/environment` neighbour — and `extractAuthToken` does fall back to `PROFILE_SESSION_COOKIE` (`middleware/auth.ts:10-16`). "Same source the desktop app uses" confirmed at `desktop-app/src/authClient.ts:296`. No deploy change needed: the NetworkPolicy already exists and documents this exact hop (`deploy/base/profiles/networkpolicies.yaml:146-174`, egress TCP 8091, *"SSR + Next.js rewrite proxy"*). All four backward-compat cases degrade to `Release unavailable` rather than breaking.

**Security.** `silentUnauthorized` is pre-existing (`lib/api.ts:33-35`), scoped per call, already used by `AuthContext` for `getMe`, and never touches the global handler; its `clearToken()` only clears legacy artifacts that nothing in profile-ui reads, while the real session is the `httpOnly` `profile_session` cookie that `document.cookie` cannot reach. No anonymous user can read the `releaseId` — the Sidebar only mounts inside `AuthGate`. Fingerprinting surface actually **decreases**: the version used to be baked into the JS bundle where any unauthenticated visitor could read it; now it's behind `requireAuth`. No XSS sinks, no new `console.*`, no secrets, and **no manifest or lockfile touched** — zero supply-chain surface.

**Tests.** `72/72` node + `22/22` vitest, `tsc --noEmit` clean, `prettier --check` clean on all 10 PR files — all `EXIT=0` read from the command, not from a pipe, on Node v24.18.0. (`format:check` across the repo is red with 156 warnings, but **not from this PR** — none of the 10 files appear, and the leading `[error]` is `run-all.mjs` passing itself `.prettierignore`. `style-rules` exits 0 with *no output*: it's scoped to `desktop-app/ui`, so it's a gate that doesn't apply here, not a green signal.)

**Mutation testing — 5 meaningful mutations, 5 caught:** the `String()`-vs-`typeof` guard, removing `publishReleaseIdentity` from the success path, retaining a failed read, the literal-template sidebar title, and breaking failure handling outright. A sixth (`cachedIdentity = identity` outside the `if`) was *not* caught, but it's an **equivalent mutant** — `loadReleaseIdentity` only reaches that body when `cachedIdentity` is already falsy (`:76`), so it can only write `null` over `null`. No test can kill it; it is not a coverage gap.

No `.only`, `.skip`, `try/catch`, `|| true`, `retry`, `bail` or `passWithNoTests` anywhere. No placeholders, no hardcoded data, no production code written to satisfy a test. `ReleaseIdentityFetcher` is legitimate dependency injection (real default on the real path; it's what lets the `node --test` lane run without `vi.mock`), not test-shaped production code.

**Your earlier decisions hold up.** Declining the move of `useReleaseIdentity` into `lib/` is **correct**: `profile-ui/lib/` has exactly 12 `.ts` files, `grep` for `from 'react'` / `from 'next'` / `use[A-Z]` returns **zero** hits, and `tsconfig.test.json:11` pulls them into the `node --test` run. All four cited hooks (`useAuth`, `useToast`, `useProfileAccess`, `useConfirmDialog`) do export from their component's `index.tsx`. Leaving the successful-but-empty read uncached is right too. And `grep -rn 'packageJson' profile-ui/app profile-ui/lib` now returns **zero** — the PR removed both and left none behind in its scope.

---

## Recommendation

- **Blocking:** nothing.
- **Before merge (cheap):** H1 + M1 as one fix with a generation token, plus a refresh-that-fails test. Then M2/M3/M4 on the body — M2 especially, since it records a false belief about the release toolchain.
- **Follow-up:** M5 (deserves its own issue), M6, M7.
- **Optional:** all Lows.

Coverage gaps worth knowing about, none blocking: `refreshReleaseIdentity` with a read in flight, refresh-that-fails, `getDesktopRelease` and `silentUnauthorized` (tests mock `@lib/api` wholesale), `SettingsContent.tsx:147` (no test file exists), the copy-on-iterate guard, StrictMode double-mount, and `ReleaseLabel`'s `className`.

Nice work overall — the diagnosis is exact, the scope discipline around `control-ui` and `DESKTOP_APP_VERSION` is right even where its stated reason isn't, and the mutation-gated tests are the reason this review could confirm behavior instead of guessing at it.

https://claude.ai/code/session_01HanTdrdL4Zp6e3BE8JiVQu

jozer-rami:
Checked the 2026-09-02 review against `d9764e32`, then landed the validated before-merge items in 09afaba98.

**Applied**

- **H1 + M1** — one fix. `refreshReleaseIdentity` no longer nulls `cachedIdentity` before the new read. It bumps a generation token, drops `inFlight`, and re-asks. A failed re-read keeps the last good value and does not publish `null`. An abandoned in-flight promise cannot clobber a newer result. Covered by `refreshReleaseIdentity keeps the last good identity when the re-read fails`, `refreshReleaseIdentity starts a new read while one is already in flight`, and `keeps a mounted label when a later refresh fails`. Those tests failed on the old cache/inFlight split, then passed.
- **Throwing subscriber (Low)** — `publishReleaseIdentity` isolates each listener. A throw no longer rewinds the cache or rejects the read.
- **M2 / M3 / M4 / L5** — PR body. `DESKTOP_APP_VERSION` is out of scope on hygiene grounds, not because `validate-release-version-bumps.mjs` would demand a desktop bump (that script's `PROJECTS` list is `external-rest-api` and `rpc-proxy` only; `desktop-app` is already absent from `packageRoots`). Labels and the trade-off section now say `v0.7.0`. Verification is `75 node + 23 vitest`. `control-ui` paths are `control-ui/components/…` and the settings line is `:485`.

**Not in this change**

- **M5** — three untyped `/desktop/release` vertices. Pre-existing, this PR added the third. Shared contract belongs in its own issue, not a drive-by type module here.
- **M6** — Sidebar failure-path assert is still weak (title is already absent on first paint). Follow-up; the load-bearing failure coverage is in `ReleaseLabel` and `releaseIdentity`.
- **M7** — `useSyncExternalStore` vs the `Set<listener>` cache. No runtime defect now that generation + last-good are in place. Switching conventions is a separate refactor.
- Remaining Lows (refresh loading placeholder, fetch `catch` observability, session-cache vs rolling update comment — the last of those is updated in code).

alfredolopez80:
**Consolidated review and historical reconciliation — PR #344**

Reviewed head: [779d20ce](https://github.com/evenfire-ai/evenfire/commit/779d20ceeee5eb14f9107d374fecba053cdb926b). Base: `e5afb0de`. Previous reviewed head: `d9764e32`.

**Historical follow-ups remain.** The review found no currently demonstrated merge blocker or new production regression in this PR, but that does not mean every earlier observation was fixed. This comment consolidates the earlier single-Low note with the complete history.

Scope: all eight discussion comments, zero formal reviews, zero inline review threads, and both relevant repository-root notes: `.local-notes/reviews/26-09-02-pr-344.md` and `.local-notes/reviews/2026-09-03-pr-344-review.md`. The sources include the [original review](https://github.com/evenfire-ai/evenfire/pull/344#issuecomment-5278627595), [initial fixes/decisions](https://github.com/evenfire-ai/evenfire/pull/344#issuecomment-5279311927), [second review](https://github.com/evenfire-ai/evenfire/pull/344#issuecomment-5372158827), [September 2 review](https://github.com/evenfire-ai/evenfire/pull/344#issuecomment-5508990710), and [author's applied/deferred response](https://github.com/evenfire-ai/evenfire/pull/344#issuecomment-5510574601).

**Disposition:** 22 distinct historical observations: **11 corrected, 8 still present with the qualifications below, and 3 accepted decisions**. Original H/M/L labels are retained for traceability; they are not a fresh severity assignment or a count of independent runtime bugs.

**Earlier observations, item by item**

| ID | Current status | Evidence / remaining scope |
|---|---|---|
| H1 — failed Refresh destroys a good label | Corrected | `releaseIdentity.ts:96,127` preserves the last good value. Permanent Node and component regression tests cover refresh failure. |
| M1 — Refresh reuses an in-flight request | Corrected | `releaseIdentity.ts:93–99` increments the generation, clears `inFlight`, and starts a new read. The test at `releaseIdentity.test.ts:192` requires two calls and protects the newer cache. The earlier characterization of the old no-op as harmless was not supported. |
| M2 — incorrect Desktop release-toolchain rationale | Corrected | The current PR body correctly describes the two-project validator and Desktop's exclusion from the automatic package counter. Scope, rather than a nonexistent mandatory bump, explains the deferral. |
| M3 — stale `v0.6.0` description | Corrected | The body says `v0.7.0`, matching the manifest. Synthetic test fixtures may still use other versions. |
| M4 — inaccurate test counts | Corrected | The current body states 75 Node + 23 component tests and no longer claims “7 new.” Compared with base 56 + 13, the increase is 29 tests. |
| M5 — three independent release contracts | Still present; deferred maintenance | Producer literal, Desktop DTO and Profile UI DTO remain separate. Current fields agree. Two alleged consequences in the old review were incorrect; see the correction below. |
| M6 — Sidebar stays green when failure handling is broken | Still present as a narrower coverage limitation | Reconstructing a hook that ignores `identity=null` leaves Sidebar 4/4 green but fails three ReleaseLabel tests. This does not demonstrate a green whole suite or an incorrect Sidebar title; see below. |
| M7 — custom pub/sub versus existing store/context patterns | Still present; optional refactor deferred | The listener Set and local hook state remain. The alternatives are available, but they are not mandatory merely because another application uses them. Preserve loading, data initiation, SSR and cleanup if revisited. |
| M8 — fetcher injection used explicitly only by tests | Accepted decision | Production uses the real default fetcher. The historical review itself accepted this as legitimate dependency injection, with no test-only production branch or weakened behavior. |
| L1 — throwing subscriber corrupts the result | Corrected | `releaseIdentity.ts:40–49` isolates each subscriber. The dedicated test protects the resolved identity and cache. |
| L2 — loading never returns to true on Refresh | Still present; UX follow-up | A React probe confirms initial loading=true, then loading=false after success and throughout a pending Refresh. The old value stays visible. Settings has a separate general spinner, which does not await the decorative request. |
| L3 — discarded error diagnostics | Still present; observability follow-up | `releaseIdentity.ts:111–116` discards the cause and exposes no distinct diagnostic state. Browser network tools may still distinguish failures; “indistinguishable from the browser” was too broad. |
| L4 — comment claims the value cannot change while the tab is open | Corrected | The misleading immutability claim was removed. Page-session caching remains deliberate and can stay stale until Refresh/reload. |
| L5 — incorrect Control UI paths/line numbers | Corrected | Current references point to `ControlSettingsPanel/index.tsx:485` and `Sidebar/index.tsx:38`. This documentation correction does not fix the deferred Control UI behavior. |
| L6 — seven-field DTO/cast, one consumed field | Still present; qualified typing observation | The cast is not runtime validation; `normalizeReleaseIdentity` validates the field actually consumed. The six unused type members have no runtime cost, so calling them dead executable code overstated the issue. Revisit with M5 if useful. |
| L7 — exported test reset helper | Still present; optional hygiene | `resetReleaseIdentityCache` remains exported without an internal marker and is called only by tests. No affected production caller was found. |
| L8 — differing optionality between Desktop and Profile UI | Still present; related to M5 | Desktop's two service-version fields are optional; Profile UI's are required. The producer emits both, so both current clients accept the response. The difference alone does not prove a current incompatibility. |
| Original: String coercion | Corrected | A `typeof raw === 'string'` guard rejects objects, arrays, numbers and booleans. |
| Original: mounted labels cannot heal | Corrected | Subscription delivery and the Settings Refresh wiring update already-mounted labels after a later successful read. This does not promise autonomous periodic retries. |
| Original: duplicated “Release” prefix | Corrected | `RELEASE_PREFIX` and `formatReleaseTitle` own the formatting; both surfaces reuse it. |
| Original: move the hook into `lib/` | Accepted rejection | Component exports match this app's hook conventions; `lib/` remains React-free. Relocation is not a required fix. |
| Original: successful empty read is not cached | Accepted behavior | An absent/invalid identity permits a later retry; a refresh with an existing good identity preserves it. This does not fetch on every render. |

**Corrections to the historical evidence**

- **M5:** the server test at `external-rest-api/test/routes.desktop.test.ts:85` fixes the response key `releaseId` literally; it does not merely follow a renamed key automatically. Desktop's `getDesktopReleaseStatus` computes `updateRequired` from `minimumDesktopVersion`, not `releaseId`. A focused probe using exact source from base, `d9764e32`, and current head rejected a renamed response shape and retained `updateRequired=true` without `releaseId` in all three revisions (exit 0). This corrects the claimed minimum-version bypass while retaining the independent-contract maintenance follow-up. It was extracted logic with synthetic boundaries, not an Electron/server suite.
- **M6:** the reconstructed mutation was only `if (!active) return` → `if (!active || !identity) return` in a temporary copy. Result: **3 expected ReleaseLabel failures, Sidebar 4/4 passing; overall exit 1**. Sidebar uses `releaseId`, not the loading state that this mutation breaks. A separate mutation that adds an incorrect title after failure **is caught by the Sidebar test**. Both results matter: my earlier dismissal was valid only for that title mutation and was too broad as a disposition of the historical M6 claim.

**Historical coverage gaps**

Temporary review probes provide evidence about this head; they do not add permanent regression protection.

| Historical gap | Current disposition |
|---|---|
| 1: Refresh while a read is in flight | Closed for M1 by the permanent test at `releaseIdentity.test.ts:192`; not proof of every possible interleaving. |
| 2: failed Refresh over a good identity | Closed by permanent Node and ReleaseLabel tests. |
| 3: real `getDesktopRelease` wrapper / `silentUnauthorized` | Still missing from the permanent tests. The component tests mock this wrapper. Focused review probes passed, but the body's blanket “each covered by a test” overstates protection of this particular choice. |
| 4: SettingsContent / Refresh button / combined surfaces | Still missing permanently. The review's real-component button/AuthGate/Sidebar probes passed, with synthetic HTTP/router boundaries. |
| 5: listener changes during publication | Still missing. Unsubscribe-before-publish and a throwing subscriber do not exercise mutation during iteration. |
| 6a: StrictMode | Still missing permanently; a temporary hydration/StrictMode probe passed. |
| 6b: successful HTTP response with an empty body at component level | Still missing; Node normalization coverage does not cover that complete client/component path. |
| 6c: ReleaseLabel className | Still missing; tests supply classes but do not assert their propagation. The implementation currently forwards the class correctly. |

These are Node/jsdom tests, not full E2E. They can remain green if an intermediate page or the actual Refresh wiring breaks, and they do not prove protected-route access. They do catch failures in the isolated label's visible transitions.

**Other previously recorded observations**

- **Global formatting gate:** the runner still passes `.prettierignore` as a formatting target. A focused check using the head configuration and the matching locked Prettier 3.8.2 reproduces “No parser could be inferred” (exit 2). This is a pre-existing tooling problem outside the PR diff. The historical count of 156 warnings was not remeasured.
- **Style-rules gate:** zero rules apply to the twelve changed files (focused check, exit 0). Its success does not certify Profile UI's style; applicable frontend guidance was reviewed manually.
- **Control UI:** its two package-counter labels remain unchanged and explicitly outside this PR's scope.
- **Desktop constant:** `desktop-app/ui/src/constants/appMetadata.ts` still exports the unused `DESKTOP_APP_VERSION`; that optional cleanup is also outside scope.

**New Low candidate — RP-001, needs-info**

In [ReleaseLabel/index.tsx:36–37](https://github.com/evenfire-ai/evenfire/blob/779d20ceeee5eb14f9107d374fecba053cdb926b/profile-ui/app/components/ReleaseLabel/index.tsx#L36), both subscription delivery and `loadReleaseIdentity().then(apply)` update local state. With the real hook/API client in React 18.3.1/JSDOM, resolving initial request A and Refresh B synthetically in the same JavaScript turn lets A's delayed continuation apply `null` after B publishes NEW. Observed: label `Release unavailable`, cache `v0.8.0-NEW` (assertion exit 1).

The same probe with native fetch and a real local HTTP server **passes** (exit 0), with label/cache NEW. Native-browser reachability has not been demonstrated. This remains **Low, needs-info, and non-blocking**, not a confirmed production regression. The cached-load variant corroborates the same mechanism and is not a second issue. If reproduced in a browser, make delivery respect the current snapshot/generation while preserving deduplication and last-good behavior.

**Verification and recommendation**

- Exact-head suites: **75 Node + 23 component tests**, all passing; base **56 + 13**.
- Typecheck passes on both; **37 GitHub checks SUCCESS**.
- Nine additional component-integration probes passed; contract/error/escaping probes passed.
- Historical H1/M1/L1 fixes and documentation corrections were verified against code, tests and the prior head.
- No product files, tracked tests or branches were changed.

No demonstrated merge blocker remains from this evidence. Retain the explicit maintenance, UX, observability and coverage follow-ups above; do not summarize them as “all previous issues fixed” or claim full browser E2E certification.

alfredolopez80 (approved):
Approved with comments

PR #333, opened 2026-08-12
fix(desktop): improve notification tray layout

josue-tech-94:
## Summary

- Keep notification and approval cards compact and aligned to the top of the in-app tray.
- Size the desktop app-header search field to the notification-tray budget, use compact copy at constrained widths, and retain full copy at wider widths.
- Preserve full-width mobile search and remove obsolete desktop utility reservations from mobile app and mounted-app headers.
- Preserve the full search scope in the input title, bound long conversation return labels while retaining their full accessible title, and wrap mounted-app controls before overlap.
- Cover the responsive contract with deterministic producer-backed Desktop Playwright setup using real mounted-app, notification, approval, and conversation producers.
- Ensure the SDK fixture grants each capability exactly the caller policy requested by the producer-backed scenario.

## Testing / validation

- `cd desktop-app && npm test` (162 files / 1,664 tests passed)
- `cd desktop-app && npm run build` (passed; existing Vite chunk-size advisory)
- `npm run style-rules` (passed)
- Focused SDK grant-request, AppHeader, and Sandbox UI coverage; Playwright discovery; direct Prettier; and `git diff --check` passed.
- The live producer-backed Playwright scenario was not run in this lane: the configured `clerum-test` Kubernetes API endpoint at `127.0.0.1:54238` was unreachable.

## Notes / risks

- The PR targets `dev`.
- The renderer build reports the existing large-chunk advisory.

## Screenshots

Before:
Before notification-tray layout

After:
After notification-tray layout

PR #347, opened 2026-08-13
fix(desktop): tighten notification tray follow-ups

josue-tech-94:
## Summary

- Report a deterministic, sanitized Sandbox UI SDK fixture failure marker when notification emission exhausts retries.
- Align the Apps header with the shared desktop utility reservation and tighten the closed tablet header gutter to the constrained search footprint.
- Pin meaningful SDK grant-request policy fields and make the producer-backed app-launch invariant explicit.

## Follow-up context

Non-blocking follow-ups identified after [PR #333](https://github.com/evenfire-ai/evenfire/pull/333).

## Changes

- Emit `E2E_SDK_SANDBOX_UI_FAIL` for rejected mounted-fixture notification emission.
- Use the shared utility-width token for Apps and preserve the full drawer budget when the drawer is open.
- Assert the complete meaningful SDK workload grant policy for default and notification-only scenarios.
- Fail clearly if the responsive producer-backed scenario has no launched Electron app.

## Testing / validation

- `cd desktop-app && npm test` — 164 files / 1,669 tests passed.
- `cd desktop-app && npm run build` — passed; existing Vite chunk-size advisory remains.
- `npm run style-rules`, direct Prettier checks, browser geometry coverage, Playwright discovery, and `git diff --check origin/dev...HEAD` passed.
- Focused SDK grant-policy and Sandbox UI fixture failure-marker coverage passed.
- The live producer-backed Playwright scenario was not run: the configured `clerum-test` API endpoint at `127.0.0.1:54238` refused connections.

## Notes / risks

- The tablet override changes only the closed-header utility budget; drawer-open page space keeps the full drawer budget.


josue-tech-94:
@claude Please perform one final independent review of this PR at its current head.

Current expected head:
5bdbd1048bccee633c293c3131f4487e3fb5984c

Review the entire final PR diff against dev, not just the latest corrective commit.

Please pay special attention to the latest CI/browser-harness correction and verify that the browser geometry coverage is now genuinely merge-gating rather than merely discoverable or locally runnable.

Specifically check:

the required Desktop CI lane actually provisions/owns the browser needed by responsiveUtilityGutter.browser.test.ts;
the browser geometry tests run, rather than skip, in the required lane;
the 901–1220px closed-header utility reservation remains observably tighter than the full drawer budget;
drawer-open still preserves the full drawer reservation;
<=900px mobile Apps and mounted-app headers retain zero right reservation;

1220px desktop reservation remains correct;

the browser harness teardown is safe if browser launch fails and does not mask the primary error;
the geometry test still uses production styles/tokens and is not duplicating the CSS formula in test logic;
the merge-base regression distinction still holds;
the Sandbox UI fixture failure marker remains deterministic and sanitized;
raw SDK failure text is not exposed to the mounted app;
SDK grant policy assertions remain correct and do not broaden authorization;
the explicit app-launch invariant still fails loudly with safe cleanup;
no new CI, CodeQL, style, type, correctness, security, or test-quality regression was introduced by the final correction.

Please also inspect the PR's current validation claims as part of the review contract. It currently reports:

cd desktop-app && npm test — 164 files / 1,669 tests passed
Desktop build passed
style rules, Prettier, browser geometry coverage, Playwright discovery, and git diff --check passed
focused SDK grant-policy and Sandbox UI fixture failure-marker coverage passed
live producer-backed Playwright was not run because the configured clerum-test endpoint remained unreachable

Treat that cluster limitation as a separate external validation gap, not automatically as a code defect.

If you find an issue, please include:

severity;
exact file/line;
concrete failure scenario;
why it is introduced or exposed by this PR;
minimal required fix and regression proof.

If there are no merge-blocking issues, please say so plainly.

This is intended to be the final independent pass on the current head only.

zach88:
Took a deep pass through this one, including reproducing the core behavior end-to-end. Good news first: the heart of the PR checks out. The `E2E_SDK_SANDBOX_UI_FAIL` marker contract works, the sanitization actually blocks injection and secret leakage, and the `import http` migration fixes a real base crash (the fixture wasn't even starting under ESM). Nothing here blocks the merge — **no Blockers/High/Critical**. What I've got is all test-quality stuff: the changes work, but a couple of the new tests don't guard what they claim to.

## Medium

1. **[R1-M1] The PR's main change isn't covered by any assert** (`desktop-app/test/browser/responsiveUtilityGutter.browser.test.ts:65-79`)
The actual line this PR changes (`styles.css:11475`, `.apps-page-header` `360px` → `var(--app-header-utilities-width)`) only shows up above 1220px, and no assert measures it there. **Reproduced:** I reverted just that one line back to `360px` and both tests stayed green — so the test suite wouldn't catch a regression on the exact thing the PR is fixing. The desktop case (1221px) asserts `mountedPaddingRight` but never `appsPaddingRight`; on tablet/mobile the header is forced to `0`. The body invariant ("align the Apps header with the shared reservation") has no executable test behind it.
Fix: add `expect(desktop.appsPaddingRight).toBe(466)`.

2. **[R1-M3] Exact `toEqual` couples the test to env vars** (`desktop-app/test/sdkWorkloadFixture.test.ts:50-91`)
Switching `objectContaining` → `toEqual` pins literals that the producer derives from **env**: `provider:'openai'`, `model:'gpt-5.4-mini'`, `credentialSlot:'openai-api-key'` come from `MODEL_PROVIDER`/`MODEL_NAME`/`CREDENTIAL_SLOT` in `sdkWorkloadFixture.ts:24-36` (read from `E2E_WORKFLOW_MODEL_PROVIDER`/`CLERUM_MODEL_PROVIDER`/`CLERUM_MODEL_NAME`/`E2E_WORKFLOW_CREDENTIAL_SLOT`), evaluated at import and not re-stubbable. So a dev/CI with `CLERUM_MODEL_PROVIDER=claude` exported gets the producer emitting `claude`/… and `toEqual` blows up with a confusing diff. The old assertion only compared env-independent fields, so it was robust; this introduces the coupling. The rest of the pinned fields (`recipeNamespace`, `quotaLimits`, `allowedEventTypes`, `allowedUserRefs`) are genuine constants and match — verified against the producer.
Fix: derive provider/model/slot from the same constants (export them from `sdkWorkloadFixture.ts`), or drop back to `objectContaining` for just those 3 fields while keeping the exact pin on everything else.

3. **[R1-M2] First browser-dependent test in the default `npm test` suite** (`desktop-app/test/browser/responsiveUtilityGutter.browser.test.ts`)
**Reproduced:** at the merge-base there's no `test/browser/` and no `chromium.launch`/`playwright` import in the non-e2e suite — `npm test` didn't need a browser. Now `beforeAll` launches Chrome/chromium, so a dev without system Google Chrome (mac) and without `npx playwright install chromium` gets a hard failure of the whole file. CI is covered by the install step you added, but this changes the `npm test` contract and local DX. Suggestion: a separate `test:browser` script, or gate on browser availability, or at least document the requirement. (The real fix lives in `vitest.config.ts`/CI, but the coupling comes from where this test sits.)

## Smaller stuff

4. **[R1-L1] macOS runs against system Chrome, not the pinned Playwright chromium** (`responsiveUtilityGutter.browser.test.ts:19-37`) — deterministic for the current asserts (padding + `clamp` over `vw`, no font metrics) and it passed, but mac and CI run different engines/versions; a future metric-sensitive assert could diverge. Prefer the bundled chromium for parity.

5. **[R1-L2] `freePort()` has a TOCTOU** (`sandboxUiFixtureFailureMarker.test.ts:24-31`) — opens a server on port 0, closes it, returns the port, and the fixture rebinds via `PORT`. Between close and bind another listener can grab it → occasional `EADDRINUSE` flake → "did not start". Low risk. Retry the start or pass the already-listening server.

6. **[R1-L3] FAIL matcher is anchored, OK matcher isn't** (`sdkWorkloadFixture.ts:309`) — `findSdkSandboxUiFailureMarker` uses `/^…$/m` while the sibling OK matcher (`:298`) is unanchored. Today `kubectl logs` doesn't use `--timestamps` so there's no bug, but if someone adds it the FAIL detection breaks silently while OK keeps working. Worth a comment pinning the "logs have no timestamp" premise.

7. **[R1-L4] Test DOM fixture combines headers that never coexist** (`responsiveUtilityGutter.browser.test.ts:29-37`) — puts `.apps-page-header` and `.sandbox-ui-mounted-header` as siblings under one `.page`, which the real app never renders (`SandboxUiPage.tsx` shows one or the other per `launch.kind`). **Verified** the critical nesting `.content-panel > section.page > .sandbox-ui-mounted-header` does match the real DOM, so the measured per-element padding is valid; the extra sibling is just cosmetic.

## What checked out clean

- **Marker contract end-to-end (reproduced):** `log()` writes without a timestamp so the waiter's `^…$/m` anchor matches. Tested against real output for `caller_not_allowed`, `http_403`, `notification_emit_failed`, `missing_endpoint_or_token` and multiline blocks. The waiter detects the failure instead of hanging.
- **Sanitization / no-injection / no-leak (reproduced):** a `code` with a space or `\n` fails the strict regex and collapses to `notification_emit_failed`. No way to inject a fake marker or leak the raw message; the response body is now constant, and `PLUGIN_WORKLOAD_SDK_TOKEN` / the endpoint never reach a sink. Security pass: clean.
- **`import http` fixes a base crash (reproduced):** at the merge-base the fixture crashes on startup (`ReferenceError: require is not defined in ES module scope`); at HEAD it starts fine. Necessary and correct.
- **Gutter math (reproduced):** at 1100px `32vw=352` inside `clamp(160,·,420)`; closed `352+10+36=398`; drawer-open `420+10+36=466`; desktop 1221 `=466`; mobile `=0`. All confirmed measuring real geometry in Chrome.
- Suites the review ran: `responsiveUtilityGutter.browser.test.ts` → 2/2; `sandboxUiFixtureFailureMarker.test.ts` + `sdkWorkloadFixture.test.ts` → 6/6 (both stand up the real fixture).

Net: fine to merge as-is on correctness. M1 and M3 are cheap and worth landing before merge since they're about the tests actually guarding the change; M2 and the small stuff can be follow-ups.

josue-tech-94:
@claude Please perform one final independent review of this PR at its **current head**.

Current expected head:

`fe23f691747e35698514d6baccacfd7aef7210ec`

Please review the **entire final PR diff against `dev`**, not just the latest commits.
Please pay particular attention to:

- the Apps desktop header reservation above 1220px:
  - `.apps-page-header` should use the shared desktop utility reservation;
  - the real-browser test should explicitly assert the Apps desktop padding, not only the mounted header;
  - reverting that production reservation to the old `360px` value should make the test fail;
- the responsive geometry contract:
  - `<=900px`: Apps and mounted headers have `0px` closed reservation;
  - `901–1220px`: closed mounted-header reservation uses the constrained tablet/search footprint;
  - drawer-open keeps the full drawer reservation;
  - `>1220px`: Apps and mounted headers use the full desktop reservation;
- the browser geometry test remains **actually merge-gating**:
  - Chromium is provisioned in the required Desktop CI lane;
  - the suite runs rather than skips;
  - launch failures remain visible and teardown does not mask them;
- the SDK grant-policy test:
  - environment-derived provider/model/credential-slot values come from the same authoritative configuration used by the producer;
  - the test does not duplicate environment precedence logic;
  - ambient provider overrides do not cause false failures;
  - stable authorization policy remains pinned exactly, including callers, event types, user refs, quotas, prompt-target structure, default target, and allowed models;
  - no production authorization scope was broadened;
- the Sandbox UI failure marker remains deterministic and sanitized:
  - raw SDK/transport error text or secrets cannot leak;
  - success, terminal failure, unsafe failure, missing-endpoint, and waiter parsing behavior remain sound;
- the producer-backed app-launch invariant still fails explicitly and cleanup is safe;
- the two recent `dev` synchronization merges did not create any semantic conflict with this PR's implementation.

Please also review the current PR description as the final-state contract and check that its validation claims remain accurate.

Current reported validation includes:

- `cd desktop-app && npm test` — 164 files / 1,669 tests passed;
- Desktop build passed;
- style rules, Prettier, browser geometry coverage, Playwright discovery, and `git diff --check` passed;
- focused SDK grant-policy and Sandbox UI failure-marker coverage passed;
- CI and CodeQL are green;
- live producer-backed Playwright was not run because the configured `clerum-test` endpoint remained unreachable.

Treat that live-cluster limitation as a documented external validation gap, not automatically as a code defect.

Also re-triage the previously noted non-blocking items only if they are still relevant:
- local `npm test` browser dependency / DX;
- macOS system-Chrome fallback;
- `freePort()` TOCTOU;
- FAIL/OK marker anchoring;
- synthetic header siblings in the geometry harness.

Do not turn those into findings unless there is a concrete current-head failure or meaningful merge risk.

If you find an issue, please include:

- severity;
- exact file/line;
- concrete failure scenario;
- why it is introduced or exposed by this PR;
- minimal required fix;
- required regression proof.

If there are **no merge-blocking issues**, please say so explicitly.

This should be a final independent review of the current head only.

zach88 (approved):

PR #350, opened 2026-08-13
feat(desktop): add keyboard shortcuts and command palette

josue-tech-94:
## Summary

- Adds a centralized Desktop command/shortcut registry shared by runtime routing, the command
  palette, and the read-only Settings → Shortcuts table.
- Adds memory-only, keyboard-driven chat view tabs with non-destructive close behavior.
- Shows shared running and awaiting-approval status indicators before open chat tab titles.
- Routes `Mod+F` to the existing global Desktop search and `Mod+Shift+F` to current-content search.
- Uses one immutable Markdown semantic model for loaded-chat matching and rendered highlights.
- Corrects native Electron app-find lifecycle/focus handling and matches Chromium's yellow/orange
  find-result treatment in chat.
- Adds `Mod+Shift+L` composer focus and a Spotlight/VS Code-style `Mod+K` command palette.
- Focuses every newly opened chat composer and provides native selection/edit context menus.

## Command palette actions

The palette includes the shortcut commands plus these approved actions:

- Open Settings — `Mod+,`
- Log out
- Go to Chat
- Go to Apps
- Go to Agents
- Open notifications and approvals
- Go to Plugins
- Go to Contexts
- Go to Teams
- Go to Connectors
- Go to Global File System
- Toggle sidebar
- Refresh current app
- Back to Apps
- Back to conversation

Only Open Settings receives a keyboard binding in this action set. The remaining actions are
palette-only and invoke existing navigation, session, notification, sidebar, or native-app owners.
The Settings shortcut table omits palette-only actions with no binding.

## Search behavior

- `Mod+F` opens and focuses the existing top-right global Desktop search from host or sandbox focus.
- `Mod+Shift+F` searches only the active content.
- Current-chat search is case-insensitive and covers only messages already loaded in renderer state.
  It performs no history pagination, backend fetch, or chat mutation.
- Loaded-chat match counts and rendered marks derive from the same normalized unified/remark/HAST
  semantic model, including Markdown links, code, GFM structures, and inert raw HTML text.
- Current-app search is scoped to the active trusted sandbox `WebContents`. The renderer/preload/IPC
  boundary uses semantic `start`/`next`/`previous` operations; main owns Electron request IDs,
  document generations, readiness, stale-result rejection, and teardown.
- No backend full-chat-search service or plugin SDK contextual-search API was added. OD-003-1 remains
  deferred/open.

## Interaction and visual behavior

- One `Mod+Shift+F` from a focused app opens contextual search with its trusted input focused and
  ready for immediate typing.
- Repeated invocation refocuses without clearing the query; dismissal restores the active app
  through the existing trusted focus path.
- The command palette owns Escape across its modal scope and dismisses only on a true backdrop click,
  preserving focus restoration and preventing accidental command execution.
- The command palette focuses its search field on open and keeps ArrowUp/ArrowDown selections visible.
- New chats opened from the command registry or sidebar focus the composer immediately.
- Right-clicking selected host or app text offers Copy; editable controls offer Cut, Copy, and Paste,
  including Paste when no text is selected.
- Registry-derived shortcut labels render `Shift` as literal text while retaining platform-correct
  Command/Ctrl matching.
- The selected chat tab and current chat panel share one responsive selected workspace across
  long/multiple tabs, local find, loading, blank, and active chat states.
- Open persisted-chat tabs show running and awaiting-approval indicators from the shared cross-chat
  session lifecycle; tabs do not start their own background processes or streams.

## Tab semantics

- `Mod+T` creates a blank chat view tab; `Mod+W` closes only the active view tab.
- Closing a tab never deletes a persisted chat or terminates its persistence/task state.
- `Mod+1..8` selects numbered tabs, `Mod+9` selects the last tab, and Control+Tab /
  Control+Shift+Tab cycles with wrapping.
- Tabs are memory-only across restart in v1.

## Architecture and security

- Uses main-process, focused-`WebContents`-aware input routing with renderer-owned application actions.
- Exposes only narrow typed trusted-renderer preload/IPC surfaces; bounded find requests target only
  the active sandbox view.
- Native text context menus execute only against the originating trusted renderer or sandbox
  `WebContents`; no clipboard content crosses into the host or another app.
- Adds no plugin-facing Node, Electron, DOM, focus, shortcut, clipboard, or find privilege. Sandbox
  partition isolation and external-navigation protections are unchanged.
- Removes the CodeQL-flagged Markdown projection regex instead of suppressing it. Markdown raw HTML
  remains inert text, unsafe URLs retain ReactMarkdown's transform, and dense matching is linear.
- Palette Log out invokes the established controller and is disabled while the existing app busy
  state prevents concurrent session cleanup.

### Passed

- Per-finding focused Vitest suites, including active-element, dismissal, formatter-consumer,
  workspace-state, and app-switch request-identity regressions.
- `cd desktop-app && npm test` — 187 files and 1,810 tests passed.
- `cd desktop-app && npm run build` — passed; existing Vite chunk-size warning only.
- `npm run style-rules` — passed.
- Changed-file Prettier and full correction-range `git diff --check` — passed.
- `cd desktop-app && npm run test:native:sandbox-shortcut-focus` — passed against actual macOS
  Electron/WebContentsView focus transfer, immediate typing, repeat invocation, close, and native
  focus restoration.
- `cd desktop-app && npm run test:native:sandbox-find` — passed against actual macOS Electron
  WebContentsView find lifecycle.
- `cd desktop-app && npm run test:native:chat-workspace-geometry` — passed wide/narrow geometry with
  multiple and long tabs.

### Skipped or unavailable

- Targeted packaged Playwright coverage was blocked before tests by the Kubernetes context guard:
  expected `clerum-test`, actual `gke_eventfire-491421_us-central1-a_clerum`. Kubernetes was not
  changed and the guard was not bypassed.
- Native Windows verification was unavailable. Windows parity is not claimed; registry
  matcher/label tests cover the intended Ctrl mapping.

## Review

- Round 3 correctness review: PASS after closing `R3-M1`, `R3-M2`, `R3-L1`, `R3-M3`, and the
  correction-caused request-identity regression `R3-M4`.
- Round 3 security review: PASS / SAFE; no new plugin, preload, Electron, or navigation privilege was
  added.
- Round 3 collateral review: CLEAN for
  `b629317a788dc728ec7eb4e6825f28dd271878cf..59af00e94840ebe72e840f41188fcfa3a084a71b`
  after `R3-M4` was reproduced and corrected.

- Post-Round-3 tab-status refinement correctness review: PASS.
- Post-Round-3 tab-status refinement security review: SAFE; the renderer-only reuse adds no IPC,
  plugin, authentication, storage, or network surface.

- Post-Round-3 interaction refinement correctness review: PASS.
- Post-Round-3 interaction refinement security review: SAFE; context-menu actions remain
  user-triggered and source-`WebContents`-scoped with no new bridge privilege.

## Scope

The PR remains Desktop-only and introduces no cross-service or backend-search implementation. It
touches 86 files, above the soft 25-file budget, because the cohesive feature and its correction
proof cross Electron main, typed preload/IPC, native WebContents lifecycle, renderer state/UI, shared
Markdown semantics, established component tests, and native/packaged coverage. Splitting those
layers would make the registry/runtime/UI contract and trust boundary harder to review consistently.


josue-tech-94:
@claude Please perform a final **full review of this PR at the current head**, including the latest sync/update and newly added change.

Review the complete PR, not only the most recent commits, with particular attention to:

* correctness and regressions
* Electron main / preload / IPC / renderer boundaries
* keyboard shortcut routing and editable-input behavior
* chat tabs and focus/state lifecycle
* global and contextual search, including native app/WebContentsView search
* command palette actions, dismissal, and focus restoration
* chat search highlighting and Markdown rendering
* responsive/visual behavior of the selected chat workspace
* security-sensitive changes and sandbox/plugin privilege boundaries
* test quality and whether the tests meaningfully protect the behavior

Please report only concrete issues attributable to this PR, with file/line evidence and a clear failure scenario. Distinguish merge-blocking issues from non-blocking suggestions.

zach88:
Took a full second pass through this one across the main/IPC/native layer, the pure renderer libs, the components, security, contract consistency, and a collateral audit of the fixes you pushed since round 1. Short version: this is in good shape. The new work (tab status indicators, composer focus, the extracted `CommandPaletteList`, palette-input focus, native context menus) is correct and well tested, the trust boundary holds, and the collateral pass came back clean (no regressions from the round-1 fixes, with the two fix-tied tests confirmed to fail against the old head). What's left is quality/debt, plus one small UX decision that's yours to make.

## Verdict

🟡 **Approve with nits**

- **Optional / non-blocking: 15** (2 Medium, 13 Low/nit)

None. Zero Blocker/Critical/High, and the collateral audit found zero open regressions (0%).

## Optional (non-blocking)

The PR can merge without these. The two Mediums have now been open two rounds, so per our own rule they either get bumped or closed out in writing rather than floating into a third round; my rec is on each.

**Mediums (please make a call, don't let them float another round):**

1. **[R1-M1] Semantic models rebuild on every streaming token.** `desktop-app/ui/src/lib/chatMessageSemantics.ts:151` + the consumer memo at `desktop-app/ui/src/App.tsx:1644`. `buildLoadedChatSemanticModels` re-parses every message through the full remark to rehype pipeline whenever `vm.activeMessages` gets a new array reference, and that memo is unconditional (not gated on search being open), so during streaming you re-parse O(total transcript) per token even with search closed. `contentRevision` (`:23,142`) looks like the scaffold for per-message memoization but nothing reads it, so the optimization was never wired. Reproduced (consumer + memo deps + grep for the dead field). Not a correctness bug (fuzzed 13k+ runs clean), it's perf debt. My rec: close this in writing and open it as a perf issue, or key a per-`messageId`+`contentRevision` cache if you'd rather fix it here. Still open since round 1.

2. **[R1-M2] The body's headline invariant has no executable test (T5).** `desktop-app/ui/src/App.tsx:337` (`handleCloseChatViewTab`). The description promises "closing a tab never deletes a persisted chat or terminates its persistence/task state". The code is correct today (it only mutates in-memory tab state and calls `revealChatViewTab`, never a delete/end-session controller), but the only test covers the pure function, which structurally can't touch persistence, so the App-level wiring is unverified. Reproduced (read the handler path). My rec: fix it here, it's cheap and it's the star invariant of the PR: render, close the active persisted chat's tab, assert the delete/end-session mock wasn't called and the tab reopens. Still open since round 1.

**Lows / nits:**

3. **[R2-L1] Sidebar "new chat" now always stacks a blank tab, no reuse, no test.** `desktop-app/ui/src/App.tsx:349` (`handleNewChatViewTab`) + `desktop-app/ui/src/components/SidebarNav/index.tsx:369`. The `nav-new-chat` button went from `handleSelect(DESKTOP_ROUTES.chat)` (navigate only) to `onNewChat()` which calls `addBlankChatViewTab` (always appends). There's a `focusBlankChatViewTab` that reuses an existing blank, and it isn't used here, so if you're already on an empty "New chat" tab, each click stacks another empty one. The `SidebarNav` test only checks `onNewChat` fired; nobody covers append-vs-reuse at the App level. This isn't a regression (the new path is a superset of the old navigation), it's a UX call: should the sidebar reuse an existing blank or always stack? Either way it wants a test.

4. **[R2-L2] Non-exhaustive eligibility fallthrough.** `desktop-app/src/desktopCommands.ts:542`. `isDesktopCommandEligible` ends the `composer-available` case with a catch-all `return context.composerAvailable`, so a future `DesktopCommandEligibility` value with no branch silently falls through to `composerAvailable`. Suggest an explicit `eligibility === 'composer-available'` check plus a `never` exhaustiveness fallthrough.

5. **[R2-L3] Allowlist sync test is one-directional.** `desktop-app/src/__tests__/desktopCommands.test.ts:42`. It asserts registry ⊆ preload but not preload ⊆ registry, so a command dropped from the registry leaves a dead id in `DESKTOP_COMMAND_IDS` undetected. Harmless today; the "synchronized" comment just overpromises. Add the reverse assertion.

6. **[R2-L4] tsc possibly-undefined in prod code.** `desktop-app/ui/src/lib/chatMessageSemantics.ts:199`, `flowFragments[fragmentCursor]?.end <= start`. Functionally inert (`undefined <= n` is `false`, the loop stops), and it doesn't gate build or CI, but it's a real strict-mode error. `(flowFragments[fragmentCursor]?.end ?? Infinity) <= start` clears it.

7. **[R2-L5] The `binding non-null ⇒ editingPolicy: 'allow'` invariant isn't pinned.** `desktop-app/src/shortcutRouter.ts` / `desktop-app/src/desktopCommands.ts`. The root of the round-1 concern is effectively fixed now (the new `isStandardEditingShortcut` bails out of `preventDefault` for standard editing chords), but nothing tests the invariant, so a future command with a binding plus a `suppress` policy would eat an editing chord with nothing red. A one-line registry assertion covers it.

8. **[R1-L2] Property-test coverage gaps (T2).** `desktop-app/ui/src/lib/chatViewTabs.ts:105` (`closeChatViewTab` is excluded from the tabs property test) and `desktop-app/ui/src/lib/chatMessageSemantics.ts:175,242` (`findChatSemanticMatches` / `annotateChatSemanticTree` have none). The code is actually correct (fuzzed 13k+ runs green), so this is coverage debt, not a latent bug. Still open since round 1.

9. **[R1-L4] Palette Escape doesn't `stopPropagation`.** `desktop-app/ui/src/components/CommandPalette/index.tsx:76`. It `preventDefault`s and closes but the native event still reaches window-level Escape listeners (`ComposerPanel.tsx:245`, `ChatThread.tsx:376`). Only matters if one of those transient UIs is live under the palette. Still open since round 1.

10. **[R1-L5] cwd-dependent style tests.** `desktop-app/ui/src/components/ChatTabs/ChatTabs.test.tsx:57` and `desktop-app/ui/src/lib/__tests__/searchHighlightStyles.test.ts:5`. `readFileSync(process.cwd() + '/ui/...')` only passes when vitest runs from `desktop-app/`, and they regex the CSS text rather than asserting observable style (brushes T4). Still open since round 1.

11. **[R1-L6] Asserting a production-dead intermediate (T4).** `desktop-app/ui/src/lib/chatMessageSemantics.ts:26,145`. `searchText` has no non-test consumer, yet the tests treat it as the primary observable instead of the match list / marks. Assert those, or drop the field. Still open since round 1.

12. **[R1-L7] Type-drift in new tests (T1).** Under `ui/tsconfig` (doesn't gate CI, which only typechecks `src/**`). The T1-flavored ones: `desktop-app/ui/src/__tests__/desktopShortcutLabels.test.tsx:63` hand-writes a `DesktopShortcutInput` fixture missing the required `isAutoRepeat`, and `desktop-app/ui/src/pages/__tests__/SandboxUiPage.focus.test.tsx:130` uses a fixture with a `title` that doesn't exist on `ActiveSandboxUiApp`. Derive these from the real producers. Still open since round 1.

13. **[R1-L3] Host shortcut-router disposer discarded.** `desktop-app/src/main.ts:306` drops the disposer the sandbox path keeps. Verified benign (the listener dies with `window.webContents`), just a latent asymmetry. Still open since round 1.

14. **[R1-N1] Palette z-index tied with confirm-dialog.** `desktop-app/ui/src/styles.css:5789` (2300 vs the confirm-dialog backdrop's 2300); if both open, stacking is source-order defined. Still open since round 1.

15. **[R1-N2] filter+sort without `useMemo`.** `desktop-app/ui/src/pages/SettingsPage.tsx:332`. Trivial cost. Still open since round 1.

## What checked out clean

- **Security: SAFE.** The three new IPC handlers (`sandboxUi:findInPage/stopFindInPage/focusActive`, `desktop-app/src/ipc.ts:1692`) all gate `assertTrustedSender` first (rejects the https embed origin). `findInPage` validates input (query 1-500, op in {start,next,previous}, positive int request id) and the `found-in-page` push carries only match counts, no page content. The native context menu binds Cut/Copy/Paste to the originating `webContents` via closure with an `isDestroyed()` recheck, so no cross-surface clipboard. The trusted preload doesn't leak find/focus/shortcut to the embed (pinned by `pluginSdkSurface.test.ts`). Markdown stays inert (raw to text), unsafe URLs blocked, the CodeQL regex stays removed, matching is `indexOf` (no ReDoS). No isolation/partition/`webPreferences` relaxation.
- **Contract aligned end to end.** All four capabilities traced layer by layer (find/stop/focus, the findResult push, `shortcuts.onCommand`, the 33=33 preload allowlist). start/next/previous semantics match between main (owns request ids, generations, stale rejection, teardown) and the renderer. The context menu is main-process only, no IPC seam. `tsc -p tsconfig.json` (main, the one that gates build) is clean.
- **Pure libs correct under fuzzing.** 13k+ fast-check runs on the semantic/search/tabs functions: count-to-marks coherence, no duplicate ids, no intra-fragment overlap, no visible-text loss, `closeChatViewTab` keeps `activeTabId` a member. Fixtures derive from real producers (T1 satisfied there).
- **Suites:** area A `vitest` 168/168, area B 21/21, area C 37/37, all green.

zach88 (approved):

PR #354, opened 2026-08-14
feat(control-ui,channel-reader): Teams setup parity with Slack, and stop silent failures in chat

jozer-rami:
Brings the Microsoft Teams channel setup flow to parity with the Slack pattern already shipped on `dev`, extends the existing unresolved-sender notice to Teams and Telegram, and stops raw tool-error envelopes reaching chat clients.

Found by running the flow end to end on dev rather than by reading code.

## Why

Teams was missing behaviour Slack already had, sometimes one line away in the same file. `edit/page.tsx` derived Slack's request URL from the unsaved draft, with a comment reasoning through exactly why. One line below, Teams still required the persisted item.

Separately, a user who messaged the bot before connecting the conversation got complete silence, and a failed tool call put `{"error":"MCP error -32603: Authentication Failed: Bad credentials"}` verbatim into a Teams channel.

## What changed

**control-ui**
- The Teams request URL is derived from the unsaved draft, matching Slack. Same reasoning: the URL depends only on namespace and name, both fixed at create.
- An annotation carve-out mirroring `hasSlackConfigForRequestUrl`, so a stale `clerum.io/teams-app-name` annotation no longer renders a copyable but fake endpoint.
- The `teams app create` command now appears on the edit page too, which is also the repair path when a channel is recreated under a new name.
- When the public origin cannot be derived, the form warns **and** hands over a placeholder command to edit, instead of silently emitting one with a fake host.
- Removed the instruction to run the command "from the project directory that has the Teams CLI project". It is false, disproved by running `teams app create` from an empty directory.

**channel-reader**
- The unresolved-sender notice now covers Teams and Telegram. Teams threads under the triggering message, because `TeamsAdapter` has no ephemeral concept and a top-level post would announce it to the whole channel. Telegram is deliberately terse with no product name and no link, because anyone who finds the bot handle can message it.
- A global cap of 50 live notice records, which now logs when it trips.
- Tool failures are classified into a closed vocabulary and rendered in our own words. No part of the tool's error string reaches chat.

## What deliberately did not change

No CRD, API, or create-gate change. A two-phase create with a pending state was designed and rejected in review: the circular dependency it existed to break was already solved by deriving the command from the unsaved draft. The Slack path is behaviourally untouched, verified line by line.

## Testing

`control-ui` 1428 tests, `channel-reader` 370, both green. Every new test was verified to fail with its change reverted.

Worth naming: eight tests in this codebase were caught passing against unfixed code during this work, several pre-existing. One asserted the raw `HTTP 404` string appeared verbatim, actively protecting the leak this PR fixes. Another would have shipped a Telegram notice that never fired in production, because real Telegram messages carry `providerWorkspaceId: null` while the test harness fakes one.

## Before merge

The spec's open risk is unverified: confirm against a real unconnected Teams user that the notice threads rather than posting top-level. That is the whole mitigation for Teams lacking an ephemeral channel.

## Known gaps

- The bot-name rule is unresolved. Create enforces kebab-case; the CLI may accept spaces, which would mean every bot has a worse display name than it needs. One command decides it.
- The provider tab still defaults to Telegram. Fixing it is a typed refactor across five call sites, not a copy change, so it was deferred.
- Telegram notices are not threaded, though group chats do have an audience. The spec pinned this; recorded as a follow-up.


jozer-rami:

alfredolopez80:
## Multi-agent review — Teams setup parity + silent-failure fixes

Reviewed at HEAD `985daf97` against `origin/dev` with four parallel agents (control-ui, channel-reader, security, tests), each finding verified against the real code in a throwaway worktree; the headline finding was additionally confirmed by direct read.

**Verdict: mergeable — no blockers.** One reproduced Medium worth closing first; everything else is follow-up or pre-existing.

### Suites (run, exit codes read unpiped)
| Suite | Executed | Result | EXIT |
|---|---|---|---|
| channel-reader (`vitest run`) | 380 / 36 files | pass | 0 |
| control-ui (`vitest run`) | 1431 / 140 files | pass | 0 |
| `tsc --noEmit` (control-ui) | — | pass | 0 |

Reported counts (370 / 1428) are slightly stale vs. actual — no issue. Security regression genuinely covered; deleted `progressFormatter` test fully migrated (nothing lost).

---

### 🟠 Medium — recommended before merge (REPRODUCED)
**`deleteConversation` persists an invalid `teamsAppName`, bypassing the validity guard** — `control-ui/app/communication-channels/[name]/edit/page.tsx:319-344`.

The guard added by the fix wave lives only in `handleSave` (`:309-313`). `deleteConversation` calls `persistDraft(nextDraft, …)` directly (`:343`), which PUTs `buildCommunicationChannelSpec(nextDraft)` — and that writes `teamsAppName` into `spec.teamsSettings.appName` whenever non-empty, no re-validation. Repro: on a channel with a Teams conversation, type `123 Bot` on the Teams tab (normalizes to `123-bot`, rejected by `isValidTeamsBotName` — leading digit), delete any conversation, confirm → the PUT carries `spec.teamsSettings.appName: "123-bot"`. The same door persists annotation-seeded or pre-PR raw invalid names.

**Fix:** hoist the guard into `persistDraft` (or the `buildCommunicationChannelSpec` caller) so one gate covers both save and delete.

### 🟡 Follow-ups (non-blocking)
- **[High — PRE-EXISTING, not introduced here]** Raw error text still reaches chat via the workflow-terminal / RPC paths: `channel-reader/src/main.ts:2073, 1955, 924, 1082, 2200, 2217` (`Error: ${...error.message}` concatenated verbatim). `git blame` → `404960c7` / `5d663a49`, both already on `dev`. This PR's own new surface (tool-step rendering) is fully closed; routing these paths through `describeToolError` is the natural next issue.
- **[Low]** `404` / `429` are unanchored in `describeToolError` while 400/401/403/413 use `\b` — `channel-reader/src/progressFormatter.ts:37-38`. Misclassification only (output stays in the closed vocabulary — no leak); no negative test for these two.
- **[Low]** An annotation-seeded invalid name (`clerum.io/teams-app-name`, e.g. `Stale Teams Name`) blocks Save from *any* tab, so a Telegram-only channel can be soft-locked by a Teams field the operator never touched. Decide intended semantics (block vs. normalize/drop on save) and pin it with a test.
- **[Low]** `buildTeamsAppCreateCommand` interpolates `botName` unescaped into a copyable bash string (`control-ui/lib/teamsSetup.ts:50-57`). Unreachable via UI typing/Copy (both gated on validity), but an out-of-band persisted name with shell metacharacters renders verbatim in the selectable `<pre>`. Render `<bot-name>` when invalid, or shell-escape.
- **[Low]** Create-page derived URL hardcodes the `channels` namespace (`new/page.tsx:196-197`) — diverges if `CONTROL_API_COMMUNICATION_CHANNELS_NAMESPACE` is overridden. Pre-existing; Slack's create URL has the identical gap; the edit page is correct.
- **[Info]** Bot-name error copy omits the length rule (min 3 / max 64), so `ab` fails with a message that looks satisfied.

### ✅ Verified clean
- **`describeToolError` is fail-closed and single-choke-point**: one producer (`progressClient.ts:150`), one chat renderer (`progressFormatter.ts:65`); every branch returns a constant; default and empty input → `'unavailable.'`. Teams/Telegram/Slack all pass through `formatProgressUpdate`. Adversarial test `describeToolError('kaboom {"secret":"hunter2"}')` → `'unavailable.'`.
- **Unresolved-sender notice is constants-only** — no sender handle / workspace id / message body echoed; no injection or spoofing vector. Teams threads via `replyToId` to the authenticated channel; Telegram deliberately terse (anti-enumeration).
- **Global cap of 50**: correct boundary (50 admitted, 51st blocked — test-reproduced), real eviction (24h TTL sweep per poll cycle), no unbounded growth, no race (synchronous check-and-set, no `await` before the record write).
- **Telegram null-workspace guard** matches the real adapter (`providerWorkspaceId: null`); Teams always carries a workspaceId (handoff validation) so the guard can't silently drop it.
- **Draft-derived request URL is server-truthful** (target id = `teams:` + base64url of `{namespace, name}`, both fixed at create; pre-credentials hit → 409 `teams_target_not_ready`). Annotation carve-out mirrors Slack exactly and is test-covered.

### On the prior deep review
- Finding 1 (raw Teams name saved from edit): **partially resolved** — typed path normalized + `handleSave` gated with tests, but the delete path above still bypasses it.
- The three nits (two `progressFormatter` test roots / non-deduped cap warning / literal `"undefined"` key prefix): **all three FIXED**, verified.

---

*Analysis only — not an approval or a change request. Posted from a local multi-agent review run.*

alfredolopez80:
## Review verdict: BLOCKED

Reviewed `evenfire-ai/evenfire#354` at the exact PR range:

- Base: `2f41f76f6658172269cd572dcf9377316ce3ac30`
- Head: `985daf97fe4d1cc777dfd13b50d0b4738cc383d3`
- 16 files changed, `+1253/-205`

The PR's existing CI is green, but exact-head negative probes found six
confirmed Medium defects and one confirmed Low defect. Two Medium findings are
merge blockers. This review is analysis only; no product code was changed.

## Findings requiring action

### [Must fix / Medium] The edit form blocks legacy and valid Teams display names

**Location:** `control-ui/app/communication-channels/[name]/edit/page.tsx:302-317`, `control-ui/lib/teamsSetup.ts:12-31`

The new `handleSave` gate applies the lowercase-kebab regex to the entire
hydrated draft. A resource with `teamsSettings.appName: "My Bot"` or
`"Legacy Teams Bot"` cannot save an unrelated change such as `CLIENT_ID`, access,
or another provider field.

Exact base/head reproduction:

- Base: PUT executed once; exit 0.
- Head: zero API calls; the form showed the Teams-name error; exit 1.

The local API accepts any non-empty `appName` up to 80 characters and the CRD
declares a plain string without a pattern. The official Teams SDK CLI documents
`--name <name>` as “App/bot name” and uses `teams app create --name "My Bot"` as
its canonical example: <https://microsoft.github.io/teams-sdk/cli/commands/app/create/>.

`deleteConversation` at lines 319-344 also calls `persistDraft` directly and
bypasses the new gate. On the exact head, Save rejects a normalized `123-bot`,
but deleting a conversation persists it. This should be fixed through one
compatible validation/build/persist boundary, not by duplicating the current
regex.

**Required direction:** Treat `appName` as a human-readable display name under
the existing API/CRD contract, preserve untouched legacy values, track
annotation/spec/user provenance, and validate all full-spec mutations through
one boundary. Introduce a separate provisioning slug only with an explicit
contract and migration decision.

### [Must fix / Medium] Teams unresolved notices reply to the leaf activity

**Location:** `channel-reader/src/main.ts:849-875,1641-1645`

The handoff already separates `providerMessageId` (the current activity) from
`providerReplyToMessageId` (the thread root), but the new notice path calls:

```ts
adapter.sendMessage(channelId, content, msg.messageId)
```

With `messageId=reply-activity-1` and `threadId=root-post-1`, the exact-head
probe observed `reply-activity-1` where `root-post-1` was required (exit 1).

Microsoft's Teams documentation says channel replies use the root message ID:

- <https://learn.microsoft.com/en-ie/microsoftteams/platform/bots/how-to/conversations/send-proactive-messages>
- <https://learn.microsoft.com/en-us/microsoftteams/platform/resources/bot-v3/bot-conversations/bots-conv-channel>

The current repository test only asserts that the third argument is truthy, so
it does not catch this regression. No live Bot Connector call was made; the
internal argument contract and exact-head reproduction are sufficient to flag
the defect.

**Required direction:** Resolve a single provider-aware delivery plan and use
the normalized Teams thread root for channel replies. Do not reimplement reply
target selection in the notice caller.

### [Should fix / Medium] HTTP is treated as a usable Teams endpoint

**Location:** `control-ui/lib/teamsSetup.ts:34-42`, `control-ui/app/communication-channels/[name]/edit/page.tsx:255-267,784-813`

`canGenerateTeamsCommand()` accepts any `http://` or `https://` URL. An HTTP
public origin therefore renders the normal Teams create/copy command instead of
an invalid-origin warning.

The exact-head contract probe confirmed:

```text
canGenerateTeamsCommand("http://webhook.example.com/webhooks/teams/x") === true
```

Microsoft requires a publicly reachable HTTPS endpoint and shows HTTPS in the
registration flow: <https://microsoft.github.io/teams-sdk/cli/concepts/local-tunnels/>.

**Required direction:** Parse the URL and require public HTTPS for the normal
command. Treat HTTP, relative, malformed, localhost, and private origins as
non-ready states. If localhost HTTP is intentionally supported, make it an
explicit development-only exception.

### [Should fix / Medium] Notice dedupe omits provider target identity

**Location:** `channel-reader/src/main.ts:492-534,781-789,1603-1608`

The limiter key contains workspace/sentinel, user, and `providerChannelId`, but
not the medium or stable `providerTarget`, even though adapter routing uses that
target.

Exact-head Telegram probe with two valid bot targets, the same user/chat, and
distinct provider events:

- Adapter A: 1 notice.
- Adapter B: 0 notices.
- Expected: one notice per outbound target.
- Exit: 1.

The case uses only two messages, so it is independent of the 50-entry cap.

**Required direction:** Use a versioned structured outbound identity containing
medium, target owner, workspace/tenant, user, and conversation. Resolve the exact
adapter before admitting the notice. Do not rely on ambiguous delimiter strings.

### [Should fix / Medium] Teams dedupe collapses separate root threads

**Location:** `channel-reader/src/main.ts:868-880,1603-1607`

For Teams, the key uses stable `providerChannelId` as the conversation. Two root
posts in one channel therefore share one 24-hour record, despite the documented
“once per user per conversation” behavior.

Exact-head probe:

```text
roots: root-post-1, root-post-2
expected notices: 2
actual notices: 1
key: tenant-1:teams-user-1:19:channel-1@thread.tacv2
```

**Required direction:** Use the normalized provider conversation/root for the
Teams conversation dimension, while retaining stable channel and target
identity as separate fields.

### [Should fix / Medium] Missing adapters consume the full 50-entry cap

**Location:** `channel-reader/src/main.ts:1608-1630`

The code records a notice before resolving `adapterForMessage`. Fifty messages
without a routable adapter therefore consume all live slots despite producing no
outbound calls. Once the adapter becomes available, the next legitimate notice
is suppressed until TTL cleanup.

Exact base/head probe:

- Base: post-recovery notice sent; PASS.
- Head: zero post-recovery sends; FAIL.
- Overall probe exit: 1.

**Required direction:** Resolve the adapter and delivery capability first, then
reserve dedupe/capacity synchronously immediately before the first `await`.
Missing routes must not consume budget; a real provider call that fails may stay
deduplicated according to the chosen retry policy.

### [Follow-up / Low] Bare 404/429 matches misclassify longer numbers

**Location:** `channel-reader/src/progressFormatter.ts:24-50`

The `404` and `429` regex alternatives lack numeric boundaries while adjacent
status codes use them.

Exact-head probe results:

- `operation finished in 4291ms` → `unavailable (rate limited).`
- `operation finished in 4041ms` → `unavailable (not found).`
- Expected for both: `unavailable.`
- Probe: 2 failures, 1 control pass; exit 1.

**Required direction:** Centralize numeric status classification with one
token-aware helper/table and add adjacent-digit negative cases. Preserve the
closed-vocabulary fallback.

## Owner decision required

The symbol/comments call the cap “global”, but `unresolvedNoticesSent` is an
instance-local Map that resets on process restart. Independent Host-owned
readers can each admit 50. Decide whether the contract is per reader/per Host
or fleet-wide; a fleet-wide guarantee requires shared atomic state.

## Verified evidence

- Exact-head focused channel-reader tests: 56/56, exit 0.
- Exact-head focused Control UI tests: 54/54, exit 0.
- Full head channel-reader suite in dependency-backed staging: 380/380, exit 0.
- Full head Control UI suite serial: 1431/1431, exit 0.
- Channel build/typecheck: base and head exit 0 in the staged harness.
- Control UI typecheck: exit 0 in the staged harness.
- Revert signal: head tests against base code produced 27 channel-reader failures,
  13 Control UI failures, and the new `teamsSetup` test failed collection on
  base.
- `git diff --check`: exit 0.
- `gh pr checks 354 --repo evenfire-ai/evenfire`: exit 0; all remote checks are
  green, including CodeQL, channel-reader build/test/typecheck, Control UI build,
  real Postgres, leak guards, registry contract, deploy manifest, shell syntax,
  and repository hygiene.

## Pre-existing items (not caused by this PR)

These were confirmed on both base and head and should be tracked separately:

- Edit navigates after a rejected PUT.
- Edit can persist a partial Teams tuple.
- Annotation-only Teams names can be promoted into `spec.teamsSettings` on an
  unrelated save.
- Edit omits the form-load `resourceVersion`, allowing stale full-spec writes.
- Existing terminal/approval paths can still concatenate upstream raw error
  messages into chat; the PR closes only the changed progress-step path.

## Validation gap

No live Teams/Bot Connector, browser, Docker, Minikube, or deployment test was
run. The PR's live unconnected-Teams acceptance remains outstanding, especially
for personal/group scopes. The negative unit/component probes above are still
deterministic exact-head failures.

## Requested next step

Please resolve the two merge blockers and the remaining Medium findings, add
regression coverage for legacy names, Delete, Teams root-vs-leaf,
cross-target/cross-root dedupe, missing adapters, HTTPS, and 404/429 boundaries,
document the cap scope, and rerun the full gates on a new exact head.

## Inline comments for the GitHub review

The following comments are anchored to the exact PR head (`985daf97fe4d1cc777dfd13b50d0b4738cc383d3`). Add them from the **Files changed** view as review comments. `RIGHT` means the changed/head side of the diff.

### The edit form blocks legacy and valid Teams display names

- **Path:** `control-ui/app/communication-channels/[name]/edit/page.tsx`
- **Line:** `309`
- **Side:** `RIGHT`
- **Comment:**

  This validation runs against the entire hydrated draft, so editing any unrelated field on a channel whose existing `teamsSettings.appName` is `My Bot` or `Legacy Teams Bot` returns here and never sends the PUT. The exact base/head probe passes on base and fails on head. The local API and CRD accept a human-readable non-empty `appName` (up to 80 characters), and the official Teams CLI example uses `My Bot`. Please avoid imposing a new slug contract at this boundary; preserve legacy values and route Save/Delete through one compatible validation and persistence path.

### Teams unresolved notices reply to the leaf activity

- **Path:** `channel-reader/src/main.ts`
- **Line:** `1645`
- **Side:** `RIGHT`
- **Comment:**

  `msg.messageId` is the leaf activity, while `teamsMessageFromHandoff` stores `providerReplyToMessageId` as `threadId`. For a Teams channel reply, the root activity ID is the required reply target. The exact-head probe supplied `messageId=reply-activity-1` and `threadId=root-post-1` and observed the leaf ID being sent. Please use the normalized thread root/provider-aware delivery plan here and add a regression test that distinguishes the two IDs.

### HTTP is treated as a usable Teams endpoint

- **Path:** `control-ui/lib/teamsSetup.ts`
- **Line:** `41`
- **Side:** `RIGHT`
- **Comment:**

  This predicate makes any `http://` origin pass the ready gate, so the UI renders and copies a normal Teams create command for an endpoint Teams will reject. The exact-head contract probe returns `true` for `http://webhook.example.com/webhooks/teams/x`. Please parse the URL and require public HTTPS for the normal command; handle HTTP, relative, malformed, localhost, and private origins as explicit non-ready states.

### Notice dedupe omits provider target identity

- **Path:** `channel-reader/src/main.ts`
- **Line:** `1606`
- **Side:** `RIGHT`
- **Comment:**

  `providerChannelId` is not the complete outbound identity: adapter routing below uses `providerTarget`, including the CommunicationChannel target. With the same user/chat routed through two valid Telegram bot targets, this key suppresses the second adapter's notice; the exact-head probe observed adapter A=1, adapter B=0, with only two messages (independent of the cap). Include medium and stable provider-target ownership in a structured dedupe key before admission.

### Teams dedupe collapses separate root threads

- **Path:** `channel-reader/src/main.ts`
- **Line:** `1607`
- **Side:** `RIGHT`
- **Comment:**

  For Teams, `conversationId` is derived from the stable provider channel, not the normalized thread/conversation root. Two different root posts in the same channel therefore share the same 24-hour notice record. The exact-head probe expected two notices for `root-post-1` and `root-post-2` but observed one. Keep stable channel/target identity separate from the Teams conversation-root dimension used for dedupe.

### Missing adapters consume the full 50-entry cap

- **Path:** `channel-reader/src/main.ts`
- **Line:** `1627`
- **Side:** `RIGHT`
- **Comment:**

  The notice is recorded before `adapterForMessage` is resolved. Fifty messages with no routable adapter therefore consume all live slots without producing an outbound call; after the adapter is restored, the next legitimate notice is suppressed. The exact base/head cap probe sends after recovery on base and sends zero on head. Resolve the adapter and delivery capability first, then reserve the cap/dedupe slot immediately before the first await.

### Bare 404/429 matches misclassify longer numbers

- **Path:** `channel-reader/src/progressFormatter.ts`
- **Lines:** `37-38`
- **Side:** `RIGHT`
- **Comment:**

  These `429` and `404` alternatives are unbounded, unlike the adjacent status-code checks. The exact-head probe classifies `4291ms` as rate limited and `4041ms` as not found; both should fall back to generic unavailable, while the control case passes. Use one token-aware status classifier and add adjacent-digit negative cases.

jozer-rami:
Thanks for this. Every finding was checked against the code before acting, and the two blockers were real. Head is now `ab36b939` (`d0c8c994` for the fixes, `ab36b939` for a doc correction they exposed).

## Taken

**The `appName` kebab rule is deleted, not propagated.** You were right that it is invented. Three sources agree it is a free-form display name: the CRD declares `appName` as a plain string with no pattern while `tenantId` beside it has one (`communicationchannel.yaml:279-284`), control-api enforces only non-empty and 80 characters while `appId` beside it has a strict UUID regex (`validateCommunicationChannelSpec.ts:65-73`), and the CLI docs you linked use `--name "My Bot"`.

Worth being plain about: the gate you found was **a regression introduced earlier in this same PR**, in response to a different review that flagged create and edit disagreeing. I resolved that disagreement by propagating the wrong rule instead of checking whether it was a rule at all. The normalization and the regex are gone from both pages and from `teamsSetup.ts`, replaced by `teamsAppNameError` matching the server contract. Validation now sits on the single persist boundary, so `deleteConversation` is covered rather than bypassing it, and `handleSave` no longer navigates away after a failed PUT.

**The notice now replies to the thread root.** `msg.threadId || msg.messageId`, matching how the file's other Teams reply paths choose a target. Your point about the test is the sharper half: asserting the third argument is truthy cannot distinguish root from leaf, so the test was guarding nothing. It now constructs a message where root and leaf differ and pins the root.

**HTTPS is now required** in `canGenerateTeamsCommand`. That change surfaced a follow-on we had to fix in the same pass: rejecting `http` routed absolute http URLs into the placeholder builder, which would have rendered `https://<public-webhook-origin>http://host/...`.

**`\b429\b` and `\b404\b` are anchored.** Also self-inflicted: those two categories were added in an earlier fix wave on this PR, immediately after anchoring `401` and `403` for exactly this reason.

Both suites green locally, control-ui 1447 and channel-reader 384. Remote checks are re-running on `ab36b939`.

## Deferred, deliberately

Three findings are not addressed here: dedupe omitting provider target identity, Teams dedupe collapsing separate root threads, and missing adapters consuming the cap.

Your own required direction for the first is a versioned structured outbound identity containing medium, target owner, workspace, user and conversation, and the second depends on that same identity work. Patching the dedupe key shape inside this PR would be guessing at a design rather than implementing one. The third means reordering record-on-attempt relative to adapter resolution, which is an invariant the original author documented at length; it deserves its own change with its own tests rather than a late edit here.

None of the three are defects this PR introduced. They are pre-existing in the Slack path and inherited by the providers this PR adds, which is a real argument for fixing them, but a separate one.

## Your owner-decision question

Agreed, and the naming oversells it. `unresolvedNoticesSent` is an instance-local Map that resets on restart, so N readers each admit 50. Per-reader or fleet-wide is a product decision rather than something to settle in review, and fleet-wide would need shared atomic state. Flagging it rather than quietly renaming the constant.

## Still open, unrelated to your findings

The Teams notice threading has not been confirmed against a real unconnected user in Teams. Code-wise it is now correct per your finding, but nobody has watched Teams render it. That needs a deploy to dev, so it is the gate for promoting to main rather than for this merge.

Issue #346, opened 2026-08-13
build-publish: one failed matrix leg silently freezes every image's revision annotation, and detection never catches up


jozer-rami:
Found while cutting **v0.7.0**. `release-images.yml` refused to promote 6 of 28 images, and the reason turned out to be a CI bug that will recur on every release.

## Symptom

Tagging `v0.7.0` produced:

```
mcp-proxy: source changed after the published image (3cbdf331). Push to dev and let build-publish run before tagging.
rpc-proxy: ...            (3cbdf331)
stdio-bridge: ...         (3cbdf331)
web-search-mcp: ...       (3cbdf331)
mock-stdio-mcp-server: ...(3cbdf331)
profile-ui: ...           (38be8570)
```

`3cbdf331` is PR #173 — roughly two weeks stale. Verified against the live registry: `ghcr.io/evenfire-ai/control-api:v0.7.0` exists, `rpc-proxy:v0.7.0` does not.

## Root cause

1. PR #298 (`feat/ux-improvements-28`) merged to `dev` as `2c5dae15`. The pre-commit hook `scripts/precommit/bump-staged-package-versions.mjs` bumped `package.json` / `package-lock.json` for those 6 packages (e.g. rpc-proxy `0.1.62` → `0.1.64`). **That version bump is the only content difference** — the formatting from the same branch was reverted by "restore format-only files from dev".

2. build-publish run [31499844320](https://github.com/evenfire-ai/evenfire/actions/runs/31499844320) for `2c5dae15` **failed — but only on `control-ui-amd64` / `control-ui-arm64`**. All 6 images built fine on both arches; 55 jobs succeeded.

3. `build-publish.yml:422`:
   ```yaml
   merge:
     name: Merge manifest (${{ matrix.image }})
     needs: [detect, build-push]
   ```
   With the implicit `if: success()`, **one failed leg of `build-push` skips `merge` for every image.** That run shows a single `skipped  Merge manifest (${{ matrix.image }})`.

4. The `merge` step is what assembles the multi-arch manifest **and writes `org.opencontainers.image.revision`** — the annotation `release-images.yml` reads. So all 6 kept their old revision even though newer per-arch images had been pushed.

5. **Detection never catches up.** `detect` uses `dorny/paths-filter` with `base: ${{ github.event.before }}` — an incremental *per-push* diff. The bump lived in an earlier push, so no later push sees a change under those paths. The images were never rebuilt again, silently, for two weeks.

[...]

PR #355, opened 2026-08-14
feat(external-rest-api): report the image build revision mid-cycle

jozer-rami:
Stacked on #344 — the profile-ui consumer needs `lib/releaseIdentity.ts` from that branch. GitHub will retarget this to `dev` when #344 merges. Review #344 first.

## The gap

#344 made the portal show the real release instead of a commit counter, which exposed the next question: between releases, `releaseId` and `desktopVersion` are **frozen at the last cut** by design (`assert: 'explicit'` and `'equals'` in `release-coordinates.mjs`), so `profile.dev` reads `Release v0.6.0` for the whole cycle no matter how much code has landed.

Nothing identified what was actually running:

| Signal | Mid-cycle | Reachable from the app |
|---|---|---|
| `releaseId`, `desktopVersion` | frozen at last cut | yes |
| per-service counters | advance, but match no tag | yes |
| `sha-<7>` image tag + `org.opencontainers.image.revision` | **the real identity** | **no** |

The revision was already being computed and stamped on every image (`build-publish.yml:388`). It just wasn't readable by anything.

## What this does

`build-publish.yml` passes `github.sha` as a `BUILD_REVISION` build-arg; external-rest-api's Dockerfile bakes it into the runtime env; `GET /api/v1/desktop/release` serves it shortened to the **same 7 characters that name the image tag**, so what the API reports can be matched against what the cluster pulled.

```
GET /api/v1/desktop/release
{ releaseId: "v0.6.0", externalRestApiVersion: "0.1.65", buildRevision: "4be949d", ... }
```

Consumer: the portal's sidebar brand title reads `Release v0.6.0 (build 4be949d)`. The settings header is unchanged and stays release-only — the build is support-facing detail, not user-facing.

## Three decisions worth checking

**Baked, not deploy-time injected.** Dev pins a moving `latest`, prod deploys from the separate `evenfire-infra` repo, and `build-publish` only rebuilds services whose paths changed. A deploy-time value would misreport the moment one image lagged. Baked, it travels with the artifact.

**Per-image, not per-platform.** For the same change-detection reason, mid-cycle a cluster is a mosaic of per-image builds. There is no single platform build sha to report, so the field means "this external-rest-api image", stated in the route comment.

**ARG placed last in the Dockerfile.** It changes every commit; any earlier position would invalidate the `COPY` layers above it and defeat the build cache.

Also: `scripts/minikube/build-images.sh` stamps local builds with `git HEAD`, passing the arg only to Dockerfiles that declare it, so the twenty-odd images that haven't adopted it don't emit unconsumed-build-arg warnings. Adopting it elsewhere later is one `ARG`/`ENV` pair per Dockerfile with no workflow change.

## Compatibility

An external-rest-api image built before this lands omits `buildRevision`; the client normalizes a missing or non-string field to empty, which renders as the release alone. Covered by tests on both sides.

## Verification

- external-rest-api: 117 tests / 25 files green, `npm run build`, `tsc --noEmit`
- profile-ui: 75 node + 24 vitest green, `next build`, `tsc --noEmit`
- `prettier --check` on every touched file, `npm run style-rules`
- `bash -n` on the build script, and the guard exercised directly: stamps `external-rest-api/Dockerfile`, skips `rpc-proxy/Dockerfile`
- Workflow YAML parsed to confirm `build-args` landed on the `docker/build-push-action` step next to the existing `labels`

**Not verified locally:** the actual `docker build`. Docker was unavailable in my environment, so the baked `ENV` is verified by inspection and unit tests, not by running the image. First CI build on this branch is the real check.

The `external-rest-api` version bump to 0.1.65 and the matching `releaseManifest.externalRestApiVersion` are the precommit hook's work, not hand-written: that field is a `counter` coordinate, so it stays in sync automatically.


jozer-rami:
Updated onto the current `dev` (via the parent branch). Two things worth flagging.

**v0.7.0 landed while this was open.** The merge conflicted in `external-rest-api/src/releaseManifest.ts`, on exactly the two fields you'd expect: `releaseId` (`v0.6.0` here vs `v0.7.0` from the cut) and `externalRestApiVersion` (my pre-commit bump to `0.1.65` vs dev's `0.1.64`). Resolved to `releaseId: 'v0.7.0'` and `externalRestApiVersion: '0.1.65'`, the latter because the merged `external-rest-api/package.json` is `0.1.65` and that field is a `counter` coordinate that must equal it. Verified with the actual CI check rather than by eye:

```
$ node scripts/release/update-desktop-release-manifest.mjs --validate-only
{ "releaseId": "v0.7.0", "externalRestApiVersion": "0.1.65", "rpcProxyVersion": "0.1.64",
  "desktopVersion": "0.7.0", "minimumDesktopVersion": "0.1.252" }
```

So the label now reads `Release v0.7.0` on dev, not `v0.6.0` as the description above says.

**Relationship to #346.** That issue reports that one failed matrix leg freezes every image's revision annotation and detection never catches up — which is precisely the condition `buildRevision` makes visible. Under #346 a deployed image can be weeks behind `dev` while nothing in the product says so; this field reports the commit that image was actually built from, so the staleness becomes readable instead of being discovered at release-cut time. It is a diagnostic for #346, not a victim of it: an image that was never rebuilt keeps its old baked revision, which is the honest answer.

It also does not add rebuilds. `build-args` do not feed `dorny/paths-filter`, so change detection is untouched, and the `ARG` sits last in the Dockerfile so a per-commit value only invalidates the trailing `ENV`/`USER`/`EXPOSE`/`CMD` layers of a build that was already going to happen.

Re-verified after the merge: external-rest-api 117 tests / 25 files, profile-ui 75 node + 24 vitest, `tsc --noEmit` clean on both.

PR #367, opened 2026-08-16
feat(minikube): add local T0/T1/T2 validation contract

alfredolopez80:
## Summary

- Add the development-only local Minikube T0/T1/T2 validation contract.
- Add fail-loud preflight, branch/profile ownership checks, transition planning,
  serialized evidence, bounded recovery, and the explicit Real PostgreSQL T1
  lane without changing the ordinary CI DSN contract.
- Add contract, scenario, and public-boundary checks plus the sanitized local
  runbook.
- Define the repository logging standard: structured logging with Pino as the
  Node-service reference, existing service wrappers preserved, and file-local
  `console.*` cleanup only when a file is already being modified.

## Scope and safety

- Development-only branch-owned Minikube profiles and explicit Kubernetes
  contexts; no production, GKE, Cloudflare, staging, shared-cluster, or
  customer-data workflow.
- No automatic PVC deletion, no secret values in output, and no committed
  runtime/profile artifacts.
- This PR is independent of PR #275 and PR #311; neither PR is modified or
  included in this branch.

## Validation

- `bash scripts/tests/test-minikube-t2-contract.sh`
- `bash scripts/tests/test-minikube-t2-scenarios.sh`
- `bash scripts/tests/test-minikube-t2-public-boundary.sh`
- `bash -n` on all new T2 scripts
- `git diff --check`

Live Minikube bootstrap, Real PostgreSQL T1, validated runtime T2, and
Control UI/Desktop Playwright evidence remain pending explicit authorization to
start the dedicated branch-owned profile. The implementation fails loudly
instead of treating those lanes as green when they are not executed.

## Base and review

The branch is rebased onto the current `origin/dev` before publication. Please
review the runtime ownership, no-skip evidence contract, secret-safety boundary,
and the file-local logging standard in `AGENTS.md` and `CLAUDE.md`.


zach88:
Took a deep pass through this one — two code reviews (core shell + the test-scripts/vitest seam) plus a security pass. The good news up front: the secret handling and the shell↔vitest fail-loud contract are genuinely solid, and I reproduced the seam end-to-end against real vitest. The problems cluster in two spots: the T1 lane has a hard bug that stops it from ever going green, and the branch is built on a stale base so a few files collide with what's already on `dev`.

### 🔴 Not ready to approve

_This is my read from the review, not a formal GH approval._

1. **[R1-B1] The T1 lane crashes with "unbound variable" on every run** — `scripts/e2e/minikube-real-postgres.sh:233`
   `main()` runs `unset PG_USER PG_PASSWORD PG_DATABASE` *before* both `run_suite` calls (lines 234-235). Every `run_suite` calls `sanitize_file`, whose first statement is `T1_REDACT_PASSWORD="$PG_PASSWORD" python3 ...` (line 105). The script runs under `set -euo pipefail` and `-u` is never cleared, so expanding the now-unset `$PG_PASSWORD` aborts with `PG_PASSWORD: unbound variable`. Net effect: the T1 lane can never complete or report PASS — it dies in the first `run_suite` regardless of the actual suite result. As a side effect the password redaction in `sanitize_file` never even runs; the only reason the plaintext value doesn't leak is that the crash happens before the `cat`.
   Repro: minimal `set -u; PG_PASSWORD=x; unset PG_PASSWORD; foo="$PG_PASSWORD"` → unbound variable, exit 1. Confirmed against the real code path.
   Fix: move the `unset PG_USER PG_PASSWORD PG_DATABASE` to *after* both `run_suite` calls (next to `unset ADMIN_DSN` on line 236), or stash the redaction value in a separate var before the unset and have `sanitize_file` use `"${T1_REDACT_PASSWORD:-}"`.

## Optional (non-blocking)

The PR can merge without these, but please look at the first one before anything else — it changes the diff itself.

1. **[R1-M3] Stale base: 4 files already live on current `dev`, and the body claims otherwise** — `control-api/test/realPostgres.requirement.ts`, `gfs-controller/test/realPostgres.requirement.ts`, both `vitest.config.ts`
   The branch's merge-base is 5+ commits behind current `origin/dev`. Against today's `dev`, both `realPostgres.requirement.ts` files and the `setupFiles: ['test/realPostgres.requirement.ts']` line in both vitest configs already exist (they landed via PR #311). On rebase the two `requirement.ts` files will conflict (added on both sides, different wording) and the config lines will conflict or no-op. The PR body says "rebased onto the current origin/dev before publication" (not the case) and "independent of PR #311" (these 4 files overlap what #311 shipped). Recommend rebasing onto current `dev` and dropping the 4 already-present files so the PR is just the new scripts/Makefile/docs.

2. **[R1-M1] Public-boundary gate passes vacuously when the base ref doesn't resolve** — `scripts/tests/test-minikube-t2-public-boundary.sh:18`
   `git -C "$ROOT" diff --no-color --unified=0 "$BASE...HEAD" 2>/dev/null || true` swallows any range-resolution failure. If `origin/dev` (or `T2_PUBLIC_BASE_REF`) isn't present — fresh clone, shallow checkout, CI that doesn't fetch dev — the scan of the committed PR diff is skipped entirely and you get `PUBLIC_BOUNDARY_PASS`. For a check that sells itself as fail-loud, silently disabling itself is the wrong failure mode.
   Repro: with `T2_PUBLIC_BASE_REF=nonexistent-ref-xyz` it prints `PUBLIC_BOUNDARY_PASS` / EXIT 0 without inspecting anything; with a resolvable base it correctly rejects a planted `postgresql://user:pass@10.0.0.5/db`.
   Fix: resolve the base first and hard-fail if it doesn't (`git rev-parse --verify --quiet "$BASE^{commit}"` → else exit 1) instead of `|| true`.

3. **[R1-M2] TOCTOU in lock acquisition lets two processes share one profile** — `scripts/minikube/t2-common.sh:513-536`
   The lock dir is created with `mkdir "$T2_LOCK_DIR"` (513) but `owner.env` with `PID=$$` isn't written until line 527. A competitor arriving in that window fails the `mkdir`, reads a missing/empty `owner.env` → empty `existing_pid` → the "alive" check is false → it falls through to `rm -rf` + re-`mkdir` and reclaims the lock the first process just took. Result: two `make minikube-t2` runs on the same profile.
   Fix: publish the lock atomically with `owner.env` already written (build in `"$T2_LOCK_DIR.$$"`, write owner, atomic `mv`), or treat an empty `existing_pid` as "lock under construction → PROFILE_BUSY" rather than reclaimable.

4. **[R1-L1] The anti-prod/GKE guard keys on the profile name, not the cluster endpoint** — `scripts/minikube/t2-common.sh:173-177`
   `t2_profile_scope` rejects `*gke*|*prod*|*staging*|...` on the profile *name* and never inspects `cluster.server`. A benignly-named context that maps to a remote cluster in kubeconfig would pass. Low because there are strong mitigations (requires local `profile.env` matching the worktree/HEAD, runs `minikube status`, dev-only self-run). Optional hardening: inspect `cluster.server` and reject anything that isn't `127.0.0.1`/`192.168.*`/`*.minikube`.

5. **[R1-L2] Ephemeral-port race in the port-forward** — `scripts/e2e/minikube-real-postgres.sh:216-225`
   `choose_local_port` binds/closes an ephemeral port and `port-forward` re-binds it later; if another process grabs it in the window, `wait_for_tcp` could connect to the intruder and T1 would run against the wrong service. Low probability. Fix: after `wait_for_tcp`, verify `kill -0 "$PORT_FORWARD_PID"` before proceeding.

6. **[R1-L3] `FAIL` literal error code is inconsistent with the stable codes** — `scripts/e2e/minikube-real-postgres.sh:142,171`
   `die_t1 FAIL "..."` serializes the literal `FAIL` as `T2_ERROR_CODE` in evidence, unlike the stable codes elsewhere (`ZERO_TESTS_EXECUTED`, `POSTGRES_NOT_READY`, …). Suggest something like `REAL_PG_SUITE_FAILED`.

7. **[R1-L4] Heuristic gaps in the public-boundary detector** — `scripts/tests/test-minikube-t2-public-boundary.sh:47-55`
   (1) Untracked new files with secrets are invisible to `git diff`/`--cached` (only the `BASE...HEAD` range sees them — and that range depends on R1-M1). (2) The "private URL" pattern is only `https?://`, so a `postgresql://private-host:5432/db` *without* credentials slips both patterns. Best-effort detector; flagging to calibrate expectations.

8. **[R1-L5] (informational) Theoretical redaction gap if a log splits the DSN across lines** — `scripts/e2e/minikube-real-postgres.sh:102-119`
   Password redaction is a literal `str.replace` (correct, metachar-safe) and the DSN is regex-based. A consumer printing the DSN with an inserted newline could leave the url-encoded half unredacted. Doesn't happen in practice (DSN only goes out as an env var; Vitest serializes on one line). Noting, not a defect.

## What checked out clean

- **The shell↔vitest fail-loud seam is robust** — reproduced against real vitest 4.1.9 in a throwaway project: the `realPostgres.requirement.ts` throw aborts the run when the DSN is missing with `REQUIRED=1`; 0 matching files makes vitest fail (no green-with-0-tests), and `run_suite`'s `find` already requires ≥1 file first; the JSON reporter field names (`numPassedTestSuites`, `numPassedTests`, `numFailedTests`, `numPendingTestSuites`, `numPendingTests`, `numTotalTests`, `success`) match exactly what's parsed; the all-skip case falls into `ZERO_TESTS_EXECUTED`. 12 `*realPostgres*.test.ts` suites exist in control-api, 2 in gfs-controller.
- **Security: safe.** Admin DSN stays in-memory and is passed as an env var (not argv); password/base64 go over stdin (not visible in `ps`); `set +x` in all four scripts with no reactivation; password redaction is a literal replace (metachar-safe); `evidence.json` holds only git/profile metadata under `.local-notes/` (gitignored, `umask 077`); the PVC reset has a real double guard (`T2_RESET_PVC=true` + exact `T2_EXPECTED_PVC_UID`) with no bypass path; no `eval` in new code.
- **`set -e` propagation** is clean — `t2_fail`/`die_t1` run as plain commands in `if…then` bodies so their `return 1` trips `set -e`; no path where fail-loud is masked. Bounded timeouts on `wait_for_tcp` and `rollout status`; no hanging loops.
- **Logging-standard doc claims check out** — control-api does expose a redacting root Pino logger (`src/observability/logger.ts`), `workspace-files-controller` uses `pino` + `pino-http`, and there's no Winston in the repo.
- All three contract scripts (`test-minikube-t2-contract.sh`, `-scenarios.sh`, `-public-boundary.sh`) run green.

zach88:
Took another pass through this after round 1, focused on the fixes in `6359a8efe` plus the dev merge. Short version: this landed well. The blocker and all three mediums from last round are closed and actually covered by executable tests now, and I couldn't find any regression the fix introduced. What's left is three Low nits, all of them consistent with the fail-loud intent of the contract.

**🟡 Approve with nits**

## Optional (non-blocking)

The PR can merge without any of these; they're follow-up or author's call.

1. **[R2-L1] Orphaned lock has no self-heal and the next-step hint doesn't say how to clear it** — `scripts/minikube/t2-common.sh:526-546`. Side effect of the M2 lock fix: if a process dies between `mkdir` (:530) and writing `owner.env` (:548) — SIGKILL, power loss — you get a `*.lock` with no `owner.env`, and every later run reads an empty `existing_pid` and fails `PROFILE_BUSY` forever, with no live PID to check. That's the deliberate fail-loud tradeoff (the contract test pins it at `test-minikube-t2-contract.sh:110-118`), so no objection there — the only nit is that `T2_NEXT_COMMAND` (:534) says "retry only after the lock is released" without telling the user *how* to release an orphaned one. Worth mentioning the manual `rm -rf "$T2_LOCK_DIR"` (or pointing at the runbook) in that hint.

2. **[R2-L2] Local IPv6 apiserver endpoint gets rejected** — `scripts/minikube/t2-common.sh:229-234`. `https://[::1]:8443` → `host="${host%%:*}"` leaves `host="["`, which matches none of `127.0.0.1|192.168.*|*.minikube` → `DEVELOPMENT_SCOPE_REQUIRED`. Minikube almost never exposes the apiserver over IPv6 and erring toward rejection is the safe direction, but a legit local IPv6 apiserver would be turned away. If you want to cover it, normalize `[::1]`/`localhost` before the `case`. Low impact.

3. **[R2-L3] Final gate runs with `T2_PLAN_MODE=true`, so it isn't strictly authoritative** — `scripts/minikube/t2.sh:194-207`. With that flag a `full-bootstrap` plan doesn't fail (`t2-preflight.sh:41`) and then `T2_T2_STATUS=PASS` is set unconditionally. In practice it's safe: `make minikube-setup` runs under `set -e` (aborts on failure) and `run_pre_gate` writes the marker, so in the final preflight `T2_BOOTSTRAP_REQUIRED=false` and the resource/postgres/deployment checks do fail loud. The gap only shows up if bootstrap "succeeded" but didn't provision and the marker were missing — very theoretical. `T2_PLAN_MODE=false` on the final gate would close it if you want it strictly authoritative.

## What checked out clean

- **R1-B1 (the blocker) is genuinely fixed.** `T1_REDACT_PASSWORD` is captured at `minikube-real-postgres.sh:216` before the port-forward and the `unset` moved to :243 after `run_suite`, so it no longer crashes under `set -u` — and the password redaction, which never actually ran before because the crash preempted it, now runs. The empty-value `str.replace("")` case is guarded by `if password:` in `sanitize_file`.
- **The public-boundary gate no longer passes on an empty scan** (R1-M1): `git rev-parse --verify --quiet` → `PUBLIC_BOUNDARY_BASE_UNRESOLVED` + `exit 1`, `|| true` gone. And it doesn't self-trip on this PR's own files.
- **Lock TOCTOU closed** (R1-M2): non-numeric/absent PID → `PROFILE_BUSY` instead of reclaiming a live lock.
- **The dev merge resolved the duplication** (R1-M3): both `realPostgres.requirement.ts` are now byte-identical to `origin/dev` and out of the changed-file set; the branch is genuinely rebased on current dev.
- Endpoint scope check, port-forward `kill -0` race guard, and the stable `REAL_PG_SUITE_FAILED` error code are all in (R1-L1/L2/L3) — verified no consumer relied on the old literal `FAIL` code.
- `shellcheck 0.11.0` clean on the four core scripts; `bash scripts/tests/test-minikube-t2-contract.sh` (with `bash -n`, boundary, scenarios) passes green, including the redaction, ownerless-lock, and remote-context regression cases.
- Security: SAFE — redaction covers log/json/port-forward output, all emitted after `sanitize_file`; password never in argv/`ps`; boundary check doesn't leak the secret it detects.

zach88 (approved):

PR #372, opened 2026-08-17
feat(guardrails): mcp-host guardrails engine + installed llm/tool hooks (§8)

andypavia:
## Summary

Adds the **mcp-host guardrails engine** and **marketplace-installed `/v1` guardrail hooks**, end-to-end (spec suite in `docs/features/mcp-host-guardrails/`). The full image-target path is live-verified on `clerum-test`: registry publish → UI install → authenticated zot pull → HCC-deployed hook pod → mcp-host dial → `may_rewrite` compaction applied to the model request byte-for-byte.

## What's included

- **CRDs** — `LlmHook` + `Host.spec.guardrails` block.
- **mcp-host** — tool-lane + LLM-lane guardrail boundary; installed-hook resolution (§8.2) into runtime descriptors; `/v1` HTTP transport (time+byte bounded, §8.6); content projection with `contentAccess` gate (§8.4); circuit breaker for advisory hooks (§8.6); SSRF-guarded remote-target dial; digest-pin quarantine + DuplicatePath fail-closed (§8.2/§1.2); deny → graceful refusal; `on_error` error redaction.
- **control-api** — org-scoped install / upgrade / uninstall-hook saga with install-time trust gates (re-run on upgrade); `llmhooks` resource endpoints.
- **control-ui** — Installed Guardrails management (per-agent + per-guardrail detail) + install-hook form.
- **host-context-controller** — `LlmHook` reconciler (digest-dedup Deployment, N8 hardening, ref-count GC, reverse-index egress NetworkPolicy); **`clerum.io/guardrails-revision` pod-template annotation** so a guardrails change auto-rolls mcp-host (it reads guardrails only at boot).
- **Network/pod isolation invariants N1–N12** (default-deny egress, scoped CoreDNS, read-only rootfs, etc.).

## Recently added / validated in this pass

- `pre_call` **apply-log** in `HookedLlmPort` (content-free) — makes a `may_rewrite` hook's rewrite verifiable end-to-end against the hook's own before/after chars.
- **HCC `guardrails-revision` annotation** + unit test — guardrails install/uninstall/upgrade now rolls mcp-host automatically (previously required a manual restart).

## Known open items (not blocking review, tracked in-repo)

- **Live/E2E coverage is partial** — see the coverage matrix in `docs/features/mcp-host-guardrails/mcp-host-guardrails-test-plan.md` (unit-strong; remote-target E2E, install/upgrade/uninstall saga E2E, a control-ui pass, and adversarial fixtures are pending). *(That test-plan doc + a draft `guardrail-input-transparency-spec.md` are staged to land separately.)*
- **Deferred tool-lane TODOs** — provenance `serverName` threading from `McpManager`, doom-loop per-task counter/persistence, prompt-shaping system-prompt-part injection.
- **Guardrail Input Transparency** (user-facing per-turn "what changed" record + desktop UI) is specced as a draft but unbuilt.

## Test plan

- **Unit** — mcp-host guardrails suites (hook projection, remote hook, breaker, SSRF, tool-lane), HCC `LlmHook` reconciler + `guardrails-revision`, control-api install/upgrade tests.
- **Live** — image-target hook install validated end-to-end on `clerum-test` (dial + `may_rewrite` compaction applied; auto-roll on guardrails change).


zach88:
Took a deep pass through this one, area by area (mcp-host LLM + tool lanes, control-api install saga, HCC reconciler + infra, control-ui, plus a full security sweep and an end-to-end contract trace). Overall this is strong work — the hard structural seams all line up field-for-field (digest-pin, podKey byte-match, DuplicatePath fail-closed, the `/v1` protocol across all 4 points, and the install/list/delete HTTP contract are all consistent end-to-end), and the security posture is solid. Where it needs work is a network-isolation invariant that contradicts what the PR itself promises, a couple of silent-failure paths around host ref updates and pod rolls, and the `contentAccess` contract.

## Verdict

🔴 **Not ready to approve**
- **Must resolve before approval: 6** (1 Blocker + 5 High)
- **Optional / non-blocking: 18** (6 Medium + 12 Low)

**Blocker**

1. **[R1-B1] `llm-hooks` in `runtimeNamespaces` opens namespace-wide DNS + HCC-API egress to every hook pod — violates the N5 "no implicit DNS" invariant this PR states.** `host-context-controller/src/config.ts:560`. Adding `llm-hooks` to `runtimeNamespaces` makes the (pre-existing) `NetworkPolicyReconciler.ensureInfrastructurePolicies()` generate `allow-dns-egress-llm-hooks` with `podSelector: {}` (`networkPolicyReconciler.ts:375`) — so *every* hook pod, including a pure `/v1` responder with no `egressBindings`, gets DNS egress to kube-system:53. And because the hook pod-template carries `clerum.io/managed-by: host-context-controller` (`llmHookReconciler.ts:577-583`), which is the default selector (`networkPolicyReconciler.ts:187`), those pods also get egress to the internal HCC gateway. This directly contradicts two of your own claims: `llmHookReconciler.ts:557-560` ("A pure /v1 responder … gets no egress at all, 'no implicit DNS'") and `deploy/base/llm-hooks/networkpolicies.yaml:14-16` ("gets NO egress — not even DNS (N5)"). A compromised third-party `/v1` hook could exfil over DNS and reach the internal gateway. This is a design call — either (a) drop `llm-hooks` from `runtimeNamespaces` and rely on the static default-deny + the per-pod-key/host-egress NPs the reconciler already emits (the only option that honors N5 as written), or (b) move it to `minimalInfraNamespaces` (drops HCC/K8s-API egress but namespace-wide DNS stays) and fix the comments/tests that claim "no DNS". Worth confirming the effective egress on-cluster.

**High**

2. **[R1-H1] Lost-update on `Host.spec.guardrails.hooks` — a hook can end up installed but unreferenced (silently inactive).** `control-api/src/routes/admin/registry.ts` (install-hook step 10) + `services/hostGuardrailRefs.ts`. The handler reads `freshHost`, computes the `hooks` map, and calls `updateResource('hosts', …, {spec})` without a `resourceVersion`; the internal precondition GET re-sends the handler's already-fixed `spec`. Two concurrent install/upgrade/uninstall ops on the same Host → one clobbers the other with no 409, leaving a CR installed but out of `Host.spec` → operator thinks the guardrail is active and it never runs. Fix: make read→compute→PUT atomic in a retry that re-reads and re-derives the map each attempt and sends that read's `resourceVersion` (`updateResourceWithConflictRetry` as-is won't do it — it re-sends the stale spec).

3. **[R1-H2] Removing the whole `guardrails` block does NOT roll mcp-host — it keeps running the old guardrails.** `host-context-controller/src/hostReconciler.ts:2394` + `utils.ts:83`. When `host.spec.guardrails` goes from present to absent, `podAnnotations` omits `guardrails-revision`, but the replace uses `preserveDeploymentAnnotations` and `mergeAnnotations` does `{...existing, ...desired}` → the old annotation survives → no diff → no rolling restart. mcp-host (reads guardrails only at boot) keeps running already-uninstalled hooks/limits until an unrelated roll. *Changing* guardrails does roll; only full removal is broken. The comment at `hostReconciler.ts:2392-2393` ("removed flips the template") is false, and no test covers the replace/merge path (T5). Fix: write an explicit tombstone (`podAnnotations[GUARDRAILS_REVISION]=''`) on absence, or exclude that annotation from the preservation merge.

4. **[R1-H3] `contentAccess`: the CRD doc contradicts the code in both consumers — content leaks vs. what's documented.** `charts/clerum-crds/crds/llmhook.yaml:178-180` documents "omit → preCall-only infers `metadata`", but control-api (`registry.ts:144,152` — absent ⇒ content-bearing) and mcp-host (`hookProjection.ts:56` — absent ⇒ includes messages) both treat **absent → content**, and `registry.ts:494` never persists an inferred `metadata`. So an operator following the doc ("omit → metadata-only, no bodies") actually gets full message bodies shipped to a third-party hook, and that same hook with `egressBindings` at low trust gets rejected with `content_egress_requires_high_trust` — the opposite of what the doc promises. Separately, the `metadata` gate only fires on `pre_call`: `hookProjection.ts:83-115` calls `projectRequest(request, true)` unconditionally for `moderate`/`post_call`/`on_error`, so a `metadata` hook that also declares `moderate` still receives all messages. Fix: align the CRD doc with the code (absent → content) or actually implement the preCall-only→metadata inference in control-api; and apply the `metadata` gate on the other lifecycle points too (or reject the combo in the resolver).

5. **[R1-H4] The "grants NO egress (not even DNS)" test asserts a single NP, not effective pod egress (T4) — it masks R1-B1.** `host-context-controller/src/llmHookReconciler.test.ts:811`. It checks that the `llmhook-<podKey>` NP has no egress, but a pod's effective egress is the union of every NP selecting it — including `allow-dns-egress-llm-hooks`. The test stays green while the pod does get DNS. Should close together with R1-B1: assert the namespace's effective egress (reconciler NPs + infra policies), or assert `llm-hooks` isn't in `runtimeNamespaces` per the B1 decision.

6. **[R1-H5] `internal` tasks hang on a guardrail `ask` (they wait for an approval nobody will give).** `mcp-host/src/agent/taskExecutor.ts:1201`. `executionMode` only maps `'cron'→'unattended'`, so an `internal` task (autonomous, no human approver) gets `'interactive'` and an `ask` takes the suspension path (`toolUseLoopToolBatch.ts:332-334`) instead of the fail-safe deny (`:321-331`) → `pending_approval` with no responder → the task hangs. Fix: treat `internal` as `unattended` (`source === 'channel' ? 'interactive' : 'unattended'`), or verify internal tasks never reach the `ask` path.

## Optional (non-blocking)

None of these block the merge — follow-ups or your call.

**Medium**

1. **[R1-M1] `maxHookTimeoutMs` (admin limit) is never wired to the fetcher — silently ignored.** `mcp-host/src/core/adapters/hookedLlmPort.ts:302-309` passes `maxOutputBytes`/`maxRewriteBytes` but not `maxHookTimeoutMs`; `LlmLaneHookDeps` doesn't even declare `timeoutMs`, so `createHookFetcher` (`hookFetcher.ts:128`) always falls back to `?? 5000`. The field is new in this PR (`config.ts:34`) and orphaned — an admin setting 1000ms gets no effect while the sibling byte caps do wire through.
2. **[R1-M2] Tool `server=` rules match against the non-authoritative `__`-derived `sourceRef` → a `deny` can fail-open.** `mcp-host/src/core/guardrails/tool/provenance.ts:50-58` + `rules.ts:90`. Documented in the body as a deferred TODO (so not a body-mismatch), but a fail-open direction on a deny rule is worth gating those rules until `serverName` is threaded from `McpManager`.
3. **[R1-M3] No integration test for the install-hook saga.** Only upgrade-hook got tests. The install rollbacks (delete Secret when `createResource('llmhooks')` fails; delete CR+Secret when the step-10 Host update fails), the end-to-end gates, and `pendingCredentials` are all uncovered — the MCP saga right next to it has exactly these. `MockGateway` already supports `llmhooks`.
4. **[R1-M4] Zero test coverage for all the new control-ui logic (~1900 lines).** `HookInstallForm/index.tsx` (all-or-nothing credential validation `:135-160`, `hostCapabilityCeiling` `:37-45`, and the `may_deny ⇒ fail-closed` rule `:141-142` which is a §8.4 security invariant — T5), `GuardrailHooksTable`, `HostGuardrailsSection`, two pages: no tests. Extract the pure functions and cover them.
5. **[R1-M5] The LLM-lane production transports have no test.** `hookFetcher.test.ts` only exercises the injected-`fetchImpl` path, which skips the streaming byte-cap (`memoryCeiling`), the real `AbortSignal.timeout`, `makeSsrfGuardedFetch`, and `makeInClusterFetch` — exactly the "time/byte bounds" the intent promises.
6. **[R1-M6] install-hook doesn't validate `body.credentials` is an object.** `const credentials = body.credentials ?? {}` with no `credential.invalidPayload` guard (the MCP saga has it). `credentials: "x"` → `Object.keys("x")` enumerates indices → spurious 400 or malformed `stringData`.

**Low**

1. **[R1-L1] `console.error` at `control-api/src/routes/admin/registry.ts:2304`** bypasses the pino redaction pipeline (control-api logging rule). Low real leak risk here (`err` is a k8s error, `hookName` validated), but the whole file uses `console.*` incl. `auditLog` — clean fix is migrating the module to `rootLogger.child({module:'admin-registry'})`.
2. **[R1-L2] `canWrite={true}` hardcoded** in `control-ui/app/hosts/[name]/page.tsx:1035` — the per-agent guardrails Edit/Add button always shows. Backend is authoritative (not a hole), but a user without permission sees controls that fail on save. Wire it to the `host:write` check.
3. **[R1-L3] doom-loop doc drift:** `doomLoop.ts:6` says "the counter is persisted with task state" but the store is ephemeral (`core/types.ts:308-314`), and `doomLoop.ts:11` / `toolLaneAdapter.ts:87` keep "implement the counter" TODOs when it's already wired (`toolUseLoopToolBatch.ts:256-258`). Update the comments.
4. **[R1-L4] `may_add_context` is declared in every layer (CRD/Host/UI/mcp-host) but consumed by no execution path** — an admin can grant it and it's a silent no-op. Fail-safe.
5. **[R1-L5] Installed tool-lane hooks are unreachable:** the LlmHook CRD `lifecyclePoints` enum (`llmhook.yaml`) only allows `preCall|moderate|postCallSuccess|onError`, not the `preToolUse`/`postToolUse` points Host/mcp-host expect → `createResource` would reject them. Fail-closed feature gap (no tool-lane reference server either).
6. **[R1-L6] doom-loop deny omits `executionMode` in `recordDecision`** (`toolUseLoopToolBatch.ts:260`) → the `execution_mode` label is lost on doom-loop denials, unlike the sibling paths.
7. **[R1-L7] Breaker half-open** lets N concurrent probes through (`hookBreaker.ts:40`, not "one probe" as the comment says) and re-emits the "tripped" warn on re-open. Functionally harmless.
8. **[R1-L8] `decision.ts aggregateDecision` (pure precedence fold) has no property tests (T2)** — missing idempotence/composability via `fast-check` (already available).
9. **[R1-L9] `hooks/reference-guardrail-hook/deploy.yaml:34-39`** omits `readOnlyRootFilesystem`/`seccompProfile: RuntimeDefault` (N8) and `resources` limits, unlike the Deployment HCC generates. It's the example hook, but worth aligning.
10. **[R1-L10] `deploy/base/llm-hooks/rbac.yaml:124-127`** — `mcp-host-hook-reader` binds to the `system:serviceaccounts:mcp-host` group → any host SA can list/watch all LlmHooks in the namespace. Fine single-tenant; flagging for when multi-tenant lands.
11. **[R1-L11] Stale/contradictory comment in `control-ui/components/PublisherView/OwnedEntries.tsx:75-78`** (says the registry doesn't send `entry_type`; the new code and `lib/api.ts:3130` say it does).
12. **[R1-L12] `GuardrailHooksTable.tsx:118-122`** — the auto-refresh `setInterval` is torn down/recreated every render (`onRefresh=loadAll` not memoized). Benign; wrap in `useCallback`.

## What checked out clean

- **Structural contract seams line up field-for-field** (traced end-to-end): digest-pin (`sha256:` identical across control-api/HCC/mcp-host), `computePodKey` byte-match between mcp-host and HCC, DuplicatePath fail-closed, the `/v1` protocol across all 4 points (request/response mapped against `server.js`), and install/list/delete HTTP (control-ui ↔ control-api). `guardrails-revision` for *changes* holds via `LlmHookWatcher` + periodic re-resolve (full-removal is the R1-H2 gap).
- **Security posture is solid**: image hooks digest-pinned + allowlisted; trust level authoritative (publisher capped in non-curated orgs); content+egress gated to `high` trust; upgrade-hook re-runs the admission gates; digest-mismatch and DuplicatePath quarantine fail-closed; remote targets always through the SSRF-guarded transport; reconciler pods hardened (non-root, drop-ALL, RO-rootfs, seccomp, no SA token); `redact.ts` untouched; `on_error` scrubs Bearer/`sk-`/token.
- **Tests/typecheck green per area**: mcp-host LLM-lane 138/138, tool-lane+net 60/60, control-api 19 + 174 (`tsc` clean), HCC reconciler+mounts 74, control-ui `tsc --noEmit` exit 0 + Sidebar 17/17. Test quality is generally good (real `http.Server` in the bounds test, the real `createHookFetcher` in the toolLaneAdapter test, observable-result assertions in guardrailGate) — the exceptions are R1-H4 and the coverage gaps in M3/M4/M5.

One process note: this is 127 files across 6 services, well over the ~25-file/1-service budget — each review round has to re-scan the whole thing. control-ui came in clean (only test coverage), so it's a natural candidate to split out if you want to shrink the surface.

jozer-rami (commented):
## Review of the `upgrade-hook` path

Context: the registry side of this contract just landed (keyper-labs/evenfire-registry #70/#71/#72). While reconciling the two halves field by field I read this branch closely, and the `install-hook` / `upgrade-hook` pair is where the remaining mismatches sit. Reviewed at `3ef4369f`; everything below is verified against that commit, and **all of it is fixable here — nothing needs a registry change.** The registry's `contentAccess`, image-target-only `requiredEgress`, and immutable versions all arrive correctly; `upgrade-hook` just never reads two of them.

Worth saying up front: `install-hook` is thorough — the trust resolution, the content/egress gate, the rollback on Host-update failure, and the per-Host floor check on upgrade being *stricter* than install's single-Host one are all good. The findings are concentrated in `upgrade-hook`, which reads like it was written before some of those pieces existed.

### One correction to my own analysis, before the list

My first pass framed the missing entry-identity check as capability/credential *laundering*. **That framing is wrong and I'm not going to state it that way.** `/admin/*` sits behind `requireAuthForControlUI` (`control-api/src/app.ts:131-141`) — a verified, non-revoked admin session. Per the guardrails spec §12.1, that caller is already in the TCB: they could call `install-hook` with any entry, or deploy an arbitrary pod. There is no privilege boundary crossed here.

What remains is still real, just different in kind — see M1.

---

## Major

### M1 — an upgraded hook's recorded identity is never updated, so installed state can silently misdescribe what is running

`registry.ts:2287-2293` writes `{ spec: upgradedSpec }` with no `metadata`. `resourceService.updateResource` merges annotations over current (`services/resourceService.ts:341-352`), so omitting `metadata` preserves them verbatim: `clerum.io/catalog-version` stays at v1, `clerum.io/catalog-id` keeps pointing at the original entry, and `clerum.io/trust-level` keeps the install-time value **even though `trustLevel` was just recomputed at `:2214`**.

Compounding it, `upgrade-hook` never compares `catalog-id` against `body.registryEntryName` — `current` is read as `{ spec?: … }` (`:2171-2175`), so `metadata` is never in scope. The only cross-check is target *kind* (`:2228`).

**Trigger** (operator error, not attack): an admin upgrading `hook-pii` fat-fingers or copy-pastes the wrong `registryEntryName`. The call succeeds. The CR now runs the other entry's image while its annotations still say `catalog-id: @evenfire/pii-redactor` and `trust-level: high`, and it inherits the original's `envSecret`, `capabilities`, `imagePullSecrets` and `egressBindings`. `getInstalledRegistryState` builds `hookKeys` from exactly those annotations (`:900-911`), so the catalog UI reports the pre-upgrade version indefinitely and keeps offering the same upgrade.

And there is no trail: `auditLog('upgrade-hook', …)` at `:2314` carries `{hookName, digest, syncedHosts}` — no `registryEntry`, no `registryVersion`, no `trustLevel`, where install carries all three (`:2101-2107`). **Nothing in the system records which version an installed guardrail hook was moved to.**

Both sibling upgrade routes already do this correctly — `:2685` (mcpservers) and `:2924` (workflowrecipes) pass `metadata: { labels, annotations: catalogAnnotations(...) }`.

*Fix:* pass `metadata: { annotations: { ...catalogAnnotations(name, version), 'clerum.io/trust-level': trustLevel } }`; require `getCatalogId(current.metadata) === body.registryEntryName` (409/422); add the version + trust level to the audit line and the 200 body.

### M2 — the §8.4 gate is evaluated on the NEW `hook_meta`; the CR is built from the OLD spec

`assertHookAdmissible(entry, hookMeta, …)` at `:2209` decides on `hookMeta.requiredEgress`, `hookMeta.contentAccess` and `hookMeta.target.remote`. The written spec takes `egressBindings` and `security` from `...curImage` and `contentAccess` from `...curSpec` (`:2270-2280`). **Neither is ever read from the new `hook_meta` on this path**, though install reads both (`:1995`, `:2007-2009`).

The comment at `:2262-2266` lumps `security` and `egressBindings` in with genuinely install-time wiring (`envSecret`, `imagePullSecrets`), but those two come from the registry entry — they are version-shaped, and preserving them is what creates the divergence.

Two concrete outcomes:

1. **Stale egress.** v2 drops `requiredEgress`; the gate computes `hasEgress = false` and admits the upgrade as egress-free, while the CR keeps the old `egressBindings` and `llmHookReconciler.ts:504` keeps programming the NetworkPolicy. The reviewed *reduction* never lands. The reverse — v2 *adds* a destination — returns 200 with no rule, so the hook's calls are blocked at runtime.
2. **Silently neutered content hook.** v1 is a `preCall` shaper with `contentAccess: 'metadata'`. v2 publishes `contentAccess: 'content'` to become a real inspector. The gate treats it as content-bearing and admits it; the CR stays `metadata`; `hookProjection.ts:55-57` (wired at `guardrailHookResolver.ts:179-186`) strips every message body. **A moderation or PII hook is upgraded into a no-op, with a 200 and no warning.** Note `contentAccess` *is* runtime-enforced today, despite `charts/clerum-crds/crds/llmhook.yaml:182-185` still calling it advisory.

The exfiltration variant — stale `egressBindings` plus a newly content-bearing point — is **not** reachable, but only because the CEL rule at `llmhook.yaml:246` rejects `contentAccess: metadata` alongside `moderate|postCallSuccess|onError`. That is an accidental backstop in a different artifact, and it surfaces as an unmapped 500 (m5), not a gate refusal.

*Fix:* build the upgraded target from the new `hook_meta` the way install does, preserving only `envSecret` and `imagePullSecrets`. Then the gate and the CR are the same object.

### M3 — a v2 image on the private registry gets no pull secret; 200 returned, pod never pulls

No `shouldAttachEvenfirePullSecret` / `ensureRegistryPullSecret` anywhere in the upgrade block; install does both (`:1969-1985`, `:1996`). `imagePullSecrets` survives only if `curImage` already had it.

**Trigger:** hook v1 uses a public image; v2 moves to `registry.evenfire.ai/@org/hook@sha256:…`. Upgrade returns 200, the Deployment has no `imagePullSecrets`, no pull Secret exists in `llm-hooks` → ImagePullBackOff. Because `status.observedDigest` never advances, `guardrailHookResolver.ts:213-217` sees no mismatch and loads the descriptor anyway; the dial then fails — fail-closed for `may_deny`, silent for advisory.

This is the exact failure the recipe route guards against, and its comment says so in as many words (`:2905-2917`): *"an upgrade is exactly how a recipe moves from a public image to a private one, and without this the new CRD persists and then fails at pull time while the request returns 200."*

---

## Minor

- **m1 — 207 is fail-OPEN for advisory hooks, and the warning says the opposite.** The body says *"referencing agents fail-closed until retried"*; `guardrailHookResolver.ts:210-212` is explicit that a quarantined advisory hook "contributes nothing", so a `may_rewrite`-only redaction hook stops redacting silently on un-synced Hosts. Also: `syncHookRefsInHosts` throws on the first failure (`hostGuardrailRefs.ts:106-115`) and discards already-synced names, so the 207 can't say which Hosts are stale. And there's a transient even on success — between `:2287` and `:2296` the Host pins the old digest while the CR carries the new `lifecyclePoints`. (Retrying the same call *is* safe and converges — that part of the recovery advice is right.)
- **m2 — capability ceiling and `may_deny ⇒ failMode: closed` are install-only** (`:1872-1890`). Upgrade can't *widen* capabilities (they come from `curSpec`), so this is missed re-validation rather than escalation — except that `withinCeiling` (`mcp-host/src/core/guardrails/capabilities.ts:19`) has no production callers, making install the only place the ceiling is ever enforced. A Host whose ceiling was tightened after install keeps its over-privileged hook through every later upgrade.
- **m3 — `reportInstall` is not called** on hook upgrade (install `:2098`, mcpserver `:2729`, recipe `:2937` all call it), so hook upgrades are invisible to registry install accounting.
- **m4 — `newImage.ref.split('@')[1]`** (`:2198`, same at install `:2040`) takes the second segment while the guard `/@sha256:[0-9a-f]{64}$/` (`:2187`) is unanchored at the start, so `foo@bar@sha256:…` pins `bar` on every Host. `.split('@').pop()` plus an anchored pattern.
- **m5 — the new version's `credentialSchema` is never read.** `envSecret` keeps pointing at the v1 Secret, unknown-key validation doesn't run, and there's no `pendingCredentials` in the response (install `:2118-2129` and mcpserver upgrade `:2732-2740` both provide it). A v2 requiring a new key boots without it.
- **m6 — `updateResourceWithConflictRetry` at `:2287` has no `try/catch`**, so a CEL rejection (e.g. M2's contentAccess conflict) surfaces as an unmapped 500; install maps `extractK8sError` (`:2029-2034`).
- **m7 — no version ordering**, so `upgrade-hook` will move an installed hook backwards to an older version — and with M1, leave no trace.
- **m8 (install-side) — an arbitrary `hookName`** (`:1909-1914`) lets the same entry+version be installed twice; identical `computePodKey` + identical `spec.path` makes the alphabetically-later CR a `DuplicatePath` loser (`llmHookReconciler.ts:446-461`), silently quarantined (`guardrailHookResolver.ts:124-136`) while install returns 201.

---

## Checked and refuted — recorded so nobody re-walks it

- **Digest pinning lost for service/remote.** Target kind is immutable across upgrade (`:2216-2234`, with a test), and install never sets a digest for non-image targets, so `newDigest === undefined` can't strip an existing pin.
- **Upgrade can change target kind** — blocked at `:2228`.
- **Upgrade skips the allowlist / digest preflight** — it runs both (`:2186-2199`). (`enforcePluginImageAllowlist` does default to `false` at `config.ts:991-993`, but equally for install.)
- **`trustLevel` isn't recomputed** — it is (`:2214`), and it drives a per-Host floor check across *every* referencing Host before any write (`:2239-2260`), which is stricter than install. Only its consumption is missing (folded into M1).
- **Setting Host refs to exactly the new `lifecyclePoints` breaks phase semantics** — mcp-host ignores the phase key: `collectRefs` flattens all phases to `{id → digest}` (`guardrailHookResolver.ts:54-66`) and runtime points come from `cr.spec.lifecyclePoints`. Adding `moderate` *is* gated at `:2209`.
- **The 207 is unrecoverable / concurrent upgrades corrupt state** — both converge. `syncHookRefsInHosts` sets refs rather than appending, and interleaved writes land in the digest-mismatch quarantine state, which is fail-closed for `may_deny` and repaired by re-running either upgrade.

Happy to open a PR against this branch for M1–M3 if that's useful — they're contained, and M3 is close to a copy of the recipe route's existing block.

andypavia:
## Review response — what was fixed, what was verified on-cluster, and what is still open

Thanks both — between the two reviews the coverage was near-complementary: @zach88 took the network/HCC/mcp-host surface plus `install-hook`, @jozer-rami took `upgrade-hook` in depth. Almost nothing overlapped, which is why the open list below is short.

Branch is at `a12e72b5`. Everything below was verified by reverting each fix and confirming its test fails, so the guards are real rather than written to match the code.

---

### Resolved — @zach88's review

| Finding | Commit |
|---|---|
| **B1** `llm-hooks` in `runtimeNamespaces` (blocker) | `62a4f59e` |
| **H1** lost update on `Host.spec.guardrails.hooks` | `a234adf5` |
| **H2** guardrails removal did not roll mcp-host | `d612a567` |
| **H3** `contentAccess` CRD doc contradicted both consumers | `d90c76e3` |
| **H4** "no egress" test asserted one policy, not effective egress | `341e8aaf` |
| **H5** `internal` tasks could hang on a guardrail `ask` | `511f0123` |
| **M1** `maxHookTimeoutMs` never reached the transport | `d364ac57` |
| **M2** tool `server=` rules matched a `__`-derived name | `14de258e` |
| **M6** malformed `credentials` payload unvalidated on install-hook | `22eb19b0` |

Notes where our conclusion differed from the report:

- **B1** — the mechanism was real, but inert as deployed: the manifest sets `CONTEXT_MAPPER_RUNTIME_NAMESPACES` explicitly and has never included `llm-hooks`. It applied wherever the env var is unset (local runs, tests). The K8s-API half was overstated — that stays closed via the opt-in label. Removed the entry anyway: it earned nothing, its stated purpose was superseded the next day by the static `deploy/base/llm-hooks/networkpolicies.yaml`, and three docs still described the pre-change default.
- **H2** — the annotation defect was exactly as described, but the stated cause ("mcp-host reads guardrails only at boot") is not why it mattered; see the next section.
- **M2** — not reachable before the fix, because `callTool` refuses any name that does not split into exactly two parts. What the fix removes is the *accident*: two files parsing one string under different rules, with safety depending on the executor happening to be stricter.
- **H4** — closed at three layers: a config guard, a new effective-egress test that composes both reconcilers against the real config, and a `np-enforce-preflight.sh` check (`PROBE_HOOK_DNS=1` execs one resolve inside a sealed hook pod and expects failure).

### Resolved — @jozer-rami's review

**M1, M2, M3 and m1/m2/m3/m6** landed via #389. Reviewed it in depth before merge: restored the pre-PR source while keeping the test file and confirmed all three new tests fail, and the spec-build fix is a character-identical mirror of `install-hook`, not an approximation.

Two follow-ups on top:

- **m4 — `626de7ae`.** #389 anchored the digest pin on `upgrade-hook`; `install-hook` still had both halves (`/@sha256:…$/` unanchored, `split('@')[1]`), so `repo@bar@sha256:…` pinned `bar` into every Host's `guardrails.hooks[].digest` — the value mcp-host compares against `status.observedDigest` to detect drift. Both sites now share `isDigestPinnedImageRef` / `imageRefDigest`, since the two drifting apart is what produced this.
- **L5 — `a12e72b5`**, which turned out to be more than the reported enum gap; see below.

### Found while fixing, not in either review

- **`f4e8559e` — both CR watches died silently and never reconnected.** `HostWatcher`/`LlmHookWatcher` restarted only on a truthy error, but `@kubernetes/client-node` reports a normal stream end as `done(null)`, and the apiserver ends watches routinely (~30–60 min). Confirmed on-cluster: a pod up 3h23m had logged exactly one watch event (the boot `ADDED`), and a live edit to its Host CR produced nothing, with no error line. Every Host-driven live update — guardrails, model, `secretRef`, personalization, approval, failover — had been frozen for hours, recoverable only by rolling the pod. **This is also the real reason H2 mattered:** the pod roll is the delivery path that does not depend on the watch.
- **L5 was a latent hole, not just a fail-closed gap.** The enum omission is fail-closed as reported. But widening it alone would have opened one: `INHERENTLY_CONTENT_POINTS` covered only the three LLM-lane points, so a `preToolUse`-only hook classified as content-**free** — a low-trust third-party hook with `egressBindings` or a remote target would have cleared the §8.4 gate while reading every tool argument (`remoteToolHook` sends `{tool, arguments}`) and every tool result. The tool lane consults `contentAccess` nowhere. Both points are now inherently content-bearing and the CEL rule refuses `contentAccess: metadata` alongside them.

---

### Verified on-cluster (`clerum-test`), not only in tests

- **N5 containment** — probed 20 targets from inside a running hook pod: apiserver, control-plane services, control-postgres, cross-namespace workloads, the sibling hook pod, cloud metadata, public internet, DNS. All blocked. Positive controls included, since "everything blocked" is also what a broken probe looks like: `nc` connects on loopback, and mcp-host still reaches the hook. Ingress is scoped — control-api and channel-reader cannot reach the hook Service; only the referencing mcp-host can.
- **H2** — removing `spec.guardrails` moved the deployment generation 92 → 93, cleared the annotation and rolled the pod; restoring moved it to 94 and put the annotation back. `kubectl.kubernetes.io/restartedAt` survived both, so only the controller-owned key is stripped.
- **H1** — three rounds of two concurrent installs and two concurrent uninstalls against the same Host: refs went 2 → 5 → 2 every time, and the Host ended byte-identical to its pre-test backup. (A green result cannot prove the writes landed in each other's window — that is what the unit tests do deterministically — but it exercises the atomic path end to end.)
- **M1** — A/B across two real turns. At the 5000 ms default the hook compressed and mcp-host applied the rewrite; at `maxHookTimeoutMs: 50` the hook pod still logged `pre_call.compressed` while mcp-host applied nothing. The hook did the work; the client stopped waiting.
- **L5** — server-side dry-run: a `preToolUse`+`postToolUse` CR is accepted (rejected before), and the same CR with `contentAccess: metadata` is refused by the CEL rule.

---

### Open — deliberately not addressed here

**Test coverage (@zach88 M3/M4/M5)**

- **M3** — no integration test for the `install-hook` saga. The most valuable of the three: it covers the rollback paths and is the seam H1 changed, which is why H1's guard sits at the helper level rather than the route.
- **M4** — no tests for the new control-ui logic, including the `may_deny ⇒ fail-closed` rule that mirrors a backend invariant.
- **M5** — the LLM-lane production transports are still only exercised through an injected `fetchImpl`, so the streaming byte ceiling and the real `AbortSignal.timeout` are untested. The SSRF primitives themselves are well covered in `core/net/__tests__/ssrf.test.ts`. This gap is why M1 went unnoticed for so long.

**@jozer-rami's deferrals, unchanged — they are product decisions**

- **m5** `credentialSchema` unread on upgrade (changes the response contract), **m7** no version ordering (may be an intentional rollback path), **m8** duplicate install names producing a silently quarantined `DuplicatePath` loser behind a 201.
- **m1 residue** — the 207 wording is fixed, but `syncHookRefsInHosts` still throws on the first failing Host and discards the names already synced, so the response still cannot say which Hosts are stale.

**Lows** — L1–L4, L6–L12 untouched. Two confirmed while working nearby:

- **L9** — the reference hook is measurably less hardened than every HCC-managed hook: no `readOnlyRootFilesystem`, no `runAsNonRoot`, and it **does** carry a mounted ServiceAccount token (1205 bytes). Not exploitable — the `llm-hooks` default SA can do nothing and the apiserver is unreachable from that pod — but it is the example others will copy.
- **L1** — two `console.*` calls remain in the admin registry route, bypassing the pino redaction pipeline.

**Noted while working, no finding raised for them yet**

- `RUNTIME_TOKEN_REVISION_ANNOTATION` has the identical stale-annotation shape H2 had: stamped only when a value exists, so it never clears.
- A timed-out hook dial logs **nothing** under the default `strict` breaker policy — only a tripped breaker warns. We observed this directly during the M1 test: an aggressive timeout silently no-ops every hook with no signal in the logs.
- #389's upgrade response does not surface a `contentAccess` widening, and its entry-identity check is skipped for CRs with no `catalog-id` annotation (legacy, self-healing, but currently implicit).
- L5 makes tool-lane hooks *installable and correctly gated*; it does not make them usable — there is no tool-lane point selector in the install UI and no tool-lane reference hook, and an installed tool hook has not been exercised end to end.

**Environment note for anyone deploying this locally:** `CLERUM_REGISTRY_URL` in the minikube overlay points at `registry.evenfire.ai` while `REGISTRY_CONNECTION_MODE=self-hosted` takes credentials from the `registry_connection` DB row. If that row was issued by a local registry, control-api presents local credentials to the hosted one and every registry read fails with `invalid_client`. Repointing the ConfigMap at the in-cluster service fixes it; the overlay itself is unchanged, so this is still a decision to make.

zach88:
Took another deep pass, this time round 2 focused on the fixes from round 1. Short version: this round is really clean. Every round-1 Blocker + all 5 Highs + 3 of the 6 Mediums are closed and verified by repro (T3 red against the previous head on B1/H1/H2/H4/H5/M1), a dedicated collateral pass over the 12 fix commits found **zero regressions**, and security came back SAFE. A few fixes even generalized the class while closing it (digest-pin unified across install+upgrade, the malformed-credentials predicate shared between both install paths, the annotation-merge helper unified). Nice work.

What's left is test coverage and process, not broken code — plus two cluster actions that genuinely gate the merge.

## Verdict

🟡 **Approve with nits**
- **Must resolve before approval: 0** — no open Blocker/High and no open regressions.
- **Optional / non-blocking: 10** (4 Medium, 6 Low) + the round-1 nits still open.
- ⚠️ **Two pre-merge cluster actions gate the deploy (not code findings, but do them before merge):**
  1. Apply the structural CRD change first — `charts/clerum-crds/crds/llmhook.yaml` adds `preToolUse`/`postToolUse` to the `lifecyclePoints` enum (`maxItems 4→6`) and extends the CEL rule. Order is **CRD → control-api → control-ui** (`kubectl apply -f charts/clerum-crds/crds/llmhook.yaml`). This isn't cosmetic: the H3 fix leans on that CEL rule being live in the apiserver — the `contentAccess: metadata` + content-bearing-point combo is blocked by admission, not by runtime projection. If the CRD isn't applied, the apiserver prunes the new fields **and** that gap reopens at runtime.
  2. Clean up the orphaned `llm-hooks` NetworkPolicies on the dev cluster. After dropping `llm-hooks` from `runtimeNamespaces`, the policies HCC already created (`allow-dns-egress-llm-hooks`, `deny-all-llm-hooks`) are no longer reconciled, so the implicit DNS egress that B1 closes lingers until you delete them by hand.

## Optional (non-blocking)

Nothing here blocks the merge — these are follow-ups or author's call. Note the three re-triaged Mediums: they've survived two rounds, so they should either move up in severity or get closed in writing with a reason, not float into a third round.

1. **[R2-M1] Watch-reconnect test doesn't cover the bug site** — `mcp-host/src/__tests__/watchReconnector.test.ts`. The `f4e8559e` fix is correct (it reconnects on a clean stream close, `err === null`, not just on error), but the test exercises the isolated `WatchReconnector` helper, not the `doneCallback` in `k8sClient.ts` (`HostWatcher` ~L215-227, `LlmHookWatcher` ~L281-291) where the bug actually lived. If someone later reverts the null branch back to `if (err) reconnector.schedule(...)`, the suite stays green and the watcher goes deaf again after a routine apiserver close. Against the previous head the new test doesn't even compile (the helper didn't exist), so it's not a real T3. Suggest injecting a fake `k8s.Watch`, firing `doneCallback(null)`, and asserting the re-list — or if the class isn't injectable, that untestability *is* the finding.

2. **[R1-M3] No integration test for the install-hook saga** — `control-api` (re-triage, open since round 1). Only upgrade-hook has route-level coverage; the step-10 rollback and the new credentials guard are exercised nowhere. Recommend escalating this one — the rollback path is exactly where a silent failure hides.

3. **[R1-M4] Zero test coverage on the control-ui guardrails surface** — `control-ui/components/HookInstallForm` and friends (re-triage, open since round 1, unchanged this round). Best home for this is its own PR: the control-ui guardrails code came in clean in round 1 and didn't change in round 2, so it fits the "extract a converged slice" call. Suggest closing this in writing when control-ui is split out.

4. **[R1-M5] Production LLM-lane transports untested** — `mcp-host/.../hooks/__tests__/hookFetcher.test.ts` (re-triage, open since round 1). Only the injected-`fetchImpl` path is covered; the streaming byte-cap, real timeout, and SSRF guard aren't. Escalate or close in writing with an issue.

5. **[R1-L1] `console.error` leaks past pino redaction** — `control-api/src/routes/admin/registry.ts:2483`. The upgrade-hook ref-sync failure logs `body.hookName` (client-controlled) through `console.*`, which both allows log-injection via CRLF and skips pino's redaction. This is now on code this PR introduced (the upgrade-hook path). `err` is a k8s error so the blast radius is small, but the fix is a one-liner: `log.error({ err, hookName: body.hookName }, 'upgrade-hook host ref sync failed')` with a module child logger.

6. **[R2-L1] `WatchReconnector.begin()` can revive a cancelled reconnector** — `mcp-host/src/k8sClient.ts:130-132`. `begin()` clears `stopped` with no guard; a `start()` in flight after `stop()` would reactivate it. Very low probability (no `start()` after `stop()` in the lifecycle today).

7. **[R2-L2] Intermediate assertion instead of the observable** — `mcp-host/.../toolRegistryAdapter.test.ts` asserts `traceDescriptor().sourceRef` rather than the observable `deny` decision of a `server=` rule (T4).

8. **[R2-L3] Reconnect timer isn't `unref()`'d** — `mcp-host/src/k8sClient.ts:145`. A pending reconnect keeps the event loop alive until `stop()`. Fine for a daemon.

9. **[R2-L4] Incomplete class-check on the annotation-removal fix** — `host-context-controller/src/hostReconciler.ts:2388,4381`. `RUNTIME_TOKEN_REVISION` would hit the same "removed key survives the merge" bug that H2 fixed if it ever went present→absent, but it's always provisioned today so it's latent, not reachable. Worth adding to the `controllerOwned` list defensively, or a comment on why it can't happen.

10. **[R2-L5] Weak T3 evidence on the H2 test** — `host-context-controller/src/__tests__/hostDeploymentAnnotationMerge.test.ts`. Copied onto the parent it fails with a `TypeError` (missing symbol), not an assertion, so the commit's "tests fail against the previous merge" is imprecise. The fix behavior itself is verified by reading + the config/egress suites, so this is just about the evidence.

The remaining round-1 nits (L2, L3, L4, L6–L13) are still open and unchanged — same as reported last round, none blocking.

## What checked out clean

- **Zero regressions** — collateral pass over all 12 fix commits (`3ef4369f..a12e72b5`) traced every changed contract to its callers. Public signatures of `syncHookRefsInHosts`/`stripHookRefFromHosts` unchanged; uninstall re-derives with `withoutHook(hooks, hookId)` so a concurrent install's ref isn't clobbered; `internal→unattended` has no production producer; the digest-pin anchor only rejects the double-`@` garbage case.
- **Security SAFE** — every r2 fix either closes a fail-open (lost-update refs → `mutateResource` with a `resourceVersion` precondition; `internal→unattended` fail-closed; anchored digest pin closes `repo@bar@sha256:…`; authoritative server identity closes the `deny server=` fail-open; watch reconnect is fail-closed). `on_error` still scrubs Bearer/`sk-`/token.
- **Suites green:** control-api 232 passed (8 files), `tsc --noEmit` clean; mcp-host 77 + 178 passed; host-context-controller 46 + 88 + 49 passed (channel-reader credentials-revision regression suite included). T3 repros confirmed red against `3ef4369f` for B1, H1, H4, H5, M1.

zach88 (approved):

andypavia:
Thanks @zach88 — went and checked the two ⚠️ pre-merge cluster actions against both clusters. **One is already done; the second doesn't apply to either cluster we run**, though the underlying warning is valid for a case neither of us was looking at.

### 1. Apply the CRD first — ✅ already applied, and the ordering advice is right

Live on `clerum-test` since the L5 commit:

```
enum     = ["preCall","moderate","postCallSuccess","onError","preToolUse","postToolUse"]
maxItems = 6
CEL      = "...moderate/postCallSuccess/onError/preToolUse/postToolUse..."
```

Verified by server-side dry-run rather than by reading the object: a `preToolUse`+`postToolUse` CR is accepted (rejected before the change), and the same CR carrying `contentAccess: metadata` is refused by the CEL rule. So nothing outstanding here for that cluster, and the CRD → control-api → control-ui order is the right sequence for anywhere it hasn't been applied.

Two corrections to the supporting reasoning, since they change what happens if someone *skips* it:

- **H3 didn't touch the CEL rule.** H3 (`d90c76e3`) was a documentation fix to the `contentAccess` field description — stating that an absent value means `content`, which is what both consumers already did. The rule that changed in this PR is L5's, extending a rule that already existed to the two tool-lane points.
- **Admission isn't the only enforcement of that combination.** control-api rejects it at install with `content_access_conflict` (`registry.ts:265-278`), independent of the CRD. So an unapplied CRD doesn't reopen a runtime gap — for the tool points specifically it means those hooks can't be created at all, which is fail-closed. The field-pruning concern is real in general; it just doesn't produce the hole described.

### 2. Orphaned `llm-hooks` NetworkPolicies — ❌ not present on either cluster

**`clerum-test`:**

```
NAME                       POD-SELECTOR
default-deny               <none>                                    ← static, deploy/base
llmhook-370a6e4adb8615e8   clerum.io/hook-pod-key=370a6e4adb8615e8   ← per-hook, reconciled
llmhook-f3f6b8690695b933   clerum.io/hook-pod-key=f3f6b8690695b933   ← per-hook, reconciled

allow-dns-egress-llm-hooks      not found
deny-all-llm-hooks              not found
allow-hcc-api-egress-llm-hooks  not found
```

**`clerum-dev`:** the namespace `llm-hooks` doesn't exist, `llmhooks.clerum.io` isn't installed, and no NetworkPolicy matching `llm-hook*` exists in any namespace. Dev is still on a pre-guardrails build.

The reason is the same thing that made B1 inert as deployed — **both clusters set the namespace list explicitly, and neither has ever included `llm-hooks`:**

```
CONTEXT_MAPPER_RUNTIME_NAMESPACES = mcp-server,sandbox-recipes,rpc-proxy,sandbox-ui   # clerum-test AND clerum-dev
```

HCC therefore never created those policies, so there is nothing unreconciled to delete. Confirmed behaviourally too, not just from a listing: a live `getent` from inside a sealed hook pod on `clerum-test` still fails to resolve. If `allow-dns-egress-llm-hooks` were lingering, that probe would succeed.

**Where the warning *is* right:** any cluster that ran this branch with `CONTEXT_MAPPER_RUNTIME_NAMESPACES` **unset** — a from-source minikube, a non-manifest deploy, someone's local checkout — did get those policies from the code default, and they are now unreconciled and need deleting by hand. That's worth keeping in the release notes; it just isn't a pre-merge action for the two clusters we operate.

### On the rest

- **R2-L5** — fair hit, and my commit message was imprecise. The H2 merge-path tests fail against the parent with a `TypeError` (the helper didn't exist yet), not an assertion. I did verify the behaviour by reverting the drop logic *in place* and watching four of six assertions fail, which is the evidence I should have described.
- **R2-M1** — agreed. The reconnect test covers the extracted helper, not the `doneCallback` in `k8sClient.ts` where the bug lived, so a revert of the `err === null` branch would keep the suite green. The bug itself was verified on-cluster (a pod up 3h23m had logged exactly one watch event, and a live Host edit produced none), but that's a one-off observation, not a guard. Injecting a fake `k8s.Watch` and firing `doneCallback(null)` is the right shape.
- **R2-L4** — matches something I flagged while fixing H2: `RUNTIME_TOKEN_REVISION_ANNOTATION` has the identical present→absent shape and is latent only because it is always provisioned. The helper takes a list, so adding it is a one-word change.
- **R1-M3 / M4 / M5** — no argument on the re-triage. M3 is the one I'd escalate too: it's the seam H1 changed, which is exactly why H1's guard sits at the helper level instead of the route.

PR #351, opened 2026-08-13
feat(plugin-sdk): platform per-minute ENV rate limits, deprecate per-run quotas (#348)

alfredolopez80:
## Summary

Replaces the operator-settable **per-run** Plugin Workload SDK quotas (`maxRequestsPerRun`, `maxNotificationsPerRun`) with **platform-level per-minute rate limits driven by ENV vars**, for both `clientNotifications` and `promptBridge`.

Fixes the stepless / SDK-only ("eager") recipe trap: those recipes have no run lifecycle, so the per-run counter fell back to an epoch-sentinel bucket (`new Date(0)`) that **never reset** — a per-run cap became a lifetime cap. A per-minute time window resets regardless of run lifecycle, which is the correct primitive for platform (anti-DoS) protection. `promptBridge` usage/cost stays governed by **Token Budgets**; the rate limit only prevents platform failure, it does not meter usage.

Closes #348. The upstream LLM provider-boundary throttle remains tracked in #349.

## What changed

- **config** (`control-api/src/config.ts`): 4 fail-loud ENV limits via `positiveIntegerFromEnv` — defaults `notifications=150/min`, `promptBridge=120/min`, `request-bucket=600`, `pre-auth=600`. Grant per-minute overrides still win (`grant ?? config`).
- **enforcement** (`pluginWorkloadSdkQuotaTracker.ts`, `pluginWorkloadSdkAuthorizer.ts`, `pluginWorkloadSdkDb.ts`): per-minute defaults now come from config; the per-run leg (`consumeQuota`) and the dead epoch-sentinel DB helpers are **deleted outright** (not stubbed — a missed caller is a compile error). Off-by-one (`> limit`), DB-error fail-closed, and replay/idempotency short-circuits preserved.
- **middleware/routes**: request-bucket (600) and pre-auth (600) limiters ENV-wired; internal/admin `120` buckets intentionally left as literals (not the plugin abuse surface) with an explanatory comment.
- **admin API** (`routes/admin/pluginWorkloadSdk.ts`): validate-but-**strip** the deprecated per-run keys (malformed input still `400`s; GET path untouched).
- **control-ui**: per-run inputs removed; the grants table shows `platform defaults` instead of an unenforced `N/run`.
- **CRD / types / docs**: per-run fields marked **deprecated, not removed** (description-only CRD change, no version bump); stale default doc-strings corrected to 120/150.
[...]

PR #379, opened 2026-08-17
fix(workflow-recipes,control-api): promptly publish Plugin Workload SDK grant re-evaluation (#375)

### LINKED ISSUE #375: workflow-recipes: Plugin Workload SDK recipe stuck in awaiting_policy — grant creation/correction not promptly re-triggered (fragile reconcile chain + frozen verifiedAt)
## Summary

A WorkflowRecipe's Plugin Workload SDK capability can stay in `status.pluginWorkloadSdk.state = awaiting_policy` for a long time (observed ~40 min) after a valid operator grant already exists, because the WRC re-evaluation of the grant is **not promptly re-triggered** when a grant is created or corrected. Neither creating/updating the operator grant nor an annotation touch on the recipe forces a re-evaluation; the recipe only re-evaluates on a slower, indirect reconcile cycle. A related observability defect leaves `status.pluginWorkloadSdk.verifiedAt` frozen, so the status actively misleads operators (it looks permanently stuck even while it may eventually self-heal).

Discovered on the clerum **dev** GKE cluster while validating recipes after the deploy that carried PR #351.

## Not caused by PR #351 (confirmed)

PR #351 (issue #348, platform per-minute rate limits) touches **zero files** in `workflow-recipes/` or `mcp-host/` — the entire bootstrap/`awaiting_policy` state machine is outside its diff. The bootstrap validation never reads quotas (`v2Ready` in `control-api/src/routes/mcp-host/plugin-workload-sdk.routes.ts` checks policyState/revision/provenance/promptTargets/defaultTargetRef only; no quota read in `pluginWorkloadSdkValidator.ts` / `pluginWorkloadSdkProvisioner.ts` / `bootstrapIdentity.ts`). The only real link is temporal: the #351 merge triggered the dev deploy that restarted the WRC ~17:03 and reset its in-memory reconcile timers — a deploy-time coincidence, not a code cause.

## Observed live behavior (dev)

Recipe `sandbox-recipes/evenfire-vulnerability-plugin` (promptBridge):

1. mcp-host bootstrap pod Ready ~17:03:46; recipe first (and only) verified at `verifiedAt = 17:24:50` → `awaiting_policy` ("awaiting an operator grant"). The operator grant was created at 17:24:51 — 1.13s AFTER that evaluation (a policy-before-workload ordering race).
2. The grant was also misconfigured (a fallback target `glm-5.3`/zai not in the recipe's `allowedModels = [gpt-5.4-mini, claude-sonnet-4-6]`) — a separate operator error, corrected manually (grant `policy_revision` 1→2→3, last update 17:59:10, targets now match the spec).
3. Between 17:24:50 and 18:04:19 the recipe stayed `awaiting_policy` with `verifiedAt` frozen at 17:24:50, despite: two grant writes, a correct grant from 17:59:10, and a metadata annotation touch (no effect after 90s).
4. It finally self-healed to `validated` at `verifiedAt = 18:04:19` — ~5 min after the grant was corrected — on an indirect reconcile cycle (closest event: a periodic "Issued mcpHost credentials" pass at 18:04:17). So the re-evaluation is eventual but slow and not triggered by the grant change itself.

Contrast: recipe `eventasks` was evaluated with a grant already bound (`bootstrap_target_mismatch`); fixing that grant re-triggered evaluation and it reached `validated` promptly. The gap is specifically for recipes evaluated with NO grant bound (`awaiting_policy`).

## Root cause (file:line)

The design intent is level-triggered re-evaluation, but it hangs on a fragile mechanism:

- The 5s re-evaluation requeue for an SDK-only `awaiting_policy` recipe is set in `workflow-recipes/src/reconciler/workflowRecipeReconciler.ts` (`requeueAfterMs: TRANSIENT_REQUEUE_BASE_MS = 5000`, `requeueFixedInterval: true`).
- That requeue is driven only by an in-memory `setTimeout` chain scheduled when a reconcile pass completes (`workflow-recipes/src/k8sClient.ts:873-890`). The catch at `k8sClient.ts:897-907` logs `Reconciliation failed` on any throw (e.g. a transient 409/patch/API blip) and **never reschedules — the polling chain can die permanently**.
- **No watchdog covers SDK recipes**: the workload-status backstop loop explicitly excludes them (`k8sClient.ts:344`: `if (recipe.spec.pluginWorkloadSdk) return false`); the external-egress forced-reconcile loop only covers exact-host egress recipes; the watch has no periodic re-LIST resync.
- **Grant creation/update emits no trigger.** The admin upsert (`control-api/src/services/pluginWorkloadSdkDb.ts`, `routes/admin/pluginWorkloadSdk.ts`) writes Postgres only — no recipe event, no WRC ping.
- **Annotation touch is inert.** A generation-unchanged MODIFIED event is classified status-only and skips the full reconcile (`k8sClient.ts:811-816` + `isStatusOnlyModifiedEvent` ~`:916-923`).
- **Observability defect:** `shouldPatchRecipeStatus` (`k8sClient.ts:105-133`) skips the status patch when phase+message are unchanged, so `verifiedAt` stays frozen even when re-evaluation passes run — a frozen timestamp cannot distinguish "loop dead" from "loop alive but grant still not found", which actively hampered live debugging.

Live discriminator on dev: no `[WR-K8s] progress retry for "evenfire-vulnerability-plugin"` lines were present while the recipe was stuck, consistent with the fast 5s chain not running for that recipe (it recovered on a slower periodic pass instead).

## Impact

- Operator creates/corrects a grant and the recipe does not reflect it for minutes (up to the slow reconcile cycle), or — if the fast retry chain died on a throw — potentially not at all until an unrelated periodic pass or a WRC restart.
- The frozen `verifiedAt` makes the status look permanently wedged and misleads debugging.
- The natural operator remedies (grant re-save, annotation touch) do not promptly work, pushing operators toward heavier actions.

## Recommended fix (priority order)

1. **Watchdog for pending SDK states.** A periodic pass (mirroring `startWorkloadStatusRefreshLoop`, `k8sClient.ts:1010+`) that `forceReconcile` every cached recipe whose `status.pluginWorkloadSdk.state ∈ {awaiting_policy, degraded, provider_unavailable}`. Makes recovery independent of the fragile timer chain and inherently covers grant-created-after-bootstrap. Revisit the SDK exclusion at `k8sClient.ts:344`.
2. **Harden the retry chain.** The catch at `k8sClient.ts:897-907` must reschedule a backoff retry instead of dropping the timer chain on a throw.
3. **Event-driven nudge (optional).** On grant upsert, control-api pokes the WRC to reconcile the affected recipe. A plain recipe-annotation patch will NOT work unless `isStatusOnlyModifiedEvent` whitelists a nudge annotation (generation-unchanged events are discarded).
4. **Observability.** Let `shouldPatchRecipeStatus` (or a low-frequency periodic patch) refresh `verifiedAt` so the status honestly reflects the last evaluation time.

## Operator workaround (today)

- Correcting the grant eventually self-heals on the reconcile cycle (~minutes), but not promptly.
- To force it now: `kubectl rollout restart deployment/workflow-recipes -n control-plane` (fresh LIST → ADDED events → full reconcile of every recipe), or make a real spec change (generation bump) to the recipe.
- Do **not** delete the per-recipe `*-mcp-host` bare Pod (no owning controller; the WRC recreates it only inside a reconcile pass, so with a dead chain it would leave the recipe hostless). It is also unnecessary — grant pickup needs a reconcile pass, not a fresh pod bootstrap.

## Related issues (distinct)

- #301 (workflow-recipes: reconciler creates workloads before applying egress NetworkPolicy — TOCTOU) — NetworkPolicy ordering, not operator grants. Distinct.
- #204 (bug(hcc): decouple readiness from initial fleet reconciliation) — HCC readiness. Distinct.

No existing issue covers the SDK grant re-evaluation re-trigger gap.

alfredolopez80:
## Summary

Fixes #375. A Plugin Workload SDK recipe could stay in
`status.pluginWorkloadSdk.state = awaiting_policy` long after a valid operator
grant existed.

**Live clerum-dev evidence reframed the root cause.** Re-evaluation never
stopped — the 30s infrastructure watchdog and the 5s transient chain both ran
(138 `reconcile()` passes over the incident window, zero `Reconciliation
failed`). The dominant defect is in **publication, not re-evaluation**:
`shouldPatchRecipeStatus` never compared the SDK capability projection, so a
computed transition with no other observable diff emitted **no** status patch and
the CRD kept proclaiming the stale state (with a frozen `verifiedAt`). The
issue's recommendation #1 (a watchdog) was already satisfied pre-fix and is **not**
duplicated here.

## What changed

- **P1 — publication (WRC).** `shouldPatchRecipeStatus` compares the SDK
  capability projection (state, message, policyRevision, policyHash,
  defaultTargetRef, bootstrapPodUid, bootstrapProvider/Model/contractVersion,
  family flags), built once in `reconcile()` and reused by `patchStatus`. A
  computed `awaiting_policy ⇄ validated` transition — **and an in-place
  model/provider re-bootstrap** — now publishes immediately. `verifiedAt` is
  excluded from semantic equality and refreshed on a 5-min throttle guarded by
  `projectedMs > persistedMs` so it cannot loop. `validatedAt` is carried forward
  across steady-state refreshes so it keeps marking the true transition time.
- **P2 — robustness (WRC).** The `handleRecipeEvent` catch re-arms the transient
  retry chain (bounded backoff) instead of dropping it on a thrown reconcile; the
  loud error log and the 404 branch are preserved.
- **P3 — event-driven nudge (cross-service).** control-api emits an
  in-transaction `pg_notify('plugin_workload_sdk_grant_update')` on grant
  upsert/delete/revoke (COMMIT-delivered, ROLLBACK-dropped; revoke gated on a real
  mutation to avoid a teardown reconcile-spin). A new WRC `grantUpdateListener`
  (dedicated LISTEN client; reconnect/stop hardened) force-reconciles the affected
  recipe through the **existing** watchdog path, cutting grant-change convergence
  from ≤30s to ~1–5s. The 30s poll remains the backstop, so a lost NOTIFY is
  never worse than today; on env drift the namespace-mismatch discard is logged
  (diagnosable, not silent).
- **P4 — observability (WRC).** A forensic log when the computed SDK state
  diverges from the published one (states only, no secrets).

## Testing

TDD RED-first for every behavior: the `awaiting_policy → validated` incident
replica, the in-place model-change publish, P2 re-arm vs 404, P4 divergence log,
the grant-listener wiring/discard/resilience, and control-api NOTIFY (unit + a
real-Postgres COMMIT/ROLLBACK integration test, env-gated per repo convention).
An **additive** minikube e2e phase asserts grant-create → `validated` in ≤10s with
no WRC restart.

- `workflow-recipes`: `tsc -b` clean · vitest **2440 passed** / 29 skipped
- `control-api`: `tsc --noEmit` clean · vitest **4243 passed** / 96 skipped
- Reviewed across a multi-agent audit (correctness, test-guardian, code-review +
  find-bugs); all findings resolved.

## Notes for reviewers

- **Deliberate tradeoff:** a steady validated SDK recipe now incurs one
  status-only `verifiedAt` refresh patch per ~5 min (bounded, ≤1 patch / 5 min /
  recipe, generation-unchanged → status-only reconcile). This is the liveness
  signal issue #375 asked for; cost is linear in SDK-recipe count.
- **Disjoint from #351:** touches none of its rate-limit middleware, quota
  tracker, or capabilities route.
- **Detection-duty:** `dbRunProcessor` shares the same latent LISTEN concurrency
  patterns this new listener hardened against (out of scope here; flagged in a
  code comment).
- **T1 minikube (live): GREEN** — the full gate ran `Total 46 | Pass 46 | Fail 0`
  on locally-built images from this branch's HEAD (with PR #205 reverted, which
  independently fixed an unrelated HCC readiness livelock in the mcp-host). The
  additive nudge phase proved the P1+P3 chain end-to-end: a grant re-upsert
  re-published `policyRevision` and `state=validated` in ~5s (≤10s) via the
  event-driven NOTIFY→listener→forced-reconcile path, with **no WRC restart**.
  (The quota-limited claude fallback was pointed at OpenAI via an opt-in run flag;
  the two harness false-greens surfaced during certification are fixed in this PR.)


zach88:
Took a deep pass through this one — traced the NOTIFY/LISTEN contract end to end on both sides, ran the WRC and control-api suites, and checked the body's invariants against the actual code. The production code is genuinely solid: the notify is parameterized and transactionally coupled, the payload carries no secrets, the listener treats it as untrusted data, and the reconnect/stop handling is careful. Nothing here blocks a merge. The two things worth a look are both about the *promise* around the fix rather than the fix itself — one body-claim gap and one test-coverage gap.

**🟡 Approve with nits**
- **Optional / non-blocking: 7** (2 medium, 5 low)

## Optional (non-blocking)

This PR can merge without these — they're follow-ups / author's call. The two mediums are the ones I'd actually think about before merge.

1. **[R1-M1] The "never worse than the ≤30s backstop" guarantee doesn't hold for steady SDK recipes.** The body (and the contract comments in `control-api/src/services/pluginWorkloadSdkDb.ts:24-32` and `workflow-recipes/src/reconciler/grantUpdateListener.ts:14-19`) claim a dropped NOTIFY degrades to the ≤30s polling backstop. But the 30s loop `refreshWorkloadStatusesForSteadyRecipes` (`workflow-recipes/src/k8sClient.ts:1271-1325`) only does a grant-reading `forceReconcile` when the recipe independently qualifies — a named secret (`recipeReferencesNamedSecrets`, only `envSecret.name`/`imagePullSecrets`), an ownership backstop, or a not-ready workload. Otherwise it runs `observeCurrentWorkloadStatus`, which the reconciler leaves with `pluginWorkloadSdkProjection` undefined (`workflow-recipes/src/reconciler/workflowRecipeReconciler.ts:513-522, 1300-1312`). A grant mutation doesn't bump CRD generation, so the raw watch never fires for it either. Net: a healthy promptBridge/clientNotifications recipe with no `envSecret` and ready workloads qualifies for *none* of the periodic triggers — if the NOTIFY is lost (reconnect gap, or old-producer→new-consumer deploy skew), its grant change isn't re-evaluated until the next unrelated CRD change or a restart, which is exactly the #375 symptom coming back for that recipe. Key question that decides the fix: do real SDK recipes reliably carry a named secret? They provision JIT credentials rather than a static `envSecret`, so they plausibly don't — if that's right, either soften the claim in the body/comments or have the 30s loop recompute the SDK projection. If they always do carry one, this evaporates and it's just a comment tweak.

2. **[R1-M2] The transactional coupling of the notify — the whole point of the PR — has no test that goes red if it breaks.** The code today is correct (all three paths fire on the in-transaction `db` client, verified). The gap is coverage of the "delivered on COMMIT, discarded on ROLLBACK" invariant: the ROLLBACK test hand-writes `client.query('SELECT pg_notify(...)')` inside `withTransaction` (`control-api/test/pluginWorkloadSdkGrantUpdateNotify.realPostgres.integration.test.ts:145-191`) rather than driving a real producer, so it proves Postgres NOTIFY semantics, not that `upsert/delete/revoke` emit on the tx client; the COMMIT test uses real `upsertGrant` but a pooled connection and the tx client both deliver on COMMIT, so it doesn't distinguish them; and the unit lane routes `withTransaction(work)` to `work({query})` with `query === pool.query` (`control-api/test/services.pluginWorkloadSdkDb.test.ts:26-34`). So a future refactor swapping `notifyGrantUpdate(db,…)` for `pool.query(…)` — which breaks the ROLLBACK-discard — keeps all three lanes green. Repro: verified by wiring analysis + running the unit lane (35/35 green even though it can't catch this). Suggested fix: in the realPostgres test, force a rollback of a real `upsertGrant` by injecting a failure into `appendControlApiPermissionEventsInTransaction` (runs before the notify in the upsert path) and assert neither the row nor the NOTIFY survives.

3. **[R1-L1] `finalizePluginWorkloadSdkRevocation` mutates a grant without a notify** (`control-api/src/services/pluginWorkloadSdkDb.ts:1009`). Class-check: 4 runtime mutation sites (upsert, revoke, finalize, delete); 3 notify, finalize doesn't. It's defensible — the WRC itself drives finalize after proving teardown, so it already knows, and notifying on teardown is exactly what the B2 guard avoids — but the body says "upsert/delete/revoke" without calling out why finalize is excluded. Worth a one-line note so it doesn't come back as a "finding" next round.

4. **[R1-L2] The P4 divergence log misses the validated→no-capability cleanup case** (`workflow-recipes/src/k8sClient.ts:1032-1039`). When the capability is removed from spec the projection is `capability: null`, so `projectedSdkState` is `undefined` and the `projectedSdkState !== undefined` guard suppresses the forensic log — but that case *does* emit a patch, so it's precisely a computed-vs-published divergence P4 was meant to catch. Observability only. Could also log when `capability === null` and there's a `persistedSdkState`.

5. **[R1-L3] The e2e "WRC not restarted" check only compares pod names, not `restartCount`** (`scripts/e2e/e2e-plugin-workload-sdk.sh`, nudge phase). An in-place container restart (CrashLoop/OOM-with-restart) keeps the pod name, so convergence could come from a re-list on container startup and the test would still call it event-driven. Covers the main risk (pod replacement), not the secondary one — add `restartCount` to the snapshot to close it.

6. **[R1-L4] The nudge `ok` message can print `~Ns (≤10s)` with N>10** (`scripts/e2e/e2e-plugin-workload-sdk.sh`). `nudge_started_at` is set before `create_grant` but `nudge_deadline` is computed after, so if the admin call is slow the `ok` line can read "in ~12s (≤10s)" even though the gate passed within its real budget. Cosmetic — measure `nudge_elapsed` from the same point as the deadline.

7. **[R1-L5] `capabilityFamily` is emitted but never consumed** (producer `control-api/src/services/pluginWorkloadSdkDb.ts:50` → consumer `workflow-recipes/src/reconciler/k8sClient.ts:691`). The producer serializes it on upsert/delete and the listener parses it, but `onGrantUpdate` drops it (whole-recipe reconcile semantics). Not a bug — just a dead payload field; a one-line note keeps a future reader from assuming family-scoped reconciliation happens.

## What checked out clean

- **Security: clean.** `pg_notify($1,$2)` is parameterized (no channel/payload injection), the payload carries no tokens/secrets, `parseGrantUpdatePayload` never throws and validates types, the forced reconcile is bounded to already-watched recipes with a namespace match (`k8sClient.ts:702-723`) so no cross-tenant nudge, logs carry only `payload_length` / state enums, the P2 retry re-arm is bounded (`TRANSIENT_RETRY_MAX_MS=60s`, name-coalesced, unref), and there are no new deps (version bumps only).
- **NOTIFY/LISTEN contract: byte-for-byte aligned.** Same channel constant both sides, `{recipeNamespace, recipeName, capabilityFamily?}` camelCase, optionality matches, `JSON.stringify`↔`JSON.parse`, extra fields ignored (forward-compat), and new-producer→old-consumer is safe (Postgres discards a NOTIFY with no listeners; the grant tx never fails for lack of one).
- **SDK projection equality** compares every persisted field; `validatedAt` is carried only on the real transition, `verifiedAt` can't loop (5-min window, NaN → single refresh, clock-back → no patch), and the `capability: null` cleanup doesn't loop (merge-patch deletes the key). Verified.
- **"Notify only on a real mutation"** holds: delete `return false` when rowCount 0, revoke gated on a real transition, upsert always mutates.
- **The two harness false-greens** the body mentions (`cleanup_on_exit`, `ensure_external_rest_api_reachable`) are actually fixed, and the new nudge phase asserts observable state (`state=validated` + `policyRevision` bumped + pod-set unchanged in ≤10s).
- **Suites:** workflow-recipes 48/48 + 91/91 green; control-api `tsc --noEmit` clean, 35/35 unit, realPostgres skipped (env-gated, not a silent false-green).

zach88 (approved):

jozer-rami:
## Blocker: the P1 publication fix is unreachable for workflow-lane SDK recipes

Deep review across five independent lenses (correctness, security-tenancy, deploy-safety, concurrency, test-integrity), each finding verified against the code or refuted. The security and rollout-skew story holds up well (details in a companion comment). This one blocks.

### The gate

`shouldPatchRecipeStatus` returns before it ever reaches the new comparison:

```ts
// workflow-recipes/src/k8sClient.ts:117-151
export function shouldPatchRecipeStatus(recipe, result): boolean {
  if (result.skipStatusPatch) return false          // :121  <-- returns here
  ...
  if (pluginWorkloadSdkProjectionChanged(recipe, result)) return true   // :136  <-- the P1 fix
```

The reconciler returns `skipStatusPatch: true` from two steady-state short-circuits for recipes in phase `active`:

- `workflowRecipeReconciler.ts:1519-1543` (`currentPhase === 'active' && awaitsTriggeredRun && !wfExecPhase`). promptBridge recipes deliberately fall through at `:1526`. **clientNotifications-only recipes do not**, and hit `skipStatusPatch: true` at `:1541`.
- `workflowRecipeReconciler.ts:1640-1657` (`currentPhase === 'active' && !awaitsTriggeredRun`). Unconditional at `:1656`, with no family condition at all.

`workflowAwaitsTriggeredRun` (`:1025-1033`) is `runtime.mcpHost.required && !workflowRunId`, and `runtimePlan.ts:286-288` sets `needsMcpHostForSdk = pluginWorkloadSdkEnabled && Boolean(spec.pluginWorkloadSdk)` without inspecting the family. So with the flag on, a clientNotifications-only recipe lands on one short-circuit or the other in every configuration.

### Repro

Drove the real `WorkflowRecipeReconciler.reconcile()` on a clientNotifications-only SDK recipe in phase `active` with persisted `state: awaiting_policy`, using the existing harness from `src/reconciler/reconciler.test.ts`:

```
mcpHost.required matrix: {"instruction-step + flag ON":true,"instruction-step + flag OFF":true,
                          "snippet-step + flag ON":true,"snippet-step + flag OFF":false}
flagON reconcile   -> phase=active skipStatusPatch=true message="Workflow trigger infrastructure registered"
run-labelled       -> phase=active skipStatusPatch=true message="Workflow completed"
persisted=awaiting_policy projected=validated skipStatusPatch=true -> shouldPatchRecipeStatus=false
same result with skipStatusPatch:false                             -> true
promptBridge reconcile -> skipStatusPatch=undefined, workflowReconciler.reconcile called=true
```

Line 121 is the sole cause. The promptBridge path is genuinely fixed; the clientNotifications-only path is not.

### Why this matters for #375 specifically

control-api emits `capabilityFamily: 'clientNotifications'` on grant upsert and delete (`pluginWorkloadSdkDb.ts:65-70`), and routing is family-agnostic, so the NOTIFY is delivered and then wasted. The operator creates the grant, the listener force-reconciles, the projection is computed, and no patch is emitted. That is the original #375 symptom, unchanged, for that family.

### The obvious fix is wrong

Hoisting `:136` above `:121` would make things worse. On the short-circuit path the reconcile never gathers a bootstrap proof, so `buildPluginWorkloadSdkStatus` falls to the `BootstrapNotReady` branch (`pluginWorkloadSdkValidator.ts:389-409`) and projects **`degraded`**, not `validated`. Publishing that would flip `awaiting_policy -> degraded` and null out the entire `CLEARED_PLUGIN_WORKLOAD_SDK_METADATA` block (all nine fields including `bootstrapPodUid`, `policyRevision`, `policyHash`, `validatedAt`).

The correct fix is to stop short-circuiting clientNotifications-only recipes at `:1526`, mirroring the promptBridge fall-through, so the reconcile actually recomputes the capability. Short-circuit `:1640` needs the same treatment or an explicit SDK carve-out.

### Neither gate can catch this today

- All six new tests in the `issue #375` describe block (`k8sClient.test.ts:387-650`) hand-build a `ReconcileResult` that omits `skipStatusPatch`. `grep -n skipStatusPatch src/k8sClient.test.ts` returns 151, 263, 363, 381, 1934 and nothing in that range. Every one of them also uses `spec: { steps: [], pluginWorkloadSdk: { promptBridge: {} } }` with `clientNotifications: false`, so the family is never exercised in `shouldPatchRecipeStatus` at all.
- Both e2e fixtures (`tests/e2e/fixtures/plugin-workload-sdk-recipe.yaml`, `plugin-workload-sdk-layout-recipe.yaml`) declare `workloads:` with no `steps:`, so they take the stepless `isSdkOnly` lane (`workflowRecipeReconciler.ts:2457-2551`), which never sets `skipStatusPatch`. The T1 nudge phase re-upserts a **promptBridge** grant, so it exercises the one shape that already works.

### Regression status

Pre-existing gap, not a regression. Verified against merge base `4f1dd1aa`: the `:121` gate is already the first line of `shouldPatchRecipeStatus`, and both short-circuits already exist (base lines 1499 and 1615). This PR's `workflowRecipeReconciler.ts` diff touches neither. So the PR does not introduce the block, it ships a fix that reaches promptBridge and stepless SDK-only recipes and silently misses the clientNotifications-only workflow lane.

The underlying design issue is that `skipStatusPatch` conflates two different things: "a transient failure occurred, do not publish" (correct) and "steady state, nothing to do" (which is exactly when a newly computed capability transition needs publishing).

### Related, same root cause

`k8sClient.ts:419-428` requeues clientNotifications-only workflow-lane recipes every interval specifically so that "a clientNotifications-only eager mcp-host whose pod dies is never recreated" cannot happen. But `:1526` short-circuits those same recipes before any pod work runs, so that requeue burns a reconcile per interval and self-heals nothing. Pre-existing and out of scope here, but the same missing fall-through fixes both.

jozer-rami:
## Five major findings (companion to the blocker comment)

All five were empirically run, not reasoned about. Mutations were applied in a scratch worktree and reverted; `git status --porcelain` and `git diff --stat HEAD` both empty afterward.

First, what held up under pressure, because it is real work: `pg_notify($1,$2)` is parameterized with a constant channel so neither channel nor payload injection is possible; the payload carries three identifiers and no secret; the force-reconcile can only reach recipes already in the single-namespace watch cache, so cross-tenant reach is refuted rather than merely unobserved; rollout skew is safe in all four producer/consumer combinations with no new endpoint, table, migration, env var, or DB privilege; there are zero dependency changes (version counters only); the B2 no-op guard at `pluginWorkloadSdkDb.ts:948` genuinely closes the revoke/NOTIFY spin; the projection compares 12 of the 13 fields the CRD declares at `charts/clerum-crds/crds/workflowrecipe.yaml:2142-2179`, with only `validatedAt` deliberately excluded; and the identity-gate test is the strongest test in the PR (mutation-killed, real interleaving, real invariant).

---

### M1. Delete the line that makes the entire P1 fix work, and 2478 tests stay green

`workflow-recipes/src/reconciler/workflowRecipeReconciler.ts:1311`

That line is the only place the projection is attached:

```ts
result.pluginWorkloadSdkProjection = this.projectPluginWorkloadSdk(recipe, result)
```

Neutered it and ran the full suite:

```
Test Files  127 passed | 4 skipped (131)
     Tests  2444 passed | 29 skipped (2473)
```

Zero failures. Without that line `pluginWorkloadSdkProjectionChanged` short-circuits at `k8sClient.ts:195` (`if (projected === undefined) return false`) on every real reconcile, and the whole fix is inert in production.

Cause: all ~550 lines of new `shouldPatchRecipeStatus` tests hand-build the projection into the `result` object literal. Not one test lets production code produce it. The suite validates a pure comparison function that, as far as it can tell, is never fed.

Four sibling wiring mutations behave the same way, all surviving the full suite:

| Mutation | file:line |
|---|---|
| remove `await this.grantUpdateListener.start()` | `k8sClient.ts:792` |
| remove `await this.grantUpdateListener.stop()` | `k8sClient.ts:1390` |
| projection never attached | `workflowRecipeReconciler.ts:1311` |
| `existingCapability: undefined` (the R1 carry-forward wiring) | `workflowRecipeReconciler.ts:6093` |
| `patchStatus` always recomputes instead of reusing | `workflowRecipeReconciler.ts:6202` |

The `Object.create(WorkflowRecipeWatcher.prototype)` harness hand-installs every collaborator, so constructor wiring is never exercised.

**Fix:** one test that drives the real `reconcile()` and asserts `result.pluginWorkloadSdkProjection` is populated.

---

### M2. All 11 projection field comparisons are individually uncovered

`workflow-recipes/src/k8sClient.ts:204-215`

Removed each field comparison one at a time. Every single one survived. Disabling all 11 at once left only 2 tests red.

Five fields never differ between persisted and projected in any test, so they have zero exercise even collectively:

- `policyHash` (`:207`), the string never appears in the test file
- `bootstrapPodUid` (`:209`), always `'pod-abc'` on both sides
- `bootstrapContractVersion` (`:210`), always `2` on both sides despite appearing in the B1 test's name
- `promptBridge` (`:214`)
- `clientNotifications` (`:215`)

The other six differ only alongside at least one other differing field, so removing any one is masked. The B1 test changes provider and model together, so neither is pinned.

Also uncovered: `:200` `if (!current) return true`, flipped to `false` and survived. That is the first-ever publish of `status.pluginWorkloadSdk`. (`:199`, the null-clear branch, is covered.)

Separately, the flagship *"patches a computed awaiting_policy -> validated transition (issue #375 incident replica)"* test passes for the wrong reason. Its persisted `verifiedAt` is `'2026-08-17T17:24:50.000Z'`, over five minutes stale, so it satisfies the throttle branch at `:230-238` rather than the `state` comparison at `:204`. Disabling all field comparisons leaves it green. The test named after the incident does not pin the incident's mechanism.

**Fix:** a table-driven case per field varying only that field (11 small cases), and give the incident-replica test a fresh persisted `verifiedAt` so it fails via the state comparison.

---

### M3. The transactional NOTIFY coupling, the centerpiece of P3, has no test that goes red

`control-api/src/services/pluginWorkloadSdkDb.ts:669, 949, 1117`

Swapped all three `notifyGrantUpdate(db, ...)` to `notifyGrantUpdate(pool, ...)`, which breaks the COMMIT-delivered / ROLLBACK-discarded guarantee the design rests on.

```
tsc --noEmit  -> exit 0, no errors
vitest run test/services.pluginWorkloadSdkDb.test.ts  -> Tests  35 passed (35)
```

It typechecks because `DbClient` is structurally `{ query }` (`control-api/src/db.ts:35-37`), so `Pool` satisfies it. Consider narrowing that parameter to the session type so this class of refactor cannot compile.

Two independent reasons no lane catches it:

1. `test/services.pluginWorkloadSdkDb.test.ts:26-34` mocks `withTransaction: (work) => work({ query })` where `query` **is** `pool.query`. The transaction client and the pool are literally the same object, so the distinction is unrepresentable in that lane.
2. The rollback test at `test/pluginWorkloadSdkGrantUpdateNotify.realPostgres.integration.test.ts:145-191` hand-writes `client.query('SELECT pg_notify(...)')` inside `withTransaction` instead of driving a real producer. Deleting all three production notify calls entirely leaves that test **still passing**. It asserts Postgres' own transactional NOTIFY semantics, not this diff.

The test's inline comment concedes the shape ("a production mutation that fails after the notify cannot be constructed"), but the substitute tests the database rather than the change.

**Fix (cheap, no Postgres needed):** stub `withTransaction` to hand out a distinguishable client and assert `notifyGrantUpdate` receives that client rather than the pool. For the realPostgres lane, force a rollback of a real `upsertGrant` by injecting a failure into `appendControlApiPermissionEventsInTransaction`, which runs before the notify in the upsert path.

---

### M4. The listener permanently pins the second of only four Postgres connections

`workflow-recipes/src/config.ts:423`, `deploy/base/control-plane/workflow-recipes.yaml`

`poolMax: getEnvInt('POSTGRES_POOL_MAX', 4)`, and `POSTGRES_POOL_MAX` appears nowhere else in the repo, including the deployment manifest. So production pool max is 4.

Two holders now keep their client for the process lifetime: `dbRunProcessor.ts:238` (pre-existing) and `grantUpdateListener.ts:185` (new). Working headroom drops from 3 to 2, a 33% cut.

Concrete contention path: the 30s tick at `dbRunProcessor.ts:605-610` fires `void pollOrphans()` and `void checkStuckRuns()` concurrently (two checkouts), and a CRD `MODIFIED` arriving in that window calls `syncFromRecipeExecution` (`k8sClient.ts:982`) as a third. With two free slots the third waits and then rejects at `connectionTimeoutMillis: 5_000` (`db.ts:39`). `processPending` also holds a client across a K8s API create inside `BEGIN...COMMIT` (`dbRunProcessor.ts:444`), so one slow apiserver call pins a slot for its full duration.

Second-order: with the new re-arm at `k8sClient.ts:1080`, a reconcile that throws on a pool timeout now schedules a retry, so the failure adds load instead of dying quietly.

**Fix (free):** multiplex both channels onto one dedicated session, `LISTEN workflow_run_update; LISTEN plugin_workload_sdk_grant_update`, costing zero extra connections. Both handlers already filter on `msg.channel` (`dbRunProcessor.ts:219`, `grantUpdateListener.ts:120`), so this is close to a no-op change. Failing that, set `POSTGRES_POOL_MAX` to at least 6 in the manifest.

---

### M5. Two paths hand a still-subscribed client back to the shared pool, and the test fake hides it

`workflow-recipes/src/reconciler/grantUpdateListener.ts:166-173` and `:178-184`

Verified against `pg-pool/index.js:384-399`: `_release` re-attaches pg's own idle listener and pushes the **same** `Client` object back into `_idle`, calling `_remove` only when an error is passed. User handlers are never removed and no `UNLISTEN` or `DISCARD ALL` is issued, so the Postgres session stays subscribed.

Both of these plain-`release()` a client that has already run `LISTEN` and has both handlers attached:

- `:166-173`, stop() landed during the in-flight LISTEN round-trip
- `:178-184`, the non-overwrite defense-in-depth guard

Reproduced with a pg-pool-faithful fake:

- stop() during LISTEN: queries were `['LISTEN ...']` only, no UNLISTEN, the client landed in the idle pool with one `notification` listener, and emitting a NOTIFY on it still dispatched `onGrantUpdate` **after `stop()` had resolved**.
- stop() then start(): pg-pool re-issued the same client, the listener registered a second handler pair, and one NOTIFY dispatched twice. Each cycle adds a pair; at 11 cycles Node emits `MaxListenersExceededWarning`.
- stop()/start() parked inside the reconnect sleep: two concurrent `attachListener()` calls, and the `:178` guard released the loser without UNLISTEN, giving two dispatches for one logical NOTIFY. Notably **zero clients left checked out**, so the monotonic pool leak the identity gate targets is genuinely closed.

Worth separating clearly: the identity gate at `:153` does close the #205-class hole it was written for. What it does not cover is the `notification` handler at `:119-138`, which has no generation check at all, so a superseded-but-not-destroyed session keeps dispatching.

On test fidelity: the T1 fake at `grantUpdateListener.test.ts:224-226` models a released client as silently non-dispatching (`if (client.released) return`) and mints a fresh object on every `connect()`. Both diverge from pg-pool in exactly the two ways that would surface this, so the "exactly-once dispatch across 25 cycles" result is a property of the fake rather than of the code.

Production reachability today is low, since `provider.stop()` is shutdown-only (`main.ts:32`). The concern is that the poisoned client goes into the **shared** pool, where `dbRunProcessor` can check it out with this module's stale `error` handler still attached; that handler calls `attached.release(err)` on a client another consumer owns.

**Fix:** before each plain `release()`, `await attached.query('UNLISTEN ...').catch(() => {})` and `removeAllListeners('notification' | 'error')`, or simply use `release(err)` on these paths so pg-pool destroys the client. Add a monotonic generation counter checked in both handlers, and make the T1 fake reuse released clients and keep dispatching after a plain `release()`.

alfredolopez80:
Thank you for the mutation-verified review — this was exactly the kind of pass that catches what a green suite hides. All six are addressed in `de3253be`, and every fix is pinned by a test that goes **RED under your exact mutation** (independently re-verified: each mutation applied to the fixed tree, run, and reverted byte-exact).

**BLOCKER — clientNotifications-only workflow-lane publication.** Fixed by the fall-through you prescribed, not the hoist you warned against. Both short-circuits now let *both* SDK families fall through to the inner reconcile so it re-gathers the bootstrap proof and computes `validated` (hoisting would have published `degraded` and nulled the metadata block). The ownership/coordinator enforcement (`ensureCoordinatorGfsNetworkPolicyIfEnabled` / `ensureSteadyWorkflowRuntimeCredentials` / `revokeOrRequeueSteadyWorkflow`) still runs before the SDK carve-out, and a workflow-in-progress SDK recipe still short-circuits to protect the 409-window — that `|| wfInProgress` term is now itself test-pinned. Verified: reverting the fall-through turns the new real-`reconcile()` lane tests RED.

**M1 — projection never produced by production in a test.** Added real-`reconcile()` tests asserting `result.pluginWorkloadSdkProjection` is populated, plus real constructor/`start()`/`stop()` wiring. Deleting the attach line (`:1311`) now fails 5 tests; the four sibling wirings (listener start/stop, `existingCapability`, `patchStatus` reuse) each fail a test.

**M2 — 11 field comparisons uncovered + incident-replica riding the throttle.** Table-driven case per field with a fresh persisted `verifiedAt` (throttle closed) + a first-ever-publish case; each field's removal turns exactly its own case RED (including the five you found with zero exercise). The incident-replica test now uses a fresh `verifiedAt` so it can only pass through the semantic comparisons.

**M3 — transactional NOTIFY coupling had no red test.** Branded the transaction client (`DbTransactionClient`, unique symbol) so `notifyGrantUpdate(pool, …)` no longer **compiles** (`TS2345`); the unit tests now assert `pg_notify` on a distinguishable tx client; the realPostgres rollback test drives a **real** `upsertGrant` (an injected `appendControlApiPermissionEventsInTransaction` failure) instead of a hand-written `pg_notify`. Two honest notes: the incident-replica lone-`state` removal is killed by the M2 per-field case rather than that test, and the realPostgres db→pool swap is killed by the compile brand + the unit `txClient` assertions rather than the rollback lane (the injected failure precedes the terminal notify) — each finding is closed by a verified sibling pin, per the test's own comment.

**M4 — listener pinning the 2nd of 4 shared connections.** Gave it a dedicated single-connection pool; it never checks out of the shared pool (test-pinned that `sharedPool.connect` is never called), restoring shared headroom to 3 with no manifest change. I preferred this over multiplexing onto the `dbRunProcessor` session (which carries the latent reconnect issues this PR documents) and over bumping `POSTGRES_POOL_MAX`.

**M5 — still-subscribed clients handed back to the pool.** `releaseSubscribedClient` now `UNLISTEN`s + `removeAllListeners('notification'|'error')`, or `release(err)` to destroy if UNLISTEN can't be delivered; and a monotonic generation counter is checked in the **notification** handler (the gap the error-side identity gate didn't cover). The test fake is now pg-pool-faithful (reuses released clients that keep dispatching), so your three interleavings go RED without the fix.

Gates: `workflow-recipes` 2471 passed, `control-api` 4243 passed, realPostgres notify suite green.

Separate, out of scope for this PR (filing a follow-up): `control-api`'s tests live in `test/` outside `src/`, and its tsconfig only includes `src/**` with `rootDir: src`, so `tsc --noEmit` (locally and in CI) never type-checks them — a pre-existing gap distinct from `workflow-recipes` (whose tests are co-located in `src/`). The `internalPrincipal`-on-`revokePluginWorkloadSdkForRecipe` type errors are pre-existing and a symptom of that gap, not introduced here.

alfredolopez80:
@claude  can you re-check a full review and validate the all issue indicate by @zach88 and @jozer-rami it was resolved with a table (resolved/no resolved/partial, etc) and review fo the rest of PR

alfredolopez80:
### Adversarial re-validation of the six non-blocking nits (before merge)

Ran an adversarial, code-grounded re-triage of the six nits from the @claude re-review (all raised by @zach88), verifying each against the source at HEAD — the goal was to find which, if any, is *actually* merge-gating. **Result: none block merge.** One correction worth recording:

**R1-M1 ("worth a decision") — NON-ISSUE, the concern traces the wrong backstop lane.**
The "never worse than the polling backstop" guarantee was analyzed against the *workload-status* 30s loop (`refreshWorkloadStatusOnly`, gated on a named secret / not-ready workload). But **SDK recipes never enter that lane**: `workflowNeedsWorkloadStatusRefresh` returns `false` for any recipe with `spec.pluginWorkloadSdk` (`workflow-recipes/src/k8sClient.ts:452`, with the comment that a status-only read "would corrupt the persisted SDK projection"). SDK recipes are instead backstopped by `workflowNeedsInfrastructureReconcile`, which returns `true` **unconditionally** for them — stepless via the `!isWorkflow && spec.pluginWorkloadSdk` branch (`k8sClient.ts:413`) and steps>0-with-no-active-run (`:428`) — driven by `refreshRuntimeCredentialsForInProgressRecipes` (`:1348`) every 5–30s (`:59-60`), which enqueues a **full** reconcile (`handleRecipeEvent('MODIFIED', …, { forceReconcile: true })`, `:1354`). So a lost `NOTIFY` for a healthy clientNotifications recipe converges in ≤30s regardless of named secrets or readiness — the comment holds; the backstop is the infrastructure-reconcile lane, not the workload-status lane.

**The other five:** all safe to defer — comment/observability/test-hardening only.
- **R1-L1** (finalize has no notify): terminal `revoking→disabled` cleanup, reachable only after teardown is proven, sole caller/consumer is the WRC; the visible `revoke` already notified.
- **R1-L2** (P4 log misses `validated→capability:null`): forensic log only — the clearing **patch** is still emitted by `pluginWorkloadSdkProjectionChanged`.
- **R1-L3** (e2e "not restarted" compares pod names, not `restartCount`): the only one with a genuine (low-probability) test-integrity gap — it can mask an in-place container restart in the *side* assertion, not the primary ≤10s gate. Left as a deferred test-hardening.
- **R1-L4** (`~Ns (≤10s)` can print N>10): cosmetic — the pass/fail decision comes from the deadline set after the upsert, not the printed elapsed.
- **R1-L5** (`capabilityFamily` unused): forward-compat field; dispatch force-reconciles the whole recipe from the DB, authz is per-invocation in control-api.

**Actions taken (this branch):**
- Swept the four one-liners (R1-L1/L2/L4/L5) + enriched the R1-M1 backstop comment to name `workflowNeedsInfrastructureReconcile` — comment/cosmetic only, no behavior change (`24a8dc89`; WRC/control-api `tsc` + the touched suites stay green).
- One honest **pre-existing** residual surfaced (not a nit, not introduced here): a `steps>0` SDK recipe **while a run is actively in progress** hits neither `:413` nor `:428`, so a lost `NOTIFY` during the run isn't backstopped until the run ends. It does not affect enforcement (authorization is per-invocation against the DB), and "never worse than the pre-existing backstop" stays literally true. Filed as #390 for a durable follow-up.

Analytical note, not a formal approval.

PR #381, opened 2026-08-17
fix(minikube): bootstrap the initial Desktop GFS operator link

alfredolopez80:
## Summary

This is Phase 2 of #310 and closes the self-hosted installation loop. A README-from-zero minimal installation now consumes the supported setup path before any login, creates the governed `initial_setup` Desktop-to-Control-Admin relationship, and verifies that relationship on reruns. A future release containing #310 plus this follow-up will not silently create an ordinary Desktop account without the operator link required by the GFS authority model.

## What changed

- Make minimal seeding setup-first for `admin@evenfire.local`.
- Allow reruns to log in only after verifying exactly one active `initial_setup` link.
- Fail closed when the link is missing or revoked; no invitation flow, direct SQL insertion, backfill, or ordinary-user fallback is used.
- Verify the Electron runtime under the supported Node 24/npm 11 toolchain before Desktop validation.
- Accept digest-only local image manifests in the local Minikube path while retaining tag requirements for registry images.
- Reuse a clearly owned branch profile across commits using the exact pre-gate marker (`worktreeId`, `gitHead`, and `clusterFingerprint`).

## Relationship to #310

PR #310 provides the lifecycle, authority, revoke, tombstone, and generation-aware enforcement. This follow-up makes the README/self-hosted bootstrap produce the governed first relationship that #310 expects, so the first Desktop user receives operator GFS authority only through the explicit setup transaction.

## Validation

- T0 contract/scenario/syntax checks and `git diff --check`: passed.
- Fresh minimal bootstrap on the branch-owned profile `clerum-fix-minikube-bootstrap-link-07f6ca5f`: one active `initial_setup` link confirmed through read-only inspection.
- Focused Control API GFS Real PostgreSQL gate: 10 suites, 66 tests, 0 skips.
- GFSC lifecycle Real PostgreSQL checks: 4/4 passed (current-user retirement, revoked link, stale link generation, and stale admin session generation).
- Exact-head T2 preflight: passed with local images, the current worktree marker, and all required deployments Ready.

## Validation boundary

The broad Real PostgreSQL aggregator executed 91 assertions (83 passed, 8 pending) but its inherited migration fixture attempts to drop runtime roles that already own 92 shared objects in the Minikube `profiles` database. That is a fixture-isolation boundary, not a failure of the bootstrap path; the focused GFS suites and exact-head T2 gate are the authoritative evidence for this follow-up.

## Release note

Deploy this change together with #310. The setup endpoint remains one-shot by design; reruns verify the existing active relationship instead of consuming setup again.


jozer-rami:
## Deep review, with fixes pushed

Reviewed this branch across correctness, security/tenancy and test integrity, verified every finding against the code, then ran the bootstrap path on a real minikube cluster. Three commits are pushed on top of `7b1c6842`.

The core change is right and I confirmed its premise directly. `setupInitialAdminCredentials` only matches a bootstrap row whose `last_login_at IS NULL` (`control-api/src/services/adminAuthService.ts:352`), and `registerAdminSuccessfulLogin` sets `last_login_at = NOW()` (`:272`). A single admin login permanently destroys setup eligibility with no recovery short of rebuilding the DB, so setup before login is load bearing rather than stylistic. The API contract the seeder reads also checks out end to end: `gfsOperatorLink{status,source,desktopUserId,controlAdminId}` at `:36-45`, envelope `{admins:[...]}` at `:678-681`, and the flag already `true` in the minikube overlay.

### What was fixed

**1. The setup-first assertion could not fail.** `assert_minimal_seed_is_setup_first_and_link_fail_closed` bounded the minimal block with a range ending at the first bare `else`, which is nested far below and swallowed 487 lines, so its presence greps were satisfied by most of the file. Two mutations proved it: moving `login_admin_only` ahead of `perform_initial_setup`, and turning the missing-link `die` into a `log`. Both left the suite green. It now compares call positions inside a bounded dispatch and requires the missing-link branch to abort. Both mutations fail it.

**2. `! cmd` is exempt from `set -e`.** Every negative assertion in the shared-contract test was a no-op. Hard-wiring `clerum_initial_setup_link_matches` to `return 0`, which is the predicate deciding whether the operator link is valid, left the suite passing. Negative assertions now run through a `refute()` helper that exits non-zero on unexpected success.

**3. A missing link left operators with a dead end.** `REUSE_DB` reruns on any cluster that ever logged in hit the fail-closed path and got a bare refusal, even though the only recovery is a DB rebuild. Both the seeder and the setup abort now say so.

**4. `allowScripts` in `desktop-app/package.json` was inert.** It is a `@lavamoat/allow-scripts` field and there is no lavamoat dependency, so npm ignored it. `AGENTS.md` and `CLAUDE.md` documented it as the enforcement mechanism. Removed, and both docs now credit `verify:electron`, which is what actually catches a missing runtime. Added `engines: node 24.x` so npm enforces the runtime contract.

**5. The quickstart was gated on Node exactly 24.x.** The platform runs in containers and 24.x only matters for Desktop/Electron validation, so `check-prereqs.sh` is back to a `>= 24` floor with a warning on other majors. `verify-electron-runtime.mjs` enforces the exact major instead, before importing Electron, since a static import hoists above the check and would report `ERR_MODULE_NOT_FOUND` rather than the runtime mismatch.

**6. `make prereqs` failed in any git worktree.** `check-prereqs.sh` tested `$PROJECT_DIR/.env` while `load-dotenv.sh` deliberately resolves a worktree to the primary checkout and ignores a worktree-local file. The gate failed on a file setup would never read. Resolution is now shared via `dotenv_canonical_dir`.

Also hoisted the duplicated identity guard into the shared contract so both call sites enforce one decision, covered the multi-node `NOT_FOUND` image sentinel, and dropped a duplicate `profile-ui` entry that was building that service twice per CI run.

### Live verification

Fresh cluster, `SEED_PROFILE=minimal`. Precondition confirmed in the DB first: one bootstrap `admin`, email NULL, `last_login_at` NULL, zero operator links.

| Scenario | Result |
|---|---|
| Fresh install | setup attempted before any login, admin plus Desktop workspace created, active `initial_setup` link verified, exit 0 |
| DB row | `initial_setup / active / gen 1`, admin and desktop both `admin@evenfire.local`, `control_admin_id` matches |
| No duplicate member | 1 user total, no `test@clerum.io` |
| Rerun after setup consumed | 409, then login, then same link verified, exit 0 |
| Link revoked | exit 1, `status=revoked` detected, ordinary-user fallback refused, still 1 user |
| Link restored | exit 0, link verified again |
| Divergent identity | exit 1 via the shared guard |

### Two notes, neither blocking

`host-context-controller NOT ready` fires at the 180s budget during setup and then settles to `1/1 Running` on its own. That is the rollout budget issue being addressed on a separate branch, unrelated to this PR.

`seed-e2e-data.sh:366` only allows `clerum-test` and a GKE example, so testing on a custom profile needs `ALLOWED_CONTEXTS`. Correct safety behaviour, just worth knowing.

### Status

29 CI checks green, 3 pending, no failures. desktop-app, Real Postgres integration, shell syntax and leak guards all pass.

The PR description's Validation section now understates coverage, since it predates these commits. Worth a refresh before merge.

jozer-rami (approved):

PR #338, opened 2026-08-12
feat(mcp-host): classify provider 404 as ModelNotAvailable and stop silent failover

zach88:
## What
Provider 404s (a retired or inaccessible model) were classified as a retryable
ApiCallFailed, so the retry loop retried and ended in a silent cross-provider failover.
This classifies 404 as ModelNotAvailable (non-retryable, not failover-eligible) and finally
emits LLM_MODEL_NOT_AVAILABLE, across all four provider error classifiers (OpenAI-compatible,
Claude, Vertex, Bedrock).

## Why
A model going away is a configuration error, not something to route around. Silent failover
hides the real problem and can send traffic to an unintended provider. The runtime should
surface a clear, non-retryable signal so the operator can fix the catalog.

## How
- errorClassification.ts: case 404 -> ModelNotAvailable; body-level model_not_found /
  invalid_api_key; 401 and 403 both -> AuthenticationFailed (identity/authorization, on the
  `auth` failover class — genuine quota has its own channel); attach httpStatus/providerCode
  to every classification.
- claude.ts: classify by the SDK-modeled error.type read at the correct envelope nesting
  (`billing_error` -> InsufficientQuota, `permission_error` -> AuthenticationFailed); on
  fallthrough, providerCode is taken from the inner modeled type, not the outer envelope
  constant.
- openaiCompatible.ts: classifyError override for provider-native codes (z.ai, Bailian),
  including billing-in-429.
- Vertex (googleGenerative.ts) and Bedrock (bedrockConverse.ts), which carry their own
  classifier, get the same 404 fix plus httpStatus/providerCode threading. Their 403
  (PERMISSION_DENIED / AccessDeniedException) already lands on AuthenticationFailed, so 403 is
  now uniform across all four arms.
- httpStatus/providerCode propagate additively through ClassifiedError, LlmError and TaskError.

## Testing
Observable tests assert, across all four classifiers (OpenAI-compatible, Claude, Vertex,
Bedrock) with fixtures derived from the real SDK error constructors (T1), that a provider
404 classifies as ModelNotAvailable / LLM_MODEL_NOT_AVAILABLE with no cross-provider
failover and no tool-use-loop retry — asserted at the LlmError boundary. The end-to-end
TaskError surfacing (TaskError.code === LLM_MODEL_NOT_AVAILABLE) is covered for the Claude
arm in taskExecutor.test.ts; the LlmError → TaskError step is provider-agnostic. Full
mcp-host suite green (build, tests, typecheck, lint).

## Risks
[...]

Issue #373, opened 2026-08-17
HCC single-writer: residual two-writer race under node-partition + forced eviction (leader-election follow-up)


alfredolopez80:
## Summary

`host-context-controller` runs `replicas: 1` with `strategy: Recreate` and **no leader election** — a deliberate single-NetworkPolicy-writer safety model. This closes the common cross-writer window but leaves one **residual** race under a pathological failure mode, flagged during the PR #205 review (jozer-rami, https://github.com/evenfire-ai/evenfire/pull/205#discussion_r3796395400).

## The residual gap

Under a **node partition + forced pod eviction**, Kubernetes can start the new HCC pod while the outgoing pod is still alive (a partitioned zombie the kubelet could not terminate). In that overlap window the outgoing pod could win a `resourceVersion` conflict-retry `replace` and land its (wider/older) NetworkPolicy spec back over a policy the new pod already hardened and certified — with `/ready` still 200.

This is **not a correctness/corruption bug** and is **not exploitable** in normal operation (`Recreate` terminates the old pod first, so there is no overlap on the standard rollout path).

## Current defense in depth (why this is low-risk today)

- `serializeByHost` orders status writers **within** a process.
- The **D3 `resourceVersion` precondition + 409-retry** closes the cross-writer window against corruption.
- `Recreate` removes overlap on the normal rollout path.
- The **authority fence (`isCurrent`)** — now **required** on all three allow-granting reconcilers (`bindingPolicyReconciler`, `reconcileContext`, `reconcileExternalEgress`) after PR #205's J3 fix — aborts a superseded pod's write.
- `statelessLifecycleExecutor.ts:168-189` documents the single-replica invariant explicitly, and the deployment manifest documents the `Recreate` tradeoff.

## What would fully close it

Leader election with lease-based fencing (e.g. a `coordination.k8s.io` Lease + fenced writes), so a partitioned old pod loses its lease and its writes are rejected. This is an architectural change deliberately omitted by the current single-writer model, hence a follow-up rather than a PR #205 blocker.

Filed as a documented tradeoff sign-off follow-up alongside #368.

alfredolopez80:
Closing as not planned, together with #392.

[...]

PR #382, opened 2026-08-18
fix(hcc): decouple readiness from watch generation — fix the clerum-dev livelock

alfredolopez80:
## Decouple HCC readiness from watch generation — fix the clerum-dev livelock

Re-introduces PR #205 (HCC readiness decoupling) **with the fix** for the livelock that took it down on clerum-dev (reverted by #380), plus a hardening pass and the minikube reproducers that would have caught it.

### The bug (verified live on clerum-dev)
The #205 readiness gate pinned its NetworkPolicy safety certificate to the **watch generation** (channel identity). On GKE the apiserver closes long-lived watches (`Premature close`) every few minutes; each reconnect re-LISTs and bumps the generation **without any content change**, so every certification pass aborted before it could persist and `/ready` stayed **503 forever** — a livelock. With `strategy: Recreate` + `replicas: 1` the previous Ready replica was already gone, so the controller was effectively down. **minikube never reproduced it** (a single local apiserver does not drop watches), which is exactly why T2 stayed green while dev broke.

### The fix — certify by content identity, not channel identity
Pin the safety certificate (and its fences) to the **desired revision** (content), not the **watch generation** (channel): `isReadinessInventoryAuthoritative` / `record…Certificate` / diffing installers (`install{Context,McpServer,Host}Snapshot`) bump the revision **iff** desired state actually changed. A `Premature close` reconnect that re-LISTs identical inventory no longer invalidates the certificate; a real change still forces re-certification (fail-closed). Sentinel `-1` keeps the gate closed until the first certification. Phase 1b applies the same content-vs-channel treatment to the Host mutation-authority fence.

### Hardening (this PR)
- **Immediate watch-close recovery + 1s floor**: the three watch doneCallbacks start recovery immediately instead of waiting the fixed 5s timer; a floor demotes a re-close within 1s to the timer (anti-busy-loop). Measured effect: the in-churn 503 blip drops from **6s → 2s**.
- **Exponential backoff with jitter** on the retry-after-failure path (base 1s, cap 30s, full jitter with a 500ms floor, reset-on-success, honors `Retry-After` on 429). Rationale: a zonal control-plane upgrade cuts all ~6 HCC streams at once; without jitter they resynchronize and hammer the apiserver in phase.

### The safety net (the lesson made executable)
- `test-e2e-hcc-watch-churn-readiness` **induces** the `Premature close` regime minikube does not exhibit (self-flapping TCP proxy + clerum-dev-scale synthetic fleet) and proves the fix **converges** under churn: bounded 503 streaks, repeated 503→200 recoveries, and post-churn in-situ re-certification of the same non-restarted pod. **Live: 11/11 GREEN.**
- `test-e2e-hcc-rollout-readiness` (new) measures the **Recreate rollout downtime window** and reproduces the botched-rollout outage (D1b). **Live: the rollout recovers in 2s** (replacement pod certifies in 3s; a fresh pod, uid changed).
- **The livelock RED is hermetic and mutation-verified**: `k8sClient.test.ts` "GKE Premature-close regime" bumps the generation DURING each authoritative pass (the interleaving minikube cannot produce) and is RED against the pre-fix tree. **The NEGATIVE is NOT reproducible in minikube by design** (the authoritative pass runs at connection age < the flapper's min-age; any min-age that cuts it also kills the re-LIST both builds need). The e2e proves the POSITIVE; the RED lives in the unit test. This split is pinned by the contract gates.

### Validation
- **T0**: HCC unit suite 256/256, `tsc` clean, contract gates (watch-churn 12/12, rollout 13/13).
- **T2** (minikube, exact image = commit): watch-churn 11/11, rollout 8/8, full stack Ready.
- Regression is **mutation-verified** end to end; final adversarial audit **CERTIFIED** (no fail-open, no weakened tests, no regressions).
- T1 (real PostgreSQL) does not apply — this PR changes no PostgreSQL-backed service.

### Architecture decisions (reviewed & approved, registered here per the audit)
- **D1 — `Recreate` + `replicas:1` (no leader election)**: the single-writer safety model is real (RollingUpdate's two-replica overlap can re-apply a wider NetworkPolicy over a hardened one with `/ready` lying 200). When the HCC is down during a rollout, **NetworkPolicies stay enforced** — only convergence pauses. Measured rollout window: **2s** certification. Leader election is a separate project (and does not help in a zonal cluster during apiserver-down). *Follow-ups*: auto-`rollout undo` in the deploy pipeline; leader-election evaluation.
- **D2 — `/ready` semantics**: 200 now means "inventory certified, no stale allow", not "fleet converged" (the old coupled semantics caused the deploy timeout AND the livelock); may drop to 503 in-life on authority loss (fail-closed). Zero broken consumers (audited).
- **D3 — API-gateway readiness decoupled**: with a single backend, coupling would give zero endpoints on every HCC blip; a per-request 503 JSON is strictly better.
- **D4 — DNS external-egress fail-static**: a transient resolver failure **freezes** entries (does not widen, does not drop); only genuine NXDOMAIN expires. *Follow-up*: alert on `retries_at_cap` and frozen FQDNs.

### Deploy prerequisites (cross-repo)
- keyper-labs/evenfire-infra **#47** (HCC in the Recreate repair `TARGETS`): **merged**.
- keyper-labs/evenfire-infra **#42** (the Recreate repair must run in `mode=diff` too): **MERGED** (2026-08-18, merge commit `bfd81ed6`) — the `mode=diff` stage now accepts the `rollingUpdate`→`Recreate` transition.

### Deferred (follow-up issues)
Phase 2 (resume watches from `resourceVersion`/bookmarks, #383); auto-rollout-undo; leader-election evaluation; egress-NP stale-IP reconnection budget (c3 — the "evil twin" of #205); egress DNS alerting.

### Evidence freshness (zach88 R1-B5)
The T0/T2 **runtime** evidence above certifies the `host-context-controller` **production** tree at `24bc09da`. Since then the only HCC change is **test-only**: the nit #1 invariant guard in `k8sClient.test.ts` (plus a `package.json`/lock version bump). `git diff 24bc09da..HEAD -- 'host-context-controller/src/**.ts' ':!host-context-controller/src/**.test.ts'` is **empty**, so the production code is byte-identical and the runtime evidence stays valid. The two later `dev` syncs (`4d080b5a` #379, `9bc54247` #338) touch no HCC path.


alfredolopez80:
@claude, do a complete review of this PR, taking into account the entire history of PR 205, the problems that arose, the follow-ups created, tests, and potential issues in the GKE deployments, so that everything is part of the analysis and we can see if we missed anything.

alfredolopez80:
> **Update (2026-08-18):** evenfire-infra **#42** is now **MERGED** (`bfd81ed6`) — the deploy precondition described below is satisfied. Original request preserved for the thread record.

---

@claude — please do a **complete re-review** of this PR, and this time explicitly **validate the resolution status of every issue raised across the whole lineage of this work — not just the items from your last review**.

### Context update — the deploy-order blocker is being closed out

The one must-do you flagged (evenfire-infra **#42** must merge before this PR is deployed) is now **in review pending approval**:

- Every review finding that was *adjustable inside #42* has been fixed and pushed (commit `926c901`): CRD-guard `MODE` validation, stderr capture so a Service-Account `Forbidden` on CRDs surfaces its real cause instead of a bare exit code, the promotion-guard tests hardened to fail-loud (a renamed Diff step no longer silently skips the ordering check), and the `mode=diff` "read-only" contract documented honestly (it is read-only except the idempotent Recreate strategy repair).
- infra CI is green (the `test-promotion-guards` suite runs in the runner), the PR is MERGEABLE, and an independent audit certified the changes do **not** break the live dev/prod GKE deploy flow — WIF Service Accounts, Cloudflare tunnel ingress, RBAC, and image overlays are all untouched; the repair script itself is byte-identical.
- The only remaining deploy-time condition is certifying the deploy Service Accounts (dev `github-actions-releaser` / prod `gha-prod-deployer`) can dry-run-patch CRDs — which #42's new stderr fix makes self-diagnosing on the first run.

### What I need from the re-review

A **resolution matrix over ALL issues from the full history**, each marked `resolved` / `not resolved` / `deferred (tracked in #…)`, so we can confirm nothing was silently dropped in the re-introduction:

1. **Your own #382 review items** — the 6 low nits (`sameHostDesiredRevision` comparing only uid+generation; churn-e2e pinning name+restartCount vs the rollout gate's uid; the Retry-After branch applying no jitter; api-gateway returning bare 502 instead of 503 JSON during the Recreate gap; the `ROLLOUT_DOWNTIME_BUDGET` loose-bound label; and the "mutation-verified" claim needing the run attached), the **Medium** operator note (Recreate + replicas:1 unavailability window during rollout), and the `sameHostDesiredRevision` guard/test you asked to track.
2. **zach88's issues from PR #205** — he approved #205 but raised 6 comments there. Confirm each concern still holds or is resolved in this re-introduction.
3. **jozer-rami's issues from PR #205** — formal review (COMMENTED) plus 4 inline comments and 1 top-level. Same treatment.
4. **Any other reviewer** across the lineage **#205 → #380 (revert) → #382**, and confirm the deferred follow-ups are still tracked (Phase 2 `#383`, and the follow-ups opened for this work `#391`–`#394`).

Please be exhaustive and cite `file:line` where a fix landed. The goal is a single, auditable answer to "did we miss anything from the entire history?" — every issue either resolved (with evidence) or explicitly deferred with a tracking issue. Call out anything that is neither.

alfredolopez80:
## Mutation-run artifact for the hermetic livelock RED (closing @claude re-review nit #6)

@claude's re-review flagged that "mutation-verified" for the hermetic RED rested on a comment claim it couldn't confirm by static reading. Here is the run.

**Test under proof:** `host-context-controller/src/k8sClient.test.ts` → describe *"McpServerWatcher readiness under sustained watch churn (GKE Premature-close regime)"* → `it('certifies readiness under sustained Context watch churn when every re-list returns identical inventory')`. It drives 5 authoritative passes, bumping `contextWatchGeneration`/`mcpWatchGeneration` by +2 each pass (a Premature-close reconnect) while the desired **revision** stays fixed (identical inventory), then asserts `isReadinessInventoryAuthoritative() === true` and `/ready` → `200`.

**Mutation (revert the fix — re-pin readiness to CHANNEL identity):** in `k8sClient.ts` `isReadinessInventoryAuthoritative()`, gate on the watch generation instead of the desired revision:

```diff
-      this.networkPolicyRevocationContextRevision === this.contextDesiredRevision &&
-      this.networkPolicyRevocationServerRevision === this.mcpServerDesiredRevision
+      // reproduces the #205 channel-identity livelock
+      this.networkPolicyRevocationContextRevision === this.contextWatchGeneration &&
+      this.networkPolicyRevocationServerRevision === this.mcpWatchGeneration
```

**RED (pre-fix / channel identity) — reproduces the clerum-dev livelock:**

```
 ❯ src/k8sClient.test.ts (257 tests | 1 failed | 251 skipped)
   × certifies readiness under sustained Context watch churn when every re-list returns identical inventory
 FAIL  … > certifies readiness under sustained Context watch churn when every re-list returns identical inventory
 AssertionError: livelock: 6 authoritative passes completed under sustained churn
   (6 reconnects, identical inventory on every re-LIST) and none certified the
   safety inventory: expected false to be true
     9273|  expect((await requestReadyOverHttp(server)).statusCode).toBe(200)
 Test Files  1 failed (1)   |   Tests  1 failed | 5 passed | 251 skipped (257)
```

Only that one test flips: the fail-closed guardrail and the revision-bump unit guards in the same describe stay green, so the mutation isolates exactly the channel-vs-content decision.

**GREEN (fix restored / content identity):**

```
 Tests  1 passed | 256 skipped (257)   —   /ready 200, certifies through the churn
```

**Conclusion:** the hermetic RED is genuinely mutation-verified — against channel-identity (the #205 code) it livelocks (`/ready` never leaves 503), and the content-identity fix certifies through the same sustained churn. The production tree was mutated locally only for this run and restored (no code change in this PR from the artifact; `git status` clean afterward).

alfredolopez80:
@claude — the branch is now **synced with `origin/dev`** (merge `9bc54247`, bringing #338's mcp-host provider-lifecycle work; zero file overlap with the HCC changes here). Please do a **final complete re-review** and confirm that everything across the lineage (#205 → #380 → #382) is either **closed with evidence** or **tracked as a follow-up issue** — i.e. nothing is silently dropped.

### What changed since your last re-review

- **nit #1 — `sameHostDesiredRevision` invariant** → *closed in-PR* (commit `903684c8`). Added the executable guard test you asked to track: annotation/label churn at the same `(uid, generation)` must not bump `hostDesiredRevision`. Mutation-verified — hashing annotations in the comparator turns it RED. Full `k8sClient` suite 257/257, `tsc` clean.
- **nit #6 — attach the mutation run** → *provided* (see the "Mutation-run artifact" comment on this PR). Re-pinning readiness to channel identity reproduces the hermetic RED (`/ready` never leaves 503; `expected false to be true`), and the content-identity fix certifies through the same churn. Production mutated locally only for the run and restored.
- **R1-B5 — evidence freshness** → *closed*. Added a note to the PR body: the T0/T2 runtime evidence certifies the `host-context-controller` tree at `24bc09da`; `git diff 24bc09da..<pre-dev-sync> -- host-context-controller/` is empty (this dev sync also touches no HCC path).
- **The "unaccounted for" #205 items you flagged** → I re-read the actual #205 threads (zach88's 6 comments + jozer-rami's review + inline, which your job's blocked `gh` couldn't open) and mapped each to the current tree. Four pre-existing, previously-untracked items surfaced; **all are now tracked**:
  - **#403** McpServer resource clamp bypass (Medium) · **#404** `getServerInfosByContext` returns `[]` for a missing Context (Medium) · **#405** api-gateway 502→503 JSON in the Recreate gap (Low, your nit #4) · **#406** jitter on the honored Retry-After branch (Low, your nit #3) · **#407** grouped test hygiene (Low, your nit #2 + zach88 R1-Lx / R1-M3 sub-gap).
  - **Two High items** (both pre-existing, not introduced by #205 or #382): one is already being resolved on a separate branch and is tracked privately (kept out of this thread deliberately); the other is filed and **assigned as #408**.
- **jozer-rami's #205 items (J1–J4)**: all resolved or decision-recorded (the 480s→900s timeout J2 is now `--timeout=900s` in all three call sites + the stub; the `isCurrent` fail-open J3 is a required arg; the Recreate sign-off J4 is D1 in the body with the race tracked in #373).

### Deploy precondition — now satisfied
> **Update (2026-08-18):** evenfire-infra **#42** is now **MERGED** (merge commit `bfd81ed6`). The `mode=diff` deploy stage accepts the `rollingUpdate`→`Recreate` transition. Its adjustable findings were fixed, infra CI was green, and an independent audit certified the changes don't break the live dev/prod GKE deploy. The cross-repo deploy precondition is satisfied.

**Question for the re-review:** is there anything in the lineage that is neither closed-with-evidence nor tracked in a follow-up issue? If not, this PR is review-ready (the evenfire-infra #42 deploy precondition is now merged).

zach88:
Took a deep pass through this one, focused on the readiness cert core, the netpol/egress safety fence, watch recovery/backoff, and the cross-service `/ready` contract. The core of the fix is solid: the revision-vs-generation switch, the fail-closed gate, the backoff/jitter, the Host mutation fence, and the consumer contract all held up under verification, and the livelock T3 is a genuine RED against the pre-fix tree. There's one real problem though, and it lives in the exact invariant this PR is about (fail-static egress).

**🔴 Not ready to approve**
- **Must resolve before approval: 1** (1 Blocker)
- **Optional / non-blocking: 5** (2 Medium, 3 Low)

1. **[R1-B1] The new DNS bounded-timeout drops accumulated egress instead of freezing it (breaks D4 fail-static).** `host-context-controller/src/networkPolicyReconciler.ts:1738-1748` (the throw) → misclassified at `:2141`. The timeout wrapper this PR adds rejects with a plain `new Error('DNS resolution timed out after …')` that carries **no `.code`**. In the catch, `classifyDnsError(err)` maps codeless errors to `'permanent'` (the NXDOMAIN/expire branch), not `'transient'` (freeze). A real node-resolver `ETIMEDOUT` is in `TRANSIENT_DNS_CODES` and would freeze; this synthetic race-timeout doesn't. At the merge-base there was no wrapper, so a slow-but-erroring resolver surfaced node's coded `ETIMEDOUT` and froze correctly — the wrapper reintroduces the drop.
   - Repro (verified two ways): `classifyDnsError(new Error('DNS resolution timed out after 5000ms')) === 'permanent'` (vs `'transient'` for `{code:'ETIMEDOUT'}`); and `reconcileEgressState(previousWithLapsedEntry, [{kind:'permanent'}], now, cfg)` keeps **0** entries where `{kind:'transient'}` keeps **1**. Under a sustained resolver stall (each attempt exceeds the 5s timeout for longer than TTL+overlap), the accumulated /32s lapse and get pruned, so the McpServer's external egress is dropped.
   - Fix: make the timeout error transient, e.g. `Object.assign(new Error('DNS resolution timed out after …'), { code: 'ETIMEDOUT' })`, so it lands in the freeze branch. Close it with the regression test in R1-M1 below.

## Optional (non-blocking)

The PR can merge without these; they're follow-up or your call. That said, I'd fold R1-M1 into the R1-B1 fix, since it's the test that should be catching this.

2. **[R1-M1] The only timeout test doesn't cover the freeze case, so it shields the R1-B1 misclassification.** `host-context-controller/src/networkPolicyReconciler.test.ts:2490-2511`. It uses a server with no pre-existing accumulated policy, so on timeout `accumulated.cidrs.length === 0` and the reconcile fails loud — the `rejects` + no-create/replace assertion holds identically whether the timeout is classified transient or permanent. The real regression test seeds an existing policy with STATE annotations whose window has lapsed, stalls `dns.resolve4`, advances timers, and asserts the rendered policy **still contains** the accumulated /32. That one must fail against the pre-fix tree.

3. **[R1-M2] The fleet-scheduler merge/precedence functions have no property tests.** `host-context-controller/src/hostFleetScheduler.ts:28-56` (`hostFleetRequestCovers`, `mergeHostFleetRequests`). This PR adds a new `inputRevision` dimension to the merge, and these are exactly the `(state,input)→state` precedence functions that want property coverage (idempotency, composability). I checked by hand that both hold today, so this is a coverage gap, not a live bug — but it's the spot where a future precedence change would regress silently. `fast-check` over total-order / no-drop / idempotency / composability.

4. **[R1-L1] The k8sClient mock hand-reimplements a safety-critical comparator.** `host-context-controller/src/k8sClient.test.ts:314-321` stands in `sameContextDesiredRevision` with a divergent version (ignores `uid`/`generation`, order-sensitive `JSON.stringify` where prod canonicalizes). Mitigated by direct unit tests on the real comparator, but `importActual` would exercise prod semantics.

5. **[R1-L2] Dead test code writes fields that don't exist in prod.** `host-context-controller/src/readinessSafetyFence.integration.test.ts:154-155` sets `networkPolicyRevocationContextGeneration`/`…ServerGeneration`, which no longer exist on the class (only the `…Revision` fields remain). Harmless (never read) but it implies a generation-based revocation still exists. Drop the two lines.

6. **[R1-L3] `revalidateHostMutationBoundary` discards a deep clone per call.** `host-context-controller/src/hostReconciler.ts:3850-4110`. Each of the ~34 boundary checks runs `revalidate()`, which recomputes the dependency vector, `isDeepStrictEqual`, then `structuredClone(current)` and throws the clone away — only the throw side-effect is used. ~34×N discarded clones per fleet pass. Splitting the guard-only check from a `resolve()` that clones would avoid it. Efficiency only, per-Host path.

## What checked out clean

- **Readiness gate is fail-closed.** `isReadinessInventoryAuthoritative()` requires all three cache lanes synced + certified fence + revision equality on both axes; every default and error path fails closed. Sentinel `-1` is correct (gate stays closed until the first certificate, no overflow), and `recordNetworkPolicySafetyCertificate` re-checks `desiredRevision === certificate.revision` synchronously before persisting, so a change landing mid-pass makes the record a no-op.
- **Comparators can't silently drop a real change.** `sameMcpServerDesiredRevision`/`sameContextDesiredRevision` hash the full spec (+labels+annotations, excluding only HCC's own `network-ready` ack), so they over-detect. A lost delete bumps no revision but is caught by `hasCertifiedSafetyInventory()` going false.
- **Backoff/jitter is correct.** Full jitter with a 500ms floor, cap `min(base*2^(n-1), 30000)` with no off-by-one, reset-on-success, `Retry-After` parsed to ms and clamped. Anti-busy-loop floor demotes a rapid re-close to the paced timer; no timer leaks.
- **Contract is consistent both rollout directions.** `types.ts` adds `authoritative?: boolean` mirroring mcp-host's optional field; the mcp-host fixture is derived from the real `reconciler.getStatus` producer; `/ready` in-life 503 pulls the pod from endpoints but never restarts it (liveness is `/health`), and rpc-proxy/api-gateway/mcp-host all handle it. The "zero broken consumers" and the `24bc09da..HEAD` empty-src-diff claims both verified true.
- **Suites/types:** full HCC suite 1543/1543, `tsc --noEmit` clean, mcp-host 54/54, rpc-proxy 3/3, contract gates green.

alfredolopez80:
Thanks @zach88 — thorough pass. I adversarially re-validated all six findings against the tree; **all six survived** (nothing refuted), and R1-B1's blast radius is actually a notch worse than described. Resolved in `5c9be2ec`:

- **🔴 R1-B1 — resolved.** The bounded-timeout wrapper now tags its error `Object.assign(new Error(…), { code: 'ETIMEDOUT' })`, so `classifyDnsError` puts it in the transient/freeze branch. Reproduced both ways you noted; also confirmed the worse case — with the accumulated set pruned to 0 the reconcile doesn't just drop, it **deletes** the egress policy (`[NetPol] Deleted external egress policy "ext-egress-frozen-mcp-api.example.com-443"` in the RED run). `networkPolicyReconciler.ts`.
- **🟡 R1-M1 — resolved.** New spec `networkPolicyReconciler.egressFailStatic.test.ts` seeds an existing policy with a lapsed STATE window, stalls `resolve4`, advances fake timers, and asserts the reconcile resolves (no throw), does **not** delete, and every write still contains the accumulated `1.2.3.4/32`. Verified **RED against the pre-fix tree, GREEN with the fix**.
- **🟡 R1-M2 — resolved.** Exported `hostFleetRequestCovers` / `mergeHostFleetRequests` and added `hostFleetScheduler.properties.test.ts` — exhaustive enumeration over the 24-shape domain (incl. the new `inputRevision`): covers-reflexivity, merge idempotency / commutativity / associativity / no-drop. No `fast-check` dependency (domain is small enough to enumerate totally, incl. the 24³ associativity triples).
- **🟢 R1-L1 — resolved.** `k8sClient.test.ts` now pulls the real `sameContextDesiredRevision` via `importActual` instead of the divergent hand-rolled stand-in; full suite stays 257/257.
- **🟢 R1-L2 — resolved.** Dropped the two dead `networkPolicyRevocation{Context,Server}Generation` writes.
- **🟢 R1-L3 — deferred to #411.** It's pure efficiency (the clone is discarded, not misused) and the fix is a structural split of the safety-critical Host mutation admission (tracing every clone consumer across ~39 boundaries). Kept out of this PR to avoid regression risk in the safety path, per your "optional / non-blocking" call.

Validation: full HCC suite **1550/1550**, `tsc --noEmit` clean.

alfredolopez80:
@claude — please do a **full closing review** of this PR: confirm zach88's review is genuinely resolved, sweep for anything superior/outstanding across the whole lineage, and give a final merge-readiness verdict.

### 1. Confirm zach88's findings are closed (commit `5c9be2ec`)
zach88 raised 1 Blocker + 2 Medium + 3 Low ([review](https://github.com/evenfire-ai/evenfire/pull/382#issuecomment-5332222703)); all six were adversarially re-validated as real, and five are fixed in-PR:

- **🔴 R1-B1 (Blocker) — D4 fail-static.** The bounded-timeout DNS wrapper threw a codeless `Error`, which `classifyDnsError` maps to `'permanent'`, so a sustained resolver stall pruned the accumulated egress /32s to zero and the reconcile then **deleted** the egress policy. Fixed by tagging the timeout error `{ code: 'ETIMEDOUT' }` (in `TRANSIENT_DNS_CODES`) so it freezes. Please verify the fix lands in the freeze branch and that the new regression spec (`networkPolicyReconciler.egressFailStatic.test.ts`) genuinely fails against the pre-fix tree (seeds a lapsed accumulated policy, stalls `resolve4`, asserts no-delete + the /32 is retained).
- **🟡 R1-M1** → the regression test above (verified RED pre-fix, GREEN post-fix).
- **🟡 R1-M2** → exported `hostFleetRequestCovers`/`mergeHostFleetRequests` + exhaustive-enumeration property spec (`hostFleetScheduler.properties.test.ts`): reflexivity, idempotency, commutativity, associativity, no-drop over the 24-shape domain incl. the new `inputRevision`.
- **🟢 R1-L1** → real `sameContextDesiredRevision` via `importActual`. **🟢 R1-L2** → dropped the dead `…Generation` writes.
- **🟢 R1-L3** → **deferred to #411** (pure efficiency; the fix is a structural split of the safety-critical Host mutation admission, kept out of this PR to avoid regression risk in the safety path — zach88 marked it optional/non-blocking).

Full HCC suite **1550/1550**, `tsc` clean.

### 2. Sweep for anything superior/outstanding
Please check the whole lineage (#205 → #380 → #382) — your own prior nits, zach88's and jozer-rami's #205 items, and the deferred set — and confirm every item is either **closed with evidence** or **tracked in a follow-up issue** (#403, #404, #405, #406, #407, #408, #411, plus #373, #383, #391–#394). Flag anything that is neither. Two High items from zach88's round-3 on #205 are tracked out-of-band (one on a separate branch, one as #408).

### 3. Merge-readiness
The cross-repo deploy precondition **evenfire-infra #42 is now MERGED** (`bfd81ed6`), so the `mode=diff` stage accepts the `rollingUpdate`→`Recreate` transition. CI here is green.

**Question:** with R1-B1 fixed and everything else closed-or-tracked, is this PR ready to merge? Call out any remaining blocker.

zach88:
Did a round-2 pass focused on the fixes landed since round 1. Short version: everything from round 1 got closed for real, and I could reproduce the important ones. The DNS fail-static blocker is genuinely fixed, its regression test fails against the pre-fix tree, and the collateral/security passes came back clean. What's left is three small test-quality nits, none of which gate the merge.

**Status: 🟡 Approve with nits**

Nothing. The round-1 blocker (`R1-B1`, DNS bounded-timeout misclassified as permanent → pruned accumulated egress) is closed in `5c9be2ec` and verified:
- The timeout reject now carries `code: 'ETIMEDOUT'`, which is in `TRANSIENT_DNS_CODES`, so `classifyDnsError` returns `transient` and the accumulated /32s freeze instead of getting pruned (`networkPolicyReconciler.ts:1746-1752` → `:2149`). Fail-static direction is correct.
- Class-check holds: the timeout is the *only* codeless throw on that path — real `dns.resolve4` rejections always carry a code, `ENOTFOUND`/NXDOMAIN still classifies permanent and still expires. No sibling left unfixed.
- **Reproduced:** reverting just the `code` tag turns the egress test red (`outcome: 'rejected'`, policy deleted); with the fix it's green. The new `networkPolicyReconciler.egressFailStatic.test.ts` seeds a lapsed accumulated policy from the *real* producer output and asserts the observable result (policy survives, no delete) — non-vacuous, fails against the pre-fix head.

Round-1 M1/M2/L1/L2 are all closed too, and the collateral pass found **0 regressions** from the two fix commits.

## Optional (non-blocking)

Merge doesn't need these — they're test-coverage follow-ups, author's call.

1. **[R2-L1] Property test uses exhaustive enumeration instead of `fast-check`** — `hostFleetScheduler.properties.test.ts`. The suite that closes `R1-M2` enumerates a finite 24-shape domain rather than using `fast-check`, which is the tool T2 names. It's defensible (finite, deterministic, `max`/equality are scale-invariant so small values exercise every branch), just flagging the deviation. Optional port to `fast-check` with generators over the request shape.
2. **[R2-L2] `hostFleetRequestCovers` only covered by reflexivity** — `hostFleetScheduler.properties.test.ts`. Its directional semantics (full covers lifecycle, rejection on differing revision, generation matching) aren't exercised. Note: the cross property "merge(a,b) covers a and b" is *not* an invariant here (merge scales to `max(revision)` and stops covering the lower-revision input, by coalescing design), so that one shouldn't be added. The `merge` coverage itself is solid.
3. **[R1-L3] `revalidateHostMutationBoundary` discards ~34×N `structuredClone` per fleet pass** — `hostReconciler.ts:3850-4110`. Efficiency only, per-Host path. No fix commit touched this file, so it's still open from round 1. Fix would split `makeHostMutationAdmission` into a guard-only (throw) path and a `resolve()` path that clones.

## What checked out clean

- Full HCC suite green: **65 files / 1550 tests**, `tsc -p tsconfig.build.json --noEmit` clean.
- **Security: no fail-open.** The freeze/prune/revoke engine (`packages/network-policy-core/index.cjs`) is untouched by the fixes; the change only tags the synthetic timeout error. Frozen entries are public CIDR-gated IPs that still get re-validated against private/metadata ranges before rendering, and removal from spec still revokes (`map.delete`). The "stale public IP held during a sustained DNS outage" property is the documented fail-static design (issue #299 / deferred c3), pre-existing for every other transient code and not worsened here.
- **Collateral: 0 regressions.** The `ETIMEDOUT` tag doesn't break the single caller (message unchanged, only `.code` added); the `hostFleetScheduler.ts` change is pure `export` + comments, no logic change; the "non-vacuous" test retouch *strengthens* assertions.
- Round-1 fixes verified individually: `R1-L1` now `importActual`s the real `sameContextDesiredRevision` comparator; `R1-L2` deletes the two phantom-field writes (confirmed absent from `src/`).

alfredolopez80:
Thanks @zach88 — round-2 read matches mine (independently re-validated: R1-B1 still RED against the pre-fix tree after the dev sync, full HCC suite 1550 green, tsc clean). Closed the two test-quality nits worth folding in, in `8349054c`:

- **R2-L2 — resolved.** `hostFleetRequestCovers` was only exercised by reflexivity. Added a property that asserts `covers(a,b)` equals the spec formula for all **576 ordered pairs** of the 24-shape domain (60 cover / 516 reject), so mode-rejection (lifecycle ⊉ full), differing-revision rejection, and generation matching are all exercised in both directions. Mutation-verified: forcing `modeCovered = true` in the impl turns just this property RED while the others stay green. I respected your guardrail — the `merge(a,b) covers a and b` property is **not** asserted (merge scales to `max(revision)` and stops covering the lower-revision input by coalescing design), and I noted that in the test.
- **R2-L1 — kept exhaustive, now documented as deliberate.** `fast-check` isn't a dependency of `host-context-controller` (it's a devDep of 5 other packages but not this one), and total enumeration over the finite 24-shape domain is a strict superset of any random sample — every branch is scale-invariant (max / equality). Porting would add a dependency and *reduce* coverage, so I left the enumeration and expanded the header comment to call out the deliberate deviation from the T2 tooling note.
- **R1-L3 — still deferred to #411.** Efficiency-only (the clone is discarded), and the fix is a structural split of the safety-critical Host mutation-admission path — out of scope for this PR, as you noted. No fix commit touched `hostReconciler.ts`.

Full HCC suite now **1551** (added the one property), tsc clean.

zach88 (approved):

PR #387, opened 2026-08-18
fix(control-ui): stop the edit page deleting spec.email from a channel

### LINKED ISSUE #386: Control UI edit page silently deletes spec.email from a CommunicationChannel
## Summary

Opening a `CommunicationChannel` that has `spec.email` in the Control UI edit page and clicking **Save** — changing nothing — deletes `spec.email` from the CR. Email polling stops permanently and silently. Nothing warns, nothing rejects it, and the diff is invisible to the operator because they did not touch email.

Found while reviewing #369; unrelated to that PR's changes.

## Why it happens

The edit page builds a spec that has no concept of email, and the save is a full-spec replace.

1. **The draft never reads email.** `control-ui/lib/communicationChannelEdit.ts` contains no reference to `email` at all — not in the draft type, not in the loader, not in `buildCommunicationChannelSpec` (`:198`), which emits only `telegram`, `slack` and `teams` conditionally.
2. **The edit page PUTs that spec verbatim.** `control-ui/app/communication-channels/[name]/edit/page.tsx:301` builds it, `:311` sends it.
3. **The PUT replaces `spec`.** `control-api/src/routes/admin/resources.ts:730` composes the update with `preserveCommunicationChannelCredentialsSecretRef(...)` — `credentialsSecretRef` is the only field carried over from the previous spec. Everything absent from the body is gone.
4. **Nothing blocks a removal.** `providerTransitionError` (`control-api/src/routes/admin/resources.ts:319`) computes `added` and returns early when it is empty (`:328`). It exists to stop a provider being *enabled* without credentials; removals are not its job.
5. **The consumer notices immediately.** `channel-reader/src/main.ts:471` selects channels via `c.spec.email && c.spec.email.length > 0`, so the inbox simply drops out of the poll set.

## Reproduce

```bash
kubectl apply -f charts/clerum-crds/examples/channels.yaml   # 'all-channels' has telegram + email + slack
kubectl get communicationchannel all-channels -n channels -o jsonpath='{.spec.email}'   # populated
```

Open `all-channels` in Control UI → External Channels → edit. Change nothing. Click Save.

```bash
kubectl get communicationchannel all-channels -n channels -o jsonpath='{.spec.email}'   # empty
```

`email` is a first-class CRD field (`charts/clerum-crds/crds/communicationchannel.yaml:133`) and appears twice in the shipped example, so this is a supported configuration, not a legacy shape.

## Impact

Any email-backed channel loses its inbox the first time anyone edits it for an unrelated reason — renaming a bot handle, changing access, toggling reply-on-mention. There is no error and no audit signal; the failure shows up later as "the agent stopped answering email".

The blast radius is the same for any provider the edit UI does not model. Email is the only one today.

## Suggested fix

Either:

- **Carry it through.** Load `spec.email` into the draft and re-emit it unchanged in `buildCommunicationChannelSpec`, the way `telegram`/`slack`/`teams` are handled. Smallest change, keeps the PUT semantics.
- **Stop the PUT dropping unmodelled fields.** Merge provider arrays the UI does not own, alongside the existing `credentialsSecretRef` preservation in `preserveCommunicationChannelCredentialsSecretRef`. Fixes this class of bug rather than this instance.

The second is safer if more providers are likely to land in the CRD before the edit UI models them.

A regression test should assert that a round trip through `buildCommunicationChannelSpec` preserves a populated `spec.email`, and would fail today.

jozer-rami:
Closes #386.

## The bug

Open a `CommunicationChannel` that has `spec.email` in the Control UI edit page, change something unrelated — a bot handle, access, reply-on-mention — and click Save. `spec.email` is deleted from the CR and the inbox stops being polled. No error, no rejection, and nothing visible in the part of the form the operator was looking at.

## Why

The edit page has no email controls, and the save is a full-spec PUT: `control-api` preserves only `credentialsSecretRef` from the previous spec. `buildCommunicationChannelSpec` never emitted `email`, so the field was absent from the body and therefore removed. `channel-reader` selects channels with `spec.email && spec.email.length > 0`, so the channel simply drops out of the poll set.

Nothing on the path objected. `providerTransitionError` exists to stop a provider being *enabled* without credentials and returns early when nothing was added, so a disappearing provider is outside its remit.

## The fix

`email` is carried through the draft untouched, the same way the modelled providers already are. It is emitted **only when populated**. Absent `email` and `email: []` are currently both treated as “no email polling” by `channel-reader` (`spec.email && spec.email.length > 0` / `spec.email?.length`). We still avoid inventing empty arrays: writing `email: []` onto every channel is the defect that recently made a Teams channel read as a misconfigured Telegram one in the logs.

## Stopping the recurrence

This bug exists because `email` was added to the CRD and the edit page was never updated, and nothing caught it. So the PR also adds:

```ts
export const COMMUNICATION_CHANNEL_SPEC_FIELD_HANDLING = {
  access: 'edited', email: 'carried', /* … */
} satisfies Record<CommunicationChannelSpecField, 'edited' | 'carried'>
```

Every top-level spec field must be declared `edited` (the draft models it) or `carried` (passed through untouched). Add a field to the spec type and `tsc --noEmit` fails here until someone decides which it is.

Verified by adding a hypothetical `spec.discord`:

```
lib/communicationChannelEdit.ts(233,3): error TS1360:
  Property 'discord' is missing … but required in type
  Record<… | "discord" | …, "edited" | "carried">
```

**Honest limit:** this protects the current TypeScript spec contract, which is the current CRD field set as modelled in Control UI types. It cannot detect a future CRD-only field before that type is updated. A CRD field never added to the type would still slip past — but the UI cannot read such a field at all, which surfaces differently.

## Why not the alternatives

- **Preserve all unmodelled keys generically** — rejected as disproportionate: the CRD has 10 spec fields and the page models 9, so `email` is the only gap. It would also add a second key list to keep in sync with the builder, where a mistake silently resurrects stale values.
- **Merge server-side in control-api** — rejected as incorrect. Removal works *because* omission means deletion: disabling Telegram omits `telegram` and the PUT drops it. If the API merged from the previous spec, disabling a provider through the UI would become impossible. The API cannot tell "the UI doesn't model this" from "the operator removed it".

## Testing

Three helper round-trip tests in `communicationChannelEdit.test.ts`; the two that assert preservation were verified to fail before the fix. The third asserts the fix does *not* over-reach by inventing `email: []` on a channel that never had one.

R1-M1: `EditCommunicationChannelPage` now has an observable Save regression. It loads a channel that already has `spec.email`, edits a modelled field (Telegram bot handle), clicks Save, and asserts the PUT body still carries the loaded email groups. Temporarily omitting `email` from `persistDraft` fails that test with `expected undefined to deeply equal [email groups]`.

control-ui: previous CI HEAD was **1813 passed**, 1 skipped (170 files). This commit adds the page-level Save test. Locally: helper + edit-page suites **45 passed**, `tsc --noEmit` clean, Prettier clean.


josue-tech-94 (approved):
Approved

Reviewed at `e1ca72272b6e91d504784fb3022aa584a75d235e`.

No merge-blocking correctness, contract, or security issues were found. The fix correctly preserves populated `spec.email`, avoids introducing `email: []` on channels without email, and keeps Telegram/Slack/Teams disable-by-omission behavior intact. Current CI and CodeQL are passing.

There is one non-blocking follow-up:

* **R1-M1 — Should fix / Medium:** add an observable edit-page regression test that exercises Save and asserts the actual PUT payload preserves the existing `spec.email` during an unrelated edit. The helper-level regression coverage is good, but it does not currently protect the page → request boundary where the original incident manifested.

Before merge, please also update the PR description so it reflects the verified final state:

* replace the stale test count with the current results;
* clarify that absent `email` and `email: []` are currently both treated as “no email polling” by `channel-reader`, even though we should still avoid inventing empty arrays;
* clarify that `COMMUNICATION_CHANNEL_SPEC_FIELD_HANDLING` protects the current TypeScript contract/current CRD field set, but cannot detect a future CRD-only field before the Control UI type is updated.

Everything else reviewed cleanly

jozer-rami:
R1-M1 is in `4d34c8d1`.

`EditCommunicationChannelPage` now clicks Save after an unrelated Telegram bot-handle edit and asserts the PUT body still carries the loaded `spec.email` groups. That is the page → request boundary the helper tests did not cover. Temporarily omitting `email` from `persistDraft` fails the new test (`expected undefined to deeply equal [email groups]`).

PR description updated for the current test results, the `channel-reader` absent-vs-`email: []` polling rule, and the TypeScript-contract limit on `COMMUNICATION_CHANNEL_SPEC_FIELD_HANDLING`.

PR #401, opened 2026-08-18
feat(gfs): make upload product limit runtime-configurable

### LINKED ISSUE #300: Desktop GFS upload cap is a hardcoded compile-time constant (make it runtime-configurable)

josue-tech-94:
## Summary

- Continue the Upload v2 architecture from PR #311.
- Make GFSC's product upload ceiling runtime-configurable.
- Make Desktop and Control UI consume the writer-advertised product policy.
- Preserve hard protocol, part, geometry, admission, and legacy boundaries.

## Why

PR #311 introduced the writer capability contract, but compatible clients still rejected files
above the compiled 200 MiB product default before a runtime increase could take effect. This closes
the remaining issue #300 gap: a new compatible client can observe later server-side product-policy
changes without being rebuilt.

## Behavior

- HCC owns `CONTEXT_MAPPER_GFSC_UPLOAD_PRODUCT_MAX_FILE_BYTES` and forwards
  `GFS_UPLOAD_PRODUCT_MAX_FILE_BYTES` to GFSC.
- The deprecated `GFS_UPLOAD_MAX_FILE_BYTES` alias remains accepted; both names must agree.
- The product policy defaults to 209715200 bytes and accepts canonical integers from 1 through the
  configured protocol maximum.
- The absolute compiled Upload v2 protocol maximum remains 1073741824 bytes.
- The hard part/request maximum remains 16777216 bytes and geometry remains capped at 1,024 parts.
- Invalid, empty, non-canonical, unsafe, conflicting, zero, negative, fractional, or above-protocol
  product configuration fails GFSC startup; values are never clamped.
- GFSC advertises and enforces the same effective product value for new-session admission.
- Exact-cap files and zero-byte files are admitted; cap-plus-one is rejected before allocation.
- Desktop and Control UI use 1 GiB only as their local protocol guard, then enforce the valid writer
  `maxFileBytes` value for new sessions.
- Enabled-v2 writers that omit `maxFileBytes` retain the exact 200 MiB compatibility fallback and
  emit a diagnostic warning through each client's existing logging seam.
- Malformed capability envelopes, non-Boolean enablement, and invalid advertised limits fail loudly
  and never select legacy.
- Existing admitted sessions remain resumable after the current product policy is lowered.
- An exact idempotent new-session retry returns its previously admitted session after a later policy
  decrease. Conflicting reuse still conflicts, while new idempotency keys obey the current policy.
- The separate legacy JSON/base64 path remains capped at 16 MiB.
- Opt-in runtime policy tests use an immutable branch-owned recovery marker written before mutation.
  The marker records both canonical and alias baselines, survives process loss, and is cleared only
  after exact restoration and convergence under the same UID-bound lease.

## Safety / version skew

- Existing PR-311 clients remain limited to 200 MiB for upward policy changes.
- New issue-300-compatible clients observe later writer product-policy changes without a rebuild.
- Older pre-v2 clients remain on legacy behavior.
- Genuine capability unavailability, canonical unsupported responses, HTTP 408/timeouts, and valid
  `enabled: false` retain the bounded fresh-upload legacy fallback.
- Explicit resume never silently creates a second legacy upload.
- Malformed writer contracts, authorization failures, caller aborts, and unexpected writer failures
  do not fall back to legacy.
- Neither client can expand beyond the compiled 1 GiB protocol bound.
- Existing part, part-count, concurrency, admission, timeout, disk, and free-space protections remain
  unchanged.
- Control API and External REST remain transparent policy relays; their static transport contract
  proves the 16 MiB part path without treating product policy as a request-body authority.

### Passed locally

- GFSC full suite: 605 passed, 8 skipped; build passed.
- HCC full suite: 1,686 passed; build passed.
- Desktop focused Upload v2 suite: 49 passed.
- Desktop full suite: 1,954 passed; main and renderer build passed.
- Control UI focused Upload v2 suite: 61 passed.
- Control UI full suite under `en_US`: 1,653 passed, 1 skipped; production build passed.
- Control API focused authority/gateway tests: 74 passed, 1 skipped.
- Control API full suite: 4,561 passed, 103 skipped; build passed.
- Shared runtime mutation/recovery helper suites: 20 passed.
- External REST build: passed.
- GFS Upload v2 geometry/static contract: passed.
- Static Minikube Upload v2 profile contract: passed.
- Repeated bounded collateral review: clean.
- Independent code/test-quality review: safe.
- Independent security review: safe.
- `git diff --check`: passed.

### GitHub checks

- CI, CodeQL, and Real PostgreSQL completed successfully for
  `f564506023c2f5d2db7579fd9711e81cdb34ed82`.

### Not run / tooling gaps

- Packaged Desktop, Control UI Playwright, Minikube T2, and live runtime policy-change journeys were
  not run because cluster mutation was not authorized.
- Local Real PostgreSQL was not run because no isolated DSN was available.
- GFSC lint was unavailable because its package does not install an `eslint` executable.

## Scope

This is one coordinated multi-service PR despite D1's one-service soft goal. GFSC is the
authoritative producer/enforcer, while Desktop and Control UI are the two PR-311 consumers. Splitting
them would leave an intermediate contract that either advertises unsupported increases or keeps
already-built clients pinned to the obsolete product default. Control API remains an opaque relay;
the External REST configuration edit only exposes its existing static relay bound to the bounded
cross-stack contract test and does not create product-policy authority.

## References

Closes #300

Canonical predecessor: #311


josue-tech-94:
@claude Please perform a fresh, thorough review of the entire current PR, not only the latest commits.

PR #401 continues the GFS Upload v2 architecture from PR #311 and makes the writer-owned product upload limit runtime-configurable while preserving the existing protocol and transport safety bounds.

Please review the complete diff against dev at the current PR head, with particular attention to correctness, regressions, security, version-skew behavior, and test quality.

Key invariants to verify:

GFSC remains the single authority for the runtime product upload policy.
The default product limit remains 200 MiB.
The absolute whole-file Upload v2 protocol ceiling remains 1 GiB.
The maximum Upload v2 part/request size remains 16 MiB and the geometry remains bounded.
GFSC advertises exactly the same maxFileBytes value it enforces for genuinely new sessions.
Desktop and Control UI use the 1 GiB value only as a local protocol guard, then honor the writer-advertised product limit.
Missing maxFileBytes retains the intended 200 MiB compatibility fallback.
Malformed capability values fail closed and cannot expand acceptance.
Existing admitted sessions remain resumable after a later product-policy reduction.
Idempotent create replay after policy lowering: if a session was admitted at 250 MiB under a 300 MiB policy and the response was lost, an identical retry with the same idempotency key after lowering the policy to 100 MiB must return the previously admitted session rather than reject it. A new key must still obey the new 100 MiB policy, and conflicting reuse must still conflict.
Authorization, ownership, and fingerprint checks must still occur before any existing session can be disclosed through idempotent replay.
Control API and External REST remain opaque relays rather than new product-policy authorities.
Legacy JSON/base64 upload behavior remains separately bounded at 16 MiB.

Please also scrutinize the updated cross-stack gateway contract test. A previous implementation could false-green because its regex accepted only a safe initializer prefix or a commented declaration. The current implementation now uses TypeScript parsing. Please verify that it is genuinely mutation-resistant and that:

only active declarations are considered;
the complete initializer is validated;
valid integer/multiplication expressions and numeric separators work;
unsupported +, /, identifiers, malformed expressions, or commented-only declarations fail;
an unsafe writer/client part-size drift or relay/gateway mismatch makes the contract test fail;
the test does not accidentally restore the 200 MiB product default as a transport/body-size authority.

Please review the tests skeptically as well: check that they assert observable behavior, exercise realistic producer/consumer contracts, and would fail for the defects they claim to prevent.

For security, pay particular attention to integer/range handling, resource exhaustion, protocol/part bounds, malformed capability values, idempotency authorization/ownership, information disclosure, and any unsafe process execution in runtime test helpers.

Please report findings in severity order with:

file and line;
concrete failure scenario;
impact;
recommended change;
test needed to prove the fix.

Separate real PR-introduced issues from pre-existing/out-of-scope concerns. If you find no merge-blocking issues, please say so explicitly and give an overall merge-readiness verdict.

alfredolopez80:
# PR #401 Review — Runtime-configurable GFS Upload v2 product limit

- **Repo:** evenfire-ai/evenfire
- **PR:** #401 — *feat: runtime-configurable Upload v2 product limit (issue #300)*
- **Base ← Head:** `dev` ← `pr-401-review` (HEAD `30b263ef`)
- **Author:** josué (`josue-tech-94`)
- **Scope:** 29 files, +1375 / −151 · continues PR #311, closes #300
- **Review method:** 5 parallel Fable-5 agents over a fetched PR-head worktree; every finding verified against real code; the AST contract parser and both client normalizers were *executed* against adversarial inputs.

## Findings by severity

| Sev | Count |
|---|---|
| Blocker | 0 |
| Critical | 0 |
| High | 0 |
| Medium | 0 |
| Low | 7 |
| Info / decision-for-owner | 1 |

**Overall verdict: no merge-blocking issues.** The core change is fail-closed at every parse boundary, does no clamping, keeps GFSC as the single policy authority, and the advertise/enforce pair is structurally incapable of diverging (both read `productMaxFileBytes` off the same in-process config object). The cross-stack AST contract test is genuinely mutation-resistant. Every Low below is a hardening nit, a UX degradation limited to the mixed-version window, or an opt-in-test cleanup.

---

## Low (follow-up, non-blocking)

**L1 — Canonical-syntax guard not applied to `GFS_UPLOAD_PROTOCOL_MAX_FILE_BYTES`**
`gfs-controller/src/config.ts:242-247` (via `uploadInteger`). The new `requireCanonicalUploadInteger` (`^(?:0|[1-9][0-9]*)$`) is applied only to the two product names. The protocol var still parses through the looser `^[0-9]+$`, so `"01073741824"` (leading zero) is accepted for the protocol var while the sibling product vars reject the same syntax. Range check still applies, no clamping — **consistency only**. *REPRODUCED (traced).* Fix: also call `requireCanonicalUploadInteger('GFS_UPLOAD_PROTOCOL_MAX_FILE_BYTES')`.

**L2 — `create()` pre-auth size bound uses the compiled 1 GiB constant, not the configurable `protocolMaxFileBytes`**
`gfs-controller/src/upload/uploadSession.ts:540-546`. If an operator lowers `GFS_UPLOAD_PROTOCOL_MAX_FILE_BYTES` (floor 200 MiB), a request claiming e.g. 900 MiB still passes the early check, computes a fingerprint, and reaches the idempotency lookup. **New sessions are still rejected** (product check is capped at `protocolMaxFileBytes`); the only reachable effect is replay of a previously-admitted session above the lowered value plus one indexed SELECT. This is consistent with the deliberate replay-across-policy-change semantics, but it makes the protocol env effectively cosmetic for replays. *READ-ONLY.* Fix (optional): `Math.min(GFS_UPLOAD_V2_PROTOCOL_MAX_BYTES, config.protocolMaxFileBytes)` in the early check, or document that the env bounds only new admissions.

**L3 — AST extractor duplicate-key bypass (defense-in-depth)**
`control-api/test/network.gateway.intent.test.ts:146-180` (`boundedEnvBytesFromSource`). A second property with the same key but a different env/value in the same object literal keeps the extractor GREEN (it matches by expected env name; `parseDiagnostics` does not flag duplicate keys) while runtime last-key-wins would use the second value. In practice `tsc` (TS1117) rejects duplicate keys in CI, so this is **defense-in-depth only**. *REPRODUCED.* Fix: count property assignments by name alone and require exactly one.

**L4 — E2E DB fixture seeded outside the `try` → leak on early throw**
`control-ui/e2e/gfs-upload-v2.spec.ts` and `desktop-app/test/e2e-playwright/gfs-upload-v2.test.ts` (runtime-policy describes). `seedGfsDirectoryFixture` / `seedGfsGrant` run before the `try`; if `createOversizedDiskUploadFixture` throws, the seeded DB fixture leaks because cleanup lives in `finally`. Narrow window, opt-in suite only. *READ-ONLY.* Fix: move seeding inside the `try` or widen the `finally`.

**L5 — Runtime e2e save/restore cannot represent the "unset" baseline**
`tests/e2e/gfsUploadV2Runtime.ts:369-377`. On a default cluster (env absent → GFSC compiled 200 MiB default) the opt-in e2e throws before mutating (fail-closed, no corruption), and `setGfsUploadProductMaxBytes` can only set, never unset — so "restore previous" cannot return a host to the unset baseline. *REPRODUCED (traced).* Fix: return a sentinel for "absent" and let restore use `kubectl set env NAME-`.

**L6 — Desktop renderer guard widening degrades error quality in the mixed-version window**
`desktop-app/ui/src/pages/FilesPage.tsx:437-457`. Widening the renderer guard 200 MiB → 1 GiB also widens what reaches the pre-existing one-shot compatibility branch (old main/preload pair). A 500 MiB file the old renderer rejected instantly now travels to the old main process and fails there with that API's own ceiling error. Fail-loud is preserved (the one-shot API keeps its own ceiling); only error *quality* degrades, only during a mixed-version deploy. *READ-ONLY.* Acceptable trade — no fix required.

**L7 — Static upload-dialog guidance text (pre-selection UX)**
`control-ui/components/GfsBrowser.tsx:1148`. The dialog copy is now static ("…up to the 1 GiB protocol maximum…"); the operator no longer sees the concrete product limit before picking a file — they discover it via the rejection message (which does interpolate the real byte ceiling). Correct behavior, mild pre-selection UX regression vs the old concrete "200 MiB" text; fixing it needs a capabilities fetch at dialog-open. *READ-ONLY.* Out of this PR's minimal scope.

---

## Info / decision for the owner

**I1 — `maxFileBytes === undefined` → 200 MiB compatibility default on both clients**
`control-ui/lib/gfsFileUpload.ts` and `desktop-app/src/gfs/upload.ts` (`normalizeUploadProductMaxBytes`, introduced by this PR). Independently surfaced by 3 of the 5 agents. This is a default-value fallback, so it is flagged per the repo's no-unrequested-fallbacks policy — **but it is not silent**: it throws a distinct, labeled error ("200 MiB compatibility limit because the writer omitted `maxFileBytes`") and both clients ship dedicated tests for it. It is the intentional wire-compatibility path for enabled-v2 writers that omit the field (older writers enforce 200 MiB themselves; current writers always advertise the field). Net effect is *more* restrictive than trusting the writer, and the server stays authoritative. **No change recommended** unless you prefer to treat an omitted `maxFileBytes` on an enabled writer as a hard `GfsUploadCapabilityError` — that is a design call for the owner.

---

## Verdict on the open `@claude` review comment
The prior automated `@claude` review (issue comment by `josue-tech-94`'s trigger) marked every invariant ✅ and found no blockers. **This multi-agent review independently corroborates that conclusion** — all nine invariants (single authority, 200 MiB default, 1 GiB protocol ceiling, 16 MiB / 1024-part geometry, advertise==enforce, canonical-never-clamped parsing, alias agreement, idempotent replay ordering with authorize-before-lookup, exact-cap/zero-byte boundaries) hold against the real code.

## What came out clean (verified, not assumed)
- **Single authority / advertise==enforce:** capability output (`uploadSession.ts:1903`) and admission (`:583`) read the same `config.productMaxFileBytes`; HCC forwards raw env strings only (no parse/default), undefined entries are dropped, empty string → GFSC startup throw. Structurally cannot diverge.
- **Integer/range handling:** strict canonical regex + `Number.isSafeInteger` + inclusive `[1, protocolMax]`, throwing at startup; rejects `1e6`, `0x10`, `100.0`, `01`, `+1`, whitespace, `0`, `-1`, `2^53`, `1 GiB+1`. No clamping anywhere; no `parseInt`/radix/BigInt-mix issues.
- **Idempotency & authorization:** `authorize?.()` runs before the DB lookup; lookup and the 23505 replay path are both scoped `owner_subject+drive+idempotency_key` and re-check `request_fingerprint`. A unit test asserts zero idempotency lookups for an unauthorized caller. No cross-owner enumeration or info disclosure.
- **Resource exhaustion:** size validated before INSERT / blob reservation / part streaming; cap+1 rejected before allocation; geometry ≤1024 parts; a malicious writer advertising a huge `maxFileBytes` makes clients throw, not allocate.
- **Supply chain:** **no dependency added** — `control-api/package.json`+lock changed only a version field (0.1.470→0.1.472); the contract test uses the pre-existing `typescript` devDependency. No postinstall scripts.
- **Unsafe process execution (author's flagged concern):** does not materialize. `tests/e2e/gfsUploadV2Runtime.ts` uses `execFile`/`execFileSync` with argv arrays only (no shell, no `exec`, no `shell:true`); the interpolated limit is numerically validated first; SQL goes through single-argv `kubectl exec -- psql` with `sqlLiteral` quoting + `::uuid` casts; strong fail-closed ownership/branch-context/dirty-worktree guards before any mutation; fixtures use `mkdtemp` + sparse `truncate`.
- **Cross-stack AST contract test:** mutation-resistant — commented/`let`/`var`/dead-branch decls are invisible; `+`, `/`, identifiers, hex, float, exponent, malformed tails all throw; underscore separators accepted (matches real source). Writer/client part-size drift, relay/gateway mismatch, and restoring 200 MiB as a transport/body authority each turn it RED (the last one twice, independently). Zero-tests guard present (`routes.length ≥ 5`).
- **Docs:** `docs/deploy/minikube.md` and `gfs-permission-store.md` match the new env-var names, 200 MiB default, 1-byte–1-GiB range, alias-must-match rule, and new-sessions-only enforcement. No doc lies.

## Pre-existing / out of scope (noted, not PR defects)
- `control-api/src/config.ts:243` — stale comment "the 200 MiB product/file limit" (now a default, not a fixed limit); relay is intentionally opaque, so untouched here.
- `uploadSession.ts:561` `authorize?.()` and `serve.ts:365` `uploadCapabilities?.() ?? {disabled}` optional seams — both fail-closed; production `index.ts` always wires them. Blame: Alfredo Lopez, PR #311 lineage.
- `routeIsAllowlisted` `catch { return false }` (`network.gateway.intent.test.ts:367-374`) — fail-closed. `GKE_INFRA_PRESENT` conditional skip — documented environmental gating.
- `fallbackConcurrency` `Number(x) || DEFAULT` coercion in the clients — pre-existing, re-sanitized downstream.

## Test-execution honesty
- The full-suite green numbers in the PR body (GFSC 594/8-skip, HCC 1085, Desktop 1936, Control UI 1631) are **author/@claude-reported**; this review did not re-run them (no isolated Postgres / cluster mutation not authorized).
- This review **did independently execute**: the AST contract parser against 26 adversarial inputs (22/22 static cases behaved exactly as claimed) and the client `normalizeUploadProductMaxBytes` pure-function against undefined/null/0/negative/NaN/Infinity/string/over-cap.
- No always-on e2e covers the live runtime-policy-change journey (opt-in `GFS_UPLOAD_V2_RUNTIME_LIMIT_E2E=1`); unit + contract layers compensate.

## Recommendation
**Merge-ready.** No blockers. Optional follow-ups: L1 (apply canonical guard to the protocol var) and L3 (dedup-key hardening) are the two worth a quick cleanup; L4/L5 are opt-in-test polish; L2/L6/L7 are documented, acceptable trade-offs. I1 (200 MiB compatibility default) is an intentional, tested design decision — confirm you're happy with it as-is.

alfredolopez80 (approved):
@josue-tech-94  indicate the all Low issue in the review will handle like follow ups after in another PR`s

approved

josue-tech-94:
@claude Please do one final fresh review of the entire current PR #401 at the latest head, not only the most recent commits.

Since your previous review, we addressed your two non-blocking observations and the selected follow-ups from Alfredo’s review, including:

legacy fallback now applies the 16 MiB client-side limit only after legacy fallback is actually selected, without restoring a global 16 MiB/200 MiB preflight;
compatibility-limit messages are derived from the actual default product-limit constant rather than duplicating "200 MiB";
GFS_UPLOAD_PROTOCOL_MAX_FILE_BYTES now uses the same canonical-integer syntax discipline as the product-limit settings;
the AST-based gateway contract extraction was hardened against duplicate authoritative property declarations;
runtime-policy E2E fixture cleanup now covers partial setup failures;
runtime product-limit save/restore now correctly preserves an originally unset environment variable instead of converting it to an explicit value.

Please review the complete diff against dev and verify that these follow-up fixes did not regress the core PR contract.

In particular, please re-check:

GFSC remains the sole runtime product-policy authority;
200 MiB remains only the default/compatibility product limit, not a transport authority;
1 GiB remains the absolute Upload v2 protocol ceiling;
16 MiB remains the part/request and legacy bound;
Desktop and Control UI still honor writer-advertised runtime limits and remain behaviorally aligned;
valid Upload v2 uploads above 16 MiB are not accidentally rejected by the legacy fallback guard;
disabled/unavailable-v2 fresh uploads fail locally at the legacy bound before an unnecessary legacy request;
explicit resume never silently falls back to legacy;
idempotent create replay after a later policy reduction still returns the previously admitted session without weakening authorization, ownership, or fingerprint-conflict checks;
the gateway AST contract remains mutation-resistant and cannot false-green on duplicate declarations, comments, initializer prefixes, or relay/writer/client drift;
runtime test helpers restore both explicit and originally-unset configuration safely and retain all ownership/non-production guards;
cleanup paths cannot leak fixtures after partial setup failure.

Please also review test quality and security implications, not just whether CI is green.

Report any findings with severity, file/line, concrete failure scenario, impact, recommended change, and required regression test. Separate PR-introduced issues from pre-existing/out-of-scope concerns.

If there are no remaining merge-blocking issues, please say so explicitly and give a final merge-readiness verdict.

alfredolopez80:
# PR #401 — Closure audit at head `ebcc68b9`

- **Base ← Head:** `dev` ← `feat/issue-300-runtime-upload-product-limit` (`ebcc68b9`)
- **Scope:** 35 files, +1933 / −251 · continues #311, closes #300
- **CI at head:** 33/33 green (including `Real Postgres integration (Postgres 16)`)
- **Method:** 6 parallel agents over a worktree at the PR head. Mutation testing with exit codes captured without pipes; pure functions and the AST extractor executed against adversarial inputs, not just read.
- **Goal of this pass:** verify that every finding from the earlier reviews was *actually* fixed, that the deferred follow-ups exist, and that no later commit reintroduced a resolved problem.

## Verdict

**No Blocker, no Critical. Zero regressions.** The PR is functionally correct and nothing was reintroduced.

Three new **High** findings, none of which affect the happy path — they are about fail-loud behaviour and the reliability of the guards themselves. One previously-closed finding turns out to be **only partially closed**.

| Severity | Count |
|---|---|
| Blocker | 0 |
| Critical | 0 |
| High | 3 |
| Medium | 8 |
| Low | ~12 |

---

## 1. Status of the earlier findings

| # | Finding | Commit | Verdict |
|---|---|---|---|
| A-OBS1 | Preflight widened 200 MiB→1 GiB, later feedback on legacy fallback | `bea46417` | **RESOLVED — but the original premise was inaccurate; the "fix" is tests only** |
| A-OBS2 | `"200 MiB"` hardcoded in message | `d4ff8728` | **RESOLVED** (exhaustive grep clean) |
| A-L1 | Canonical guard missing on `GFS_UPLOAD_PROTOCOL_MAX_FILE_BYTES` | `e28aa2bb` | **RESOLVED** (3/3 vars; mutation → RED) |
| A-L3 | AST extractor green on duplicate key | `9ce13db9` | **PARTIAL — see H2** |
| A-L4 | DB fixture seeded outside the `try` | `6f335e6e` | **RESOLVED** (10 sites, both files) — the only fix **without** mutation verification (needs a live cluster) |
| A-L5 | "Absent" baseline not representable | `c9d008b4` | **RESOLVED** — but its tests **run in no lane** (see M3) |

Deferred with owner approval — **all four confirmed unchanged, none made worse**: L2 (compiled pre-auth guard), L6 (renderer guard widening), L7 (static dialog copy), I1 (200 MiB compatibility default — but see M1).

### Note on A-OBS1

The original observation stated that with Upload v2 disabled an oversized file "clears the client preflight and is rejected by the server". That is not accurate: the 16 MiB guard **already existed on `dev`** and rejects **locally** — `control-ui/lib/gfsFileUpload.ts:1475` (before `file.arrayBuffer()` and before `apiSend`) and `desktop-app/src/appService.ts:351` (`stat()` before reading, before `createResource`). Conclusive proof: `git diff origin/dev...HEAD -- desktop-app/src/appService.ts` is **empty**.

`bea46417` touching only tests is therefore the *correct* outcome — there was nothing to fix, and pinning the behaviour is valuable. The only real degradation is one extra `GET /capabilities` before the rejection. Flagging it because the follow-up comment described it as a behavioural change, and the verification chain closed over a premise that did not hold.

---

## 2. No regressions — including the merge nobody reviewed

The last automated review ran against `c9d008b4`. The current head is `ebcc68b9`, a merge of `origin/dev` bringing **140 commits**, created *after* every review. Audited explicitly:

Intersection with the PR's 35 files: **8 files**, 7 of them noise (`package.json`/`package-lock` ×3, one doc). The single substantive one, `control-api/test/network.gateway.intent.test.ts`, had a real conflict, and **the resolution is strictly better than either side**:

| Version | Regex |
|---|---|
| merge-base `1b845a3d:358` | `value:\s*"?(\d+)"?` (unbalanced quotes) |
| PR head `c9d008b4:604` | identical to base — the PR never touched that line |
| dev `3b9ff0f4:358` | `value:\s*(["']?)(\d+)\1` (hardened) |
| **merge `ebcc68b9:604`** | **dev's**, with the capture index correctly reindexed to `![2]` |

No hunk from the PR side was dropped (the combined diff emits no body for that file → every hunk matches at least one parent).

**One silent risk worth recording:** `control-api/package.json` resolved to dev's version. The AST contract test does `import ts from 'typescript'`, added by `fa51e287`. Had that resolution dropped the entry, the whole contract would stop compiling. `"typescript": "^5.9.2"` **survives** — verified by *running* the test, not by reading the lockfile.

No fix undid another (`572010d6`→`fa51e287`→`9ce13db9` are cumulative; `d39f2eaa` and `e28aa2bb` coexist at `config.ts:248-250`).

**The critical ordering holds** — independently confirmed twice:

```
uploadSession.ts:561-565   await this.deps.authorize?.(...)
uploadSession.ts:570-573   SELECT ... WHERE owner_subject/drive/idempotency_key
uploadSession.ts:583-588   sizeBytes > productMaxFileBytes
```

`65997d8e` moved **only** the product check. No information-disclosure issue. Credit where due: `uploadSession.test.ts:767-786` instruments `db.idempotencyLookupCount` and asserts it stays flat on an unauthorized attempt — that mechanically pins authorize-before-lookup and would go red if the ordering regressed. That is a real guard, not decoration.

**Mutation verification: 6 of 7 fixes go RED** (`e28aa2bb`, `9ce13db9`, `c9d008b4`, `d4ff8728`, `bea46417`, `65997d8e`). No vacuous fix. The exception is `6f335e6e` (r3-l3), which needs a live Playwright cluster.

---

## 3. New High findings

### H1 — Malformed writer capability silently degrades to the legacy path with a misleading message

`control-ui/lib/gfsFileUpload.ts:190-202` + `components/GfsBrowser.tsx:394-402,498-511`
`desktop-app/src/gfs/upload.ts:139-151` + `src/appService.ts:2227-2237,2261-2271` · **REPRODUCED**

This is a semantic collision from reusing an error class. `normalizeUploadProductMaxBytes` throws `GfsUploadCapabilityError` for an invalid ceiling — but that class has meant *"the writer does not support Upload v2"* since #311, and it is exactly the predicate both callers use to switch to legacy.

```
normalizeUploadProductMaxBytes(2_147_483_648)   // writer advertises 2 GiB, above protocol
  └─> throw new GfsUploadCapabilityError('...advertised an invalid file ceiling.')
GfsBrowser.tsx:394  if (!(error instanceof GfsUploadCapabilityError)) throw error
        │                                    ↑ IS a CapabilityError → does not rethrow
        └─> showToast('Resumable upload is unavailable; using the legacy 16 MiB path.')
            └─> uploadGfsFileLegacy(input)    // ≤16 MiB → THE UPLOAD SUCCEEDS
```

**Scenario:** writer advertises `maxFileBytes: 2147483648`, or `"500"` as a string. A 300 MiB file produces a toast saying *"Resumable upload is unavailable"* — which is false, v2 *is* available; what is broken is the advertised value — followed by *"legacy GFS is limited to 16777216 bytes"*. A file ≤16 MiB **uploads successfully** and the writer's contract violation is buried entirely.

**Impact:** the root cause never reaches the operator. This is silent degradation rather than fail-loud. It does **not** widen acceptance (legacy is stricter), which is why it is High and not Critical.

**Suggested fix:** an invalid ceiling is not "capability absent" but "capability corrupt". Introduce a distinct class (e.g. `GfsUploadContractError`) that callers do **not** treat as a legacy switch, or explicitly exclude the ceiling error from the fallback predicate.

**Regression test needed:** prove that a malformed `maxFileBytes` originating in `normalize` never reaches `uploadGfsFileLegacy`. The current `appService` test constructs the `DesktopUploadCapabilityError` by hand rather than originating it in `normalize`, so it does not cover this path.

### H2 — The duplicate-key fix (A-L3) is blind to computed keys, spread and shorthand

`control-api/test/network.gateway.intent.test.ts:155-158` · **REPRODUCED**

`boundedEnvBytesFromSource` counts only `ts.isIdentifier` and `ts.isStringLiteral`:

| duplicate form | extractor |
|---|---|
| `gfsUploadMaxPartBytes: ...` | REJECT ✓ |
| `'gfsUploadMaxPartBytes': ...` | REJECT ✓ |
| **`['gfsUploadMaxPartBytes']: ...`** | **ACCEPT ✗** |
| **`...override`** | **ACCEPT ✗** |
| **`gfsUploadMaxPartBytes,`** | **ACCEPT ✗** |

Executed against the real object evaluated at runtime:

```
runtime effective value : 209715200 (200 MiB)
extractor reports       : {"fallback":16777216,"ceiling":16777216}
```

Last property wins in JS, so the relay would run at 200 MiB — well above the writer's 16 MiB part maximum — and **the contract test stays GREEN**. That is precisely the drift this test exists to prevent. The PR's own test at `:435` only covers the identifier form.

Worth stating plainly, because it is a methodological point rather than a criticism: static inspection of this fix reads as complete — the `properties.length !== 1 → throw` guard is right there. What the inspection cannot show is that those three syntactic forms never enter `properties` in the first place, so `length` is 1 and the guard never fires. Only executing it surfaced that.

**Suggested fix:** also count `ShorthandPropertyAssignment`; treat any `SpreadAssignment` in the same object literal as unresolvable (throw); resolve `ComputedPropertyName` whose expression is a string literal.

Related, lower probability: `staticBytesFromSource` (`:121-140`) accepts a `const` shadowed by `export { OTHER as SYMBOL }` — reports 16 MiB while the exported binding is 200 MiB. **REPRODUCED.**

### H3 — Runtime restore can fail and leave the cluster pinned; baseline poisoning

`tests/e2e/gfsUploadV2Runtime.ts:435` + `control-ui/e2e/gfs-upload-v2.spec.ts:441` + `desktop-app/test/e2e-playwright/gfs-upload-v2.test.ts:496`

The restoring `finally` calls `setGfsUploadProductMaxBytes(previous)`, which re-runs the full `assertOwnedRuntimeContext()` — including `git status --porcelain` (`:160-168`) and a live `kubectl get configmap` (`:174-182`). If the worktree becomes dirty or the apiserver blips between the mutation and the `finally`, **the restore throws and the cluster stays at 300 MiB**, with no compensating action.

Worse: a SIGKILL between `setGfsUploadProductMaxBytes(lowerMax)` and the `finally` leaves 300 MiB behind, and the **next** run adopts 300 MiB as its `previous` — baseline poisoning that neither `worktreeId` nor `gitHead` detects. There is no in-flight mutation sentinel in the cluster; the sentinels added by A-L4 are local variables.

---

## 4. Selected Medium findings

**M1 — The 200 MiB default is applied with no signal at all on the accept path.** `control-ui/lib/gfsFileUpload.ts:191`, `desktop-app/src/gfs/upload.ts:140`. **REPRODUCED.** The justification that closed I1 — *"not silent: it throws a labelled error"* — covers only the **reject** branch. Writer omits `maxFileBytes` + a 50 MiB file → `normalize(undefined)` returns 200 MiB and the upload proceeds with **no log, no toast, no telemetry**. The fallback is silent in the majority case and only becomes visible when it also rejects. Neither client has a test covering the accept branch.

**M2 — The 200 MiB default is pinned by no test. Confirmed by mutation.** `control-ui/lib/__tests__/gfsFileUpload.test.ts:205-207` (identical in `desktop-app/src/gfs/upload.test.ts`) derives its expectation by calling the SUT:

```ts
const compatibilityMaxFileBytes = normalizeUploadProductMaxBytes(undefined)
const compatibilityLimit = `${compatibilityMaxFileBytes / (1024 * 1024)} MiB`
Object.defineProperty(file, 'size', { value: compatibilityMaxFileBytes + 1 })
```

Expected string, file size and production message all trace to the same constant, so they shift together. Changing `200 * 1024 * 1024 → 512 * 1024 * 1024` leaves the suite **green: 44/44, EXIT=0**. No other control-ui test references it, and the AST contract test pins only protocol and part mirrors. One-line fix: `expect(GFS_FILE_UPLOAD_DEFAULT_PRODUCT_MAX_BYTES).toBe(209_715_200)`.

**M3 — The five new unit tests run in no lane.** `tests/e2e/gfsUploadV2Runtime.test.ts` (4) and `.read.test.ts` (1) are not in `DEFAULT_VITEST_SUITES` (`scripts/e2e/run-vitest-e2e.sh:13-42`, 21 explicit suites) and no other lane points at them. Consequently `buildGfsUploadProductMaxEnvArgs` (the entire `absent`/`NAME-` fix), `parseRenderedGfsUploadProductMax` and `parseGfsUploadProductMaxState` have **zero always-on coverage**. The runner itself is correctly fail-loud (`passWithNoTests: false`, `Tests N>0` floor, `PIPESTATUS`) — but no guard saves a file nobody lists.

**M4 — `assertUploadV2TransportBounds` is vacuous on empty arrays.** `:205-214`. `[].some() === false`, so `protocolMirrors: [], partMirrors: [], relayBounds: []` passes GREEN. This repo already has that guard **in the same suite** (`:1258 expect(routes.length).toBeGreaterThanOrEqual(5)`); the new helper did not inherit it.

**M5 — One env is read, two are written.** `gfsUploadV2Runtime.ts:412-424` vs `:450-457`: the read covers only `GFS_UPLOAD_PRODUCT_MAX_FILE_BYTES`, while the write/unset also touches the `CONTEXT_MAPPER_GFSC_UPLOAD_MAX_FILE_BYTES` alias whose baseline is never read. Safe on the standard overlay (both set to the same value), but it is coupling without an assertion.

**M6 — Desktop does not bound `partBytes` on resumed sessions; Control UI does.** `desktop-app/src/gfs/upload.ts:1438` vs `control-ui/lib/gfsFileUpload.ts:1276-1281`. Pre-existing, but it is a contract divergence between the two clients.

---

## 5. Test numbers in the PR body

Suites were executed for real (exit codes captured without pipes, `total > 0` verified):

| Suite | EXIT | Actual | Declared |
|---|---|---|---|
| gfs-controller | 0 | 605 passed / 8 skipped | ✅ exact |
| desktop-app (after `verify:electron`, Node 24) | 0 | 1937 passed | ✅ exact |
| control-api gateway contract | 0 | 57 passed / 1 skipped | ✅ exact |
| control-api full | 0 | 4544 passed / 103 skipped | 4461 — stale |
| **control-ui (system locale, es-ES)** | **1** | 1633 passed / **2 failed** | **1633** |
| control-ui (`LC_ALL=en_US.UTF-8`) | 0 | **1635** passed | — |
| host-context-controller | 0 | **1686** passed | 1085 — off by 601 |

The two control-ui failures are **locale-only** (`'12,50 US$'` vs `'12.5'` in `budgets.test.ts`; `/33\.33%/` in `GovernedTraceSurface.test.tsx`), both in files this PR does not touch, and both pass under `en_US`. **Not a PR regression.** But `1633 passed, 1 skipped` is precisely the passed count of that red run — a genuinely green run yields 1635. Worth correcting so the recorded evidence matches a green run.

---

## 6. What came out clean (verified, not assumed)

- **`advertise == enforce` is structurally incapable of diverging.** `index.ts:231` passes `config.uploadV2` to `uploadCapabilities` and `:240` passes **the same object** as `deps.config`; advertisement (`uploadSession.ts:1903`) and admission (`:583`) read the same field of the same in-process object.
- **Integer handling: 30/30 adversarial inputs behaved exactly as required, zero deviations.** Rejects `""`, `"0"`, `"-1"`, `"+1"`, `"01"`, `" 1"`, `"1 "`, `"1\n"`, `"1e3"`, hex, `"1.5"`, `9007199254740992`, `1073741825`, **and Arabic-Indic / fullwidth digits**. No `Math.min`/`Math.max`/`clamp` anywhere on the product path; the only `??` is default-selection over `undefined`.
- **HCC is a genuinely opaque forwarder**: raw `process.env` read, `undefined` dropped, `""` forwarded so GFSC throws at startup.
- **Geometry cannot overflow**: 1 GiB ÷ 1 MiB minimum part = exactly 1024, the `maxPartCount` ceiling.
- **The AST extractor is robust on its main surface** — rejects `+`, `/`, `-0`, `<<`, `**`, identifiers, calls, `parseInt`, parentheses, comma operator, hex, float, exponent, BigInt, `as const`, `satisfies`, unary and ternary; accepts only decimal integers (with `_`) and `*`. Comments, `let`/`var`, strings, template literals and `declare const` are invisible; dead-branch and `namespace` declarations produce a second match → throw. **8 mutations against the real sources → 8 REDs**, baseline green.
- **`NAME-` syntax verified offline** with `kubectl set env --local`: unsets while preserving unrelated vars, unsetting a missing var is an idempotent no-op, and mixing `A=v B-` applies each correctly. The code never mixes them.
- **Process execution is safe**: `execFile`/`execFileSync` with argv arrays everywhere, no `exec`, no `shell:true`, no shell-string interpolation, `timeout` and `maxBuffer` set. `validateGfsUploadProductMaxBytes` runs before the value reaches `kubectl set env` on both branches. SQL uses `sqlLiteral` quoting, `::uuid` casts, `psql -v ON_ERROR_STOP=1`, and a single argv element.
- **`assertOwnedRuntimeContext()` is the first `await` in all four mutating helpers** (`:369`, `:435`, `:460`, `:472`); read-only helpers correctly omit it.
- **No widened preflight regression** (REPRODUCED): 100 MiB + v2 writer advertising 500 MiB passes; `1 GiB + 1` still rejected.
- **Docs are accurate** — `minikube.md` and `gfs-permission-store.md` match the new names, default, range, alias-agreement rule and new-sessions-only semantics.
- **Supply chain**: no new dependency; the contract test uses the pre-existing `typescript` devDependency; no postinstall scripts.
- Zero placeholders, zero `|| true`/`set +e`/`continue-on-error`, zero `.only`/`xit`, zero unconditional skips, zero `testMode`/`NODE_ENV` branches in production code across all 35 files.

---

## 7. Coverage that only exists behind an opt-in flag

`GFS_UPLOAD_V2_RUNTIME_LIMIT_E2E=1` appears **only** in the two specs and in `docs/deploy/minikube.md:326` — never in a workflow, Makefile or script. **The PR's headline claim — lowering the writer to 100 MiB and raising it to 300 MiB takes effect without rebuilding the deployed bundle — is verified there and nowhere else**, so it never runs automatically. Combined with M3, the pure helper behind that journey has no CI coverage either.

Mitigating: config parsing, writer admission/replay semantics, both clients' capability negotiation and limit enforcement, and the e2e helper's validation logic *are* covered by tests that do run. What is uniquely lost with the flag unset is the **live wiring**: HCC env → rendered writer deployment → advertised capability → browser behaviour. The sibling gates `GFS_UPLOAD_V2_E2E` and `GFS_UPLOAD_V2_NEGATIVE_E2E` are gated identically on the base commit, so this follows house convention rather than introducing a new anti-pattern.

---

## 8. Follow-up issues do not exist

The approval on 2026-08-19 was conditional: *"indicate the all Low issue in the review will handle like follow ups after in another PR's"*.

**No issue currently captures L2, L6, L7 or I1.** This repo does have that practice — e.g. #426 is explicitly *"NP-08 PR1 follow-up, R1-L5/NEW-1"* — so this is a genuine gap rather than a convention mismatch. On merge, those four findings plus the ~23 from this pass stop existing in the tracker.

Minor traceability note: the fix commits renumbered the findings after L2 was deferred — `r3-l1`→L1, `r3-l2`→**L3**, `r3-l3`→**L4**, `r3-l4`→**L5**. Searching the history for "L3" does not find its fix.

---

## What I'd like from you, @josue-tech-94

Please reply indicating what you intend to do with each item. My suggested split, for reference — feel free to disagree with any of it:

**Worth fixing in-PR (~15 lines total, high return):**

1. **H2** — complete the duplicate counting (shorthand + spread + computed). This closes for real a finding already reported as closed, and it is the guard protecting the cross-stack transport contract.
2. **M2** — add `expect(GFS_FILE_UPLOAD_DEFAULT_PRODUCT_MAX_BYTES).toBe(209_715_200)`. One line; restores the lost pin on the default.
3. **M4** — add a length floor to `assertUploadV2TransportBounds`. One line; the same suite already has this guard elsewhere.
4. **M3** — add the two test files to `DEFAULT_VITEST_SUITES`. Two lines, and it activates CI coverage for the whole A-L5 fix.
5. **PR body** — refresh the HCC / control-api numbers and re-run control-ui under `en_US` so the recorded evidence is a green run.

**Follow-up issues (please create them — today none exist):** H1, H3, M1, M5, M6, the `staticBytesFromSource` shadowing variant, plus the still-live deferrals L2 / L6 / L7 / I1 and the Lows (message drift between the two clients including the `'GFS resimable'` typo, raw bytes vs MiB in adjacent messages, the duplicated 200 MiB constant, the temp-dir leak in `gfsUploadV2Fixtures.ts`, the hand-pinned `209715200` literal in two e2e specs, the parametrised row that exercises the protocol guard rather than the product guard).

**Your call:**

- **H1** — fix it, or accept that a writer publishing a corrupt contract degrades to legacy? My view is that it is worth fixing, because a ≤16 MiB file *succeeds* and buries the violation entirely — but the design call is yours.
- Do the follow-up issues get created before merge, or do we accept losing them?

**One outstanding certification gap:** `6f335e6e` (r3-l3, Playwright fixture ownership) is the only fix without mutation verification — it needs a live cluster. If you want 100% closure, that is the remaining hole.

Everything else in this PR holds up well. The idempotent-replay reordering is genuinely hard to get right and it is correct, with a test that mechanically pins the authorize-before-lookup ordering; the config parser is one of the more rigorous ones in the repo; and the AST contract test is a real improvement over the regex it replaced, even with H2 outstanding.

josue-tech-94:
@alfredolopez80 
## Addressed in PR #401

The round-four design stop was resolved through Spec 007 and the following items are now implemented and validated:

- **H1** — malformed writer capability is now fail-loud and no longer silently degrades to legacy; genuine unavailable/disabled outcomes retain the bounded fallback.
- **H2** — the Upload v2 source-authority extraction was replaced with the bounded structural resolver, including computed/shorthand/spread/export-shadow handling.
- **H3** — runtime policy mutation now has durable branch-owned recovery state rather than relying only on `finally`.
- **M1** — the missing-`maxFileBytes` 200 MiB compatibility path is preserved but now emits a non-disruptive diagnostic.
- **M2** — the 200 MiB compatibility default is independently pinned by tests.
- **M3** — the runtime helper suites are now registered in the always-on Vitest lane.
- **M4** — empty protocol/part/relay authority sets now fail closed.
- **M5** — both canonical and alias product-limit env inputs are snapshotted and restored exactly, including asymmetric/absent states.

The final implementation pass also caught and fixed two same-round recovery issues:

- recovery/deletion is bound to the immutable lease and UID;
- recovery takeover now requires explicit matching UID + holder evidence rather than being inferred automatically.

The final collateral, correctness, and security reviews are clean. CI, CodeQL, and Real PostgreSQL are green on the reviewed head, and the final review result is **Approve with follow-ups / no must-fix items**.

## Intentionally preserved decisions

These are not being treated as bugs in #401:

- the compiled pre-auth protocol guard / replay behavior remains the accepted contract;
- missing `maxFileBytes` continues to mean the intentional 200 MiB compatibility fallback;
- legacy fallback and explicit-resume semantics remain as defined by PR #311 / Spec 007.

## Follow-up PR work

I am keeping the remaining non-blocking work explicitly tracked rather than considering it closed by #401.

### Client version-skew / guidance follow-up
This will cover:
- mixed-version Desktop error quality;
- Control UI pre-selection guidance versus the actual resolved writer product limit.

### Upload v2 test / copy hygiene follow-up
This will cover:
- message spelling/drift and byte-vs-MiB consistency;
- Upload v2 temp-directory cleanup;
- avoidable duplicated 200 MiB literals in E2E plumbing while preserving the literal contract assertions;
- the parameterized test whose description does not match the guard actually exercised.

### Desktop resumed-session geometry follow-up
The Desktop resumed-session `partBytes` validation divergence was confirmed as pre-existing rather than introduced by #401, so it stays outside this PR and will be addressed separately to bring Desktop into parity with Control UI.

## Live certification

The same-built Desktop / same-deployed Control UI **100 MiB → 300 MiB runtime-policy journey** is still not executed because it requires explicit cluster mutation.

That remains separate promotion certification rather than a known code defect or merge blocker. It will only be run against a branch-owned non-production profile with exact baseline restoration.

So nothing from the review is being dropped: the correctness/contract items are closed in #401, the accepted design choices remain explicit, and the remaining UX/test/pre-existing items are preserved as follow-up work.

PR #410, opened 2026-08-18
fix(#408): generation-aware network-ready handshake between WRC and HCC

### LINKED ISSUE #408: WRC: waitForNetworkReady lacks an observedGeneration check and the network-ready annotation is never cleared
### Summary
The WRC readiness gate `waitForNetworkReady` gates on a flat `clerum.io/network-ready === "true"` annotation with **no `observedGeneration` check**, and the annotation is **never cleared** on a spec change. A stale ack from a previous generation can therefore satisfy the gate, letting a recipe proceed while the network for the *current* generation is not actually ready.

### Where
- `workflow-recipes/src/reconciler/mcpDelegation.ts:898` — `waitForNetworkReady` (flat annotation gate).
- Contrast with `waitForExternalEgressReady` (`:969`) which correctly checks `condition.observedGeneration === currentGeneration` (`:1000`).
- The annotation is merged/carried-forward on replace (`:442`) and never cleared.

### Impact
A recipe can observe a stale `network-ready: "true"` left over from a prior generation and proceed before the current generation's network is ready. **High**, pre-existing — surfaced by zach88's round-3 review on #205, re-confirmed by the adversarial re-review of #382 (#205 → #380 → #382).

### Fix direction
Mirror `waitForExternalEgressReady`:
1. Require `observedGeneration` to match the current generation before treating the ack as valid.
2. Clear (or invalidate) the `network-ready` annotation on a spec change so a stale ack cannot satisfy a new generation.

### Provenance
Deferred follow-up from the #382 lineage re-review. Filed and assigned to attack directly.

### LINKED ISSUE #413: WRC 7b/9a spec-hash divergence: preDeploy vs delegate manifests never converge (2 PUTs/pass; #408 makes the network-ready ack flap across WRC↔HCC)
## Summary

`preDeployMcpServers` (step 7b) and `delegateTransportWorkloads` (step 9a) build **divergent desired manifests for the same McpServer in the same reconcile pass**, so their `spec-hash` annotations never agree. The idempotency gate in `ensureMcpServer` therefore fires a `replace` **twice every pass, forever** — a permanent 2-PUT/pass ping-pong that the spec-hash gate's docstring explicitly promises to prevent (it claims to skip unchanged writes to stop HCC's watch-driven NetworkPolicy recompute storm).

This churn is **pre-existing**, but it becomes materially worse after #408 (PR #410): that PR drops the `network-ready` ack pair on every `replace`, so the ack now **flaps across two controllers each pass** (WRC drops it → HCC re-acks it), coupling the generic stdio readiness gate's latency to HCC's re-ack turnaround. Filing this so the 7b/9a divergence is fixed rather than left open-ended, since #408 is what turns the latent write churn into real cross-controller ack writes.

## Root cause (verified against the code, HEAD of `dev`)

- **7b** injects the annotation into the manifest before ensure: `workflow-recipes/src/reconciler/mcpDelegation.ts:833-838` sets `meta.annotations = { ..., [PRE_DEPLOY_ANNOTATION]: 'true' }`, then `ensureMcpServer` at `:846`.
- **9a** builds the manifest via `buildMcpServerManifest` (`:1148`) and calls `ensureMcpServer` (`:1150`) **without** that annotation; `buildMcpServerManifest` does not add it.
- Both target the same McpServer name (`mcpServerName`), same namespace, same pass: `workflowRecipeReconciler.ts:2212` (7b) → `:2242-2243` (7c `waitForNetworkReady`) → `:2436` (9a).
- `computeSpecHash` **includes** `metadata.annotations` (only strips the spec-hash key itself) — `specHash.ts:54-74`, strip at `:65`. So `hash(7b manifest) ≠ hash(9a manifest)`.
- Each pass, `specHashUnchanged` (`mcpDelegation.ts:468-475`) returns false for the other path's stored hash → both `replace` paths run → 2 PUTs/pass, permanently (the live object never converges: 9a carries `pre-deploy:true` over via the merge but computes its hash without it).

## Interaction with #408 (PR #410)

- **Pre-#408:** the replace merged `{ ...existingAnnotations, ...manifestAnnotations }`, so the ack survived every replace, and HCC's old guard (`network-ready !== 'true'`) never re-acked. The ack was **write-stable** despite the ping-pong.
- **Post-#408:** the replace drops the ack pair (`mcpDelegation.ts:477-479`). Because 7b replaces immediately before 7c, the ack is dropped **right before `waitForNetworkReady` runs**, so the generic gate must wait for HCC to re-ack (event-driven on the rv bump; typically 1–3s, up to the 30s `NETWORK_READY_TIMEOUT_MS` ceiling if HCC lags). Still fail-open — worst case is latency, never a block — but the ack now flaps across two controllers every pass.

## Fragile dependency worth fixing together

`workflowRecipeReconciler.ts:2427` comments "remove pre-deploy annotation" on the 9a path, but 9a **does not** remove it — the replace carry-over preserves it. That inaccuracy is **load-bearing**: HCC's re-ack guard requires `pre-deploy === 'true'` (`host-context-controller/src/reconciler.ts:1940`). If someone "fixes" 9a to honor that comment, HCC would stop re-acking and every generic-stdio pass would eat the full 30s timeout. Fix the comment and make the dependency explicit.

## Proposed fix

Make 7b and 9a hash identically so the idempotency gate actually holds (zero PUTs in steady state, as its docstring promises), which also makes the #408 ack-drop a steady-state no-op:

1. Exclude `PRE_DEPLOY_ANNOTATION` from `computeSpecHash` (alongside the spec-hash key) so 7b's and 9a's manifests hash the same, **and** keep carrying `pre-deploy` over in the replace (it already survives via the merge). Net: the object converges, the gate skips, HCC does not re-recompute NetworkPolicies each pass.
2. Correct the misleading `:2427` comment and document that `pre-deploy` intentionally persists on the live object and that HCC's re-ack keys off it.
3. Add a targeted T1/E2E confirmation on a **`managed:true` stdio, no-external-egress** McpServer (the only profile that enters `genericServers` via `waitForNetworkReady`), measuring drop→re-ack latency per pass, so the convergence fix is proven and can't regress.

## Severity

- **Medium.** Not a correctness bug and not a merge blocker for #408 (fail-open; worst case +30s latency on generic stdio readiness). But it defeats the spec-hash gate's stated purpose (no-op writes) and, post-#408, turns latent churn into recurring cross-controller ack writes + gate latency.

## Scope note

Surfaced during the #408 review (PR #410, @claude review + adversarial verification). #408's fix is correct and safe on its own merits; this is the follow-up it explicitly defers. Verified only on the reconcile-loop path; independent of PR #275 and issue #412.

### LINKED ISSUE #414: WRC↔HCC handshake annotation keys are duplicated string literals across two services (no shared constant)
## Summary

The three annotation keys of the WRC↔HCC pre-deploy readiness handshake are defined as **independent string literals in each service**, with no shared source of truth:

- WRC: `workflow-recipes/src/reconciler/mcpDelegation.ts:790-794` — `PRE_DEPLOY_ANNOTATION`, `NETWORK_READY_ANNOTATION`, `NETWORK_READY_GENERATION_ANNOTATION` (exported).
- HCC: `host-context-controller/src/reconciler.ts:76-80` — the same three keys as local (non-exported) constants.

A typo on one side breaks the handshake **silently**: the gate is never fresh → `waitForNetworkReady` hits its 30s fail-open timeout → the recipe proceeds with a `console.warn`, nothing goes red. Raised by @zach88 in the #410 review (R1-L2).

## Why not fixed in #410

Both services already depend on `@clerum/network-policy-core` (`file:../packages/network-policy-core`), which is the natural home for these handshake keys (they are part of the NetworkPolicy-readiness contract). But moving them touches the build of two services plus the shared package, and #410 already carries #408 + #413. An in-repo cross-service equality test is not a clean substitute — HCC and WRC are separate npm packages with their own tsconfigs, so importing one's constants into the other's test breaks build isolation.

## Proposed fix

1. Add the three handshake annotation keys as exported constants in `@clerum/network-policy-core`.
2. Import them in both `workflow-recipes` and `host-context-controller`, removing the per-service literals.
3. The shared constant makes a divergence a compile error instead of a silent 30s-timeout degradation.

## Scope

- **Severity: Low** — no current bug (the keys are byte-identical today, verified); this is defense against future drift.
- Surfaced during the #410 review (issue #408 / #413). Independent of #412.

Partial mitigation already in #410: `workflow-recipes/src/reconciler/specHash.test.ts` has a drift guard tying `HASH_EXCLUDED_ANNOTATIONS` to `PRE_DEPLOY_ANNOTATION` — but only within WRC, not cross-service.

alfredolopez80:
## Summary

Fixes #408. The WRC readiness gate `waitForNetworkReady` trusted a flat
`clerum.io/network-ready: "true"` annotation with **no `observedGeneration`
check**, and the annotation was **never cleared on a spec change**. A stale ack
left over from a previous generation could therefore satisfy the gate, letting a
recipe proceed before the **current** generation's network was actually ready.

This mirrors the freshness discipline already used by `waitForExternalEgressReady`
(`condition.observedGeneration === currentGeneration`), applied minimally to the
network handshake — **without** introducing a new status condition.

## Root cause (verified against the code)

- HCC `setNetworkReadyAnnotation` wrote only the flat `network-ready: "true"`
  (`host-context-controller/src/reconciler.ts`), and its re-ack guard
  (`network-ready !== 'true'`) never re-acked once the flag was set — even across a
  generation change.
- WRC `waitForNetworkReady` treated `annotations['network-ready'] === 'true'` as
  ready, with no generation comparison (`workflow-recipes/src/reconciler/mcpDelegation.ts`).
- On a spec change, WRC `replaceFromExisting` merged existing annotations as the
  base (`{ ...existingAnnotations, ...manifest }`), carrying a stale
  `network-ready: "true"` from generation N-1 into generation N. The manifest never
  re-specifies it, so it was never cleared.

## The fix (the two points from the issue)

**#1 — generation-aware ack (defense in depth):**
- HCC now stamps `clerum.io/network-ready-observed-generation: "<metadata.generation>"`
  alongside the ack, and the re-ack guard re-acknowledges when the stamped
  generation is stale (spec changed since the last ack). Non-numeric generation
  keeps the legacy behavior.
- WRC `waitForNetworkReady` trusts `network-ready` only when the stamped generation
  equals `metadata.generation` (non-numeric generation tolerated as fresh, same as
  the mirror). The 404 = resolved path is unchanged.

**#2 — clear the stale ack on a spec change:**
- WRC `replaceFromExisting` drops the network ack pair
  (`network-ready` + `network-ready-observed-generation`) from the carry-over. This
  branch runs only when the desired manifest's spec-hash differs from the stored one
  (the idempotency gate skips the replace when they match), so a byte-identical
  desired object is never rewritten.

**Complementarity:** #2 removes the carry-over path; #1 makes the gate reject any
stale ack regardless of how it arrived.

## Behavior note (fail-open, not fail-closed)

The gate remains **fail-open**: `waitForNetworkReady`'s result only feeds a
`console.warn` in `workflowRecipeReconciler.ts` and the recipe proceeds after the
30s timeout regardless. This fix removes the **stale-ack false positive** (the gate
no longer returns `ready:true` for a prior generation); it does **not** turn the
gate into a hard block. The security posture ("proceed after 30s") is unchanged.

## Rollout / version skew

WRC and HCC are separate images from the same release. In the skew window
**WRC-new + HCC-old** (no stamp), the gate is never fresh → it waits 30s → warns →
proceeds: only +30s of latency per pass on stdio servers, identical posture to
pre-fix. **WRC-old + HCC-new** is inert (old WRC ignores the stamp). Recommended
deploy order: **HCC first, then WRC** (the apply order is controlled by
evenfire-infra and not enforced here).

## Out of scope

- Variant 2b (a full `NetworkReady` status condition) — intentionally not done; the
  annotation pair closes the hole with a minimal change.
- The pre-existing 7b/9a spec-hash ping-pong (see "Follow-up") is **not** fixed here.

## Tests (TDD, red → green)

Five tests written red-first (fail against old code), then green after the fix; no
regressions.

- WRC `mcpDelegation.test.ts`:
  - gate rejects an ack stamped for an older generation
  - gate rejects an ack that lacks the generation stamp
  - replace drops the carried-over ack pair on a spec change
  - (+ regression pins: current-gen stamp accepted, non-numeric tolerated, 404 resolved)
- HCC `reconciler.test.ts`:
  - stamps the observed generation alongside the ack
  - re-acks when the stamped generation is stale
  - (+ pin: does not re-ack when the stamp already matches — no write storm)

Results: **WRC 2478 passed / 0 failed** (29 skipped), **HCC 1088 passed / 0 failed**.
`tsc` build green for both packages.

## Docs

`docs/architecture/platform-topology.md` — annotation table (§11.4.3), Option C row,
and the handshake diagram updated for the ack + generation pair.

## Follow-up (not in this PR)

A pre-existing spec-hash divergence between step 7b (`preDeployMcpServers`, stamps
`pre-deploy: true`) and step 9a (`delegateTransportWorkloads`, does not) makes the
replace path recur on each reconcile pass, so the ack drop can recur too. This is
pre-existing churn (not introduced here) and is safe (the re-ack does not bump
`generation` — the CRD has a status subresource). It should be verified in a T1 live
run and, if confirmed, unified in a separate issue.

---

## Also fixes #413 — WRC 7b/9a spec-hash convergence

Surfaced by the @claude review of this PR: the ack-drop introduced here interacts with a pre-existing structural divergence. WRC step 7b (`preDeployMcpServers`) stamps `clerum.io/pre-deploy: "true"` on the manifest; step 9a (`delegateTransportWorkloads`) builds the manifest without it. Because `computeSpecHash` hashed `metadata.annotations`, the two manifests for the same McpServer hashed differently → the idempotency gate fired a `replace` twice every pass forever, and (post-this-PR) that dropped the `network-ready` ack pair each pass, flapping it across WRC↔HCC.

**Fix (cherry-picked from @claude's #413 branch, verified here):** exclude `clerum.io/pre-deploy` from `computeSpecHash` (`HASH_EXCLUDED_ANNOTATIONS`) so 7b and 9a hash identically → the gate converges to **zero PUTs in steady state**, which also makes this PR's ack-drop a steady-state no-op (it only runs on a real spec change, as intended). `pre-deploy` still persists on the live object via the replace merge (HCC's re-ack guard requires it); only the hash excludes it. The misleading Step 9a "remove pre-deploy annotation" comment is corrected, and a drift-guard test ties `HASH_EXCLUDED_ANNOTATIONS` back to `PRE_DEPLOY_ANNOTATION`.

**Verification (locally, this branch):** `tsc` build green; `specHash.test.ts` 12/12 (convergence + drift-guard); full `workflow-recipes` suite **2480 passed / 0 failed**.

### Correction to the R1 evidence (from the @claude review)

The earlier R1 note ("recipe converges, zero write-churn") was measured on a `streamableHttp` / `managed:false` McpServer, which never enters `waitForNetworkReady` (generic gate is `stdio`-only) and never receives the `network-ready` ack (HCC diverts `managed:false` before the ack block). So that observation did **not** exercise the ack-drop path — its stable `resourceVersion` shows the 7b/9a path simply didn't run in that window. The real severity is: fail-open, +1–3s typical / +30s ceiling per pass on generic-stdio readiness — now resolved by the #413 convergence above.

## Note on the diff

The `package.json` / `package-lock.json` version bumps in both packages are automatic from the repo's `.githooks/pre-commit` `version:staged` step (one patch increment per commit that touches each package), not manual changes.

Fixes #413

## Also fixes #414 — shared handshake annotation keys

Following @zach88's R1-L2, the three handshake keys (`pre-deploy`, `network-ready`, `network-ready-observed-generation`) were duplicated as independent string literals in WRC and HCC, so a typo on one side degraded silently (never fresh → 30s fail-open → nothing red). Moved them into `@clerum/network-policy-core` (already the home of the egress-fqdn annotation contract, already a dependency of both services); WRC re-exports them for existing consumers, HCC imports them, and `specHash`'s `HASH_EXCLUDED_ANNOTATIONS` uses the shared constant too. **A divergence is now a compile error, not a silent runtime drift** — a stronger guarantee than any cross-service test, which is why the "cross-controller coverage" residual is closed here. Verified: `tsc` green (WRC+HCC), WRC suite 2480/0, HCC #408 green, `network-policy-core` self-test 42/0.

Fixes #414

### DIFF
--- a/docs/architecture/platform-topology.md
+++ b/docs/architecture/platform-topology.md
@@ -875,7 +875,7 @@ WRC                                    McpServer CRD                          HC
 | Annotation                                   | Set by | Value                  | Meaning                                                                                                              |
 | -------------------------------------------- | ------ | ---------------------- | ------------------------------------------------------------------------------------------------------------------- |
 | `clerum.io/pre-deploy`                       | WRC    | `"true"`               | McpServer CRD created before workload pods exist; HCC should apply NPs proactively                                   |
-| `clerum.io/network-ready`                    | HCC    | `"true"`               | L2/L3 NetworkPolicies applied; safe to start workload pods. Valid only together with a matching generation stamp     |
+| `clerum.io/network-ready`                    | HCC    | `"true"`               | L2/L3 NetworkPolicies applied; safe to start workload pods. Honored only when the generation stamp matches the current generation (a non-numeric generation is tolerated as fresh without a stamp — defensive; `metadata.generation` is always numeric in a real cluster) |
 | `clerum.io/network-ready-observed-generation` | HCC    | `"<metadata.generation>"` | Generation HCC observed when acking (Issue #408). WRC honors `network-ready` only when this equals the current generation, so a stale ack from a prior generation cannot satisfy the gate. HCC re-acks (re-stamps) when the generation advances |
 
 #### 11.4.4 Timeout and Graceful Degradation
--- a/docs/architecture/platform-topology.md
+++ b/docs/architecture/platform-topology.md
@@ -840,7 +840,7 @@ Three approaches were evaluated:
 | --------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | **A — ReadinessGate**                   | HCC patches `pod.status.conditions` with `clerum.io/network-isolated: True` after applying NPs                                                    | Standard K8s pattern, Service won't route until ready                                                    | HCC must watch pods in all namespaces, direct pod patching violates Invariant #1 (CRD-only communication), adds Pod RBAC to HCC, complex lifecycle (pod restart = re-patch) |
 | **B — Init container**                  | Sidecar init container blocks until NetworkPolicy exists                                                                                          | No HCC changes needed                                                                                    | Requires shared ServiceAccount, polling from within pod is fragile, no guarantee NP is _applied_ vs just _created_                                                          |
-| **C — Pre-deploy annotation handshake** | WRC creates McpServer CRD with `clerum.io/pre-deploy: true` → HCC applies NPs → HCC sets `clerum.io/network-ready: true` → WRC creates Deployment | 100% CRD-based (Invariant #1 preserved), no pod patching, no RBAC expansion, reuses existing watch loops | Adds ~1-5s latency to first deployment, requires timeout handling for HCC unavailability                                                                                    |
+| **C — Pre-deploy annotation handshake** | WRC creates McpServer CRD with `clerum.io/pre-deploy: true` → HCC applies NPs → HCC sets `clerum.io/network-ready: true` (+ `network-ready-observed-generation`, Issue #408) → WRC creates Deployment | 100% CRD-based (Invariant #1 preserved), no pod patching, no RBAC expansion, reuses existing watch loops | Adds ~1-5s latency to first deployment, requires timeout handling for HCC unavailability                                                                                    |
 
 **Decision**: **Option C** — the annotation handshake preserves the CRD-only communication invariant while closing the vulnerability window. ReadinessGate (Option A) is deferred indefinitely as Option C provides equivalent security guarantees with simpler implementation.
 
@@ -857,7 +857,7 @@ WRC                                    McpServer CRD                          HC
  │                                           ├── watch triggers ───────────────►│
  │                                           │                                  │
  │                                           │◄── HCC applies L2/L3 NPs ───────┤
- │                                           │    + sets annotation:            │
+ │                                           │    + sets ack + generation:      │
  │                                           │    clerum.io/network-ready: true │
  │                                           │                                  │
  │◄── Step 7c: Poll for network-ready ──────┤                                  │
@@ -872,10 +872,11 @@ WRC                                    McpServer CRD                          HC
 
 #### 11.4.3 Annotations
 
-| Annotation                | Set by | Value    | Meaning                                                                            |
-| ------------------------- | ------ | -------- | ---------------------------------------------------------------------------------- |
-| `clerum.io/pre-deploy`    | WRC    | `"true"` | McpServer CRD created before workload pods exist; HCC should apply NPs proactively |
-| `clerum.io/network-ready` | HCC    | `"true"` | L2/L3 NetworkPolicies applied; safe to start workload pods                         |
+| Annotation                                   | Set by | Value                  | Meaning                                                                                                              |
+| -------------------------------------------- | ------ | ---------------------- | ------------------------------------------------------------------------------------------------------------------- |
+| `clerum.io/pre-deploy`                       | WRC    | `"true"`               | McpServer CRD created before workload pods exist; HCC should apply NPs proactively                                   |
+| `clerum.io/network-ready`                    | HCC    | `"true"`               | L2/L3 NetworkPolicies applied; safe to start workload pods. Valid only together with a matching generation stamp     |
+| `clerum.io/network-ready-observed-generation` | HCC    | `"<metadata.generation>"` | Generation HCC observed when acking (Issue #408). WRC honors `network-ready` only when this equals the current generation, so a stale ack from a prior generation cannot satisfy the gate. HCC re-acks (re-stamps) when the generation advances |
 
 #### 11.4.4 Timeout and Graceful Degradation
 
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.171",
+  "version": "0.1.172",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.171",
+      "version": "0.1.172",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.170",
+  "version": "0.1.171",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.170",
+      "version": "0.1.171",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.172",
+  "version": "0.1.173",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.172",
+      "version": "0.1.173",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.171",
+  "version": "0.1.172",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.170",
+  "version": "0.1.171",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.172",
+  "version": "0.1.173",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/src/reconciler.test.ts
+++ b/host-context-controller/src/reconciler.test.ts
@@ -149,6 +149,89 @@ describe('Reconciler managed:false guard (Risk 1.7)', () => {
     )
   })
 
+  it('Issue #408: stamps the observed generation alongside the network-ready ack', async () => {
+    const server = {
+      ...makeServer({ name: 'stdio-mcp', managed: true }),
+      generation: 3,
+      annotations: { 'clerum.io/pre-deploy': 'true' },
+    }
+
+    await reconciler.reconcile(server)
+
+    expect(customApi.patchNamespacedCustomObject).toHaveBeenCalledWith(
+      expect.objectContaining({
+        plural: 'mcpservers',
+        name: 'stdio-mcp',
+        body: {
+          metadata: {
+            annotations: {
+              'clerum.io/network-ready': 'true',
+              'clerum.io/network-ready-observed-generation': '3',
+            },
+          },
+        },
+      }),
+      expect.objectContaining({ middleware: expect.any(Array) })
+    )
+  })
+
+  it('Issue #408: re-acks when the stamped generation is stale (spec changed since last ack)', async () => {
+    const server = {
+      ...makeServer({ name: 'stdio-mcp', managed: true }),
+      generation: 3,
+      annotations: {
+        'clerum.io/pre-deploy': 'true',
+        'clerum.io/network-ready': 'true',
+        'clerum.io/network-ready-observed-generation': '2',
+      },
+    }
+
+    await reconciler.reconcile(server)
+
+    // Old guard (network-ready !== 'true') would skip the patch; the fix must re-ack
+    // for the current generation, re-stamping observed-generation to '3'.
+    expect(customApi.patchNamespacedCustomObject).toHaveBeenCalledWith(
+      expect.objectContaining({
+        plural: 'mcpservers',
+        name: 'stdio-mcp',
+        body: {
+          metadata: {
+            annotations: {
+              'clerum.io/network-ready': 'true',
+              'clerum.io/network-ready-observed-generation': '3',
+            },
+          },
+        },
+      }),
+      expect.objectContaining({ middleware: expect.any(Array) })
+    )
+  })
+
+  it('Issue #408: does NOT re-ack when the stamped generation already matches (no write storm)', async () => {
+    const server = {
+      ...makeServer({ name: 'stdio-mcp', managed: true }),
+      generation: 3,
+      annotations: {
+        'clerum.io/pre-deploy': 'true',
+        'clerum.io/network-ready': 'true',
+        'clerum.io/network-ready-observed-generation': '3',
+      },
+    }
+
+    await reconciler.reconcile(server)
+
+    expect(customApi.patchNamespacedCustomObject).not.toHaveBeenCalledWith(
+      expect.objectContaining({
+        body: expect.objectContaining({
+          metadata: expect.objectContaining({
+            annotations: expect.objectContaining({ 'clerum.io/network-ready': 'true' }),
+          }),
+        }),
+      }),
+      expect.anything()
+    )
+  })
+
   it('should create Deployment when managed: undefined (default: true)', async () => {
     const server = makeServer({ name: 'mongo-mcp' })
     await reconciler.reconcile(server)
--- a/host-context-controller/src/reconciler.test.ts
+++ b/host-context-controller/src/reconciler.test.ts
@@ -232,6 +232,39 @@ describe('Reconciler managed:false guard (Risk 1.7)', () => {
     )
   })
 
+  it('Issue #408: re-acks when the ack predates the stamp (upgrade path: stamp absent)', async () => {
+    // A pre-#408 HCC set network-ready:'true' without an observed-generation stamp.
+    // The re-ack guard's stale-generation branch (typeof gen === 'number' && stamp !==
+    // String(gen), i.e. undefined !== '3') must fire so the ack gains a generation.
+    const server = {
+      ...makeServer({ name: 'stdio-mcp', managed: true }),
+      generation: 3,
+      annotations: {
+        'clerum.io/pre-deploy': 'true',
+        'clerum.io/network-ready': 'true',
+        // no clerum.io/network-ready-observed-generation
+      },
+    }
+
+    await reconciler.reconcile(server)
+
+    expect(customApi.patchNamespacedCustomObject).toHaveBeenCalledWith(
+      expect.objectContaining({
+        plural: 'mcpservers',
+        name: 'stdio-mcp',
+        body: {
+          metadata: {
+            annotations: {
+              'clerum.io/network-ready': 'true',
+              'clerum.io/network-ready-observed-generation': '3',
+            },
+          },
+        },
+      }),
+      expect.objectContaining({ middleware: expect.any(Array) })
+    )
+  })
+
   it('should create Deployment when managed: undefined (default: true)', async () => {
     const server = makeServer({ name: 'mongo-mcp' })
     await reconciler.reconcile(server)
--- a/host-context-controller/src/reconciler.ts
+++ b/host-context-controller/src/reconciler.ts
@@ -74,6 +74,9 @@ type McpServerReconcilerDeps = {
 // G2: Pre-deploy handshake constants
 const PRE_DEPLOY_ANNOTATION = 'clerum.io/pre-deploy'
 const NETWORK_READY_ANNOTATION = 'clerum.io/network-ready'
+// Issue #408: stamp the generation observed when acking so WRC can reject a stale
+// ack carried over from a previous generation.
+const NETWORK_READY_GENERATION_ANNOTATION = 'clerum.io/network-ready-observed-generation'
 
 // Issue #223: sha256 of the envSecret content, carried on the POD TEMPLATE (not
 // the Deployment object) so that changing it changes the pod template hash and
@@ -1499,6 +1502,11 @@ ${authHeaderLines ? '\n        # ── Credential auth headers (envsubst-resolv
           metadata: {
             annotations: {
               [NETWORK_READY_ANNOTATION]: 'true',
+              // Issue #408: stamp the acked generation so WRC's waitForNetworkReady
+              // can reject a stale ack from a previous generation.
+              ...(typeof server.generation === 'number'
+                ? { [NETWORK_READY_GENERATION_ANNOTATION]: String(server.generation) }
+                : {}),
             },
           },
         },
@@ -1930,7 +1938,11 @@ ${authHeaderLines ? '\n        # ── Credential auth headers (envsubst-resolv
     // NetworkPolicies are in place and workload can proceed.
     if (
       server.annotations?.[PRE_DEPLOY_ANNOTATION] === 'true' &&
-      server.annotations?.[NETWORK_READY_ANNOTATION] !== 'true'
+      (server.annotations?.[NETWORK_READY_ANNOTATION] !== 'true' ||
+        // Issue #408: re-ack when the stamped generation is stale (spec changed since
+        // the last ack). A non-numeric generation keeps the legacy behavior.
+        (typeof server.generation === 'number' &&
+          server.annotations?.[NETWORK_READY_GENERATION_ANNOTATION] !== String(server.generation)))
     ) {
       try {
         await this.setNetworkReadyAnnotation(server)
--- a/host-context-controller/src/reconciler.ts
+++ b/host-context-controller/src/reconciler.ts
@@ -13,6 +13,11 @@ import * as k8s from '@kubernetes/client-node'
 import { IntOrString } from '@kubernetes/client-node/dist/types.js'
 import { createHash } from 'crypto'
 import { classifyPluginImage } from '@clerum/image-policy'
+import {
+  NETWORK_READY_ANNOTATION,
+  NETWORK_READY_GENERATION_ANNOTATION,
+  PRE_DEPLOY_ANNOTATION,
+} from '@clerum/network-policy-core'
 import { isWorkflowRecipeDefaultAllowedCapability } from '@clerum/workflow-recipe-capability-policy'
 import { config } from './config'
 import { MANAGED_BY_LABEL, MANAGED_BY_VALUE, MCPSERVER_LABEL } from './constants'
@@ -71,12 +76,10 @@ type McpServerReconcilerDeps = {
   networkingApi?: k8s.NetworkingV1Api
 }
 
-// G2: Pre-deploy handshake constants
-const PRE_DEPLOY_ANNOTATION = 'clerum.io/pre-deploy'
-const NETWORK_READY_ANNOTATION = 'clerum.io/network-ready'
-// Issue #408: stamp the generation observed when acking so WRC can reject a stale
-// ack carried over from a previous generation.
-const NETWORK_READY_GENERATION_ANNOTATION = 'clerum.io/network-ready-observed-generation'
+// G2: Pre-deploy handshake constants — imported from @clerum/network-policy-core
+// (single source of truth shared with WRC; a divergence is a compile error rather
+// than a silent 30s fail-open timeout). clerum.io/network-ready-observed-generation
+// (Issue #408) is the generation HCC stamps when acking, so WRC can reject a stale ack.
 
 // Issue #223: sha256 of the envSecret content, carried on the POD TEMPLATE (not
 // the Deployment object) so that changing it changes the pod template hash and
--- a/packages/network-policy-core/index.cjs
+++ b/packages/network-policy-core/index.cjs
@@ -9,6 +9,14 @@ const STATE_ANNOTATION = 'clerum.io/egress-fqdn-state'
 const TARGETS_ANNOTATION = 'clerum.io/egress-fqdn-targets'
 const RESOLVED_AT_ANNOTATION = 'clerum.io/egress-fqdn-resolved-at'
 
+// Pre-deploy network-readiness handshake keys (WRC producer/consumer, HCC producer).
+// Shared here so WRC (workflow-recipes) and HCC (host-context-controller) derive the
+// exact same literals — a divergence becomes a compile error, not a silent 30s
+// fail-open timeout. Consumed by both services' reconcilers.
+const PRE_DEPLOY_ANNOTATION = 'clerum.io/pre-deploy'
+const NETWORK_READY_ANNOTATION = 'clerum.io/network-ready'
+const NETWORK_READY_GENERATION_ANNOTATION = 'clerum.io/network-ready-observed-generation'
+
 const DEFAULT_PROTOCOL = 'TCP'
 const LEGACY_PORT = 443
 // Upper bound on a DNS TTL (ms) folded into the window (audit M2). A hostile or
@@ -400,6 +408,9 @@ module.exports = {
   STATE_ANNOTATION,
   TARGETS_ANNOTATION,
   RESOLVED_AT_ANNOTATION,
+  PRE_DEPLOY_ANNOTATION,
+  NETWORK_READY_ANNOTATION,
+  NETWORK_READY_GENERATION_ANNOTATION,
   emptyState,
   reconcileEgressState,
   stateHash,
--- a/packages/network-policy-core/index.d.ts
+++ b/packages/network-policy-core/index.d.ts
@@ -130,6 +130,11 @@ export declare const STATE_ANNOTATION: string
 export declare const TARGETS_ANNOTATION: string
 export declare const RESOLVED_AT_ANNOTATION: string
 
+/** Pre-deploy network-readiness handshake annotation keys (shared by WRC and HCC). */
+export declare const PRE_DEPLOY_ANNOTATION: string
+export declare const NETWORK_READY_ANNOTATION: string
+export declare const NETWORK_READY_GENERATION_ANNOTATION: string
+
 /**
  * Classify a DNS failure as 'transient' (resolver/upstream temporarily
  * unavailable → fail-static freeze, H1) or 'permanent' (genuine no-records →
--- a/workflow-recipes/package-lock.json
+++ b/workflow-recipes/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "workload-recipes-controller",
-  "version": "0.1.296",
+  "version": "0.1.297",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "workload-recipes-controller",
-      "version": "0.1.296",
+      "version": "0.1.297",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1009.0",
--- a/workflow-recipes/package-lock.json
+++ b/workflow-recipes/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "workload-recipes-controller",
-  "version": "0.1.297",
+  "version": "0.1.298",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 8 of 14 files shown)


alfredolopez80:
@claude please do a full review of this PR (issue #408).

This fix makes the WRC↔HCC `network-ready` pre-deploy handshake generation-aware, mirroring `waitForExternalEgressReady`'s `observedGeneration` freshness check, so a stale ack from a prior generation can no longer satisfy the gate. Production diff is 2 files (`workflow-recipes/src/reconciler/mcpDelegation.ts`, `host-context-controller/src/reconciler.ts`) plus their tests and the `platform-topology.md` doc. It has already been through heavy validation — please verify it independently and flag anything below that you disagree with.

### Validation performed (evidence)

**Unit tests (TDD, red→green, mutation-validated)**
- `workflow-recipes`: `npm test` → **2478 passed / 0 failed** (29 skipped).
- `host-context-controller`: `npm test` → **1088 passed / 0 failed**.
- Mutation check: reverting the two source files to pre-fix code makes exactly the 5 new tests go red (3 WRC: gate rejects stale/missing stamp, replace clears the ack pair; 2 HCC: stamps generation, re-acks on stale generation). The pre-existing HCC test asserting the exact ack body stays green untouched.
- `npm run build` (tsc) green for both packages. (`npm run lint` is blocked locally by an ESLint 9 flat-config gap in these two packages — pre-existing, not introduced here; CI enforces lint.)

**Automated review — 4 independent agents, all clean**
- *Plan fidelity + correctness*: the two issue points implemented as specified; edge cases verified (`generation === 0` uses `typeof`, no falsy bug; `annotations === undefined` safe; non-numeric generation tolerated as fresh, matching the mirror). Protected regions byte-identical (`waitForExternalEgressReady`, the idempotency gate, the 404 path, timeouts).
- *Security*: gate stays **fail-open** (result only warns; recipe proceeds after the 30s timeout) — the fix removes the stale-ack false positive, it does not turn the gate into a hard block. The non-numeric tolerance is unreachable by an attacker (`metadata.generation` is apiserver-managed) and never weaker than the pre-fix baseline. No `console.*` added; the ack-pair drop is exact-key.
- *Contract WRC↔HCC + version skew*: the annotation key string is byte-identical between producer (HCC) and consumer (WRC); `String(metadata.generation)` serialization is symmetric. Backward-compatible both ways (WRC-new+HCC-old degrades to timeout+warn+proceed; WRC-old+HCC-new is inert). Recommended deploy order: HCC before WRC.
- *Tests & regressions*: no existing test modified; coverage of the changed branches is real.

**Adversarial pass — 3 attacker-lens agents, zero surviving findings**
- Technical (TOCTOU / race / coercion): `annotations` + `generation` read from the same GET; the ack-pair `delete` operates on a copy; CAS-protected replace; `String()` symmetric even beyond 2^53.
- Trust-boundary / forgery: annotation forgery is pre-existing (RBAC traced — no tenant SA can patch `mcpservers`); the generation stamp is a freshness token, not an authenticity one — the fix neither introduces nor worsens the surface.
- Logic / crescendo: cross-generation attacks fail-toward-waiting (safe); the stamp can only go backward, never ahead of the live generation.

**Live minikube T1/T2 (branch profile, `IMAGE_SOURCE=local` HEAD build)**
- Fix deployed live and verified in the running pod (`network-ready-observed-generation` present in the compiled `mcpDelegation.js`).
- **T0** (static/contract): PASS. **T2 preflight**: PASS.
- **T1 Real PostgreSQL**: **92/92 real-postgres tests passed, 0 failed**.
- **R1 (churn observation)**: a recipe-owned transport McpServer converges (READY, `resourceVersion` stable across ~3 min); the WRC reconcile loop's replaces are absorbed by the idempotency gate — reconcile-loop churn, not etcd/write churn.
- One orthogonal caveat, filed separately as #412 and **not** part of this change: the control-api `db.realPostgresMigration` suite can't run against the shared `control-postgres` (its `beforeAll` drops runtime roles that are in use) and needs an isolated postgres in the local T1 lane. It is a T1-harness gap, unrelated to this reconciler fix.

### Ask
Please independently verify the two production edits against the mirror pattern, confirm the fail-open framing is accurate, and check the version-skew reasoning. Flag any correctness, security, or contract issue you find.

alfredolopez80:
Thanks @claude — this is a sharp review, and you're right on the substantive point. I re-verified all of it against the code (adversarial pass) and I'm conceding the framing, not just the follow-up.

**On the 7b/9a ack-flap — confirmed structural, and you're right that #408 activates it.** All five steps check out (7b stamps `pre-deploy` at `mcpDelegation.ts:833-838`, 9a omits it at `:1148-1150`, `computeSpecHash` includes annotations at `specHash.ts:65`, both replace paths fire every pass, each dropping the ack at `:477-479`). Pre-#408 the merge preserved the ack so it was write-stable; this PR's drop is what makes it flap. Filed as **#413** with the fix direction (exclude `PRE_DEPLOY_ANNOTATION` from the spec-hash so 7b/9a converge, which also makes the ack-drop a steady-state no-op) — plus a fragility your trace implies: the `:2427` "remove pre-deploy annotation" comment is inaccurate **and load-bearing** — 9a never removes it, and HCC's re-ack guard requires `pre-deploy === 'true'` (`reconciler.ts:1940`), so a naive "fix" of that comment would make every generic-stdio pass eat the 30s timeout. #413 covers correcting that too.

**On my R1 evidence — you're right, and I'm retracting it.** My R1 observation was on `mongodb-mcp-stack-mongodb-mcp-server`, which is `streamableHttp` / `managed: false`. That server **never enters `waitForNetworkReady`** (generic gate requires `stdio`, `workflowRecipeReconciler.ts:869-871`) and **never receives the `network-ready` ack** (HCC diverts `managed:false` to `reconcileWrcOwnedServer` and returns before the G2 ack block). So its stable `resourceVersion` proves the 7b/9a path *didn't run* in that window — not that it can't. The ack-drop path was simply not exercised. "R1 converges, zero write-churn" was false comfort on the wrong server type; the real confirmation must be on a **`managed:true` stdio, no-external-egress** McpServer, which is tracked in #413.

**Corrected severity framing:** not "safe / write-stable" but *fail-open, +1–3s typical / +30s ceiling per pass on generic-stdio readiness; a pre-existing 2-PUT/pass churn that this PR turns into recurring cross-controller ack writes.* Not a blocker (worst case is latency, never a block; the external-egress gate is condition/status-based and untouched), but not "merely pre-existing churn" either.

**Minor notes — addressed:**
1. *No cross-controller integration test* — correct; all new tests are single-controller unit tests, and an in-repo harness is a real gap (HCC and WRC are independent packages). Folded into #413's T1/E2E confirmation rather than bolted on here.
2. *Ack carry-over comment* — fixed in `5fe3e10d`: a note that only the ack pair is dropped and `pre-deploy` carries over intentionally (HCC keys off it).
3. *Version bumps* — the `package.json`/`package-lock` bumps (HCC +1, WRC +2/+3) are automatic from the repo's `.githooks/pre-commit` `version:staged` step — one patch increment per commit that touches each package, not a manual change.

Net: keeping the #408 fix as-is (you and my adversarial pass agree it's correct, minimal, and safe on its own merits), with the 7b/9a convergence + T1 confirmation tracked in #413 rather than left open-ended. Thanks for catching the R1 blind spot.

zach88:
Took a deep pass through this one, tracing the annotation handshake end to end (HCC producer → WRC consumer) plus a security and contract sweep. The core fix is genuinely solid — the generation-aware freshness check, the string↔number typing, the mirror of `waitForExternalEgressReady`, and HCC's re-ack guard all check out, and security is clean. There's one real thing though, and it's a self-inflicted one: the ack drop reintroduces exactly the churn the handshake is trying to kill, on the common path.

**🔴 Not ready to approve**

1. **[R1-H1] The ack drop fires on *every* reconcile of the common path, not just on real spec changes → perpetual ack flap.** `workflow-recipes/src/reconciler/mcpDelegation.ts:445-484`

   The justifying comment (`:447-453`) says the drop "only runs when the desired manifest's spec-hash differs from the stored one, so a byte-identical object is never rewritten." True per-call, but it misses that `preDeployMcpServers` (7b) and `delegateTransportWorkloads` (9a) never share the same desired manifest: `preDeployMcpServers` adds `clerum.io/pre-deploy: "true"` to the manifest (`mcpDelegation.ts:818-822`), `delegateTransportWorkloads` doesn't, and `computeSpecHash` hashes annotations (`specHash.ts:64-70`). So the two paths stamp different spec-hashes permanently.

   Verified in the code: before this PR the ping-pong replace *preserved* `network-ready` (the merge). Now each replace **deletes** it, so on every reconcile of the common path the ack flaps present→absent→present: preDeploy drops it → `waitForNetworkReady` waits for HCC to re-ack → delegate drops it again. Generation never changes across these replaces (the divergence is annotation-only, not `.spec`), so HCC re-applies NetworkPolicies and re-acks for the *same* generation every loop.

   It's fail-open (just a `console.warn`, recipe proceeds after the timeout), so it's not a correctness break — but it defeats the stability goal of this very PR on the common path and adds recurring churn + latency. The "Follow-up" note calls this "safe (the re-ack does not bump generation)"; that's true but incomplete — the cost isn't the generation bump, it's the NP re-apply and the gate wait each loop.

   Options: (a) merge and accept the churn, opening a separate issue for the 7b/9a ping-pong (the body already flags it as follow-up); (b) drop the ack **only when the generation actually changed**, not on any spec-hash diff — smallest change that kills the flap at the root; (c) unify the 7b/9a desired manifest or exclude annotations from the spec-hash (bigger, better as its own issue). I'd lean (b) if you want it closed here.

## Optional (non-blocking)

The PR can merge without these — follow-up / author's call.

1. **[R1-L1] The "does NOT re-ack when the stamp already matches" test is a guard, not a regression test.** `host-context-controller/src/reconciler.test.ts:206`. It passes against the parent (`ac6972ab`) — verified — because the old code didn't re-ack once `network-ready === "true"` either, so the negative assertion already held without the fix. Fine as a guard against a future write-storm, but it doesn't cover the bug. Worth adding a case where the numeric generation matches *and* `-observed-generation` is missing (the upgrade case), which actually exercises the re-ack.

2. **[R1-L2] The handshake annotation keys are duplicated as string literals in both services.** `workflow-recipes/src/reconciler/mcpDelegation.ts:794` + `host-context-controller/src/reconciler.ts:79`. All three keys (`pre-deploy`, `network-ready`, `network-ready-observed-generation`) live as independent literals per service with no shared package — this PR extends the pattern with the third key. A future typo on one side breaks the gate silently (never fresh → 30s timeout → proceeds degraded, nothing goes red). A shared constant or a cross-service equality test would close it.

3. **[R1-L3] The annotation doc is incomplete.** `docs/architecture/platform-topology.md:878` says `network-ready` is "valid only together with a matching generation stamp," but the code tolerates a non-numeric generation as fresh without a stamp (`mcpDelegation.ts:952`, `reconciler.ts:1507`). Harmless (that branch is unreachable in a real cluster — `metadata.generation` is always an integer), but the statement is incomplete.

## What checked out clean

- **Freshness core** (`mcpDelegation.ts:949-956`): conservative — a gen-N ack against a spec at N+1 is rejected. Types cross correctly (`=== String(currentGeneration)`).
- **Mirror of `waitForExternalEgressReady`** (`:998-1001`): the new check faithfully replicates the condition-based freshness pattern, line for line.
- **HCC re-ack guard** (`reconciler.ts:1939-1946`): all four cases correct (first ack, stale, match → no re-ack, non-numeric → legacy). No oscillation.
- **No stamp-induced write-storm**: `mcpserver.yaml:457-458` declares `subresources.status`, so writing the annotation via merge-patch doesn't bump `metadata.generation` — the guard converges after a single write.
- **Drop removes the full pair** (`:482-483`), no orphan; the consumer requires `network-ready === "true" && ackFresh`, so an orphan is ignored fail-closed.
- **Contract**: keys match byte-for-byte across both services; HCC is the sole producer, WRC a pure consumer; no out-of-sync third parties anywhere in the repo. Version skew safe both ways.
- **Security: SAFE** — the "non-numeric = fresh" branch isn't exploitable (`metadata.generation` is system-managed; annotation writers are already inside the RBAC boundary). External-egress path stays fail-closed and untouched.
- **Suites**: HCC `reconciler.test.ts` 36 passed, WRC `mcpDelegation.test.ts` 90 passed. T3 against `ac6972ab`: the new-behavior tests fail red as they should (they cover the bug); `tsc` green both packages.

alfredolopez80:
Follow-up @claude: rather than defer the 7b/9a convergence, I pulled your #413 fix into this PR (it's this PR that activates the flap, so it belongs here). Cherry-picked `85faf62c` as `2d2adfee` (your authorship preserved) and ran the verification your issue branch couldn't: **`tsc` green, `specHash.test.ts` 12/12** (your convergence + drift-guard tests), **full `workflow-recipes` suite 2480 passed / 0 failed**. With 7b/9a converged, this PR's ack-drop is a steady-state no-op (only runs on a real spec change). PR body updated with `Fixes #413`, the corrected R1 framing, and the version-bump note. Thanks again for the catch.

alfredolopez80:
@claude — requesting a **final closing review**. Every point from your review is now addressed **in this PR** (not deferred), so please confirm merge-readiness or flag anything remaining.

**What changed since your review:**

1. **7b/9a convergence — fixed here (was your substantive concern).** Cherry-picked your `#413` fix (`85faf62c` → `2d2adfee`, your authorship preserved): `clerum.io/pre-deploy` is excluded from `computeSpecHash` via `HASH_EXCLUDED_ANNOTATIONS`, so step 7b and step 9a build identically-hashing manifests → the idempotency gate converges to **zero PUTs in steady state**, which makes this PR's ack-drop a steady-state no-op (it only runs on a real spec change). `pre-deploy` still persists on the live object via the replace merge (HCC's re-ack guard requires it); only the hash excludes it. The misleading Step 9a "remove pre-deploy annotation" comment is corrected, plus a drift-guard test tying `HASH_EXCLUDED_ANNOTATIONS` to `PRE_DEPLOY_ANNOTATION`.
   - **Verification your issue branch couldn't run:** `tsc` green; `specHash.test.ts` **12/12** (your convergence + drift-guard tests); full `workflow-recipes` suite **2480 passed / 0 failed**.

2. **R1 evidence — retracted and corrected.** You were right: my R1 was on a `streamableHttp`/`managed:false` server that never enters `waitForNetworkReady` nor receives the ack, so it never exercised the drop path. PR body now states the real severity (fail-open, +1–3s typical / +30s ceiling on generic-stdio readiness) and no longer claims "zero write-churn."

3. **Minor notes — done.** Ack carry-over comment (`5fe3e10d`); version-bump provenance documented in the PR body (`.githooks/pre-commit version:staged`); the cross-controller integration-test gap is folded into the T1/E2E confirmation in the PR body's follow-up.

**Net state of the PR:** the original #408 fix (generation-aware handshake) plus the #413 convergence that makes it a no-op in steady state; unit suites green (WRC 2480/0, HCC 1088/0), `tsc` green, mutation-validated. Local `make minikube-t2` reached T0 PASS / T2-preflight PASS / T1 Real PostgreSQL 92/92 passed (the one unrelated skip is filed as #412).

Please do a final pass over `2d2adfee` (the convergence change) and the corrected framing, and let me know if it's good to mark ready-for-review / merge, or if anything still needs work.

alfredolopez80:
Thanks @zach88 — thorough trace, and R1-H1 is exactly right. All four are addressed; here's the point-by-point.

## R1-H1 (must-resolve) — resolved in this PR

You diagnosed the 7b/9a flap precisely: `preDeployMcpServers` stamps `pre-deploy`, `delegateTransportWorkloads` doesn't, `computeSpecHash` hashed annotations → permanent hash divergence → a `replace` every pass, and (post-this-PR) the ack-drop turned that into a present→absent→present ack flap.

This is now fixed **in the PR** (commit `2d2adfee`, your option **(c)** in the "exclude annotations from the spec-hash" variant): `computeSpecHash` excludes `clerum.io/pre-deploy` (`HASH_EXCLUDED_ANNOTATIONS`), so 7b and 9a hash identically → the idempotency gate returns `specHashUnchanged === true` → **zero PUTs in steady state**, which also makes the ack-drop a steady-state no-op (it only runs on a real spec change, as intended). `pre-deploy` still persists on the live object via the replace merge, so HCC's re-ack guard (which keys off it) is unaffected — excluded from the hash, not the object.

On your (b) vs (c): I went (c) because (b) can't actually see the generation change pre-PUT (the replace is what would bump it), and (b) leaves the two PUTs/pass intact — the `resourceVersion` churn that feeds HCC's watch-storm, which is the gate's whole reason to exist. (c) kills both at the root. Swept the manifests: `recipe-bindings` is identical on both paths, `spec-hash` and `pre-deploy` are excluded — no other divergent annotation remains. Verified locally: `specHash.test.ts` 12/12 (convergence + drift-guard), full `workflow-recipes` suite 2480/0.

Detail write-up in #413 (closes with this PR via `Fixes #413`).

## R1-L1 — fixed (commit `21881506`)

Added the upgrade-path test: `network-ready:'true'` + `generation: 3` + **no** `observed-generation` → must re-ack. It's red against the parent (old guard skips on `network-ready==='true'`) and exercises the `stamp !== String(gen)` branch (`undefined !== '3'`) that nothing covered. (Minor note: the existing `re-acks when the stamped generation is stale` test at `:178` *is* already red-vs-parent — but you're right that the stamp-absent branch was the uncovered one.)

## R1-L2 — real, filed as follow-up #414

The three keys are duplicated literals per service and a typo degrades silently (30s fail-open). Both services already depend on `@clerum/network-policy-core`, which is the right home — but moving them touches two service builds + the shared package, and an in-repo cross-service equality test breaks build isolation (separate npm packages). Filed as **#414** rather than expanding this PR further.

## R1-L3 — fixed (commit `21881506`)

`platform-topology.md` now states the `network-ready` cell is honored only with a matching generation stamp, **and** that a non-numeric generation is tolerated as fresh without a stamp (the defensive branch), matching the code.

Net: R1-H1 closed at the root here, R1-L1/L3 fixed, R1-L2 tracked in #414. Happy to take another look if anything's still off.

alfredolopez80:
@claude — **final closing review please**, now that both your review and @zach88's are fully addressed in-PR. Please validate the complete state and confirm merge-readiness (or flag anything remaining).

## Full state since the last reviews

**Your substantive concern + @zach88's must-resolve (same 7b/9a ack-flap) — fixed at the root in this PR** (commit `2d2adfee`, your #413 branch cherry-picked, authorship preserved). `computeSpecHash` excludes `clerum.io/pre-deploy` (`HASH_EXCLUDED_ANNOTATIONS`) → step 7b and step 9a hash identically → the idempotency gate converges to **zero PUTs in steady state**, making this PR's ack-drop a steady-state no-op (only runs on a real spec change). `pre-deploy` still persists on the live object (HCC's re-ack needs it); only the hash excludes it. Chose this over @zach88's alternative (drop the ack only when generation changed) because that can't see the change pre-PUT and leaves the 2-PUT/pass churn that feeds HCC's watch-storm; the hash convergence kills both. Misleading Step 9a comment corrected; drift-guard test added.

**@zach88's optional notes — addressed** (commit `21881506`):
- R1-L1: added the HCC upgrade-path test (`network-ready:'true'` + numeric generation + **no** stamp → must re-ack) — red against the parent, covers the `stamp !== String(generation)` branch nothing exercised before.
- R1-L3: `platform-topology.md` now documents that the ack is honored only with a matching generation stamp **and** that a non-numeric generation is tolerated as fresh without one (the defensive branch).
- R1-L2 (duplicated handshake-key literals across two services): real but cross-package; filed as follow-up **#414** (move the keys into `@clerum/network-policy-core`) rather than expanding this PR.

**Your earlier minor notes — done:** ack carry-over comment (`5fe3e10d`); R1 evidence retracted/corrected (it was measured on a `managed:false` server that never exercises the gate); version-bump provenance documented.

## Current verification (this branch, HEAD `21881506`)

- `tsc` build green (WRC + HCC).
- `workflow-recipes`: **2480 passed / 0 failed**; `specHash.test.ts` 12/12 (7b/9a convergence + drift-guard).
- `host-context-controller`: #408 suite green incl. the new upgrade-path test (**4 passed**); full suite was 1088/0 before these test-only additions.
- Local `make minikube-t2`: T0 PASS, T2-preflight PASS, T1 Real PostgreSQL **92/92 passed** (one unrelated skip filed as #412).

Scope in this PR: #408 (generation-aware handshake) + **Fixes #413** (7b/9a convergence). Please do the final pass and let me know if it's good to mark ready-for-review / merge.

alfredolopez80:
@claude — one more validation pass to **confirm the complete PR**, please. Since your last closing review (which said merge-ready) I added one commit that closes the residual you'd flagged, so I want your sign-off on the full state before I ask @zach88 for a re-review.

## What changed since your closing review: `857e6b93`

Your closing review noted one non-blocking residual — cross-controller coverage still unit-level, with the handshake keys duplicated as literals per service (also @zach88's R1-L2 / #414). I closed that at the **contract** level: moved the three keys (`pre-deploy`, `network-ready`, `network-ready-observed-generation`) into `@clerum/network-policy-core` (already the home of the egress-fqdn annotation contract, already a dependency of both services). WRC re-exports them so existing consumers are unchanged; HCC imports them (dropping its local literals); `specHash`'s `HASH_EXCLUDED_ANNOTATIONS` uses the shared `PRE_DEPLOY_ANNOTATION` too (no more hardcoded literal). **A key divergence is now a compile error, not a silent 30s fail-open drift** — a stronger guarantee than a cross-service test. `Fixes #414`.

Please check specifically:
1. The re-export in `mcpDelegation.ts` preserves every existing consumer (e.g. `specHash.test.ts` still imports `PRE_DEPLOY_ANNOTATION` from `./mcpDelegation`) — no behavior change, keys are byte-identical to before.
2. `HASH_EXCLUDED_ANNOTATIONS = [PRE_DEPLOY_ANNOTATION]` (from `@clerum/network-policy-core`) still excludes pre-deploy from `computeSpecHash` exactly as before — the #413 convergence is intact.
3. `@clerum/network-policy-core` is CJS with `index.d.ts` types; the three new `export declare const` entries type-check the imports in both TS services.
4. Nothing else in the diff shifted the #408 gate / HCC re-ack / #413 ordering.

## Full verification (HEAD `857e6b93`)

- `tsc` build green: WRC + HCC.
- `workflow-recipes` suite: **2480 passed / 0 failed** (incl. `specHash.test.ts` convergence + drift-guard).
- `host-context-controller`: #408 suite green incl. the new upgrade-path test.
- `@clerum/network-policy-core` self-test: **42 passed / 0 failed**.
- Local `make minikube-t2` (earlier HEAD): T0 PASS, T2-preflight PASS, T1 Real PostgreSQL **92/92 passed**.

## Full disposition
- #408 generation-aware handshake — the fix.
- **Fixes #413** — 7b/9a spec-hash convergence (your + @zach88's must-resolve), root fix.
- **Fixes #414** — shared handshake constants (this commit), your residual + @zach88's R1-L2.
- #412 — isolated-postgres for the migration T1 suite — unrelated, tracked separately.
- @zach88's R1-L1 (upgrade test) + R1-L3 (doc) — done.
- R1 evidence retracted/corrected.

If you agree the whole PR is good, please say so clearly — I'll then ask @zach88 for a re-review to move toward approval.

alfredolopez80:
@zach88 — re-review please. All four of your points are resolved **in-PR**, and I took your R1-H1 root fix rather than deferring it. Quick map to the commits:

- **R1-H1 (must-resolve) — the 7b/9a ack-flap.** Fixed at the root via your option (c): `computeSpecHash` excludes `clerum.io/pre-deploy` (`HASH_EXCLUDED_ANNOTATIONS`) so 7b and 9a hash identically → the idempotency gate converges to zero PUTs in steady state → the ack-drop is now a steady-state no-op (runs only on a real spec change). I went (c) over (b) because (b) can't see the generation change pre-PUT and leaves the 2-PUT/pass `resourceVersion` churn that feeds HCC's watch. Commit `2d2adfee`, `Fixes #413`.
- **R1-L1 — HCC upgrade-path test.** Added: `network-ready:'true'` + numeric generation + **no** stamp → must re-ack, exercising the `stamp !== String(generation)` branch nothing covered (red against parent). Commit `21881506`.
- **R1-L3 — doc.** `platform-topology.md` now documents the generation-stamp requirement **and** the non-numeric-tolerated-as-fresh branch. Commit `21881506`.
- **R1-L2 — duplicated key literals.** Went past a test: moved the three keys into `@clerum/network-policy-core` (shared by both services), so WRC and HCC import the same constant — **a divergence is now a compile error, not a silent 30s fail-open**. Commit `857e6b93`, `Fixes #414`.

Verification at HEAD `857e6b93`: `tsc` green (WRC+HCC), `workflow-recipes` 2480/0, HCC #408 suite green incl. the new upgrade test, `@clerum/network-policy-core` self-test 42/0. Local `make minikube-t2`: T0 PASS, T2-preflight PASS, T1 real-postgres 92/92.

@claude re-reviewed the full state and called it merge-ready. Would appreciate your re-read to see if it's approvable now — and thanks again for the R1-H1 catch, it was the real one.

zach88:
Took another pass on this now that it's grown to cover #413 and #414 alongside the original #408 fix. Good news: the three commits in this round close all four round-1 findings at the root, and the collateral pass found zero regressions. The convergence fix (excluding `pre-deploy` from `computeSpecHash` so 7b/9a hash identically) is the right call and lands cleanly. What's left is all test-quality stuff, nothing that gates the merge.

## Verdict

🟡 **Approve with nits**
- **Optional / non-blocking: 3** (1 Medium, 2 Low)

None. Zero blockers, zero open regressions.

## Optional (non-blocking)

The PR is mergeable without these — they're follow-up / your call.

1. **[R2-M1] Convergence test uses a hand-built fixture instead of the real producer** — `workflow-recipes/src/reconciler/specHash.test.ts:66-77`. The "ignores the pre-deploy handshake annotation" test hand-rolls `nineA`/`sevenB` and only proves `computeSpecHash` ignores `pre-deploy`. But the actual 7b/9a convergence (the load-bearing claim behind R1-H1) depends on `preDeployMcpServers` adding *only* `pre-deploy` on top of the same `buildMcpServerManifest` that `delegateTransportWorkloads` uses — and that's guaranteed today only by reading the code, not by the test. If 7b ever injects another annotation/label, or `buildMcpServerManifest` goes non-deterministic, 7b and 9a diverge again and the perpetual flap comes back *without turning this test red*. Convergence is correct today (full suite green); this is a future-coverage gap. Suggested fix: derive both manifests from the real producer — call `buildMcpServerManifest(...)` for 9a, apply the same `pre-deploy` injection `preDeployMcpServers` does for 7b, then assert `computeSpecHash(sevenB) === computeSpecHash(nineA)`.

2. **[R2-L1] Tests hardcode the annotation literals instead of importing the shared constants** — `workflow-recipes/src/reconciler/mcpDelegation.test.ts:1350-1353, 2075-2160`. A bit ironic given the whole point of the refactor is "single source of truth so a divergence is a compile error": these tests re-introduce hand-typed copies of `clerum.io/network-ready`, `clerum.io/network-ready-observed-generation`, `clerum.io/pre-deploy`. If a constant's value ever changed, the test would keep asserting the old key silently. Fix: import them from `@clerum/network-policy-core` (or the `mcpDelegation` re-export), the way `specHash.test.ts` already does for `PRE_DEPLOY_ANNOTATION`.

3. **[R2-L2] Drift-guard assertion is basically tautological** — `workflow-recipes/src/reconciler/specHash.test.ts:82-84`. `expect(HASH_EXCLUDED_ANNOTATIONS).toContain(PRE_DEPLOY_ANNOTATION)` — `HASH_EXCLUDED_ANNOTATIONS` is literally `[PRE_DEPLOY_ANNOTATION]` and both sides come from the same import, so it can only fail if someone empties the list. Harmless but near-zero value; if you fix R2-M1 that already covers the real invariant.

## What checked out clean

- **Security: safe.** The fix is strictly hardening — `waitForNetworkReady` now rejects a stale ack from a prior generation, closing the false-positive. Doesn't touch authn/authz, secrets, NetworkPolicy YAML, or RBAC. The non-numeric-generation tolerance isn't exploitable: `metadata.generation` is apiserver-managed (not writable by an actor), and the generic gate is fail-open anyway.
- **HCC clean.** `server.generation` is populated from `metadata.generation` in both real producers (LIST + WATCH); the CRD has a status subresource so generation only advances on spec changes → no write-storm in steady state. `setNetworkReadyAnnotation` uses `merge-patch+json`, so `pre-deploy` and other annotations survive the ack.
- **Contract end-to-end.** Single `index.cjs` entrypoint (no ESM/CJS drift possible), both consumers declare the `file:` dep and resolve, `tsc --noEmit` clean on both. All three version-skew claims in the body verified by tracing: WRC-new + HCC-old = +30s fail-open (doesn't break), WRC-old + HCC-new = inert, `generation` nullability handled symmetrically on both sides.
- **No regressions.** `computeSpecHash` stays pure (deep clone before the `delete`, confirmed by running) and idempotent; every caller traced — only the McpServer carries `pre-deploy`, StatefulSet/Deployment hashes unchanged.
- **Repro tests are genuine (T3).** The new #408 tests fail against the parent `ac6972ab` and pass at head (HCC: base worktree → 3 failed | 34 passed). They catch the bug rather than pinning it.
- **Suites:** WRC `npm test` → 2480 passed / 29 skipped, `tsc` clean. HCC `reconciler.test.ts` → 37 passed, `tsc` exit 0. network-policy-core self-test → 42/42.

One operational note (not a finding): the body's recommended deploy order **HCC first, then WRC** is correct — in the skew window, WRC-new against HCC-old pays the full 30s timeout per reconcile on generic transport servers until HCC updates. Fail-open, doesn't break, just +30s latency.

zach88 (approved):

PR #416, opened 2026-08-19
Fix(#391): pipeline-owned HCC Recreate recovery + e2e undo path

### LINKED ISSUE #391: Auto rollout-undo in the deploy pipeline (mitigate HCC botched-rollout outage)
From PR #382 (decision D1b). With `strategy: Recreate` + `replicas: 1`, a botched HCC rollout (new pod never certifies — bad image, or a dead watch) leaves the control plane down until a MANUAL `kubectl rollout undo`. `progressDeadlineSeconds: 1200` is detection-only. Add automatic `kubectl rollout status --timeout` + `rollout undo` on failure in the keyper-labs/evenfire-infra deploy workflow, so a stuck rollout self-heals instead of becoming an outage. Priority: P1 (closes the worst mode of D1 without touching the single-writer model). Measured healthy rollout window is 2s (e2e-hcc-rollout-readiness).

### LINKED ISSUE #412: T1 local: db.realPostgresMigration suite needs an isolated PostgreSQL (fails dropping in-use runtime roles on shared control-postgres)
## Summary

The control-api T1 "Real PostgreSQL" lane suite `control-api/test/db.realPostgresMigration.integration.test.ts` **cannot pass in the local minikube T1/T2 gate** (`make minikube-t2`) because its `beforeAll` drops the runtime roles that are **in use by the live shared `control-postgres`**. The suite needs an **isolated/throwaway PostgreSQL** to drop and recreate cluster-global roles; against the shared cluster database it fails deterministically.

This is a **local T1 harness limitation**, independent of any service code under test. It is separate from PR #275 (which fixes deploy-script fail-loud/timeout bugs — `sync-auth-key.sh`, `verify-gfs.sh`, `full-setup.sh` — and does not touch the Real PostgreSQL migration lane).

## Symptom

Running `make minikube-t2` reaches `T1: Real PostgreSQL suites` and the control-api lane reports:

```
numTotalTestSuites: 40, numPassedTestSuites: 38, numFailedTestSuites: 2,
numTotalTests: 100, numPassedTests: 92, numFailedTests: 0, numPendingTests: 8, success: false
```

The 8 pending tests are all in `db.realPostgresMigration.integration.test.ts`. Per the T1 fail-loud contract in `AGENTS.md` (a Real PostgreSQL run must fail rather than silently skip), the whole gate fails with exit non-zero even though **92/92 real-postgres tests pass with 0 failures**.

## Root cause (verified empirically)

`control-api/test/db.realPostgresMigration.integration.test.ts:130-135`:

```ts
beforeAll(async () => {
  adminPool = new Pool({ connectionString: adminUrl })
  await adminPool.query(`DROP ROLE IF EXISTS ${runtimeRoles.join(', ')}`)  // control_api_runtime, trace_maintenance_runtime, workflow_recipes_runtime
  await adminPool.query(`CREATE DATABASE ${quoteIdent(database)}`)
  dbPool = new Pool({ connectionString })
})
```

Run against the shared live `control-postgres`, the `DROP ROLE` fails:

```
error: role "control_api_runtime" cannot be dropped because some objects depend on it
```

because those roles own live objects in the running services' databases. The suite is written to run against a **dedicated postgres** with no live dependents (as in CI), where dropping/recreating cluster-global roles is safe.

## What it is NOT

It is **not** a missing `CONTROL_API_REAL_PG_ADMIN_URL`. The T1 runner already provides it (`scripts/e2e/minikube-real-postgres.sh:235` derives `ADMIN_DSN` from the `control-postgres` Secret + a profile-owned port-forward and passes it at `:134`). Verified:

- A debug test proves `process.env.CONTROL_API_REAL_PG_ADMIN_URL` reaches the vitest process (`realPostgres.requirement.ts` does not throw).
- Running the suite manually with a verified non-empty `ADMIN_URL` against a live-postgres port-forward still yields `8 skipped` and the `role ... cannot be dropped` error.

So wiring/persisting `ADMIN_URL` does not fix this.

## Reproduction

```bash
# with a healthy branch-owned profile already up (control-postgres Running)
PGU=$(kubectl --context <profile> -n control-plane get secret control-postgres -o jsonpath='{.data.POSTGRES_USER}' | base64 -d)
PGP=$(kubectl --context <profile> -n control-plane get secret control-postgres -o jsonpath='{.data.POSTGRES_PASSWORD}' | base64 -d)
kubectl --context <profile> -n control-plane port-forward svc/control-postgres 35972:5432 &
cd control-api
CONTROL_API_REAL_PG_ADMIN_URL="postgresql://$PGU:$PGP@127.0.0.1:35972/postgres" \
  CONTROL_API_REAL_PG_REQUIRED=1 npx vitest run db.realPostgresMigration
# -> 8 skipped; beforeAll: role "control_api_runtime" cannot be dropped because some objects depend on it
```

## Proposed fix (harness only — no service code change)

Give the migration lane an **isolated PostgreSQL** in the local T1 gate. Options:

1. Spin up a throwaway/dedicated `postgres` instance (a separate pod or an ephemeral container) for the migration lane, and point `CONTROL_API_REAL_PG_ADMIN_URL` at it — never the shared `control-postgres`.
2. Or make the suite's `beforeAll` resilient to in-use roles when it must reuse a shared server: `DROP OWNED BY <role>` / reassign before `DROP ROLE`, guarded so it only touches its own test roles/database. (Riskier on a shared server; option 1 is cleaner.)

Either way, the T1 lane should either run the migration suite against an isolated server or explicitly exclude it from the shared-server local lane (documented, not silently) — consistent with the fail-loud contract in `AGENTS.md`.

## Impact / severity

- **Severity: Medium** — blocks a green `make minikube-t2` locally even when everything else passes; masks nothing about product correctness (0 test failures). CI (dedicated postgres) is unaffected.
- **Scope:** local T1/T2 developer tooling only. Development-only, per the `AGENTS.md` T0/T1/T2 contract.

## Relationship to other work

- **Not covered by #275** (deploy-script fail-loud/timeout fixes; different files, different lane).
- Surfaced while validating the WRC↔HCC generation-aware network-ready fix (issue #408) with a local `make minikube-t2` run; unrelated to that reconciler change.

alfredolopez80:
## Summary
- Keep Recreate, replicas 1, and `progressDeadlineSeconds: 1200` (detection only). Recovery is owned by evenfire-infra deploy workflows via `rollout undo --to-revision`, not by Kubernetes and not by HCC runtime.
- Add `EXPECT_RECOVERY=1` so the rollout e2e proves undo as the **test action**, not cleanup `set image`. `EXPECT_STUCK` and `EXPECT_RECOVERY` are exclusive.
- Pin the D1b premise set in unit tests and execute fail-closed guards in CI (not grep-only).
- Unblock local T1: role-reset Real PostgreSQL suites run on throwaway `postgres:16-alpine`, never shared `control-plane` `control-postgres` (#412).
- Align the T0/T1/T2 orchestrator with the documented contract (planner vs T2 verdict). That harness is separate evidence from this product e2e.

Closes #391
Closes #412

### Offline / CI (PR 416 HEAD including `7e647c19`)
- [x] `npx vitest run src/deploymentStrategy.test.ts` (host-context-controller) — 8/8
- [x] `bash scripts/tests/test-hcc-rollout-readiness-gate.sh` — 16/16
- [x] `bash scripts/tests/test-hcc-rollout-readiness-guards.sh` — 10/10 (gate is executed, not grepped)
- [x] CI public: contract + executable guards wired in `.github/workflows/ci-public.yml`
- [x] `make test-e2e-hcc-rollout-readiness` entry point pins ack, `profile=context`, exclusive modes

### Product e2e live (#391) — owned profile `clerum-feat-issue-391-hcc-auto-rollout-undo-58dc455e`
Marker `gitHead=7e647c19`, gate `e2e-hcc-rollout-readiness`. Ran **after** coverage commit, **before** the `origin/dev` sync merge. `WITH_SYNTHETIC_FLEET=0` because the default 24/115/8 fleet materializes enough McpServer pods to exhaust the profile's 110-pod allocatable and leave HCC `Pending` (not `ImagePullBackOff`). That is a profile-capacity/harness limit, not a Recreate or undo defect.

- [x] `EXPECT_STUCK=1 EXPECT_RECOVERY=0 WITH_SYNTHETIC_FLEET=0` — **8/8 PASS** (ErrImagePull, 206×503, zero 503→200, 210s streak > 120s budget)
- [x] `EXPECT_STUCK=0 EXPECT_RECOVERY=1 WITH_SYNTHETIC_FLEET=0` — **12/12 PASS** (`rollout undo --to-revision=32`, image restored, revision 33→34, new pod uid, 29-sample hold with zero 503)

### T0/T1/T2 harness (separate lane)
- [x] Canonical `make minikube-t2` on HEAD `8d639ecd` before the coverage commit: T0=PASS T1=PASS T2=PASS (Playwright=NOT_RUN)
- [ ] Do not treat this as live on clerum-dev until the evenfire-infra PR is merged

### Sync
- [x] Branch rebased/merged with current `origin/dev` (`fee7c037` merge commit; 0 commits behind `origin/dev`)

## Review
Please do a full review of product #391 / PR 416 only (HCC Recreate detection vs evenfire-infra undo, e2e STUCK/RECOVERY, guards, Makefile/CI). T0/T1/T2 harness docs are out of the product-recovery scope unless they regress the e2e gate.


alfredolopez80:
@claude please do a complete review of this PR (evenfire#391 / PR 416).

Focus on the product contract, not the Minikube T0/T1/T2 harness:

1. Detection vs recovery split: Recreate + replicas:1 + progressDeadlineSeconds:1200 is DETECTION ONLY; recovery is `kubectl rollout undo --to-revision` owned by evenfire-infra, never Kubernetes self-heal and never HCC application code.
2. E2E `EXPECT_RECOVERY=1` must prove undo as the test action (not cleanup `set image`), refuse a no-op undo, keep IMAGE_BROKEN set unless every post-undo proof passed, and require the same sustained D1b outage as EXPECT_STUCK before undo.
3. Executable guards (`test-hcc-rollout-readiness-guards.sh`) must actually run the gate; grep-only contract is not enough.
4. Live evidence on the branch-owned profile (HEAD `7e647c19`, before the origin/dev sync merge):
   - EXPECT_STUCK=1 WITH_SYNTHETIC_FLEET=0 → 8/8 PASS (ErrImagePull, sustained 503, zero 503→200)
   - EXPECT_RECOVERY=1 WITH_SYNTHETIC_FLEET=0 → 12/12 PASS (undo --to-revision=32, image/revision/uid/hold)
   Default synthetic fleet (24/115/8) overflowed this profile's 110 allocatable pods and is a harness-capacity issue, not a product-contract failure.
5. Confirm the branch is synced with origin/dev (merge `fee7c037`) and that CI on the new HEAD is the one to trust for unit/guards.

Please review the full diff vs `dev`, not only the latest commit.

alfredolopez80:
@claude please do a complete re-review of this PR (evenfire#391 / PR 416) against current origin/dev.

Review the full three-dot diff vs dev, not only the tip commit. HEAD is now cadfea82 (fix: restore T2 port-forward conflict detection) on top of the previously reviewed fee7c037 sync.

Focus:

1. Detection vs recovery: Recreate + replicas:1 + progressDeadlineSeconds: 1200 stay DETECTION ONLY. Recovery is kubectl rollout undo --to-revision owned by evenfire-infra, never Kubernetes self-heal and never HCC application code.
2. E2E EXPECT_RECOVERY=1 must prove undo as the test action (not cleanup set image), refuse a no-op undo, keep IMAGE_BROKEN set unless every post-undo proof passed, and require the same sustained D1b outage as EXPECT_STUCK before undo. The two flags stay exclusive.
3. Executable guards (test-hcc-rollout-readiness-guards.sh) must actually run the gate; grep-only is not enough.
4. New since the last review: t2_process_check in scripts/minikube/t2-common.sh. Confirm PORT_FORWARD_CONFLICT can fire again (default IFS so UID/PID/PPID split; awk matches argv0 kubectl / .../kubectl port-forward, not a wrapper that only mentions those words). Contract + scenario pins must stay fail-loud.
5. #412 isolation remains a separate T1 lane (throwaway postgres:16-alpine, never shared control-postgres). Do not treat T1 or T2 harness green as #391 undo proof on clerum-dev (evenfire-infra PR 52 owns that).

Please list any remaining merge blockers. This is a re-review request only; do not merge.

zach88:
Took a deep pass through this one. The product change is tiny (the HCC manifest edit is comment-only, fields are untouched), so almost all the risk lives in the bash harness. The security surface, the STUCK/RECOVERY exclusivity, the #412 postgres isolation, and every body claim (8/8, 16/16, 10/10, "guards executed not grepped", CI wiring) all check out. The one real problem is concentrated in the T2 port-forward conflict detector: the commit that says it restores that check actually leaves it fail-open, and a new test locks the broken behavior in as green.

### 🔴 Not ready to approve
- **Must resolve before approval: 2**
- **Optional / non-blocking: 6**

1. **[R1-B1] Port-forward conflict detector is fail-open** (`scripts/minikube/t2-common.sh:533`) — *regression from `cadfea82`* ("fix: restore T2 port-forward conflict detection"). The new awk filter requires `port-forward` to sit immediately after `kubectl` (`kubectl[[:space:]]+port-forward`) and anchors on a BSD/macOS TIME column. Two failure modes, both fail-open:
   - **macOS:** every real launcher puts flags between the two tokens (`kubectl --context=… -n … port-forward …` in `pf-all-stack.sh:139`, `pf-control-stack.sh:20`, the `run-e2e-*.sh` scripts), so none match. The only thing that matches is the test's fabricated form.
   - **Linux:** the `HH:MM:SS` ps format (`00:00:00`) never satisfies the TIME anchor, so it matches zero lines regardless of command shape.

   **Repro:** ran all four command forms through the exact head regex. On BSD `ps` only the fixture form matched (the 3 real launcher forms did not); on Linux `ps` nothing matched at all. The merge-base filter (`/[p]ort-forward/ && /kubectl/`) matched all four. So worktree B running `pf-all-stack.sh` on the shared profile is invisible to worktree A's `t2_process_check`, which then shares/collides the port-forward — exactly what the guard exists to stop. Blast radius is the local T2 harness, not production, but it silently corrupts the T2 evidence this PR is about.

   **Fix:** drop the adjacency requirement and the TIME anchor. Match `kubectl` (argv0 basename) plus a standalone `port-forward` token on the argv, or go back to the merge-base `/kubectl/ && /port-forward/` plus the argv0/`comm` hardening.

2. **[R1-H1] Test fixture certifies the dead guard as green** (`scripts/tests/test-minikube-t2-scenarios.sh:246-257`) — *regression from `cadfea82`*. The `bare_pf`/`path_pf` fixtures place `--context` after `port-forward` (`kubectl port-forward --context=… svc/x`), which is the one shape the broken regex matches. The real producer (`pf-all-stack.sh`) emits `--context`/`-n` before the verb, so the fixture isn't derived from it, and the suite passes with the bug present.

   **Repro:** the suite runs green today against the broken guard. Fix it together with R1-B1: build the conflict fixture from the real launcher form (`kubectl --context=<ctx> -n <ns> port-forward svc/<x> <ports>`) and assert `t2_process_check` fails with `PORT_FORWARD_CONFLICT`. That test must fail against the current head, and go green once B1 is fixed.

## Optional (non-blocking)

1. **[R1-M1] Duplicated outage-proof logic** (`scripts/e2e/e2e-hcc-rollout-readiness.sh:434-471`). The 6-assertion D1b outage block is copy-pasted near-verbatim from the `EXPECT_STUCK` branch (`:392-433`) into the `EXPECT_RECOVERY` branch. Two copies of the same contract will drift. Extract a shared `assert_botched_outage()` and let each branch keep its own action (STUCK: nothing, RECOVERY: `rollout undo`). This also folds in the R1-L2 nit below.
2. **[R1-L1] `T2_PLAN_MODE` defaulted twice** (`scripts/minikube/t2-common.sh:46-47` vs `:76`). Harmless, just drop one.
3. **[R1-L2] `grep -cxF … || true` swallows a kubectl failure** (`scripts/e2e/e2e-hcc-rollout-readiness.sh:445,455-457`). On a kubectl error the empty input makes `grep` print `0`, so the "old pod terminated" assert passes on a query failure. Sibling of an existing idiom in the STUCK branch, inherited by the copy. Best fixed inside the R1-M1 helper.
4. **[R1-L3] Tests assert on manifest comment prose** (`host-context-controller/src/deploymentStrategy.test.ts:99-124`). The three string-match tests guard documentation text, so an equivalent reword breaks CI. This is the intended anti-deletion guard, so fine as-is; flagging only so it's a conscious choice. I verified all seven substrings exist verbatim in the manifest at head, and the `.toEqual` premise-set test (real field values) is the behavioral one.
5. **[R1-L4] Recovery-mode evidence artifact reads `restarts=?` and zeroed `roll_*`** (`scripts/e2e/e2e-hcc-rollout-readiness.sh:487-488,264`). Diagnostics only; real evidence is in the `stuck:`/`post-hold:` lines.
6. **[R1-L5] `LAST_GOOD_REVISION` now gates the healthy path too** (`scripts/e2e/e2e-hcc-rollout-readiness.sh:347-348`). Moved out of the botched branch, so the healthy mode now reads the revision annotation it never used to. Benign (the annotation is always present after a rollout), just a new precondition.

## What checked out clean

- **Security (full diff):** no vulnerabilities introduced. Throwaway `postgres:16-alpine` uses trust-auth with no password, binds `127.0.0.1` only, `--rm` plus a cleanup trap, and never touches the shared `control-postgres`. The new CI job is `pull_request` (not `_target`), with no `${{ }}` in `run:`, no `secrets.*`, no added `permissions`.
- **Body claims all true at head:** `deploymentStrategy.test.ts` 8/8, `test-hcc-rollout-readiness-gate.sh` 16/16, `test-hcc-rollout-readiness-guards.sh` 10/10 (all run). The guards test genuinely executes the gate binary (`env -i … "$BASH_BIN" "$GATE"`, asserts exit≠0 plus the guard needle) rather than grepping for it, and CI wires both guard scripts as plain `run: bash …` steps with no `continue-on-error`/`|| true`, so a non-zero exit blocks merge.
- **EXPECT_STUCK/EXPECT_RECOVERY exclusivity:** defaulted to `{0,1}`, validated, both-set rejected, junk values exit 1, neither-set is the healthy path. Every outage/undo proof routes through `fail()`, so a non-materialized outage or a no-op undo yields an overall FAIL. No fail-open path.
- **#412 isolation:** the two realPostgres suites doing global-role DDL (`pg_authid`/`DROP ROLE`/`ALTER ROLE`) are exactly the ones routed to the isolated lane against a throwaway DSN on an ephemeral port; they're `continue`d out of every shared lane, with an empty-DSN guard and a fail-closed reporter check. No path points them at `control-postgres`.
- **Manifest:** comment-only. `replicas:1` / `Recreate` / `progressDeadlineSeconds:1200` identical to the merge-base. Version bump is the service's own field (`0.1.228 → 0.1.230`), no dependency change.
- `bash -n` clean on all edited scripts.

alfredolopez80:
@claude please do a complete re-review of this PR (evenfire#391 / PR 416) against current `origin/dev`.

Review the **full three-dot diff vs `dev`**, not only the tip commit. HEAD is now `dd755abe` (`fix: background the T1 port-forward as a direct kubectl child`). Branch is 0 commits behind `origin/dev`. Reply on this PR with comments. **Do not merge.**

This is a close-out review: confirm every issue **you** previously flagged, and every item **@zach88** flagged, is either **closed** on this HEAD or still **open** (cite file:line). Do not wave at "harness only" — if a prior finding still fails, say so.

## Your prior findings (must re-check)

From your re-review on `cadfea82` ([comment](https://github.com/evenfire-ai/evenfire/pull/416#issuecomment-5347032872)):

1. **`t2_process_check` TIME-format-dependent / Linux `HH:MM:SS`** — you said the awk TIME anchor matched macOS `N:NN[.nn]` and missed Linux `ps -ef` `HH:MM:SS`, so `PORT_FORWARD_CONFLICT` never fired on Linux. Confirm the current awk is a loose `kubectl` argv0/path token **plus** a later standalone `port-forward` token (flags allowed between them), with **no** TIME-column anchor, and that `test-minikube-t2-scenarios.sh` pins a real Linux `00:00:00` launcher line.
2. **Doc drift (e2e header / cleanup vs manifest)** — you said the e2e still called undo "manual" while the manifest said evenfire-infra owns automatic undo. Confirm `scripts/e2e/e2e-hcc-rollout-readiness.sh` header now names evenfire-infra automatic `rollout undo --to-revision` and that cleanup `set image` remains leftover restore only.

Also re-confirm the product contract you already passed (do not skip it because HEAD moved):

3. Detection vs recovery: Recreate + `replicas:1` + `progressDeadlineSeconds:1200` stay **DETECTION ONLY**. Recovery is `kubectl rollout undo --to-revision` owned by evenfire-infra, never Kubernetes self-heal and never HCC application code.
4. E2E `EXPECT_RECOVERY=1` proves undo as the **test action** (not cleanup `set image`), refuses a no-op undo, keeps `IMAGE_BROKEN` set unless every post-undo proof passed, requires the same sustained D1b outage as `EXPECT_STUCK` before undo, and the two flags stay exclusive.
5. Executable guards (`test-hcc-rollout-readiness-guards.sh`) actually run the gate; grep-only is not enough.
6. #412: role-reset Real PostgreSQL suites stay on throwaway `postgres:16-alpine`, never shared `control-postgres`.

## @zach88 findings (must re-check)

From [zach88's pass](https://github.com/evenfire-ai/evenfire/pull/416#issuecomment-5347942215):

### Must-resolve before approval

7. **[R1-B1] Port-forward conflict detector fail-open** (`t2-common.sh`) — adjacency `kubectl[[:space:]]+port-forward` + TIME anchor. Real launchers put flags between tokens (`pf-all-stack.sh`, `pf-control-stack.sh`, e2e). Confirm that is gone and a foreign **real** launcher still trips `PORT_FORWARD_CONFLICT`.
8. **[R1-H1] Fixture certified the dead guard** — fixtures used `kubectl port-forward --context=…`. Confirm fixtures are derived from real launchers (flags between `kubectl` and `port-forward`, including Linux TIME) and that the suite would fail on the old adjacency regex.

### Optional / non-blocking (still say closed vs open)

9. **[R1-M1]** Duplicated D1b outage block → shared `assert_botched_outage()`.
10. **[R1-L1]** `T2_PLAN_MODE` defaulted twice in `t2-common.sh`.
11. **[R1-L2]** `grep -cxF … || true` treating a kubectl failure as "old pod terminated" (should be fail-closed inside the shared helper).
12. **[R1-L3]** Unit tests matching manifest comment prose (conscious anti-deletion guard — confirm still intended).
13. **[R1-L4]** Recovery evidence `restarts=?` / zeroed `roll_*` (diagnostics only).
14. **[R1-L5]** `LAST_GOOD_REVISION` now read on the healthy path too.

## Verdict format

For each item 1–14: **CLOSED** or **OPEN**, one sentence, file:line.

Then:

- Any **new** merge blocker on current HEAD (product #391/#412 **or** T2 harness that still fail-opens evidence).
- Confirm you reviewed `origin/dev...HEAD`, not only `dd755abe`.
- Do **not** merge. A human must Approve before merge.

zach88:
Took another deep pass through this one for round 2. Good news up front: both of last round's serious findings (the fail-open port-forward conflict detector and the fixture that was certifying the dead guard) are genuinely fixed, and I verified the closes by mutation, not just by reading. The awk detector now matches every real launcher form on both macOS and Linux, and the scenarios fixture derives from the real producer and extracts the real awk program instead of re-copying the regex.

What's holding it back this round isn't correctness so much as blast radius: the new GFS self-heal block (`settle-gfs-reader-rollout.sh`, `gfs-rollout-shim/kubectl`, `wait-gfs-reader-ready.sh`, `converge_gfs_reader_after_restore`) landed as a set of `fix:` commits and it brought two regressions with it, plus it's decision logic that grew without a design to review against. Honestly my strongest recommendation is to pull that GFS self-heal out into its own PR (with a small decision table) and merge the #391 detection/undo work on its own, since that part converges. The PR is at 33 files now and two test files have picked up 13 `fix:` commits each, which is the usual sign the harness design wants a beat before more patches.

### Verdict
🔴 **Not ready to approve**
- **Must resolve before approval: 2** (both open regressions)

1. **[R2-M1] `t2_deployment_check` stopped checking the python3 exit status** (`scripts/minikube/t2-common.sh:475`) — regression from `9e2c0e53`. The rewrite went from `if ! python3 …; then t2_fail PROFILE_UNHEALTHY` to `unready="$(python3 … )"` with only a `[ -n "$unready" ]` check afterward (`:494`), so the python exit code is never evaluated. When `kubectl get deployments -A -o json` emits non-JSON but non-empty stdout (a warning line prepended, or `minikube kubectl` printing "Downloading kubectl…" on first use), `json.loads` throws. Under `set -eo pipefail` (how `t2-preflight.sh:39` calls it, plainly), the failed command substitution aborts the script with a raw exit 1: no `t2_fail`, no stable code, no `T2_NEXT_COMMAND`. The normal valid-JSON-unready path still emits `PROFILE_UNHEALTHY` correctly, so this is only the unreadable-inventory path. Reproduced (shell mechanics). This also breaks the "stable failure codes keep their meanings" invariant in `.cursor/rules/minikube-t2-harness.mdc`. Fix: check the python3 rc explicitly and emit `PROFILE_UNHEALTHY` on the failure path like before.

2. **[R2-M3] the GFS self-heal now runs on every security gate, not just T2** (`scripts/minikube/pre-gate-sync.sh:575` and `:649`) — regression from `9e2c0e53`. That fix made `provision_gfs_serving` run in every `pre-gate-sync` pass, including the "No cluster sync required" fast path that used to be near-noop. `pre-gate-sync.sh --gate` gets invoked once per gate from `run-platform-security-gates.sh:41`, so now every gate in the suite runs `sync-auth-key.sh` + `reconcile-gfs-deploy-credentials.sh` (fail-loud, `exit 1`) + `settle` + `converge`, which can delete CrashLoop pods, scale ReplicaSets, and restart `gfsc-reader`. Net effect: the whole platform-security-gate suite now mutates GFS credentials and can perturb the reader on each gate, and a GFS reconcile failure now aborts any gate, not just T2. Confirmed by reading the call graph. Fix: scope the provision to the T2 gate, or confirm in writing that mutating GFS on every gate is intended.

## Optional (non-blocking)

3. **[R2-M2] the GFS self-heal is decision logic born inside `fix:` commits, and the same rule lives in 3 copies** (`scripts/minikube/settle-gfs-reader-rollout.sh`, `gfs-rollout-shim/kubectl`, `wait-gfs-reader-ready.sh`, `pre-gate-sync.sh:319`). "Delete unready reader pods and judge by `readyReplicas`, not generation" shows up three times with different conditions: `settle` deletes only CrashLoopBackOff when the reader is Ready; `converge` deletes all unready-live pods when it's unready and also does `rollout_restart_with_retry`; `wait` has its own readiness judgment. Worth a half-page decision table (for each reader state — Ready / unready / CrashLoop / terminating / stale-RS — what wins and what happens when HCC rewrites the generation) before stacking more cases on it. This is the same reason I'd suggest splitting it into its own PR.

4. **[R2-M4] the new GFS orchestration is only covered by source-substring asserts, not behavior** (`scripts/tests/test-minikube-t2-contract.sh:60` onward). The tests `grep -Fq`/`grep -c` the script text (e.g. `grep -c 'gfs-rollout-shim' full-setup.sh >= 2`, `grep -Fq 'settle-gfs-reader-rollout.sh'`), which asserts the string exists, not the observable result. It'd stay green if `settle` ran after the reconcile, if the shim's PATH prefix didn't take effect, or if `sync-auth-key.sh` ran in the wrong order. The behavioral tests only cover the leaf scripts in isolation, never the composition. Worth at least one test that runs `provision_gfs_serving` with kubectl/reconcile stubbed and asserts the order (sync-auth-key → settle → reconcile-with-shim-on-PATH → converge).

5. **[R1-M1] the STUCK/RECOVERY outage-proof duplication is still open** (`scripts/e2e/e2e-hcc-rollout-readiness.sh:245-292`), still open since round 1. The `assert_botched_outage()` helper deduped the call sites but reintroduced the same 6 conditions byte-for-byte inside it; the only thing that differs between the `flavor=stuck` and `flavor=recovery` branches is the ok/fail message strings. A future tweak to one condition still won't reach the other branch. Heads up that this is a Medium that survived a fix round, so it either goes up to blocker or gets closed in writing with a rationale, not floated another round. Fix: one assert sequence with the message suffix selected by `flavor`.

6. **[R2-L1] settle fixture doesn't match the real producer form** (`scripts/tests/test-minikube-settle-gfs-reader-rollout.sh:87`). It uses `readyReplicas=0` in the RS rows, but kubectl jsonpath emits empty for `.status.readyReplicas==0` (omitempty), so the real row is `…|1||12`. The code tolerates both via `${rs_ready:-0}` so it hides nothing today, but a regression that dropped the default wouldn't get caught. Add a row with the empty field.

7. **[R2-L2] the shim silently floors the caller's timeout to 600s** (`scripts/minikube/gfs-rollout-shim/kubectl:78`) even when the shared lib passes `--timeout=30s`. Intentional and documented (CrashLoop backoff reaches 5 min), just flagging that it can mask a genuinely dead reader for 10 min in the harness.

8. **[R2-L3] the new `sync-auth-key.sh` calls don't pass `--require-gfs`** (`scripts/minikube/pre-gate-sync.sh:369`, `full-setup.sh:990,1067`) — low confidence. The empty-source guard only fires with `REQUIRE_GFS=true`, so if the platform Secret exists but its `jwt-public-key` is transiently empty while `gfs-config` has a legit key, `sync_gfs` could patch `gfs-config` to empty (the exact thing the commit means to prevent). Not confirmed as a regression since the pre-existing `sync_mcp_host_auth_key` already calls it the same way; worth a look at whether these should carry `--require-gfs`.

9. **[R1-L4 / R1-L5] two round-1 nits still open** in `scripts/e2e/e2e-hcc-rollout-readiness.sh`: the RECOVERY evidence artifact prints `restarts=?` and `rollout: samples=0` (`:334`, cosmetic), and the `LAST_GOOD_REVISION` capture with `die` now also gates the healthy path even though it's only consumed in RECOVERY (`:420`, low risk since the annotation always exists after the first rollout — move it inside the recovery branch).

## What checked out clean

- **R1-B1 (blocker) and R1-H1 (high) are both properly closed**, verified by mutation: the new port-forward awk matches all four real launcher forms on BSD and Linux `ps`, doesn't self-match (`kubectl(` with a paren), and false positives like `grep` fall out via `comm=kubectl` + the profile/context filter. Reverting the awk to the broken form turns `test-minikube-t2-scenarios.sh` red, and the fixture now extracts the real awk program. R1-L1 and R1-L2 also closed.
- **Security: clean, no findings.** The shim only intercepts `rollout status deployment/gfsc-reader` and execs the real kubectl for everything else; it's prepended to PATH only inline in specific subprocesses, never globally. The JWT key is public and its handling lives in `sync-auth-key.sh` (untouched here); the new scripts only touch annotations/metadata, never Secret `.data`. No injection (python patches by positional argv, `--patch-file=/dev/stdin`; `safe_profile` sanitized). Throwaway Postgres still binds `127.0.0.1` only, `--rm`, trust without a password; DSNs redacted and passed by env, not argv. CI is `pull_request`, no `${{}}`/`secrets.*`.
- **Manifest is comment-only** (`deploy/base/control-plane/host-context-controller.yaml`): `replicas:1`/`Recreate`/`progressDeadlineSeconds:1200` identical to the merge-base, 0 non-comment lines, and the 7 asserted comment substrings exist verbatim at head so CI isn't lying.
- **Body claims hold at head**: `deploymentStrategy.test.ts` 8/8, `test-hcc-rollout-readiness-gate.sh` 16/16, `test-hcc-rollout-readiness-guards.sh` 10/10 (it executes the gate under `env -i`, doesn't grep it). STUCK/RECOVERY exclusivity guard is in place.
- **#412 isolation is complete**: the only two tests doing global-role DDL (`db.realPostgresMigration`, `gfsReaderRole`) are exactly the two routed to the isolated lane; no shared-lane test mutates roles. Recovery flow is fail-closed (any pre/post-undo failure leaves `IMAGE_BROKEN=1` so cleanup restores the image; no-op undo = FAIL, covered by an executable test).
- Audited contract changes that don't break callers: `1fa12a26` (T1 verdict from the JSON reporter, strict gates, empty json → `ZERO_TESTS_EXECUTED`), `dd755abe` (T1 port-forward: `t2_kc` is `kubectl --context`, same binary, cleanup kills the right PID), and `provision_gfs_serving` runs exactly once per invocation (mutually exclusive branches, no double reconcile).

alfredolopez80:
@claude Please perform a final full close-out review of PR #416 at the pushed head `41d60fa136caba1d0a1134fb9a63a0acc52c8ec1`.

Review the complete three-dot diff `origin/dev...HEAD`, the PR description, every prior Claude review, and every Zach review/comment. Do not review only the tip commit. Do not merge; this is a final review request.

Please produce the result entirely in English and include a complete issue-preview disposition table with these columns:

| ID/source | Issue preview | Scope | Disposition: FIXED / DEFERRED / DOCUMENTED / OPEN | Evidence: file:line, test or CI job | Follow-up issue/action | Merge impact |
|---|---|---|---|---|---|---|

The table must explicitly cover:

1. PR #391 detection-versus-recovery contract: Recreate, replicas=1, progressDeadlineSeconds=1200, detection-only semantics, and ownership of automatic `kubectl rollout undo --to-revision` by evenfire-infra.

2. PR #391 E2E contract: `EXPECT_STUCK`, `EXPECT_RECOVERY`, sustained D1b outage, undo as the test action, no-op undo rejection, pod UID/revision/image/readiness/stability proofs, `IMAGE_BROKEN` fail-closed behavior, and flag exclusivity.

3. Executable guard coverage: confirm that `test-hcc-rollout-readiness-guards.sh` executes the real gate and is wired into CI, rather than relying only on grep/source assertions.

4. Issue #412: real PostgreSQL role-mutating suites must use throwaway `postgres:16-alpine`, never the shared `control-postgres`; treat this as a separate evidence lane.

5. T0/T1/T2 harness changes: profile ownership, port-forward conflict detection, fail-closed evidence semantics, planner-versus-verdict separation, lock handling, GFS recovery ordering, auth-key convergence, and the fact that T0/T1/T2 evidence is not product undo proof on clerum-dev.

6. Zach R1 findings: R1-B1, R1-H1, R1-M1, R1-L1, R1-L2, R1-L3, R1-L4, and R1-L5.

7. Zach R2 findings: R2-M1, R2-M2, R2-M3, R2-M4, R2-L1, R2-L2, and R2-L3.

8. The previous SOL MAX findings: EXIT-trap restore failure masking, auth ConfigMap/consumer convergence, nonexistent GFS deployment targeting, stale-lock reclaim TOCTOU, and exact PEM trailing-newline handling.

9. All prior documentation/design observations: manual-versus-automatic undo wording, GFS decision-table/documentation requirements, source-only versus executable test coverage, deferred evenfire-infra ownership, and any remaining test/evidence limitations.

For every item, independently determine whether it is actually fixed, intentionally deferred, documented with a valid follow-up, or still open. Do not mark an item as documented unless the repository or PR contains the relevant documentation. If a follow-up issue does not exist, explicitly mark that follow-up as required instead of inventing an issue number.

Also report:

- final verdict: MERGE / MERGE WITH CONDITIONS / NO MERGE;
- all remaining merge blockers;
- all deferred items and their concrete follow-up actions;
- whether the current remote SHA and CI correspond to the audited commit;
- T0/T1/T2 evidence status;
- Health, Playwright/UI/Desktop, and fresh product `EXPECT_RECOVERY=1` E2E status;
- confirmation that evenfire-infra was kept out of scope but recorded as a separate follow-up;
- confirmation that the full `origin/dev...HEAD` diff was reviewed.

Use exact evidence and line references. Do not merge.

zach88:
Took another deep pass through this one for round 3. The product side keeps holding up — the manifest change is still comment-only, security is clean across the expanded GFS credential surface, and a lot of the prior harness debt is genuinely closed and verified (R2-M4 now has a real behavioral order test, the `41d60fa` PEM-newline fix checks out byte-for-byte). The problems this round are concentrated in two of the fix commits: one of them breaks the T2 gate on macOS, and the "harden recovery" commit trades a bootstrap soft-skip for a `die` in the wrong spot.

🔴 **Not ready to approve**
- **Must resolve before approval: 3**

1. **[R3-B1] `make minikube-t2` fails deterministically on any UTF-8 locale** (`scripts/tests/test-minikube-t2-contract.sh:36`, `scripts/tests/test-minikube-t2-public-boundary.sh:32`) — *regression from `76b8441b`*. That commit added `python3 -m py_compile "$GFS_FILTER"`, which drops an untracked binary `scripts/minikube/__pycache__/filter-gfs-resources.cpython-*.pyc` into the tree. The public-boundary scan then runs `sed 's/^/+/'` over every untracked file, and BSD/macOS `sed` aborts on the binary with `sed: RE error: illegal byte sequence`; under `set -euo pipefail` that kills the whole gate. The stray `.pyc` may also trip `t2_repo_metadata`'s dirty-worktree guard.
   - **Repro:** removed `__pycache__`, ran `bash scripts/tests/test-minikube-t2-contract.sh` fresh → RC=1 under `en_US.UTF-8` and `C.UTF-8`; passes only under `LC_ALL=C`. Reproduced independently by two reviewers.
   - **Fix:** stop littering the tree — compile with `PYTHONDONTWRITEBYTECODE=1` or send the `.pyc` to `$TMPDIR`; and independently harden the boundary scan to skip binaries (`grep -Iq . "$file"` before the `sed`).

2. **[R3-H1] `t2_lock_validate_inherited` returns 0 even after a guard fires — fail-open** (`scripts/minikube/t2-common.sh:737-774`). `e3da2f00` fixed exactly this in `t2_lock_acquire` by adding `return 1` after each `t2_fail` (since `t2_fail` returns 1, it does not exit). The sibling function has the identical `t2_fail`-inside-`if` structure at `:744,750,764,770` but ends with the truthy assignments `T2_LOCK_HELD=false; T2_LOCK_RELEASED=false`, so it returns 0 after firing a guard. It's fail-closed today *only* because every call site invokes it as a plain statement under `set -e` — the moment anyone writes `if t2_mutation_lock; then` or `t2_mutation_lock || …`, `set -e` is suppressed and a mismatched inherited token proceeds to mutate. This is the class-check sibling that the `e3da2f00` hardening left behind (F2).
   - **Repro:** fail-open under a conditional call, fail-closed under a plain call — so latent, not currently live.
   - **Fix:** add `return 1` after each `t2_fail` in `t2_lock_validate_inherited` (and after `t2_lock_profile_id_check` at `:739`), matching the `t2_lock_acquire` idiom.

3. **[R3-M1] Bootstrap soft-skip became a `die` ahead of the DSN producer** (`scripts/minikube/sync-auth-key.sh:397-409`) — *regression from `e3da2f00`*. `sync_gfs` used to soft-skip (`return 0`) when `gfs-controller-db` is absent or its `connection-string` is empty — a deliberate transient bootstrap state (still visible at `a656c313:sync-auth-key.sh:156-157`). `e3da2f00` turned those three branches into `die`. But in `provision_gfs_serving` the order is `sync_gfs_auth_key → settle → reconcile`, and `reconcile-gfs-deploy-credentials.sh:91` is what creates/populates the DSN — so the new `die` runs *before* the step that would fix the condition.
   - **Scenario:** a partial T2 cluster where `gfs-config`+`gfsc-reader` exist (guards pass) and the auth key re-derived (`force_rollout=true`) but `gfs-controller-db` has no `connection-string` yet. Before: skip → reconcile populates the DSN → converges. Now: `sync_gfs` dies and aborts pre-gate-sync before reconcile can self-heal. Confirmed statically (order + DSN producer + the `die` branch); not run against a live cluster.
   - **Fix:** keep the soft-skip on the T2 convergence path (where reconcile runs after), or move the `die` to after reconcile. The `e3da2f00` hardening is right for the `full-setup` callers that already document FAIL-LOUD — just not in this seam.

## Optional (non-blocking)

The PR can merge without these; follow-up or author's call.

1. **[R2-M2] GFS reader-convergence logic is documented but still physically duplicated** (`settle-gfs-reader-rollout.sh`, `wait-gfs-reader-ready.sh`, `pre-gate-sync.sh:459-468`). Credit where due: the 8-row decision table you added at `docs/testing/minikube-t2-runbook.md:185-201` is real and substantive — it closes the "no design to review against" concern (D3) that made me flag this last round. What's left is the code dedup (the same predicate lives in 3 places with divergent conditions). This no longer blocks; recommend opening a tracking issue to consolidate `settle`/`converge`/`wait` behind the table before more cases stack on it.
2. **[R3-L1] `|| true` swallows a kubectl failure in the converge pod-delete loop** (`scripts/minikube/pre-gate-sync.sh:461`). A transiently-failing `get pods` leaves `pod_rows` empty, the delete loop no-ops, and it falls through to the wait. It fails closed (the wait retries and times out → `return 1`), but the original failure is invisible. Distinguish NotFound from a real error like the other paths in this same diff do.
3. **[R3-L2] `shellcheck` over deleted paths** (`scripts/minikube/t0.sh:16`). `shell_files+=("$PROJECT_DIR/$path")` has no `[ -f ]` guard, but `git diff --name-only` includes deleted files, so a future PR that removes/renames a `.sh` would make T0 fail spuriously. Not triggered here (this PR deletes no `.sh`).

Carried-over by-design items from earlier rounds (R1-L3 comment-prose test guard, R1-L4 diagnostics artifact fields) stay as-is — conscious choices, no action needed.

## What checked out clean

- **Security (full diff): SAFE.** `sync-auth-key.sh` only touches the *public* JWT key (reads `.data.RPC_PROXY_JWT_PUBLIC_KEY`, writes ConfigMaps — never a private key); the "preserve auth key bytes" fix validates base64/UTF-8, so it's integrity of public-key bytes. The `gfs-controller-db` DSN is read for a non-empty check then `unset` immediately — never in `die`/`log`/argv (it travels by stdin to node/python). `reconcile-gfs-deploy-credentials.sh` is deny-by-default (remote path needs `GFS_REMOTE_RECONCILE_AUTHORIZED=true` *and* exact `ALLOWED_CONTEXTS` membership; `provision-gfs-runtime.sh` removes the `*example-dev*` wildcard). Shim only intercepts `rollout status` of `gfsc-reader`, `exec`s for the rest, prepended inline per-subprocess never global. `filter-gfs-resources.py` has no `yaml.load`/`eval`. Throwaway postgres is `127.0.0.1`-only, `--rm`, trust no-password, shared `control-postgres` untouched. CI is `pull_request` (not `_target`), no `${{ }}`/`secrets.*`.
- **Product:** manifest is still comment-only (0 non-comment lines vs merge-base); `deploymentStrategy.test.ts` 8/8; version bump only, no dependency change.
- **Verified closures from round 2:** R2-M1 (python3 rc now evaluated, `t2-common.sh:488-514`), R2-M3 (GFS scoped to the T2 gate), **R2-M4** (real behavioral order test `sync→settle→reconcile→converge`, T3 verified red against base), R2-L2, R2-L3, R1-M1 (single assertion sequence, flavor-selected messages), R1-L4/L5. And `41d60fa` "preserve auth key bytes" is correct — `source_hash` and `configmap_key_hash` match byte-for-byte, and its newline test fails against `41d60fa^` (T3 OK).
- **Suites:** `deploymentStrategy.test.ts` 8/8; `test-minikube-gfs-provision-order.sh`, `-sync-auth-key-durable.sh`, `-t2-lock-race.sh`, `-pre-gate-restore.sh`, `-settle-gfs-reader-rollout.sh` and the rest green. Only `test-minikube-t2-contract.sh` / `-t2-public-boundary.sh` fail — that's R3-B1, green under `LC_ALL=C`.

Makefile
```diff
@@ -362,12 +429,29 @@ minikube-apply-secrets: ## Apply all secrets to cluster (LLM keys read from .env
...
+		T2_PROJECT_DIR="$(CURDIR)" T2_PROFILE="$(MINIKUBE_PROFILE)" T2_CONTEXT="$(MINIKUBE_PROFILE)" T2_SKIP_LOCK="$(T2_SKIP_LOCK)" T2_LOCK_TOKEN="$(T2_LOCK_TOKEN)" \
+		bash scripts/minikube/sync-auth-key.sh --context=$(MINIKUBE_PROFILE); \
+	else \
+		T2_PROJECT_DIR="$(CURDIR)" T2_PROFILE="$(MINIKUBE_PROFILE)" T2_CONTEXT="$(MINIKUBE_PROFILE)" T2_SKIP_LOCK="$(T2_SKIP_LOCK)" T2_LOCK_TOKEN="$(T2_LOCK_TOKEN)" \
+		bash scripts/minikube/sync-auth-key.sh --context=$(MINIKUBE_PROFILE) --skip-gfs --require-mcp; \
+	fi
 
 .PHONY: minikube-sync-auth-key-if-present
 minikube-sync-auth-key-if-present: ## Sync JWT public key only when minikube auth resources already exist
+	@if [ -z "$(T2_LOCK_TOKEN)" ]; then echo "Skipping auth key sync (canonical T2 profile lock is not held)."; exit 0; fi
```
jozer-rami: **Blocking — this makes `make minikube-start` stop syncing the JWT public key entirely.**

`T2_LOCK_TOKEN` has no `?=` default anywhere in this Makefile; it is only ever forwarded from a parent. The sole caller of this target is `minikube-start` (`Makefile:148`), which never sets it. So for every ordinary developer this guard hits `exit 0` unconditionally and the target becomes a permanent no-op.

That matters because `minikube-start` is the documented top-level entry point — `scripts/minikube/dev.sh:91` invokes it, and `t2-common.sh:344` tells users to run `make minikube-start` on an unbootstrapped profile. The result is that the JWT public-key drift this target exists to repair silently stops being repaired on the main path. That behavior change isn't mentioned in the PR description.

Worth noting the existing coverage can't catch this: `test-minikube-makefile.sh:66` asserts via `make -n`, which prints recipe text without executing the guard, so `rpc-proxy-secrets` still appears in the output and the test stays green.

If the intent is genuinely "never mutate outside the canonical lease", then `minikube-start` needs to acquire or forward a lease, or this target needs a non-lock path for the plain dev flow. Skipping silently is the worst of the three.

---
_Generated by [Claude Code](https://claude.ai/code)_

Makefile
```diff
@@ -289,11 +328,30 @@ minikube-restart-all: ## Restart all Clerum deployments
...
+		$(MAKE) --no-print-directory minikube-deploy-crds-body
+
+minikube-deploy-crds-body:
+	@T2_PROJECT_DIR="$(CURDIR)" T2_PROFILE="$(MINIKUBE_PROFILE)" T2_CONTEXT="$(MINIKUBE_PROFILE)" \
+		T2_SKIP_LOCK=true T2_LOCK_TOKEN="$(T2_LOCK_TOKEN)" \
+		bash scripts/minikube/require-t2-mutation-lock.sh
 	kubectl --context=$(MINIKUBE_PROFILE) apply -f deploy/base/namespaces.yaml
-	helm upgrade --install --kube-context=$(MINIKUBE_PROFILE) clerum-crds ./charts/clerum-crds
-	kubectl --context=$(MINIKUBE_PROFILE) apply -f ./charts/clerum-crds/crds/
+	@if [ "$(MINIKUBE_GFS_MUTATION)" = "false" ]; then \
```
jozer-rami: **Blocking — inverted comparison relative to every other `MINIKUBE_GFS_MUTATION` site.**

Everywhere else in this file the test is `= "true"` for mutation-on, with anything else treated as off — lines 243, 244, 268, 274, 292, 444. This one tests `= "false"` to mean off.

So any third value takes the *mutating* branch: `MINIKUBE_GFS_MUTATION=0`, `=no`, or an explicit empty `MINIKUBE_GFS_MUTATION=` on the command line (which overrides the `?= false` default on line 161 and survives the pass-through on line 335). In that state `minikube-deploy-crds` applies `globalfilesystem.yaml` while `minikube-deploy-all` is concurrently filtering every GFS object out of the same sync — the two halves of the non-T2 path disagree about whether GFS exists.

```suggestion
	@if [ "$(MINIKUBE_GFS_MUTATION)" != "true" ]; then \
```

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/minikube/filter-gfs-resources.py
```diff
@@ -0,0 +1,56 @@
...
+
+def read_documents(paths: list[str]) -> str:
+    if not paths:
+        return sys.stdin.read()
+    return "\n---\n".join(Path(path).read_text() for path in paths)
+
+
+def main() -> int:
+    source = read_documents(sys.argv[1:])
+    documents = re.split(r"(?m)^---\s*$", source)
```
jozer-rami: **Regex-based YAML splitting is unsafe for the input this actually receives.**

Two problems compound here:

1. `re.split(r"(?m)^---\s*$", source)` splits on *any* line that is `---`, including one inside a block scalar. `kubectl kustomize` output routinely contains ConfigMaps whose `data:` values hold embedded YAML or Markdown, and a `---` in that payload splits one document into two fragments.
2. Each fragment is then handed to `is_gfs_owned`, whose `^  name:` / `^  namespace:` patterns (lines 23-24) match *any* two-space-indented key, not just `metadata.name` / `metadata.namespace`. A fragment torn out of a block scalar can easily present an indented `name:` or `namespace: gfs` that has nothing to do with the object's metadata.

The failure mode is either dropping a non-GFS document or re-emitting a corrupted one into `kubectl apply` — and because `is_gfs_owned` is deliberately fail-open for unknown documents, a wrong drop won't announce itself.

`python3` is already a hard dependency of this path, so `yaml.safe_load_all` would remove the whole class of problem. If PyYAML isn't guaranteed present, at minimum anchor the name/namespace lookup to the `metadata:` block rather than to indentation depth alone.

Two smaller things in the same function:

- When `kept` is empty the script writes nothing and returns 0, so the downstream `kubectl apply -f -` fails with `error: no objects passed to apply` rather than being a clean no-op.
- The new `kubectl kustomize | python3 filter | kubectl apply -f -` pipelines (Makefile:268-272, 277-281) run without `pipefail`, so a kustomize or filter failure is masked by a successful `apply`. The two-stage version of this pre-dates the PR, but inserting a third stage widens the window.

---
_Generated by [Claude Code](https://claude.ai/code)_

autoresearch.last-run.json
jozer-rami, line 1: **Blocking — this file and 33 siblings look like accidentally committed local tooling output.**

`autoresearch.md`, `autoresearch.ideas.md`, `autoresearch.jsonl`, this file, and 30 files under `autoresearch.runs/reduce-minikube-startup-overhead-.../` are unrelated to #391 and #412, and they embed the author's absolute local paths:

- `/Users/alfredolopez/.codex/worktrees/f3d2/evenfire` (this file, `cwd` and `generation.files.*`)
- `/Users/alfredolopez/Documents/GitHub/codex-ralph-vault-loop/config/scorecards/ralph_autoresearch_v1.yaml` (`autoresearch.jsonl`)

That publishes a developer's filesystem layout and the location of a separate private tool into the repo history. `.gitignore` picked up `.ralph/plans/` in this PR (line 98) but not these paths — worth adding them in the same change and dropping the files from the branch.

---
_Generated by [Claude Code](https://claude.ai/code)_

Makefile
```diff
@@ -323,9 +381,18 @@ minikube-restart-deploy: ## Restart a single deployment without rebuilding (usag
...
+	@T2_PROJECT_DIR="$(CURDIR)" T2_PROFILE="$(MINIKUBE_PROFILE)" T2_CONTEXT="$(MINIKUBE_PROFILE)" \
+		T2_SKIP_LOCK="$(T2_SKIP_LOCK)" T2_LOCK_TOKEN="$(T2_LOCK_TOKEN)" \
+		bash scripts/minikube/with-t2-mutation-lock.sh -- \
+		$(MAKE) --no-print-directory minikube-gen-keys-body
+
+minikube-gen-keys-body:
+	@T2_PROJECT_DIR="$(CURDIR)" T2_PROFILE="$(MINIKUBE_PROFILE)" T2_CONTEXT="$(MINIKUBE_PROFILE)" \
+		T2_SKIP_LOCK=true T2_LOCK_TOKEN="$(T2_LOCK_TOKEN)" \
+		bash scripts/minikube/require-t2-mutation-lock.sh
+		@scripts/minikube/generate-keys.sh
```
jozer-rami: Nit: this is indented with two tabs while the surrounding recipe lines use one, so it reads like a continuation of the `require-t2-mutation-lock.sh` invocation above it — but line 394 has no trailing backslash, so it's a separate recipe line.

I checked the behavior and GNU make strips the first tab, then skips remaining whitespace before recognizing the `@` prefix, so it still runs silently and correctly. Purely cosmetic, but the extra tab is misleading to read.

---
_Generated by [Claude Code](https://claude.ai/code)_

jozer-rami (commented):
## Review of PR 416

Static review only — nothing was executed. Diffed against `origin/dev` at `9e20e575`, head `70378f5e`.

### Scope

GitHub reports 181 files / +15,583 because the PR's base sha (`05a52ce2`) is stale. Against current `dev` HEAD the real delta is **95 files, +7,305 / −653, 49 commits**.

The description asks for a review of "product #391 / PR 416 only" and puts the T0/T1/T2 harness out of scope, but that split doesn't match the diff. The product change is under 400 lines (`host-context-controller.yaml`, `deploymentStrategy.test.ts`, `e2e-hcc-rollout-readiness.sh`). The rest is harness: `t2-common.sh` +581, `sync-auth-key.sh` +539, `pre-gate-sync.sh` +325, a `kubectl` PATH shim, and a new mutation-lock protocol threaded through six Makefile targets. Every defect below is in that harness half, so it can't be reviewed as out of scope. I'd suggest splitting the harness and mutation-lock work into its own PR — it would also make the #391 evidence much easier to see.

### Blocking

1. **34 local agent-tool artifacts committed**, embedding the author's absolute paths (`/Users/alfredolopez/.codex/worktrees/...`, `/Users/alfredolopez/Documents/GitHub/codex-ralph-vault-loop/...`). Inline on `autoresearch.last-run.json`.
2. **`make minikube-start` silently stops syncing the JWT public key** — `T2_LOCK_TOKEN` has no default and `minikube-start` never sets it, so the new guard makes `minikube-sync-auth-key-if-present` a permanent no-op on the main dev path. Not covered by the existing `make -n` test. Inline on `Makefile:454`.
3. **Inverted `MINIKUBE_GFS_MUTATION` comparison** in `minikube-deploy-crds-body` — `= "false"` where all six other sites use `= "true"`. Inline on `Makefile:344`.

### Worth fixing

4. **`filter-gfs-resources.py` parses YAML with regex** — splits on any `---` line including inside block scalars, and matches `name:`/`namespace:` by indentation depth rather than by position under `metadata:`. Can drop or corrupt a non-GFS document, silently. Also: empty output breaks the downstream `apply`, and the new three-stage pipelines run without `pipefail`. Inline on line 48.

### Nits

5. `deploymentStrategy.test.ts` asserts on four prose fragments inside a YAML comment (`'DETECTION ONLY'`, `'it does not roll back'`, `'Kubernetes will not self-heal'`, `'The deploy job still fails after a successful undo'`). The intent is documented and defensible, but it already forced string churn in this PR and will fail on any reword. One stable marker line would hold the same contract more cheaply.
6. `Makefile:395` — stray second tab. Cosmetic; verified make still handles it. Inline.
7. `scripts/minikube/gfs-rollout-shim/kubectl` shadows `kubectl` on `PATH`. It's tightly scoped (intercepts only `rollout status` for `$GFS_READER_DEPLOYMENT`, `exec`s the real binary otherwise) and the rationale is well documented, but it changes kubectl semantics for its entire subtree, so it's worth being explicit that it must never reach a non-dev path. One edge: real-binary discovery takes the first `kubectl` on `PATH` outside `SELF_DIR`, so with two shim copies present (a second worktree, say) it could chain rather than reach the real binary.

### What's good

The e2e work is the strongest part of this PR, and several changes close real latent holes rather than just adding a mode:

- `assert_botched_outage` now **fails closed when the pod-UID listing itself fails**. The previous `kctl get pods ... | grep -cxF "$OLD_POD_UID" || true` counted a query failure as "old pod terminated" — a silent false pass on the central D1b assertion.
- Replacement-pod matching is much stricter: excludes terminating / `Failed` / `Succeeded` pods, requires a ReplicaSet owner whose revision matches the Deployment's current revision, and requires the container image to equal the injected botched image. The old version took the first pod with a different UID.
- `[ "$e2e_fail" -eq 0 ] && IMAGE_BROKEN=0` is the right call given `fail()` records and continues — a failed post-undo proof leaves cleanup's image restore armed.
- The guards suite is a genuine step up from grep-only contract tests: it runs the gate under `env -i` with a throwaway kubeconfig and a fake kubectl that refuses every mutating verb, then asserts the refusal originated in the guard block rather than later. The post-hoc check on the invocation log in case 7 (no mutating verb recorded before the node-label guard) is a nice touch.
- The #412 isolated-PostgreSQL lane is thorough, with fail-closed `die_t1` codes on every path, and the cleanup guard's De Morgan rewrite (`!= *port-forward* || != *svc/control-postgres*`) is correct.

Filed as a comment rather than a formal change request — the three blocking items are the maintainers' call.

---
_Generated by [Claude Code](https://claude.ai/code)_

zach88:
Took another deep pass, this time scoped to the fix batch pushed since the last close-out review at `41d60fa1`. Good news first: the three behavior findings from round 3 (R3-B1 the CI blocker, R3-H1 the fail-open lock, R3-M1 the premature `die`) all genuinely closed on this head, verified with evidence. The problem is what the batch that closed them dragged in.

All of this round's blockers are regressions from this round's own `fix:` commits, and they're concentrated in the GFS self-heal module (`settle-gfs-reader-rollout.sh`, `t2_deployment_check`). One of them, R4-B1, is the same class of macOS-determinism bug as R3-B1, reintroduced in the same round that closed R3-B1. This is round 4, and the module concentrating the findings is exactly the one flagged a couple rounds back for extraction into its own PR with a decision table. My honest recommendation is to do that split now: pull the GFS self-heal out, land the #391 detection/undo work (which converges) on its own, and give the self-heal a small design pass before more patches stack on it.

## Verdict

🔴 **Not ready to approve**

- **Must resolve before approval: 4** (2 Blockers, 1 High, 1 open regression)

1. **[R4-B1] Empty-array crash breaks `make minikube-t2` on macOS** (`scripts/minikube/settle-gfs-reader-rollout.sh:171`). `for rs_name in "${scaled_rs_names[@]}"` runs with the array empty on the healthy common path (no leftover ReplicaSet to scale), which aborts with `scaled_rs_names[@]: unbound variable` under `set -u` on bash 3.2 (the macOS default). **Regression from `1bc24dcc`.** Repro: `test-minikube-settle-gfs-reader-rollout.sh` (scenario with no `FAKE_RS_ROWS`) fails, and it takes `test-minikube-t2-contract.sh` down with it; on `41d60fa1` the settle test passes, on this head it fails. Same class as R3-B1 (macOS-local determinism). Fix: guard with `if (( ${#scaled_rs_names[@]} )); then ... fi`. The repo already knows this gotcha (`full-setup.sh:926`, `sync-auth-key.sh:429,503,529` all guard it), so please apply the same guard to every `for x in "${arr[@]}"` in the module, not just this one line.

2. **[R4-B2] 30 AI-scratch files committed, will land in dev on merge** (`autoresearch.*`, `autoresearch.runs/**`). Added in `6db0071e`. They're in the branch tree at this head but absent from `origin/dev` (`git ls-tree HEAD | grep -c autoresearch` = 30 vs 0 on dev), so a merge carries them into the monorepo. They're an autoresearch tool's run artifacts (`candidate.patch`, `scan_report.json`, `benchmark.*.preview.txt`, `decision.json`, run logs). They don't show up in the curated product diff, which is why they're easy to miss. Fix: `git rm -r autoresearch.*` off the branch and add it to `.gitignore`.

3. **[R4-H1] Same empty-array crash in the non-T2 branch** (`scripts/minikube/settle-gfs-reader-rollout.sh:110`). `for allowed_context in "${allowed_contexts[@]}"` after `IFS=',' read -r -a allowed_contexts <<<"${ALLOWED_CONTEXTS:-}"`: with `ALLOWED_CONTEXTS` empty the array has 0 elements and crashes with `unbound variable`, masking the intentional `die` on 113-114 with a confusing error instead. **Regression from `1bc24dcc`**, class sibling of R4-B1. Lower blast radius (needs `GFS_READER_ROLLOUT_AUTHORIZED=true` and ends in a `die` anyway), but it should be fixed in the same commit as B1. Correct guard already exists at `sync-auth-key.sh:72` (`[ -n "${ALLOWED_CONTEXTS:-}" ] || return 1`).

4. **[R4-M1] T2 preflight readiness gate silently narrowed to 5 deployments** (`scripts/minikube/t2-common.sh:555`, `t2_deployment_check`). It used to iterate every Deployment in the inventory and flag `PROFILE_UNHEALTHY` on any `ready<desired`; the batch rewrote it to inspect only `T2_REQUIRED_DEPLOYMENTS` (`control-api`, `host-context-controller`, `external-rest-api`, `rpc-proxy`, `mcp-host/chatllm`). Anything outside that set (`control-ui`, `profile-ui`, `workflow-recipes`, `gfsc-*`) that's crash-looping no longer fails the preflight, and there's no compensating check (`t2_resource_checks:467` only asserts the Service object exists, not pod readiness). Net: a broken frontend or worker certifies the profile as healthy, which is the exact recovery flow this PR exercises. **Regression from `5b17324a`** (whose title, "read per-deployment readiness metadata", only advertises adding `observedGeneration`/`unavailableReplicas`; the scope narrowing is unannounced). Repro: fixture with `control-plane/control-ui` at `0/1` fails under the R3 logic, passes under this head's. Flagging this as gating because it's an open, reproduced regression, but the call is yours: if narrowing to 5 is intentional, it needs a compensating readiness check for the rest plus a contract test pinning it (the "X must not pass while Y is unhealthy" invariant should be a test, not prose).

## Optional (non-blocking)

1. **[R4-L1] Lease-identity revalidation duplicated in 3 entrypoints** (`scripts/minikube/sync-auth-key.sh:82`, `require-t2-mutation-lock.sh:29-41`, and one more). Same `T2_BRANCH`/`T2_HEAD`/`T2_WORKTREE_ID`/`t2_lock_validate_inherited` snippet copy-pasted. It's a guard, not a state-producing business rule, so it's not a hard D4 violation, but it's a candidate to extract into a single function in `t2-common.sh`.

2. **[R4-L2] New fail-closed edge in the sync-auth-key profile/context guard** (`scripts/minikube/sync-auth-key.sh:66`). The `[[ PROFILE==T2_PROFILE && PROFILE==T2_CONTEXT ]] || die` runs before the remote/local branch split, so on the authorized remote path, if the env carries `MINIKUBE_PROFILE`/`CONTROL_API_REAL_PG_CONTEXT` with a value different from `--context`, it `die`s in a case that worked on `41d60fa1` (when `KCTL=--context=$PROFILE`). Clean prod env has those empty so it passes. Read, not reproduced; it's a new fail-closed, not a bad write.

3. **[R4-L3] deploymentStrategy test asserts YAML comment prose** (`host-context-controller/src/deploymentStrategy.test.ts:110-124`). `toContain('DETECTION ONLY')`, `'it does not roll back'`, etc. Fragile against comment reformatting. It's the continuation of R1-L3 and it's defensible (recovery is owned by an external pipeline so there's no code behavior to assert, and the real fields `replicas`/`strategy.type`/`progressDeadlineSeconds` do get pinned against the manifest). Not blocking.

## What checked out clean

- **Round 3 blockers all closed and verified**: R3-B1 (`.pyc` now written to an `mkstemp` temp with `os.unlink` in `finally`, plus the consumer scan skips binaries via `LC_ALL=C grep -Iq`), R3-H1 (`return 1` after every `t2_fail` in `t2_lock_validate_inherited`, and its class siblings are pre-existing and fail-closed under `set -e` today), R3-M1 (the `GFS_AUTH_SYNC_ALLOW_STAGED` soft-skip `return 0` runs ahead of the DSN producer, traced end to end).
- **Security: clean.** No new vulns. The synced key is the public `jwt-public-key` (not a private secret), never echoed to logs; the Postgres DSN is only read for a presence check and `unset`, never printed; base64 goes through `python3 b64decode(validate=True)` with no truncation; locks are fail-closed and TOCTOU-safe (`mkdir` atomic, `umask 077`, full identity revalidation); the `kubectl` shim is array-exec pass-through with no injection.
- **Auth-key bytes (`ca1920cb`)**: no corruption or encoding change; hashing matches byte-for-byte on both sides, optimistic-concurrency JSON patch with `test` ops on `resourceVersion` + data key.
- **The good half of `5b17324a`**: the per-item `metadata` rebind is a real fix (metadata used to stay pinned to the last item of the first loop, comparing each deployment against a foreign generation), and its `multi_ready_json` test fails against the parent (valid T3).
- **Callers of the newly lock-gated `sync-auth-key.sh` / `converge-control-db-after-reset.sh`** were all updated consistently in the same diff.
- **Body claims hold**: `EXPECT_STUCK=1 && EXPECT_RECOVERY=1` exclusivity is enforced (`e2e-hcc-rollout-readiness.sh:100-101`); the guard tests run the real gate in CI via `bash` (not grep-only).
- **Suites**: `deploymentStrategy.test.ts` 8/8; 12 PR-touched shell tests green; the only two failures (`test-minikube-settle-gfs-reader-rollout.sh`, `test-minikube-t2-contract.sh`) both trace to R4-B1.

zach88:
Did a focused round-5 pass on this one. Scope was narrow on purpose: verify the four findings that were still open from round 4, and audit what the new batch of fix commits (`70378f5e..430f130d`) might have broken. Good news up front, this is the first round where a batch of fixes didn't reintroduce a blocker.

## Verdict

🟡 **Approve with nits**

All four round-4 items (the two array-unbound blockers, the autoresearch scratch, and the narrowed readiness gate) are closed and verified, and the collateral pass came back clean, no behavioral regressions reproduced. What's left is coverage/hygiene follow-up that doesn't need to block the merge. This is my read from the review, not a formal GH approval.

None. The gating items from round 4 are all closed:

- **[R4-B1] `scaled_rs_names` array unbound under `set -u`** (`scripts/minikube/settle-gfs-reader-rollout.sh:181`) - fixed with `[ "${#scaled_rs_names[@]}" -gt 0 ]`. Reproduced on bash 3.2.57: without the guard it dies `arr[@]: unbound variable`, with it it's clean.
- **[R4-H1] `allowed_contexts` array unbound** (`scripts/minikube/settle-gfs-reader-rollout.sh:110`) - fixed with `if [ -n "${ALLOWED_CONTEXTS:-}" ]`. Reproduced: with the var empty it now reaches the intentional `die` instead of the unbound error.
- **[R4-B2] committed `autoresearch.*` scratch** - `git ls-tree` on HEAD shows 0 `autoresearch*` files, net-removed plus a `.gitignore` entry. Doesn't land in dev.
- **[R4-M1] readiness gate narrowed to 5 deployments** (`scripts/minikube/t2-common.sh`) - fixed with `T2_REQUIRED_DEPLOYMENTS_INCLUDE_ADDITIONAL=true` (default when the scope is empty). The python checker now selects the required scope plus the whole inventory, so a broken `control-ui`/`profile-ui`/`workflow-recipes`/`gfsc-*` with `replicas>0` fails the gate again. Callers passing an explicit scope keep the narrow behavior.

## Optional (non-blocking)

The PR can merge without these; they're follow-up, ideally as separate issues rather than another round here.

1. **[R5-L1] Weakened contract tripwire in `deploymentStrategy.test.ts:105`** (regression from `4d54b794`). The test used to assert 7 concrete strings of the recovery contract in the manifest (the anti-deletion tripwire). The fix replaced them with a single marker assertion (`# CONTRACT: hcc-rollout-recovery=...`). I grepped `deploy/base/control-plane/host-context-controller.yaml`: all 7 strings are still present and the manifest only added the marker line, so the old test would still have passed, the rewrite was a free relaxation. No runtime impact, but the contract prose can now be deleted without any test failing. Suggest restoring the 7 assertions alongside the marker.
2. **[R5-M-proc] YAML parser grown inside a `fix:` in `scripts/minikube/filter-gfs-resources.py`.** ~137 lines of structural YAML parsing (document splitting with block-scalars, `structural_mappings`, `inline_metadata`, indentation precedence) landed in a fix commit with no mini-spec. It works on the inputs it's tested against and the new fixture fails against base, but its decision table (what counts as a document separator, which `name`/`namespace` wins) has no reviewable design. Worth a short mini-spec before any further patching on it.
3. **[R5-L2] `t2_worktree_id` not fully generalized.** The function was extracted but the identical `printf '%s' ... | shasum | awk '{print $1}'` idiom still lives in ~4 more spots (`scripts/minikube/image-mode.sh:231`, `scripts/e2e/e2e-np08-hcc-authorization.sh:131`, `scripts/e2e/_lib/hcc-watch-recovery-lock.sh:10`, `scripts/e2e/_lib/hcc-watch-recovery-fixture.sh:31`). The function is byte-identical to the inline form so there's no functional divergence, purely class hygiene.

## What checked out clean

- All four round-4 items closed; three reproduced directly, one (M1) verified by reading the python selection logic plus the collateral pass.
- Zero production behavior regressions across the 6 new commits. Candidates I chased and cleared: the `2>/dev/null` -> `2>&1` + `die` changes in `settle-gfs-reader-rollout.sh` (fail-loud on a path where readiness is already verified), the removed `owner_revision` fallback in `gfs-credential-rollout.sh` (dead code, both sides died anyway), and `provision-gfs-runtime.sh` switching to `$ROOT`-anchored paths (hardening).
- `t2_worktree_id` is byte-identical to the old inline idiom, IDs match. `filter-gfs-resources.py` output is YAML-equivalent to before on the real producer (`kubectl kustomize`), differing only in blank lines; where it diverges on adversarial input it removes more gfs resources, not fewer.
- Test alignments in the `test(pr-416):` commits are legit: the settle stub match `*gfs-template-hash*` -> `*creationTimestamp*` tracks the real producer dropping that jsonpath field, and the pre-gate/makefile ones are structural.

zach88 (approved):

PR #417, opened 2026-08-19
chore(desktop): harden shortcut lifecycle follow-ups

josue-tech-94:
## Summary

Post-merge follow-ups from #350:

- Remove unused `editingPolicy` metadata from the Desktop command registry.
- Dispose host shortcut routing when the owning BrowserWindow closes.
- Verify the existing sandbox palette focus ownership and pending sandbox-find behavior.

## Why

The command metadata was decorative rather than enforced, and the host shortcut listener relied on
implicit WebContents cleanup. This change makes both ownership boundaries explicit without changing
shortcut routing, palette destinations, or sandbox privileges.

## Validation

- `cd desktop-app && npm test` (193 files, 1,924 tests)
- `cd desktop-app && npm run build`
- `cd desktop-app && npm run test:native:sandbox-shortcut-focus`
- `cd desktop-app && npm run test:native:sandbox-find`
- `npm run style-rules`

## Review

- Code review: SAFE after closing R1-B1 with producer-backed host shortcut lifecycle coverage.
- Security review: SAFE


josue-tech-94:
@codex Please perform a final **full review of this PR at the current head**.

Review the complete PR diff against `dev`, not only the latest R1-B1 fix. This is a small post-merge Desktop follow-up from #350.

Please pay particular attention to:

* correctness and regressions
* host shortcut listener ownership and cleanup
* BrowserWindow / WebContents lifecycle behavior
* exact `before-input-event` listener registration/removal identity
* window close and recreation behavior
* duplicate-listener / duplicate-command-dispatch risks
* removal of the unused `editingPolicy` metadata
* preservation of standard editable-input shortcuts
* host vs sandbox shortcut routing
* Electron main-process trust boundaries
* test quality, especially whether the new R1-B1 lifecycle coverage would actually fail if:

  * routing were registered twice
  * teardown removed the wrong callback
  * the old source could still dispatch after close
  * a recreated window accumulated listeners
* whether the tests assert observable behavior rather than source text or implementation details

Also verify that the PR does not unintentionally change:

* shortcut matching semantics
* command eligibility
* palette behavior
* sandbox focus/find behavior
* preload / IPC / plugin privilege surfaces

Please separate findings into:

1. **Merge-blocking issues**
2. **Non-blocking follow-ups**

For every issue, include concrete `file:line` evidence, a failure scenario, impact, and the required fix/test.

If there are **no merge-blocking issues**, please say so clearly.

zach88:
Took a pass through this one. Clean, tightly-scoped follow-up: the `editingPolicy` removal is safe and the shortcut dispose-on-close is a real fix, not just cosmetics. Nothing here blocks a merge; the only two things I'd flag are test-quality nits.

**🟡 Approve with nits**
- Optional / non-blocking: 2

## What checked out clean

- **`editingPolicy` was genuinely dead metadata.** `grep -rn editingPolicy` across the whole repo (both base and head) returns only the test assertion — zero consumers. The `'suppress'` semantics were never wired to this field; editing-chord suppression lives entirely in `isStandardEditingShortcut` in `shortcutRouter.ts`, independent of what got removed. On top of that, every command carrying `editingPolicy:'suppress'` has `defaultBinding: null`, and `matchDesktopCommand` only returns commands with a non-null binding, so those were never keyboard-triggerable anyway. Removing the field loses no behavior — the body's "decorative rather than enforced" is accurate.
- **The dispose fix is a real repro.** Patching out `host.once('closed', dispose)` turns `mainShortcutLifecycle.test.ts` red (`removeListener` expected once, `listenerCount` expected 0), so the test genuinely covers the leak. At base the dispose returned by `wireDesktopShortcutRouting` was discarded and the `before-input-event` listener leaked for the window's lifetime; this closes it.
- **Sandbox path untouched.** The old `wireDesktopShortcutRouting` is still used by `sandboxUiDriver.ts:482`, which already disposed via `cleanupShortcutRouting`. The route object the new host wrapper passes is byte-identical to the old inline call, so routing/palette destinations/sandbox privileges are unchanged.
- **Double `'closed'` handling is fine.** `main.ts:306` (dispose) and `main.ts:308` (`mainWindow=null`) have no ordering dependency, and the `window.destroy()` error path also emits `'closed'`, so dispose runs there too — no leak on the throw path.
- Suites green: 11/11 on the touched files, 29/29 on the broader shortcut/command suite, and `tsc` on the main process is clean.

## Optional (non-blocking)

The PR can merge without these — author's call / follow-up.

1. **[R1-L1] Test title over-claims; the main.ts integration path isn't exercised** — `desktop-app/src/__tests__/mainShortcutLifecycle.test.ts:65`. It's named "owns one routable listener until close and gives a recreated window a fresh listener," but it only calls `wireHostDesktopShortcutRouting` twice on hand-built `FakeBrowserWindow`s. It never drives `createWindow`, so the real recreation path (re-wiring on each call, plus the two `'closed'` handlers at `main.ts:306`/`308`) isn't covered. The router unit itself is well covered. Either rename to reflect it's a router-unit test, or add a thin main-process integration assertion.
2. **[R1-L2] `FakeWebContents.isDestroyed()` hard-returns `false`** — `desktop-app/src/__tests__/mainShortcutLifecycle.test.ts:32`. In real Electron the WebContents is already destroyed when `'closed'` fires, which is exactly when `dispose` calls `removeListener`. No real bug — `removeListener` is Node-EventEmitter-level and safe on a destroyed WebContents — but because the fake reports `isDestroyed() === false` throughout, it could never surface a "throws on destroyed webContents" regression. Minor faithfulness gap in the fake.

zach88 (approved):

PR #419, opened 2026-08-20
feat(desktop-app): dock a resizable chat drawer beside the live app

zach88:
## What
Adds a Cursor-style chat drawer that docks to the right of a running app on the
apps route, so you can chat about an app while it stays open.

- Drawer mounts beside the live app embed, with an open-chats switcher in its
  header and composer reference overlays confined to the drawer.
- The drawer seeds its conversation from conversationOrigin without a
  destroy/remount round-trip, and keeps the embed alive when starting a new chat
  while the app is in the foreground.
- Notifications behave sanely with the drawer open: the tray stays a popover,
  the drawer only opens for notifications that actually surface in it, and the
  notification popover renders above the drawer.
- The drawer is resizable via a left-edge drag handle — by mouse or keyboard
  (the handle is a focusable ARIA separator: arrow keys nudge the width, Home/End
  jump to the bounds) — with compact drawer-only styling. When the window is too
  narrow to fit the drawer beside a legible app embed, the drawer is suppressed
  and the app reclaims full width, reappearing when the window re-widens.

## Why
Acting on an app and chatting about it used to mean leaving the app. Docking the
chat keeps both in view. The resize makes the drawer usable for dense content
such as tables, and narrow-window suppression keeps it from crowding the app on
small screens.

## How
- The drawer and the fullscreen chat share the same ChatPage; the drawer only
  changes the CSS ancestor (.chat-drawer) and the active route, so fullscreen
  chat is left untouched.
- Width is a dedicated --chat-drawer-width, decoupled from the notification
  drawer variable; the app embed gutter and the header column derive from it.
  The two drawers are mutually exclusive, so there is no collision.
- Width is clamped to [340, min(820, panel − 460 − 46)] and is session-only by
  design (resets on launch, not persisted).
- Fit is decided from a ResizeObserver on the content panel: once the panel can
  no longer host the drawer at its 340px minimum plus the app embed at its 460px
  floor plus the fixed 46px gutter (i.e. below 846px), the drawer is suppressed —
  it stops rendering and the embed takes the full width — until the panel
  re-widens. The user's open intent is kept, so it reappears on re-widen; only a
  manual close clears it.
- The drag teardown is leak-safe: it ends on window blur and on hide/unmount
  mid-drag, restoring the body cursor and user-select.
- The handle is keyboard-operable per the ARIA window-splitter pattern: it is
  focusable (tabIndex 0), ArrowLeft/ArrowRight resize by a fixed step and
  Home/End jump to min/max (all through the same clampWidth as the drag), and
  aria-valuemin/max/now report the width to assistive tech.

## Testing
- Drawer component and App integration tests pass.
- Resize hook pure logic (clamp / narrow-width suppression / drag mapping) and
  keyboard resize are covered by dedicated unit tests; full renderer UI suite
  green.
- Renderer typecheck: no new errors introduced by this branch.
- Renderer build (vite): green.
- Pre-commit gates (prettier format, style-rules, commitlint) pass.

## Risks / Follow-ups
- The embed floor (460px), width floor (340px), and compact paddings are
  pending live visual calibration; each is a single constant.
- The resize drag may need a pointer-capturing overlay while the cursor crosses
  the native app embed; pending a check on a real desktop build.


zach88:
Took a review pass through this one and landed fixes for what came up on the branch, so this is more of a "here's what I found and how it got resolved" note than an approval. Everything lands in the desktop-app renderer; the drawer state wiring, the IPC contract for `chat.switcher`, and the resize/CSS all held up well. The findings were all low-severity, mostly test quality, and I've already closed them. One resize behavior is worth a manual check before merge. (The fix commits referenced below are on the branch and go up with the next push.)

**🟡 Approve with nits**
- Optional / non-blocking: **1**

## What I found and fixed

All of these were low-severity and are already addressed on the branch (one commit each):

1. **[R1-L1] Prettier line over printWidth in `desktop-app/src/__tests__/desktopCommands.test.ts`** — turned out to already be resolved on the current head; the `chat.switcher` editingPolicy cleanup reformatted the file and `prettier --check` is green. No separate fix needed.
2. **[R1-L3] Hand-written fixture in `desktop-app/ui/src/components/ChatSwitcher/ChatSwitcher.test.tsx`** — the `tabs` array is now derived from the real `chatViewTabs` producers (`createChatViewTabsState` / `openPersistedChatViewTab` / `addBlankChatViewTab`) instead of a hand-rolled literal, so it stays honest if `ChatViewTab`'s shape changes. Fixed in `7e1a0f8e`.
3. **[R1-L5] Missing coverage for the cross-team notification branch in `desktop-app/ui/src/__tests__/App.chatDrawer.test.tsx`** — `handleOpenNotificationInDrawer` (`App.tsx:449-464`) ejects to full-screen when `notification.teamId !== getCurrentTeamId()` and must not open the drawer. Added a test with a differing `teamId` asserting the drawer stays closed and `handleOpenNotification` is called without `keepNavItem`. I verified it fails against a broken team-check, so it's a real guard, not a false green. Fixed in `9570c5af`.
4. **[R1-L6] Hardcoded `4000` in `desktop-app/ui/src/__tests__/chatDrawerHeaderStacking.test.tsx:52`** — that literal is the real `.toast-stack` z-index, so the stacking assertion now reads it from the actual CSS rule (and throws if the rule moves to a token, rather than passing vacuously). Fixed in `d34fd868`.
5. **[R1-L4] Weak pin in `desktop-app/ui/src/hooks/__tests__/useAppController.chatDrawer.test.tsx:143`** — the full-screen route test only checked routing, not that the chat was selected. Added `activeChatId === 'chat-1'` as a lightweight observable pin (it sets synchronously before the harness's rpc-less `switchToChat` throws, so no harness surgery needed). Fixed in `2a1df28a`.

One finding I closed in writing rather than patching:

- **[R1-L2] `recomputeOverlay` leaves `isOverlay` stale when `panelWidth <= 0`** (`useChatDrawerResize.ts:84`) — this is unreachable in production (the content panel is the always-mounted ~100vw ancestor while the drawer is active) and has no executable repro at any level: in jsdom `clientWidth` is 0, so the overlay computation is a no-op and the stale-true state can't even be constructed in a test. Changing it would be unverifiable churn in the same file as the one open item below, so I'd fold it into that if we ever touch it. Happy to drop in a one-line guard if you'd rather have it.

## Optional (non-blocking)

1. **[R1-M1] Resize drag may freeze when the cursor crosses the native embed** (`desktop-app/ui/src/hooks/useChatDrawerResize.ts:113-157`) — the drag handle is on the drawer's left edge, so widening drags the cursor left toward the `SandboxUiPage` `WebContentsView`, which paints above all DOM. Drag tracking relies on `mousemove`/`mouseup` on `window`; if the native view captures pointer events once the cursor enters its rect, the width stops updating mid-drag and only the `blur` listener would end it. I couldn't confirm this in jsdom (no native view), so it needs a check on a real desktop build: dock the drawer over a foregrounded app and drag the handle left, watching whether the width keeps following the cursor once it's over the embed. If the OS keeps pointer capture on the main window while the button is held, this is a non-issue. If it does freeze, the fix is a transparent pointer-capturing overlay during `isResizing` (same shape as the existing shell-overlay freeze). Non-blocking; fine as a follow-up.

## What checked out clean

- **Security**: no isolation changes to the embed (no webview/nodeIntegration/sandbox/CSP/postMessage touched), the only bridge change is adding `chat.switcher` to the inbound command allow-list (non-privileged, just opens/focuses the drawer), no unescaped HTML, and the resize width is clamped and non-evadable.
- **IPC contract**: `chat.switcher` is consistent end to end (same name in `desktopCommands.ts` / `preload.ts` / `App.tsx`, types derived from the shared registry, effectively registered, sync test green). The renderer signature extensions (`keepNavItem`, `options`) are additive and backward-compatible; no old caller passes the new arg.
- **Reconciler** (`reconcileChatViewTabs`): correct by manual trace plus property tests (fast-check) for order, dedup, no-drop, and idempotence.
- **Resize teardown** is leak-safe: idempotent teardown across mouseup / hide / unmount, symmetric listener add/remove, `ResizeObserver.disconnect` on cleanup, clamp with no inversion.
- **CSS**: no variable collision with the notification drawer (mutually exclusive), correct z-index ordering, fullscreen chat untouched.
- **Suites**: full renderer ui suite green (945 tests), `desktopCommands` 11/11, main/preload typecheck clean. After the fixes, the collateral audit over the fix range came back with zero regressions.

alfredolopez80:
@zach88 — thanks for the thorough self-review pass. We independently verified all five of your claimed fixes (R1-L1, R1-L3, R1-L4, R1-L5, R1-L6) against HEAD and every one is genuine and non-vacuous — in particular R1-L5's cross-team guard really does assert `drawer: false` + no `keepNavItem` for a mismatched team, and R1-L6 reads the real `.toast-stack` z-index rather than a hardcoded copy. Nice work.

**Verdict: no blockers. One item needs a manual check on a real desktop build before merge; everything else is follow-up or optional.**

---

## Verify before merge

### 1. Resize drag may freeze when the cursor crosses the native embed (your R1-M1 — we agree, and elevate it from "optional" with code evidence)

`desktop-app/ui/src/hooks/useChatDrawerResize.ts:113-157`

The drag listeners (`mousemove`/`mouseup`/`blur`) live on `window`, but the sandbox `WebContentsView` is a native surface painted above the DOM. Widening the drawer drags the cursor **left, into the embed's rect** — the primary gesture direction — and if the native view captures the pointer there, the host renderer stops receiving `mousemove`/`mouseup`: `pointerX` freezes and the drag can neither advance nor end, with recovery depending on `blur` firing. A DOM overlay can't help because it can't cover a native view.

Two honest caveats, so this isn't overstated:
- **Not reproducible in jsdom** — none of our agents could exercise it in CI; it needs a real desktop build.
- OS/Chromium implicit mouse capture on mousedown *may* keep routing drag events to the host view even over the sibling native view, in which case this never manifests. That's exactly what the manual check should settle.

**The fix is already in the codebase if needed:** `desktop-app/ui/src/pages/SandboxUiPage.tsx:498-538` hides the native view via `sandboxUi.setVisible(false)`, but it's only gated on `shellOverlayOpen` (`headerShellOverlayOpen || commandPaletteOpen`, `App.tsx:2065-2067`). Feeding `chatDrawerResize.isResizing` into that same path (treat an active resize like a shell overlay, optionally with `setIgnoreMouseEvents`) would close this with minimal new machinery.

Related: if the drag does freeze, a lost teardown lets a second mousedown orphan the first listener set plus a lingering rAF and stuck `body` cursor/userSelect (`useChatDrawerResize.ts:139-149`). It self-heals on drawer hide/unmount and disappears entirely with the fix above, so we're not tracking it separately.

## Follow-ups (non-blocking)

- **Unit tests for the resize hook's pure logic** — flagged independently by two reviewers. `clampWidth` (`useChatDrawerResize.ts:43-51`), the overlay flip at `embedWidth < 460` (`:87`), and the drag mapping `clampWidth(rightEdge - pointerX)` (`:129`) have no direct test; `ChatDrawer.test.tsx:95-103` only asserts the handle fires `onResizeHandleMouseDown`. If min/max were swapped or the drag mapping inverted, all 69 scope tests would still pass. This logic is pure and trivially testable — a small suite here buys a lot.
- **`.composer-reference-submenu` can be clipped by the drawer** — `desktop-app/ui/src/styles.css:12810` (context `:12590/:12604`). The submenu is `position:absolute` and sits inside `.chat-drawer { overflow: hidden }`, so it's clipped at the drawer box, not just repositioned; the spec-§6 comment reasons about embed occlusion but not drawer clipping. The common case (~248px) fits, so this only bites with long reference names at/near the 340px min width. Worth a quick manual look in the narrow column — we could not reproduce it without an Electron runtime.
- **Keyboard resize support** — `desktop-app/ui/src/components/ChatDrawer/index.tsx:30-36` has `role="separator"` + `aria-orientation` but only `onMouseDown`; the ARIA separator pattern expects arrow-key resizing when focusable. Minor a11y gap.
- **Non-null assertions in `desktop-app/ui/src/lib/chatViewTabs.ts:14,148`** (`tabs[0]!`, `tabs.at(-1)!.id`) — no live bug (empty is unreachable by construction: create seeds one tab, close guards `length === 1`), but a guard or invariant comment would make the file safer against future refactors.

## Optional

- **R1-L2 (`isOverlay` stale when `panelWidth <= 0`, `useChatDrawerResize.ts:84`)** — we agree with your call: unreachable in production (`.content-panel` is always mounted at ~100vw while the drawer is active, and the effect runs post-paint). An invariant comment or one-line guard would document it; no action required.
- The reconciler tests don't explicitly assert relative tab-array order (`chatViewTabs.test.ts`) — order holds by construction (map/append only), so this is purely test hardening, not a gap in correctness.
- `App.chatDrawer.test.tsx:81-108` uses faithful hand-mocks of `handleSelectChatAgent`/`handleOpenNotification`; the real path is covered in `useAppController.chatDrawer.test.tsx`, so this is just a note that a drifting `keepNavItem`/eject contract could theoretically be masked at the App layer.

## Verified clean

- **Electron isolation intact** — `desktop-app/src/main.ts:298-302` (`contextIsolation: true`, `sandbox: true`, `nodeIntegration: false`) untouched by the diff.
- **IPC `chat.switcher` consistent end-to-end** — type-derived from the registry, preload allowlist pinned by test, dispatch fail-closed at both boundaries, zero-arg and non-privileged.
- **Reconciler correct** — held up under a 20,000-run randomized property check across 6 invariants (order, dedup, no-drop, unique ids, active resolvable, same-ref idempotence). This was the standout of the PR.
- **Backward compatible** — `keepNavItem?`/`options?`/`focusRequestId? = 0` all additive; no old caller breaks.
- **CSS layering sound** — drawer at 90, header lifts to 120, toast at 4000; chat and notification drawers mutually exclusive via `App.tsx:534`; fullscreen chat untouched.
- **Clamp math has no off-by-one; no stale closures; teardown is leak-safe on all normal paths.**
- **Suites green** — 69/69 in-scope (exit 0); full run 1953 pass / 1 fail, where the single failure (`sandboxUiFixtureFailureMarker.test.ts`) is unrelated to this PR (out-of-scope sandbox-UI fixture startup).

---

*This is an automated multi-agent review (four independent reviewer agents + Fable 5 adversarial verification), checked against the code at HEAD. It is advisory input for the human reviewers — not a formal GitHub approval.*

zach88:
@alfredolopez80 — thanks for the multi-agent pass. Landed the three actionable follow-ups on the branch, plus checked the runtime-only one. Collateral audit over the fix range came back clean (0 regressions), full renderer UI suite is green (965), and typecheck has no new errors.

## Fixed

- **R2-M1 — resize hook pure logic now under test** (`8245ed4e`). Pulled `clampWidth`, the overlay flip, and the drag mapping into pure exported helpers (`computeChatDrawerOverlay`, `widthForCursor`) — behavior-preserving, the hook just calls them now — and pinned each at its boundary. Verified non-vacuous by mutation: inverting the clamp min/max, flipping the `< OVERLAY_EMBED_MIN` comparison, or reversing `rightEdge - pointerX` turns the new tests red (the exact failure you called out where all DOM-level tests would stay green).
- **R2-L2 — keyboard resize** (`c51817c6`). The handle is now a focusable ARIA separator (`tabIndex=0`): ArrowLeft/ArrowRight nudge the width by a step, Home/End jump to max/min (all through the same `clampWidth` as the drag), and `aria-valuemin/max/now` report the width to AT.
- **R2-L3 — non-null assertions guarded** (`db6a7313`). Replaced `tabs[0]!` / `tabs.at(-1)!` in `activeChatViewTab` / `selectLastChatViewTab` with named `firstTab`/`lastTab` accessors that throw a diagnosable invariant error instead of silently returning `undefined` / a cryptic `.id of undefined`. The two other `!` sites (`cycleChatViewTab`, `closeChatViewTab`) are locally guarded by length checks, so they were left as-is.

## Couldn't reproduce

- **R2-L1 — composer submenu clipping.** Tried to reproduce the `.composer-reference-submenu` clipping in a real Electron runtime, including at the narrow column near the 340px min, and couldn't get it to clip — the submenu stays inside the drawer interior. Leaving it as-is, no change.

## Still open (unchanged this round)

- **R1-M1 — drag freeze over the native embed.** Still the one manual check on a real desktop build; nothing changed here. If it does freeze, the fix is the `sandboxUi.setVisible(false)` path gated on `isResizing`, as you noted.

alfredolopez80:
@zach88 — thanks for the three fixes and for the honest "couldn't reproduce" notes. We re-ran the review at HEAD `c83c78a2` (after your merge with dev today), with four independent agents plus mutation testing.

**Your three fixes are genuine.** We didn't take the mutation claim on faith — we reproduced it: 9 mutations applied to production code, 9 reds. Details and the two caveats below.

**One finding gates the merge, and a single measurement on a real desktop build settles it.** One other is a new bug we reproduced with a failing test. Everything else is follow-up.

---

## 1. Overlay mode may render the drawer *behind* the live app — needs one measurement

`styles.css:12568-12570` · `App.tsx:2001-2002` · `SandboxUiPage.tsx:487-493`

This one doesn't rest on our judgement — it follows from invariants this repo states itself:

1. The repo asserts **three times** that the native view paints above all DOM. Verbatim, `styles.css:12801-12802`: *"The native app view paints above ALL DOM, so any chat overlay that crosses the embed's rect is hidden behind it."* (Also `SandboxUiPage.tsx:495-497`, `App.tsx:604-605`.)
2. `isOverlay` has **exactly one production consumer** — the CSS class at `App.tsx:2001-2002`. Nothing hides, parks, or clips the native view in overlay mode.
3. Embed parking (`parked` in `useEmbedBounds`, `SandboxUiPage.tsx:487-493`) is wired to `shellOverlayOpen` = header ∪ command palette ∪ sidebar ∪ toasts. **The chat drawer is not in that set.**
4. In overlay, `.content-panel--chat-drawer-overlay > .page:has(.sandbox-ui-mounted-header) { padding-right: 0 }` expands the embed slot to **full width**, and `useEmbedBounds` pushes that rect to main — so the `WebContentsView` covers the rect where the drawer lives (`position: fixed; right: 18px`).

**Failure scenario:** live app + panel narrower than ~846px (340 drawer + 46 gutter + 460 threshold) → user opens the drawer → the drawer is invisible behind the embed.

The comment introducing the rule (`styles.css:12563-12565`) says *"stop pushing the embed … and let the drawer float over it"*, and the PR body repeats it. That is the opposite of what the repo's own invariant allows — one of the two is wrong.

**Honest caveat:** the code contradiction is verified; the final visual fact is not — we have no Electron runtime. **The measurement:** live app, window narrow enough to trip overlay, open the drawer. Is the drawer visible?
- **If not** → this is Critical and blocks. The fix is feeding `isOverlay` into the same `sandboxUi.setVisible(false)` / `parked` path that `shellOverlayOpen` uses.
- **If yes** → the three invariant comments are wrong and should be corrected, plus a note on why the native view doesn't win here.

Related, same root: `ChatDrawer/index.tsx:7-9` claims *"the embed slot is shrunk with padding-right so the native WebContentsView **never** overlaps it"* — only true in docked mode.

> Context: your Risks section already lists the overlay threshold under "pending live visual calibration", so this is the same area you flagged — we just have a concrete mechanism to check now.

## 2. A second `mousedown` during an active drag leaks the first drag's listener permanently — REPRODUCED

`useChatDrawerResize.ts:159-176`

The idempotency guard `if (endDragRef.current !== endDrag) return` (`:160`) runs **before** the `removeEventListener` calls. `onResizeHandleMouseDown` neither checks `event.button` nor ends the previous drag: `:170` overwrites `endDragRef.current` without invoking the old `endDrag`. From then on `endDrag1` **cannot deregister itself** (its own guard prevents it) and its `onMove1` stays on `window` forever. The cleanups at `:215-219` can't repair it — they only call `endDragRef.current`, already `null`.

React's `onMouseDown` fires for **any** button, so a right-click on the 8px handle strip while the left button is held is enough to enter twice.

Reproduced with a failing test against this HEAD:

```
expected 700 to be 420
```

After releasing the button, a bare `mousemove` resized the drawer from 420 to 700px, with `isResizing === false` and no button held. It survives navigation (listeners are on `window`).

Not a vacuous red: with the fix applied the test goes green and the 25 resize/keyboard/ChatDrawer tests stay green.

**Minimal fix:** call `endDragRef.current?.()` on entry to `onResizeHandleMouseDown`, before installing the new drag. Adding `if (event.button !== 0) return` also stops a bare right-click from starting a drag.

**Sturdier alternative:** move each closure's own `removeEventListener`/`cancelAnimationFrame` **before** the guard, and let the guard protect only the shared state (`body`, `setIsResizing`, `endDragRef`). A closure can only remove its own listeners, so this is safe, kills the leak regardless of runtime, and also covers the escalation path of R1-M1 below.

---

## Verification of your three fixes

| | Verdict | Evidence |
|---|---|---|
| **R2-M1** `8245ed4e` | **Genuine** | Diff `8245ed4e^`→`8245ed4e` compared line by line: `clampWidth` only gains `export`; `computeChatDrawerOverlay` is a literal extraction; `widthForCursor` wraps exactly `clampWidth(rightEdge − pointerX, panelWidth)`. No change to min/max order, `<` vs `<=`, sign, rounding or coercion — behavior-preserving confirmed. 6 mutations → 6 reds. |
| **R2-L2** `c51817c6` | **Genuine** | Proved empirically that keyboard and drag share one clamp path: mutating **only** `clampWidth` reddens the pure drag tests *and* all 3 keyboard tests simultaneously. `preventDefault` fires only on handled keys (`default: return` exits first). |
| **R2-L3** `db6a7313` | **Genuine** | Mutation M9 (reverting to the silent `!`) → 2 exact reds. Your guard claim audited line by line: `cycleChatViewTab` is solid (modulo over `length ≥ 2`); `closeChatViewTab` is correct today, with a caveat below. |

**Two honest caveats on the above:**

- **The `min = 340` boundary is not pinned by a literal** in the pure-logic suite R2-M1 adds. Those tests import `CHAT_DRAWER_MIN_WIDTH` and assert against the symbol, so mutating 340→341 leaves **all 17 hook tests green** — the only red comes from a literal `'340'` in the ARIA test of a *different* commit (`c51817c6`). `max = 820` *is* literal-pinned in the pure file. Worth closing that asymmetry, since this is exactly the drift the commit is meant to catch.
- **`closeChatViewTab:191-196`** is the guard-on-input / index-on-filtered pattern. It doesn't fire today (ids come from a monotonic counter and `openPersistedChatViewTab` dedups), but with duplicate ids `filter` would drop more than one → `tabs[-1]` → the same cryptic TypeError this commit set out to eliminate. So "locally guarded by explicit length checks" is slightly imprecise here: the guarantee also depends on a **non-local** id-uniqueness invariant, not the length check alone.

## Your two "couldn't reproduce" calls

- **R2-L1 (submenu clipping) — correct, but for a different reason than stated.** The drawer clipping is real: `.chat-drawer { overflow: hidden }` (`styles.css:12604`) plus `.chat-drawer__surface { overflow-y: auto }` (`:12725`, which computes `overflow-x: auto`, clipping both axes). No `createPortal` anywhere in the composer tree, so nothing escapes. What actually saves it: the submenu has no `width`/`max-width`, only `min-width: 210px`, and shrink-to-fits against a 34px containing block with ellipsised labels — so it measures 210px flat. At 340px the right edge lands at 268px of 338px usable. **The guarantee is fragile and undocumented**, though: the `agentSelector` slot at `ComposerPanel.tsx:639` is dead (no call site passes it), which is what keeps the trigger left-anchored. Pass it someday and the trigger moves toward the centre → `170 + 40 + 210 = 420 > 338` → clip. Worth a comment or a test.
- **R1-L2 (`isOverlay` stale) — your premise holds.** `.content-panel` (`App.tsx:1992`) is the ancestor of the drawer itself (`App.tsx:2078`), with no `display:none` anywhere in its chain, so it is genuinely always mounted while the drawer is active. The early-return would be transient and the ResizeObserver reconverges. No action.

## R1-M1 — still open, and your read is exact

Confirmed `chatDrawerResize.isResizing` is consumed **only** at `App.tsx:2104` as a cosmetic prop; it does not reach `shellOverlayOpen` or `SandboxUiPage`'s props. Also confirmed the `SandboxUiPage.tsx` changes in this PR come from the original feature commits (`94196740` Phase 1, `7828e0fb` Phase 4), not a silent fix.

Anatomy, verified in code: no `setPointerCapture` (window listeners only); no DOM shield during the drag (`is-resizing` only paints the handle accent); if the drag freezes, listeners plus `body.cursor: col-resize` / `userSelect: none` and `isResizing = true` all stay stuck. **And it escalates into finding 2**: a second `mousedown` after a frozen drag produces the permanent leak.

The keyboard resize you added is a real workaround for "can't widen by drag" — but it does **not** repair the orphaned listener, which overwrites the width on every mouse move. The sturdier teardown fix in finding 2 closes this regardless of what the runtime does.

---

## Follow-ups (non-blocking)

- **`composer.focus` (Mod+Shift+L) is inert in exactly the state this PR introduces.** `composerAvailable` (`App.tsx:1271-1274`) requires `navItem === chat`, and `executeDesktopCommand` bails at `isDesktopCommandEligible` (`:1326`). With the drawer open there *is* a visible composer, the drawer flow bumps `composerFocusRequestId` (`:398`, `:425`), and the executing branch is already drawer-aware (`:1454-1458`) — it's just never reached. `chat.newTab` and `chat.switcher` were made drawer-aware; this one was missed.
- **Home/End are inverted relative to the ARIA value semantics the handle itself exposes.** The code does Home → MAX (820), End → MIN (340), which is spatially coherent (handle on the left edge, Home = "all the way left" = widen), and the arrows are right. But the handle reports `aria-valuenow` = drawer width with `aria-valuemin=340`/`aria-valuemax=820`, and the separator/slider convention binds Home → valuemin. A screen-reader user pressing Home hears the value jump to its **maximum**. Either flip Home/End, or redefine the reported axis (splitter position rather than drawer width). The "(WAI-ARIA window-splitter pattern)" label at `:181-182` doesn't hold for Home/End as written.
- **Ghost focus before ready.** `styles.css:12590-12614` hides the not-ready drawer with `opacity: 0; pointer-events: none`, not `display`/`visibility`. A `tabIndex=0` under `opacity: 0` is **still focusable**, so between mount and the embed's bounds-ack, Tab can land on the handle and the invisible CTAs. The test named *"stays hidden and inert"* asserts opacity/pointerEvents but not non-focusability. `visibility: hidden` for the not-ready state, or `inert` on the aside, closes it. (No ghost when the drawer is closed — `App.tsx:2084` unmounts it.)
- **The "does this notification surface in the drawer?" predicate lives in two files.** `handleOpenNotificationInDrawer` (`App.tsx:449-464`) re-encodes the controller's internal routing. If the controller gains a `kind` that navigates elsewhere, App won't know and will open the drawer for nothing — the exact failure the comment at `:448` aims to avoid. Exporting `notificationSurfacesInDrawer(notification, currentTeamId)` from the controller module would make it an invariant instead of hand-synced convention.
- **Two unrelated files carry format-only churn from an evil merge.** `git blame` attributes both hunks in `hookFetcher.ts:51-53` and `k8sClient.ts:43` to merge `c69a4f25`, with content present in **neither parent** (a formatter appears to have run during conflict resolution). Both are semantically inert — we checked `sanitizeForLog` byte for byte (same two regexes, same order, same replacements, so the CRLF/log-injection defence is intact), and the moved import only crosses an `import type` block, which is erased at compile time. But the resulting format diverges from **both** dev and `feat/prettier-diff-ci`, so it'll churn on future merges and may redden that gate. Re-resolving both hunks to dev's byte form would avoid it.
- Minor: three constant pairs are hand-duplicated between TS and CSS (340/420, 46 = `--space-2` + 36, 18 = `--space-4`) with no test tying them — all three are documented with cross-referencing comments, which is why this is low. And `useChatDrawerResize.ts:21` calls it "the **old** CSS clamp", but that clamp is still live as the pre-mount fallback at `styles.css:2166` — someone trusting the comment and deleting it would break the closed state.
- The body's *"fullscreen chat is left untouched"* is slightly inexact: CSS-wise yes (everything is scoped under `.chat-drawer`), but the PR does modify `lib/chatViewTabs.ts`, whose own doc-comment says *"Used by both the full-screen chat route and the app drawer"*, and changes a silent fallback to a throw there.

---

## Suites — measured, not asserted

Captured with `> log 2>&1; echo EXIT=$?` so the exit code is the command's, never a pipe's. Node 24, `npm ci` without `--ignore-scripts`, `verify:electron` green first.

| | Exit | Result |
|---|---|---|
| `npm test` @ `c83c78a2` | 0 | **200 files · 1974 tests · 0 failures** |
| `npm test` @ `origin/dev` | 0 | 193 files · 1924 tests · 0 failures |
| `build:main` (tsc) / `build:renderer` (vite) | 0 / 0 | clean |

**+7 test files, +50 tests, zero regressions.** Your "965 renderer UI suite" doesn't conflict — that's the renderer subset, 1974 is all of `desktop-app`. Worth noting the `sandboxUiFixtureFailureMarker.test.ts` failure from the last round now passes, so there is no unrelated failure left.

- **Security**: the new invariant message is a static string with no chat/team ids or titles, so nothing leaks into logs or crash reports. All **four** width write paths go through `clampWidth`, and the only writer of `--chat-drawer-width` is `App.tsx:2007` reading hook state. Zero persistence, matching the "session-only by design" claim. The preload allow-list grew by **exactly** `chat.switcher` (`actionOwner: 'renderer'`, non-privileged); no new IPC channels; `desktop-app/src/main.ts` isn't in the diff at all, so Electron isolation is untouched. The keydown handler reads only `event.key`, and there is no `stopPropagation` anywhere in the diff, so global shortcuts and composer focus are unaffected.
- **Repo prohibitions**: no placeholders; **zero `catch` blocks added** in the whole diff; no test-only branches in production. The direction is the right one — tests derive expectations from production (the z-index test reads the real `tokens.css`; the ChatSwitcher fixture is derived from the `chatViewTabs` producers).
- **Hook correctness**: no stale closures (`widthRef` refreshed in render; `rightEdge` deliberately frozen and documented); listener add/remove symmetry is correct with identical references; rAF cancelled in `endDrag`; `ResizeObserver.disconnect()` on cleanup. The only crack is the re-entrancy in finding 2, not the symmetry. Observer and drag can't fight: the drawer is `position: fixed` and the gutter is `padding-right` on the child `.page`, so resizing the drawer doesn't change the panel's `clientWidth`.
- **Mutual exclusion of the two drawers is a real invariant, not a convention** — `notificationTrayUsesDrawer` and `appNotificationDrawerOpen` (`App.tsx:534-535`) are *derived*, not stored, so `appNotificationDrawerOpen && chatDrawerVisible` is unsatisfiable in every render. No future caller can break it without editing that one expression. Nice.
- The reconciler property tests, IPC contract, z-index layering and backward compat were verified clean in round 2 and are untouched by this round's commits — not re-audited.

---

*Automated multi-agent review (four independent agents plus mutation testing) against the code at HEAD `c83c78a2`. Finding 2 is reproduced by a failing test; finding 1 is a verified code contradiction whose visual outcome still needs one check on a real build. Advisory input for the human reviewers — not a formal GitHub approval.*

alfredolopez80:
@zach88 — round 4, re-reviewed at head `3b53eba0` **merged with `origin/dev`** (`1040d53f`, clean, zero conflicts; the merge touches only `control-ui/**`, so the desktop-app scope is exactly this PR's). Seven agents in parallel plus adversarial mutation testing in isolated worktrees.

**Both of the round-3 blockers are closed, and not on your word — every claimed fix was mutated until its test went red.** Two independent agents reproduced the whole ledger separately: baseline 46/46 green, 9 mutations, 9 nominal reds, post-restore 46/46 green.

**What gates the merge now is new: one HIGH we reproduced, which is the side effect of the very fix that closed the Critical.** Plus a formatting gate that is genuinely failing, and a PR body that now documents a design that no longer exists.

---

## 1. A new chat can become unreachable — REPRODUCED, confirmed twice

`App.tsx:399-437` (`handleNewChatViewTab`, `openChatDrawer`) · `App.tsx:549`

`App.tsx:549` publishes `chatDrawerAvailableRef.current = chatDrawerAvailable` — **not** `chatDrawerVisible`. The width suppression (`!panelTooNarrow`) lives only in the last hop of the `available → desired → visible` chain, but the ref the callbacks read is published from the **first** hop.

**Failure scenario:** live app, panel narrower than 846px, drawer closed → `Mod+T` creates the blank tab, reveals it *in the drawer*, and skips `leaveSandboxForChat()`. The drawer is suppressed, so `navItem` stays `sandbox-ui`, there is no drawer and no chat route. The chat exists in state and no surface can reach it. The `SandboxUiPage` toggle does not rescue it either: it calls `openChatDrawer()`, the intent was already `true`, so it stays suppressed.

**Evidence:** probe → `navItem=sandbox-ui, switcherPresent=false`. Independently confirmed by reading `App.tsx:543/547/549` from `HEAD`.

The three-level chain is the right design — it is what makes suppression atomic and what closed the Critical. What is unfinished is that a `ref` deliberately bypasses React's data flow, so nothing in the compiler or the tests distinguishes `chatDrawerAvailable` from `chatDrawerVisible` there.

**Recommended fix — fold the suppression into `chatDrawerAvailable` itself, not into each opener.** `chatDrawerAvailable` answers *"is the drawer a valid surface to divert this gesture to?"*, and it has **four** consumers: `:402` (newTab), `:411` (`openChatDrawer`), `:465` (notification diverter) and `:864` (`launchSandboxUiApp`). Patching openers one by one covers the two we reproduced and leaves the contract broken for the other two and for every future caller.

> ⚠️ Fold it into `chatDrawerAvailable`, **never** into `chatDrawerDesired`. `chatDrawerDesired` is the `active` that feeds the measuring hook (`App.tsx:546`); suppressing it there would stop the `ResizeObserver` from observing the re-widen, and the drawer would never come back. That turns "does not appear when it should" into "never returns", which is worse.

## 2. `composer.focus` (Mod+Shift+L) is inert in exactly the state this PR introduces — REPRODUCED

`App.tsx:1323-1326`

`composerAvailable` still requires `vm.navItem === DESKTOP_ROUTES.chat`, so with an app open and the drawer showing a loaded chat, `isDesktopCommandEligible` returns `false`, `executeDesktopCommand` bails at `:1378`, and the already-drawer-aware branch at `:1506` is never reached. There is a visible composer the shortcut cannot focus.

**Evidence:** A/B probe in the same test — chat route `composerFocusRequestId 1 → 2`; with the drawer open `3 → 3`, `isEligible('composer.focus') === false`.

`git blame` puts the line at `7349a8186` (pre-existing), but this PR creates the state in which it becomes wrong. `chat.newTab` and `chat.switcher` were made drawer-aware; this one was missed. Same shape as the item we flagged last round — it just now has a repro.

## 3. The duplicated notification predicate has already diverged — REPRODUCED

`App.tsx:462-475` vs `useAppController.ts:1035`

Last round this was a design risk ("the predicate lives in two files"). It is no longer a risk. `handleOpenNotificationInDrawer` re-encodes the controller's routing but is **missing the `if (!targetAgent) return` guard** that `useAppController.ts:1035` has.

**Failure scenario:** a conversation notification with an empty `agentName` → the drawer opens with nothing loaded. **Evidence:** probe → `drawerOpen=true`.

What *is* sound and was checked: `AppNotificationKind` has exactly 4 values (the workflow/sdk filter is exhaustive today), and `vm.getCurrentTeamId()` and `currentTeamIdRef.current` are the same reference (`useAppController.ts:729`), so the team check cannot drift. Exporting `notificationSurfacesInDrawer(notification, currentTeamId)` from the controller would make the whole thing an invariant instead of hand-synced convention.

## 4. `prettier --check` fails — and CI cannot tell you

`desktop-app/ui/src/components/ChatSwitcher/index.tsx:101` (blame `6c9036d1`) — 1 of 28 files, `EXIT=1` with prettier 3.8.2. The `ChatStateBadge` JSX exceeds the print width. Mechanical fix.

Worth knowing *why* this survived: `grep -rln 'prettier|format:check|style-rules' .github/workflows/` returns **empty**. No CI workflow runs prettier, so the 37/37 green checks are not evidence of formatting compliance — the only gate is the local pre-commit hook. The PR body's *"Pre-commit gates (prettier format, style-rules, commitlint) pass"* is false at head.

## 5. The PR body now contradicts the code in five sentences

After `7d239f73` and `844449f9`, these are false today:
- *"with an overlay fallback for small windows"* (What)
- *"Overlay is decided from a ResizeObserver … the drawer stops pushing the embed and floats over it"* (How, the whole paragraph)
- *"Home/End jump to max/min"* (How — it is min/max now, deliberately)
- *"Width is clamped to [340, min(820, panel-80)]"* (How — it is `panel − 460 − 46` now)
- *"The overlay threshold (460px)"* (Risks — there is no overlay mode)

A reviewer or QA validating against the body will read correct behaviour as broken.

---

## Verification of your fixes — every one genuine, proven by mutation

| Item | Verdict | Mutation → result |
|---|---|---|
| **Overlay behind the embed** `7d239f73`+`40ef83fc` | **GENUINE** | Zero residue of `isOverlay` / overlay CSS in `ui/src`. The replacement cannot reproduce the invisibility: below 846px the drawer is **not rendered at all**. `CHAT_DRAWER_MIN_PANEL_WIDTH` is derived (340+46+460), never a literal |
| **Re-entrant drag leak** `b1f90d78` | **GENUINE — and you took the structural fix, not the narrow patch** | Repro against `b1f90d78^` under head's tests → **2 reds** (`expected 820 to be 420`); head green. `:202-212` removes listeners and rAF **unconditionally before** the guard, so each closure detaches itself even if a drag ends abnormally |
| `f0314177` non-left-button guard | **GENUINE** | Caught in both directions |
| `c067ba71` literal 340 pin | **GENUINE** | `340→341` → 1 red. Closes the asymmetry we flagged |
| `844449f9` Home/End vs ARIA | **GENUINE** | Re-inverting → 1 red. Head: Home→340 / End→820, consistent with `aria-valuemin/max/now` |
| `b40a8110` inert + deferred focus | **GENUINE** | Removing `inert` → 1 red; removing the `chatDrawerReady` gate → 1 red |
| `5d6d31f5` `widthForCursor` removed | **GENUINE** | Genuinely dead (0 uses); inverting the drag mapping still yields 1 red, so removing it did not drop the pin |
| `8245ed4e` / `c51817c6` / `db6a7313` | **NOT REGRESSED** | — |

One note on the leak fix, useful for future refactors: it has **two redundant defences** (moving the removals ahead of the guard, and calling `endDragRef.current?.()` on entry). Reverting *either one alone* leaves the test green; only reverting both reproduces the bug. That is real robustness, but it means neither line is individually pinned.

**R1-M1 (drag freeze over the native embed) is still open, and your read remains exact** — `isResizing` reaches only `resizing={...}` at `App.tsx:2164` → a CSS class; nothing reaches `sandboxUi.setVisible` or parking. It still needs a real desktop build, as your Risks section says. **Its worst consequence is now closed**, though: the structural teardown means a frozen drag can no longer escalate into the permanent listener leak.

Two of your earlier calls, re-adjudicated: the **submenu clipping** is now partially addressed (the flight direction is documented at `styles.css:12908-12916` and pinned by a test) — what remains undocumented is the width shrink-to-fit, plus the dead `agentSelector` prop that would break it if ever passed. And the **evil-merge format churn** in `hookFetcher.ts`/`k8sClient.ts` is gone from the net diff after the dev sync.

---

## Other findings

**MEDIUM**
- **`closeChatViewTab` still carries the `!` that `db6a7313` announced it removed** — `chatViewTabs.ts:194-197`. Guard-on-input / index-on-filtered: with duplicate ids, `filter` drops more than one entry. Two branches, and the silent one is worse — it returns `{ tabs: [], activeTabId }` with no noise, deferring the crash to the next render where `activeChatViewTab` (`App.tsx:1325`, `ChatSwitcher/index.tsx:32`) throws **on the render path**. Reproduced against the unmutated module: `ALL-DUP THROWN: TypeError ... reading 'id'` and `PARTIAL-DUP RESULT: ["C"]` (3 tabs in, 1 out). The invariant does hold today, but it spans two files, is written down in neither, and no test asserts id uniqueness for the producers that mint ids. The function itself is pre-existing (`993bf46d8`); what this PR adds is the docblock's claim of completeness.
- **The switcher can open itself and steal focus** — `App.tsx:296` + `:839-844`. `pendingChatSwitcherFocusRef` is only cleared inside the effect requiring `chatDrawerVisible && chatDrawerReady`. Narrow panel + `Mod+E` → nothing happens (correct); the user widens the window later for an unrelated reason → the dropdown opens by itself. Reproduced: `options=1`, expected `0`.
- **Auto-select diverges between drawer and full-screen** — `useAppController.ts:1051-1059` vs `:1086`. The `stayInDrawer` branch passes `selectLatest: false` and omits `chatId`, so `handleSelectChatAgent` computes `suppressAutoSelect: true` and runs `clearActiveChat()`; the full-screen branch passes no options. An agent notification without `chatId` lands on the agent's last chat full-screen, but shows a blank composer in the drawer.
- **Two of the four sidebar auto-collapse tests are vacuous** — `App.deepLinks.test.tsx:299` and `:325`. Deleting the guard at `App.tsx:674` leaves **34/34 green**. The fixture starts expanded, so the simulated manual gesture is *expand*, whose value coincides with the memorised pre-app state — the assertion cannot tell "respected the user" from "restored a stale ref". The discriminating case is the opposite gesture, verified both ways: guard deleted → `collapsedAfterClose=false` (the app re-expands over the user); head → `true`. **The production guard is correct and load-bearing; what is missing is the test.**
- **No coverage for any teardown that is not `mouseup`** — `useChatDrawerResize.ts:218, 256-260`. Nothing covers ending a drag via window `blur`, via `active→false` (hide mid-drag), or via unmount, nor the restoration of `document.body.style.cursor/userSelect`, nor `observer.disconnect()`. Deleting the `blur` listener or both cleanups leaves **26/26 green**.

**LOW** (12, abbreviated): `aria-valuemax` is the static 820 while the effective max is the dynamic cap, so End announces an unreachable target; the drag tests only ever exercise the right-edge fallback branch, never `getBoundingClientRect().right`; the three TS↔CSS constant pairs (340/420, 46, 18) still have no CSS-side test; `.sandbox-ui-chat-drawer-btn` has no rule in any stylesheet, so the "Chat" toggle renders unstyled next to its styled siblings; a 1-frame premature-ready race when an in-flight `setBounds` resolves after the anti-flash gate rearms; the preload↔registry sync check is one-directional, so a stale extra allow-list entry would pass forever (pre-existing `e109e12a5`); neither `App.chatDrawer.test.tsx:330-332` nor `ChatSwitcher.test.tsx:69-91` asserts `document.activeElement`, so neither layer verifies the focus that `b40a8110` exists to guarantee; `aria-haspopup="listbox"` without `aria-controls` and a listbox without `id`; neither app-close nor identity reset clears `chatDrawerOpen`/`pendingChatSwitcherFocusRef`, so opening a different app without `conversationOrigin` unfurls the drawer with the previous conversation; `void chat.switchToChat(...)` at `useAppController.ts:995` discards the promise with no UI signal (the full-screen path awaits inside try/catch); the anti-flash gate has no exit, so if the embed never acks bounds the drawer stays `opacity:0` + `inert` indefinitely with the gutter reserved; and `3b53eba0`'s *"Repro-test: none — CSS media-query specificity is not evaluated by jsdom"* overlooks that the repo already asserts on stylesheet text (`ChatDrawer.test.tsx:95-99`).

**Detection duty — pre-existing, outside this PR, not for you to fix here:** `App.tsx:628,647` `.catch(() => undefined)` swallows `sandboxUi.setVisible` failures (blame `3559031b6`), and `useAppController.ts:1082-1084` has a silent `catch {}` fall-through after a failed `switchToChat` (blame `5d663a49`).

---

## Suites — measured, not asserted

Isolated worktree, Node v24.18.0, `npm ci` `EXIT=0` and `verify:electron` `EXIT=0` confirmed **from file** before running anything. Every exit code captured with `cmd > log 2>&1; echo EXIT=$?` — never through a pipe.

| | Exit | Result |
|---|---|---|
| `npm test` (desktop-app) | **0** | 222 files · **2157 tests · 2157 passed · 0 failed · 0 skipped** |
| `build:main` (tsc) | **0** | clean |
| `build:renderer` (vite) | **0** | 569 modules; one non-blocking 958KB chunk warning |
| `prettier --check` (28 PR files) | **1** | **1 file fails** |
| `style-rules` (same scope) | **0** | 0 violations |
| `commitlint` (head commit) | **0** | clean |
| GitHub CI | — | **37/37 pass** — but it does **not** include prettier |

**+183 tests over the 1974 we measured last round, zero regressions.**

- **Security: zero findings across nine dimensions**, all re-derived from `git show HEAD:` / `git grep … HEAD` rather than the working tree. `desktop-app/src/main.ts` is **absent** from the 28 files; the isolation grep (`contextIsolation|sandbox:|nodeIntegration|webSecurity|setWindowOpenHandler|executeJavaScript|insertCSS|CSP|ipcMain|loadURL|eval(`) returns **zero hits in code**. The preload allow-list grew by **exactly one line**, zero-arg and non-privileged, fail-closed at both boundaries. Note that `sources: ['host','sandbox']` does **not** hand the embed an IPC channel — `shortcutRouter.ts:78` emits only to `route.trustedRenderer`. No HTML/URL sinks; the CSS custom properties are `Math.round()` of measured numbers, never external strings. The cross-team guard is enforced in **production** at `App.tsx:468`, not merely asserted in a test. The invariant error is an identical literal at `:12` and `:18` with no interpolation, and the diff adds **no** `console.*`/logger call at all. All width write paths go through the clamp (drag `:187`, keyboard `:251`, render `clampWidth` at `:125`), one writer of the CSS variable, and no persistence — session-only confirmed. **No `stopPropagation` anywhere in the diff.**
- **Repo prohibitions:** zero `catch` blocks added, zero placeholders, zero production branches gated on `NODE_ENV`/`isTest`.
- **Geometry:** overlay removal is complete — `styles.css` is 3 net hunks, no overlay classes, no `isOverlay`, no full-width expansion rule, no dead variables. `padding-right = var(--chat-drawer-width) + 46px` on `.page` against a drawer fixed at `right: 18px` gives a constant 46px clearance in every state, and suppression is atomic: `chatDrawerVisible` governs class **and** mounting in the same render. Fullscreen chat genuinely untouched (everything scoped under `.chat-drawer`). The stacking test reads the real values from `tokens.css` and the literal `z-index: 4000` from the actual `.toast-stack` rule — the right pattern.
- **IPC contract:** `'chat.switcher'` is the identical literal in the registry (`:101`), the preload allow-list (`:15`) and the call site (`App.tsx:1468`); types are **derived** from the registry with no hand-written union. No `(group, order)` pair is duplicated anywhere in the registry. The additive signatures are genuinely inert for old callers (`keepNavItem?` enters purely as `&& !options.keepNavItem`; `focusRequestId = 0` returns early on `<= 0`).
- **Throw semantics of `db6a7313`:** unreachable today — no persistence/hydration path, both seeds are literals, `reconcileChatViewTabs` only maps or appends. And no caller relied on the old silent `undefined`: all thirteen sites dereference on the next expression, so every one already threw — the commit only names the error.
- **`contentPanelRef` mount ordering:** cannot be `null` with `active = true`. Refs attach in the commit phase before passive effects, and the `<section ref={contentPanelRef}>` at `App.tsx:2046-2047` is unconditional with `isAuthenticated` as its only ancestor gate.
- **`ChatSwitcher` fixtures** are genuinely derived from the real producers, so a change to `ChatViewTab`'s shape breaks compilation rather than passing silently.

---

## Recommendation

**Gates the merge:** finding 1 (with the `chatDrawerAvailable` fix and the `chatDrawerDesired` caveat), the prettier gate, and correcting the five body sentences.

**Worth doing in this PR:** findings 2 and 3, plus the self-opening switcher — all three are the same pattern (inherited guards reasoning about `navItem === chat`, or a predicate re-encoded by hand), all three reproduced, all three cheap.

**Follow-up:** the remaining MEDIUMs and the 12 LOWs. R1-M1 still needs one measurement on a real desktop build, exactly as you flagged.

**Honest caveats:** we have no Electron runtime, so nothing visual was confirmed on a real build — R1-M1 and the submenu width behaviour remain unmeasured by us. Everything marked REPRODUCED above is a failing probe against this head; everything else is stated as read.

---

*Automated multi-agent review (seven agents plus adversarial mutation testing) against head `3b53eba0` merged with `origin/dev`. Advisory input for the human reviewers — not a formal GitHub approval.*

zach88:
Round-4 fixes are in — all three merge gates are closed. Here's what changed in code, mapped to your numbers.

**1 · Unreachable new chat — fixed, but deliberately not the "fold into `chatDrawerAvailable`" version.**
That fold reintroduces the exact cycle you flagged for `chatDrawerDesired`: `chatDrawerAvailable` feeds `chatDrawerDesired`, which is the `active` input to `useChatDrawerResize` — suppressing it there stops the `ResizeObserver` from ever seeing the re-widen ("never returns"). So instead of touching the chain I added a parallel signal:

```ts
const chatDrawerDivertable = chatDrawerAvailable && !chatDrawerResize.panelTooNarrow
```

published to its own ref (`chatDrawerAvailableRef` → `chatDrawerDivertableRef`) and read by the two divert-decisions, `handleNewChatViewTab` and `handleOpenNotificationInDrawer`. Below 846px they fall to their existing `else` branch and **eject to the full-screen chat route**. That's the intended UX, not a regression: at that width you operate the app *or* full-screen chat — there's no room for both — so a new-chat gesture goes full-screen instead of into a suppressed drawer. The invariant `chatDrawerVisible ⟹ chatDrawerDivertable` (both derived in the same render pass) guarantees the `else` never lands in a hidden drawer. Repro-test added in `App.chatDrawer.test.tsx` (Mod+T at width 800 → `navItem === chat`); it fails on the parent commit. — `bbbab008`

**2 · `composer.focus` inert in the drawer — fixed.**
Widened `composerAvailable` to `(navItem === chat || chatDrawerVisible)` and added `chatDrawerVisible` to the memo deps. The command's reveal was already drawer-aware; the eligibility gate was the only thing pinning it to the chat route. — `b032d7aa`

**3 · Duplicated notification predicate — guard restored.**
Added the missing `!targetAgent` guard to `surfacesInDrawer`, mirroring the controller's `if (!targetAgent) return` with the same `String(agentName || '').trim()` normalization. I did the mirror guard, **not** the full `notificationSurfacesInDrawer` extraction you suggested — that's the right structural move, but it's exactly that (structural), so I kept it as a follow-up rather than doing it inside a fix commit. — `a28753c9`

**4 · `prettier --check` — fixed.** Reflowed the `ChatStateBadge` JSX; the whole renderer scope is `prettier --check` clean now. — `715aef9f`

**5 · PR body — corrected.** Rewrote the five stale sentences: dropped the overlay mode, clamp is now `[340, min(820, panel − 460 − 46)]`, Home/End jump to min/max, and "embed floor (460px)" instead of "overlay threshold". Updated in the description above.

**Not touched this round (being upfront):**
- The **self-opening switcher** (MEDIUM) — same root cause as #1 and reads the same `divertable` signal, so it's a small follow-up; not in yet.
- The other MEDIUMs (`closeChatViewTab`'s `!`, the drawer-vs-full-screen auto-select divergence, the two vacuous auto-collapse tests, non-`mouseup` teardown coverage) and the 12 LOWs.
- **R1-M1** (drag freeze over the native embed) — still needs the one measurement on a real Electron build, exactly as you called out.

Each fix is one commit carrying its own `Class-check` / `Repro-test`, and code + security review came back clean on each.

alfredolopez80:
@zach88 — round 5, verified at head `47b191c9` (which already contains `origin/dev` `0c8551cf` as an ancestor, so this time no synthetic merge on our side). Adversarial mutation verification plus measurement, in isolated worktrees.

**Four of the five merge gates are closed and mutation-verified. The fifth has a residual hole we reproduced — and per discussion it is accepted as a documented risk rather than a blocker. Nothing blocks the merge.**

---

## First: you were right and our round-4 recommendation was wrong

We told you to fold the suppression into `chatDrawerAvailable`. You rejected that and gave the reason. The code confirms your reason, not ours:

```
:544  chatDrawerAvailable = navItem === apps && Boolean(activeSandboxUiApp)
:554  chatDrawerDesired   = chatDrawerAvailable && chatDrawerOpen
:557  useChatDrawerResize(contentPanelRef, chatDrawerDesired)   ← the hook's `active`
```

`chatDrawerAvailable` is **upstream** of `chatDrawerDesired`, so folding there is folding into `chatDrawerDesired` transitively: `tooNarrow=true` ⟹ `desired=false` ⟹ the observer is disconnected ⟹ `panelWidth` freezes narrow ⟹ `tooNarrow` latches true forever. That is exactly the "never returns" failure our own caveat warned about — we named the wrong node. Our advice would have turned "does not appear" into "never comes back."

Your parallel-signal design is the correct call, and the invariant holds by construction in every render, not just the common case:

```
chatDrawerVisible    = available && open && !narrow
chatDrawerDivertable = available         && !narrow
⟹  visible = divertable && open  ⟹  visible ⟹ divertable  ✓
```

## Verification of your four fixes — all genuine, each mutated to red

| Fix | Verdict | Mutation → result |
|---|---|---|
| **1 · `bbbab008`** divertable guard | **GENUINE** (with a residual hole, below) | `App.tsx:564` → parent semantics → **1 RED**: *"routes a new-chat gesture to full-screen chat when the panel is too narrow…"* `expected 'sandbox-ui' to be 'chat'` |
| **2 · `b032d7aa`** `composer.focus` in drawer | **GENUINE** | Dropping `\|\| chatDrawerVisible` at `:1345` → **1 RED**. The memo dep array genuinely includes `chatDrawerVisible` — checked, because a missing dep would go stale rather than fail on first render, and no test would catch that |
| **3 · `a28753c9`** `agentName` mirror guard | **GENUINE — and an exact mirror** | Removing the guard at `:478` → **1 RED** (`expected true to be false`). Predicates diffed line by line: the controller's falsy check on `String(x \|\| '').trim()` and the wrap's `!== ''` on the same expression are logically identical. We checked specifically because a *near*-mirror is worse than none — it looks synchronised while drifting |
| **4 · `715aef9f`** ChatStateBadge reflow | **GENUINE** | Pure JSX attribute reflow, same props, zero behaviour change |
| **5 · PR body** | **CORRECTED** | Zero remaining hits for `overlay fallback`, `floats over`, `stops pushing the embed`, `overlay threshold`, `panel-80`, `max/min`. The arithmetic now reads 846 = 340 + 460 + 46, which matches the code |

We also checked `059f8e7c` ("soften over-promising comments"), because *"softening comments"* is exactly how a weakened assertion would be described. It is **prose only** — the `waitFor`/`expect` lines are byte-identical, and the added comment in `SandboxUiPage.test.tsx` correctly documents that the multi-overlay test does not subsume the single-overlay ones. Nothing weakened.

---

## The residual hole — REPRODUCED, and accepted as a documented risk

`App.tsx:564` + `useChatDrawerResize.ts:123-126,145`

```
:123  const [panelWidth, setPanelWidth] = useState(0)
:126  panelTooNarrow = panelWidth > 0 && panelWidth < CHAT_DRAWER_MIN_PANEL_WIDTH
:145  if (!active) return        ← active = chatDrawerDesired = available && chatDrawerOpen
```

`chatDrawerDivertable` consumes `panelTooNarrow`, but the panel is only measured while the drawer is *desired*. If the drawer was **never opened**, `chatDrawerOpen` is false ⟹ `active` false ⟹ the measuring effect returns early ⟹ `panelWidth` stays `0` ⟹ `panelTooNarrow` reads **false** ⟹ `chatDrawerDivertable` is true **at any window width**.

**Failure scenario:** launch an app from the picker with the window under 846px, having never opened the drawer → `Mod+T` → the chat is diverted into a drawer that the very next render suppresses. `navItem` stays `sandbox-ui`, the chat is unreachable until a re-widen. The notification diverter has the same exposure, and **zero coverage** — the mutation reddened only the `Mod+T` test.

**Verified:** a probe against **unmutated** production code fails `expected 'sandbox-ui' to be 'chat'` while all 12 shipped tests pass.

**Root cause, and why nobody should feel bad about it:** the `panelWidth > 0` clause was correct and benign while `panelTooNarrow` had a single consumer (`chatDrawerVisible`, which only matters once the drawer is already desired). It became a defect the moment `panelTooNarrow` gained a **second consumer with different timing** — one evaluated *before* the measurement exists. The design is right; what went unaudited is that the datum the new signal depends on only exists under the condition the new signal is meant to anticipate.

**Fix direction (for the follow-up):** measure while `chatDrawerAvailable` rather than only while desired, or take a direct measurement when `panelWidth === 0` — keeping the hook's `active` semantics for exactly the reasons your fix-1 comment gives.

> **Decision: accepted by implementer and reviewer as a known, documented risk. Not a merge blocker.** Recorded here with scenario, evidence and fix direction so it is a decision rather than something forgotten.

**One new LOW, also accepted as pending:** `App.tsx:452-457` — with the drawer suppressed, `toggleChatDrawer` (reading `chatDrawerVisibleRef`) calls `openChatDrawer()` again: a visual no-op that also re-runs an in-drawer reveal into a hidden drawer, and the user cannot "close" a suppressed drawer via the toggle.

---

## Measured, not asserted

Isolated worktree, Node v24.18.0, `npm ci` `CI_EXIT=0` and `verify:electron` `VERIFY_EXIT=0` confirmed **from file** before anything ran. Every exit code captured with `cmd > log 2>&1; echo EXIT=$?` — never through a pipe.

| | Exit | Result |
|---|---|---|
| `npm test` (desktop-app) | **0** | 222 files · **2165 tests · 2165 passed · 0 failed · 0 skipped** |
| `build:main` (tsc) | **0** | clean |
| `build:renderer` (vite) | **0** | 569 modules |
| `prettier --check` — 28 diff files | **0** | clean |
| `prettier --check` — **all** of `desktop-app/ui/src` (457 files) | **0** | clean |
| `style-rules` — 28 diff files | **0** | 0 violations |

Your *"the whole renderer scope is `prettier --check` clean now"* is accurate under both readings — we checked the 457-file tree with an explicit `find`-generated list precisely to rule out a false green from an empty glob. **+8 tests over last round's 2157, with no drop in executed count.**

Repo prohibitions clean across all five commits: no placeholders, no unrequested fallbacks or silent `catch`, no production code shaped to pass a test, no assertion masking.

---

## What remains open — nine follow-up issues recommended

None of this blocks the merge. Grouped by root cause rather than by finding, so each issue is one coherent piece of work:

| | Issue | Contents |
|---|---|---|
| **1** | Drawer-divert decisions read state at the wrong time | The accepted HIGH above + the new `toggleChatDrawer` LOW + the self-opening switcher (`pendingChatSwitcherFocusRef` never cleared when the drawer never becomes visible). **Same root cause**, two of the three reproduced |
| **2** | `closeChatViewTab`: close the surviving `!` and pin id uniqueness | Reproduced (`TypeError` plus a silently dropped tab). The invariant spans two files and is written down in neither |
| **3** | Close vacuous-coverage gaps | The two sidebar auto-collapse tests, non-`mouseup` teardown, the switcher-focus tests that never assert `document.activeElement`, and the drag tests that only exercise the right-edge fallback branch — four places where deleting production code leaves the suite green |
| **4** | R1-M1: measure the drag freeze on a real Electron build | The one item that needs a real build, open since round 1. Its worst consequence is already closed |
| **5** | Drawer vs full-screen auto-select divergence | Same notification lands on the agent's last chat full-screen, but a blank composer in the drawer |
| **6** | Submenu clipping fragility + dead `agentSelector` prop | The dead prop clips the submenu silently the day someone passes it |
| **7** | Drawer not cleared on app close / identity reset | Opening a different app unfurls the drawer with the previous conversation |
| **8** | Silent UI failures | `void chat.switchToChat(...)` with no user-visible signal, and the anti-flash gate with no exit if bounds are never acked |
| **9** | Drawer accessibility polish | Static `aria-valuemax` of 820 against the dynamic cap, and `aria-haspopup` with no `aria-controls` on an `id`-less listbox |

**Four LOWs we explicitly do *not* think deserve an issue**, so they do not come back as noise: the hand-duplicated TS↔CSS constants (documented with cross-references, bounded consequence), the 1-frame premature-ready race (a pre-existing pattern shared with the notification drawer — fixing it only here would be inconsistent), the one-directional preload↔registry sync check (pre-existing, `e109e12a5`), and `3b53eba0`'s "Repro-test: none" note (process, not a defect).

Separately, and **not for this PR**: two pre-existing silent catches surfaced by detection duty — `App.tsx:628,647` `.catch(() => undefined)` (blame `3559031b6`) and `useAppController.ts:1082-1084` (blame `5d663a49`).

---

## Verdict

**Nothing blocks the merge.**

- Round 3's two Criticals (drawer invisible behind the native embed; permanent listener leak): **closed, mutation-verified**.
- Four of round 4's five gating items: **closed, mutation-verified**.
- The fifth: **accepted and documented** above, with a reproduced scenario and a fix direction.
- Suites 2165/2165, typecheck and build green, all three format/style gates clean, security clean across nine dimensions in the previous round and untouched by these commits.

Remaining: one accepted risk, six MEDIUMs and thirteen LOWs, folded into the nine issues above. None is a regression — all are pre-existing debt or polish.

Thanks for the round-4 turnaround, and in particular for pushing back on the `chatDrawerAvailable` recommendation instead of implementing it. That pushback was correct and it avoided a worse bug than the one it was meant to fix.

---

*Automated multi-agent review with adversarial mutation testing, against head `47b191c9`. Advisory input for the human reviewers — not a formal GitHub approval.*

alfredolopez80 (approved):
approved with comments

PR #421, opened 2026-08-20
fix(hcc): bind MCP discovery and auth to Host Context

jozer-rami:
## Summary

Bind HCC MCP inventory and `/auth` to the calling Host's current Context so a compromised first-party mcp-host cannot list or pull credentials for another Context (NP-08).

Production identity is Kubernetes TokenReview of the per-Host ServiceAccount (`host-<name>-sa`). HCC then uses `Host.spec.contextRef` from its own cache. Path Context and unsigned `x-clerum-*` headers are not authority. `/auth` does not read a Secret until the server is on that Host's Context allowlist.

mcp-host sends the projected ServiceAccount bearer on discovery/auth (not `/ready`). Hosts pin `automountServiceAccountToken: true`. The HCC gateway forwards `Authorization`. ClusterRole `host-context-controller-tokenreview` grants HCC `tokenreviews/create` only.

**Deploy order:** apply TokenReview RBAC → restart the HCC api-gateway → roll Hosts onto the mcp-host image that sends the SA bearer → then this HCC image. HCC is `Recreate`; applying HCC first 401s discovery until Hosts roll (running Hosts keep last fleet; cold start can sit on `/ready` 200 then 401).

`MCP_PROXY_ENABLED` remains false by default. mcp-proxy still polls unauthenticated global list-all and will 401 if that flag is enabled — follow-up, not a Host Context bypass.

## Testing

- [x] `npm test` passes for changed services
- [x] `npx tsc --noEmit` clean

- host-context-controller: `npx vitest run` **1624 passed** (70 files); `npx tsc -p tsconfig.build.json --noEmit` **exit 0**
- mcp-host: `npx vitest run src/contextMapperClient.test.ts` **15 passed**; `npx tsc --noEmit` **exit 0**
- control-api: `npx vitest run test/network.gateway.intent.test.ts` **41 passed / 1 skipped**

## Checklist

- [x] This is not a new third-party MCP server / WorkflowRecipe (those go to the registry)

PR #422, opened 2026-08-20
fix(hcc): close NP-08 cross-context MCP token disclosure

alfredolopez80:
## Summary

This PR implements the selected NP-08 PR1 mitigation for cross-Context MCP credential disclosure.

The credential authorization flow is:

```text
valid HCC JWT
  -> live caller Host
  -> live Host Context
  -> Context-owned MCP server
  -> Secret read
  -> post-read authority fence
  -> token response
```

## Security boundary

PR1 closes the credential path: a compromised mcp-host from Context A cannot use its Host-bound JWT to retrieve a credential for a server granted only to Context B. HCC derives the Host and Context server-side, checks the Context allowlist before reading a Secret, fails closed, and revalidates the complete authority grant after the read.

The existing `GET /api/v1/mcpservers` global inventory remains a deliberate, metadata-only compatibility residual for the currently deployed mcp-proxy. It returns server topology metadata but no Secret, auth selector, or credential value. It is not the mcp-host authorization contract and is not claimed as closed by PR1. PR2 must introduce an authenticated system-inventory principal/route for mcp-proxy, migrate the poller, and then retire the v1 global route. Adding a Host JWT directly to this global route would break its cross-Context discovery contract rather than solve the identity problem.

## Main changes

- Added strict HCC JWT verification with issuer, exact audience, capability, Host reference, Host UID and lifetime checks.
- Added live Host and Context resolution and an allowlist check before any Secret value read.
- Added pre-read and post-read fences for Host, Context, McpServer and Secret lifecycle/version changes.
- Retired the vulnerable v1 Context and per-server auth routes with fail-closed tombstones.
- Added mcp-host authority revocation, retained-state withdrawal and bounded staleness handling after HCC authorization failures.
- Preserved L2 NetworkPolicy isolation and added render, RBAC, evidence and security contract gates.
- Kept mcp-proxy out of the PR1 data-plane expansion; its authenticated system-inventory migration and v1 retirement are PR2 work.

## Validation

- Current synced base: `origin/dev` `05a52ce2045dfe26a7e26a2337d01ba962d46cb3`; PR head: `140c2bfc622edd6001236c7da3b5639ec99acd7b`.
- Local full suites on the synced head: HCC 1,681 passed; Control API 4,461 passed / 103 skipped; mcp-host 3,526 passed / 1 todo.
- GitHub CI is green for all build/test/typecheck jobs, Real Postgres, CodeQL, manifest, leak, hygiene and shell gates.
- NP-08 focused guards and security contracts pass locally, including positive same-Context, negative cross-Context, no-token-disclosure, v1 tombstone, cleanup and provenance cases.
- T2 is a separate branch-owned local Minikube evidence lane; no GKE, Clerum-dev or Clerum-prod deployment was performed. Re-run the final local T2 profile against this exact synced head before any deployment.
- No Secret values, bearer tokens or tenant identifiers are included in this PR.

## Review and disclosure boundaries

- This is a private security fix; no public security issue was opened.
- Please review rollout ordering between Control API token issuance, HCC, gateway and mcp-host.
- PR2 owns authenticated mcp-proxy system inventory, its cutover, and v1 global-route retirement.
- This PR does not authorize production deployment. Deployment and coordinated disclosure require separate maintainer approval.

## Follow-up folded into this PR — HCC staleness ceiling (issue #425)

Review found that `mcp-host` capped `HCC_AUTHORITY_MAX_STALENESS_MS` with a hard
`Math.min(env, 60_000)` that operators could only lower. A normal HCC `Recreate`
rollout in clerum-dev was measured at ~77s to readiness (startupProbe budget up
to 120s), so once this PR's staleness/revocation machinery deploys, a routine HCC
rollout would fire the staleness deadline on every `mcp-host` and tear down the
whole MCP fleet, aborting in-flight tool calls — a cluster-wide availability
regression with no operator escape hatch.

Fixed in this PR (commit `9497625d`):

- `mcp-host/src/config.ts`: the staleness window is now bounded-raisable — default
  `60000` → `180000`, operator-tunable up to a `600000` hard ceiling. The existing
  startup validation (poll interval strictly below the window) is unchanged.
- The window governs only the `unavailable` failure class (HCC unreachable / 5xx).
  Identity/401 revocations remain immediate and are not gated by it, and a grant
  change while HCC is healthy is applied by the next successful poll — so raising
  the ceiling does not delay any real authority revocation and does not widen the
  NP-08 credential-disclosure surface (which lives entirely in HCC per-request
  authorization).
- `mcp-host/src/config.hccAuthority.test.ts`: asserts the raisable window and the
  600s hard cap.
- `deploy/overlays/minikube/configmaps/mcp-host-config.yaml`: the e2e lane keeps
  its `60000` pin (fast, deterministic revoke-on-staleness) — documented inline.
- `scripts/tests/test-np08-render-contract.sh`: assertion message clarified as the
  minikube e2e-lane pin, not a global invariant.
- `docs/architecture/overview.md`: staleness semantics and the config table updated.

Validation: `mcp-host/src/config.hccAuthority.test.ts` 8/8 passed; `tsc --noEmit`
clean. Re-run the local T2 minikube NP-08 revoke-on-staleness scenario (still pinned
to 60s) before deployment to confirm the security lane is unchanged.


alfredolopez80:
@claude Please perform a complete security and correctness review of the current PR #422 head (67aa5f44) against the NP-08 threat model.\n\nReview the actual diff and surrounding code, not only the PR description. Check:\n- JWT/caller -> live Host -> live Context -> Context-owned MCP server -> Secret read -> post-read fence;\n- v1 route retirement and gateway behavior;\n- mcp-host revocation and stale authority;\n- NetworkPolicy/RBAC boundaries;\n- mcp-proxy PR2 boundary and alternate attack paths;\n- Secret/token/log leakage, TOCTOU, identity recreation, cache and rollout risks;\n- unit, shell, PostgreSQL and local Minikube evidence claims;\n- the recent CI/CodeQL fixes.\n\nFor every finding, include severity, exact file/line, exploit or failure scenario, and a minimal remediation. Explicitly say when an item is not a blocker. Do not modify code or deploy anything; this is a review request only.

jozer-rami:
Heads-up: overlapping NP-08 bind in #421 (`fix/hcc-discovery-context-bind`, TokenReview of `host-<name>-sa`) is closed without merge so this PR stays the selected path.

Audit write-up (internal): Engineering → Evenfire Specs → Security-Assessments. N8/C3 stay High until this lands; we will not re-score against #421.

jozer-rami:
Closing without merge for now. Jose is making changes on the NP-08 path before we take another PR.

Branch `codex/np-08-cross-context-mcp-token-plan` is left in place. Reopen or open a follow-up when the updated approach is ready.

alfredolopez80:
## NP-08 PR1 follow-up — commits `09421d02` and `61cbd242`

These commits close the inexpensive, non-blocking review items without changing the selected PR1 security boundary.

### Resolved

- The NP-08 evidence scanner now uses portable `grep -E` and fails closed on scanner absence or execution errors instead of treating them as zero findings. Regression fixtures cover both failure modes and the test is wired into public CI.
- The HCC readiness E2E no longer calls the retired v1 Context-scoped route. It uses the metadata-only global inventory and binds the uniquely named fixture to the expected Context and Ready state.
- HCC-qualified access-token issuance is capped at the verifier ceiling of 600 seconds, while workflow-only token TTL behavior remains unchanged. The test verifies the signed `exp - iat` value.
- The mcp-host authentication boundary now requires a synchronous caller-authorization failure callback; optional callback invocation was removed from the authenticated path so revocation cannot silently be skipped.
- The local T2 runbook now documents the manual NP-08 Gate E/Gate F evidence boundaries and the required redacted status recording.

### Deferred follow-ups (not PR1 blockers)

- **Refresh/reissue lineage:** remove HCC audience/capability from the longer-lived refresh token while preserving the access-token contract. This needs a separate auth-contract and rollout change.
- **mcp-proxy global inventory:** the current route remains metadata-only for compatibility. PR2 must introduce an authenticated system-inventory principal/route, migrate the poller, bind grants to the live destination revision, and retire v1.
- **Anonymous route rate limiting:** add bounded, compatibility-safe limits to diagnostic and metadata routes in a separate gateway/HCC rollout so readiness, metrics, and the retained inventory consumer are not disrupted.

### Validation

- Control API focused suite: 47 passed; build passed.
- mcp-host focused suite: 30 passed; typecheck and build passed.
- NP-08 scanner, v1 reference inventory, E2E cleanup/provenance guards, and HCC watch-recovery contract: passed.
- ShellCheck, actionlint, Prettier, and `git diff --check`: passed.
- No production, GKE, Clerum-dev, or Clerum-prod deployment was performed; no Secret or token values were included.

The repository E2E static auditor still emits its known Playwright-only network-wait heuristic for this Bash/Node HTTP gate; the repository-specific shell contract validates the actual request/status/readiness assertions.

jozer-rami:
Not a blocker for this PR's scope, raising it because this PR is the place the behavior gets written down.

The README change documents `GET /api/v1/desktop/:hostRef` as "requires the `CONTEXT_MAPPER_DESKTOP_API_TOKEN` bearer token — the check is skipped when that variable is empty." That sentence is accurate, but it promotes a fail-open default to documented-intended, and I think the default is worth a second look before it hardens.

On `dev` today:

- `host-context-controller/src/server.ts:200-201` — `const token = config.desktopApiToken; if (!token) return true // skip auth in dev mode`
- `host-context-controller/src/config.ts:670` — `desktopApiToken: getEnv('CONTEXT_MAPPER_DESKTOP_API_TOKEN', '')!`, so the default is empty
- `deploy/base/control-plane/host-context-controller.yaml` does **not** set the variable
- the only place that sets it is `deploy/overlays/minikube/patches/dynamic-images.yaml:53`

Two things that make this worth a decision rather than a doc line:

1. **The comment says "dev mode" but nothing checks dev mode.** The condition is only "is the env var empty." There is no `CLERUM_DEV_MODE` guard on this path, so the skip is not scoped to development in any enforceable way.

2. **The polarity is inverted from what the comment implies.** The dev overlay is the one that sets a token. Base does not, and base is what non-minikube overlays build from. I can't see the prod overlay from this repo (it lives in evenfire-infra), so I'm not asserting prod is unauthenticated — but the default that ships from this repo is the fail-open one, and that's the wrong direction for a route that returns Host status.

Suggestion: fail closed when the token is unset (401 rather than allow), and let dev opt out through an explicit `CLERUM_DEV_MODE` check rather than through absence of configuration. Absence of config is indistinguishable from misconfiguration, which is exactly when you want the 401.

Happy to be told this is deliberate and adequately covered by the deploy layer — in that case the README line is the right place to say *why* it is safe, not just that it happens.

Context: surfaced while verifying the overclaim set from the 2026-08 whitebox audit. It was not scored there; NP-08 itself is out of scope for this comment.

jozer-rami (commented):
Deep review of the NP-08 change (reviewed at `61cbd242`, HCC suite run locally: 1,681/1,681 pass, `tsc --noEmit` clean).

The design holds up — I could not find a live bypass. `isLive()` requiring both uid and resourceVersion, `authorizationMetadata()` defaulting to `''` so a malformed object is dead rather than trusted, the post-read fence comparing `context.mcpServers`, and re-verifying the issuer's own response with the same strict verifier (`mcpHostRuntimeTokenIssuerClient.ts:132`, which closes the rolling-update downgrade) are all genuinely fail-closed. Signer/verifier wiring checks out end to end.

Two findings below that I think are worth addressing before merge / before rollout. Both were verified against the code, not inferred from the description.

host-context-controller/src/mcpAuthorization.test.ts
```diff
@@ -0,0 +1,536 @@
...
+  )
+  expect(error).toMatchObject({ code })
+  expect(error).not.toHaveProperty('token')
+  expect(JSON.stringify(error)).not.toContain(
+    Buffer.from(secret.data.token, 'base64').toString('utf8')
+  )
+}
+
+describe('McpAuthorizationService', () => {
+  it('keeps two Host/Context grant sets disjoint in both directions', async () => {
```
jozer-rami, line 145: **This test cannot fail on the vulnerability the PR closes.**

It builds two independent `FakeStore`s, each holding exactly one server. So `serviceA.getCredential(principal, 'server-b')` is denied because `storeA.readMcpServer()` returns `server-a` and `resolveServerGrant`'s `server.name !== serverName` trips — **not** because of the Context allowlist at `mcpAuthorization.ts:236`. The real attack is one cluster where `server-b` genuinely exists.

I mutation-tested it. Deleting the allowlist gate (`mcpAuthorization.ts:236-238` — the core of NP-08):

```
Tests  1 failed | 1680 passed (1681)
FAIL  rejects a credential when the Context allowlist changes during the post-read fence
AssertionError: expected { code: 'authorization_unavailable' } to match { code: 'not_found' }
```

One failure out of 1,681, and it is an **error-code** assertion on the post-read fence, not a disclosure.

I then wrote a single-store fake — `host-a → context-a → [server-a]`, with a live `server-b` owned by `context-b` — and ran it against that same mutated build:

```
RESULT: {"token":"TENANT-B-SECRET","credentialRevision":"MQDHwBaiHWO29D2..."}
```

Full cross-tenant credential disclosure, suite still green.

The neighbouring test at line 360 (`enumerates no foreign server…`) doesn't cover it either — with `mcpServers: []` the `listServers` loop is vacuous regardless of the gate.

Real coverage does exist, at `scripts/e2e/e2e-np08-hcc-authorization.sh:462`, but that's the branch-owned manual minikube T2 lane, not CI. So CI currently cannot catch a regression of the exact bug being fixed.

**Suggested fix:** one fake store holding both Contexts and both servers, asserting `getCredential(principalA, 'server-b')` → `not_found` **and** `readSecret` call count == 0. Happy to push that test to a branch if useful.

For contrast, the recreated-Host UID fence is properly covered — mutating `host.metadata.uid !== principal.hostUid` fails 2 tests. One adjacent gap while you're in here: replacing `capabilities[0] !== MCP_CREDENTIAL_READ_CAPABILITY` (`mcpApiAuthentication.ts:165`) with `false` leaves all 1,681 tests green. Removing the whole `mcpCapabilities` block *does* fail 2, so it's specifically the value equality that's untested.

mcp-host/src/config.ts
```diff
@@ -604,6 +608,14 @@ export const config: Config = {
   // Context Mapper poll interval (default 30 seconds)
   contextMapperPollInterval: parseInt(getEnv('CLERUM_CONTEXT_MAPPER_POLL_INTERVAL', '30000')!, 10),
 
+  // Bound transient HCC/Kubernetes authority outages. Identity failures revoke
+  // immediately; 5xx/transport failures may preserve the last good fleet only
+  // within this finite window.
+  hccAuthorityMaxStalenessMs: Math.min(
```
jozer-rami: **`Math.min` here is a hard ceiling, and I think it makes any HCC outage >60s a fleet-wide MCP outage.**

An operator can lower `HCC_AUTHORITY_MAX_STALENESS_MS` but never raise it. The deadline armed at `main.ts:815` fires independently of the poll loop and calls `revokeMcpAuthority`, which nulls `mcpManager` and closes it; `McpClient.retire()` aborts the retirement controller, so in-flight tool calls die with it.

**Trigger A (the one I'd worry about):** `kubectl rollout restart deploy/host-context-controller`. While HCC is unready, protected routes return 503 `authorization_unavailable`, which classifies as `unavailable` — not an identity failure — so the staleness deadline is what fires. If HCC needs more than 60s to become ready (image pull + watch warm-up, and `getReadinessState()` gates on `providerAuthoritativeFn()`), **every** mcp-host in the cluster tears down its entire MCP fleet and aborts in-flight tool calls. A routine HCC deploy becomes a user-visible MCP outage, and there's no operator escape hatch.

**Trigger B:** `CLERUM_CONTEXT_MAPPER_POLL_INTERVAL` is documented as operator-tunable (`docs/architecture/overview.md:820`) and unbounded (`config.ts:609`). Set it above 60000 and you get permanent flapping: connect → deadline revokes at 60s → reconnect at the next poll → revoke again, forever. I grepped — nothing validates the relationship between the two values.

**Suggested fix:** let the ceiling be raised (or at minimum scale it to the poll interval), and fail startup — or clamp with a warning — when `hccAuthorityMaxStalenessMs <= contextMapperPollInterval`. Note the render-contract test pins the minikube value to `"60000"` (`test-np08-render-contract.sh:34`), so that assertion would need to move with it.

alfredolopez80:
Follow-up for jozer-rami's inline review findings (commit 1f3559b6398c341cbccaae89a4f0fed4230e459b):

- Reworked the cross-Context authorization test to use one FakeStore containing both live Hosts, Contexts, McpServers, and Secrets. It now proves that Host A cannot request Host B's existing server, performs zero Secret metadata/value reads for that denial, and proves the same server succeeds for Host B.
- Added the missing negative JWT capability-value case (wrong capability is rejected).
- Added a production startup guard: CLERUM_CONTEXT_MAPPER_POLL_INTERVAL must be strictly lower than HCC_AUTHORITY_MAX_STALENESS_MS; invalid configurations fail closed instead of flapping. Documentation now records the invariant.
- evenfire-infra, CRDs, NetworkPolicies, and RBAC were not changed by this follow-up.

Validation:
- host-context-controller: 71 files / 1,682 tests passed; build passed.
- mcp-host: 324 files / 3,530 tests passed (1 todo); typecheck/build passed.
- Prettier and git diff --check passed.

The separate desktop-token fail-open observation remains outside NP-08 PR1 scope and is not included in this change.

host-context-controller/src/mcpAuthorization.test.ts
alfredolopez80: Resolved in 1f3559b6398c341cbccaae89a4f0fed4230e459b. The regression test now uses one map-backed FakeStore containing host-a/context-a/server-a and host-b/context-b/server-b. It denies both cross-Context credential requests, asserts no additional Secret metadata/value reads, and proves the same-Context positive credential path. The capability exact-value gap is covered by a malformed mcp:credential:write case. HCC focused/full evidence is green (71 files, 1,682 tests).

mcp-host/src/config.ts
alfredolopez80: Resolved in 1f3559b6398c341cbccaae89a4f0fed4230e459b. Production mcp-host now validates at startup that CLERUM_CONTEXT_MAPPER_POLL_INTERVAL is strictly below the effective HCC_AUTHORITY_MAX_STALENESS_MS ceiling; invalid/equal/greater combinations fail closed instead of flapping. The 60-second ceiling remains intentional fail-closed policy, with the default 30s/60s relation documented and tested. The public repository GitHub Actions owns the rollout/restart path; no evenfire-infra change is required for this PR.

alfredolopez80:
@claude conducts a thorough review of this PR, taking into account all previous reviews by jozer-rami, claude, and anyone else. And validates that everything is resolved or justified.

jozer-rami (approved):

zach88:
Took a deep pass through this one across all three services (HCC authz core, mcp-host consumer/lifecycle, control-api token issuance) plus the JWT and wire contracts end to end. Good news up front: the credential-disclosure fix is sound and fails closed everywhere I pushed on it. Everything left is test coverage or minor notes — nothing that reopens the NP-08 path, and nothing that blocks.

## Verdict

🟡 **Approve with nits**
- **Optional / non-blocking: 9** (2 Medium, 7 Low)

## Optional (non-blocking)

The two Mediums are both coverage gaps on the exact seams this PR is about — the fix itself is correct, but a future regression of it would ship green. Cheap to close, and worth it given what this PR protects.

1. **[R1-M1] Real store happy-path field mapping isn't bound to the producer** — `host-context-controller/src/k8sClient.test.ts:385-425`. Every service/route test drives a hand-written `FakeStore`. The only tests over the real `createMcpAuthorizationStore` assert 404→null normalization and the `readSecretMetadata` `.data` drop — nothing asserts that `readHost` maps `spec.contextRef→contextRef`, `readContext` maps `spec.mcpServers→mcpServers`, or the `readMcpServer` identity/auth/transport projection on a *successful* apiserver response. So a field-mapping regression in the real store (wrong `spec` path, dropped `mcpServers`) leaves the whole authz fence broken but the suite green, because the FakeStore fixtures encode the intended shape by hand. Suggest one happy-path test over `createMcpAuthorizationStore` with a mocked `getNamespacedCustomObject`/`readNamespacedSecret` asserting the full `AuthorityHost`/`AuthorityContext`/`AuthorityMcpServer` projection — the k8sClient mock harness is already in that file.

2. **[R1-M2] The main.ts wiring seam isn't asserted end to end** — `mcp-host/src/main.ts:930-973`. The pure functions `handleMcpAuthorityPollFailure` and `revokeMcpAuthorityState` are well covered, but the ~15-line options object that binds them in `pollContextMapper`/`revokeMcpAuthority` (`maxStalenessMs`, `revoke`, `hasPublishedManager`, the four callbacks) isn't touched by any test — confirmed by grep, they appear in no test file. A miswire there (`revoke` pointed at a no-op, wrong staleness source, `hasPublishedManager` inverted) would let an HCC 5xx tear down a legit fleet, or let a revoked authority survive, with nothing turning red. Worth one test that drives `pollContextMapper` end to end and asserts the observable outcome (fleet withdrawn / bearer gone).

3. **[R1-L2] 600s ceiling is a hardcoded const in control-api but configurable in HCC** — `control-api/src/utils/auth/mcpHostJwtToken.ts` (`MCP_HOST_HCC_ACCESS_MAX_TTL_SECONDS=600`) vs HCC `HCC_MCP_HOST_JWT_MAX_TTL_SECONDS`. If an operator lowers HCC's max below 600 while control-api's access TTL stays ≥600, HCC rejects every HCC token. Fail-closed, not a disclosure — but it's a two-service coupling with no shared source.

4. **[R1-L3] The clamp test doesn't lock in the refresh-uncapped / boundary behavior** — `control-api/test/mcpHostJwtToken.hostRefs.test.ts`. It asserts the access clamp and WRC-unchanged, but nothing asserts the HCC refresh token stays uncapped (by design) nor that `exp-iat===600` is accepted. Add positive assertions so a future change to the clamp condition can't silently start capping refresh or stop capping access.

5. **[R1-L5] Host-namespace equality rides on two independent literals, not a shared key** — HCC `CONTEXT_MAPPER_HOST_NAMESPACE` (`deploy/base/control-plane/host-context-controller.yaml:75-76`) and control-api `CONTROL_API_HOSTS_NAMESPACE` (`deploy/base/control-plane/configmaps.yaml:13`). Unlike issuer/max-TTL/signing key (which HCC reads from the same control-api ConfigMap keys, so they can't drift), these are two separately-authored literals — both `mcp-host` today, but changing one overlay without the other silently 401s every HCC MCP call. Consider sourcing both from one key like the others.

6. **[R1-L7] The new `hccTargetNamespace !== hostNamespace` FATAL guard is a startup-behavior change** — `host-context-controller/src/main.ts:97-101`. HCC previously only warned when `HCC_TARGET_NAMESPACE` was outside `{mcp-host, sandbox-recipes}`; it now crash-loops unless it equals `CONTEXT_MAPPER_HOST_NAMESPACE`. Base overlay is fine (both `mcp-host`) — just worth checking any non-base overlay that set these differently before rollout, or the pod crash-loops.

7. **[R1-L4] `McpServerInfo` type is looser than the wire the decoder accepts** — `mcp-host/src/types.ts:176-197` keeps `auth?`/`contextRef?` optional, but `contextMapperClient.ts:84` rejects any v2 response carrying `contextRef|auth|secretRef|secretKey`. Not a runtime break (the decoder is the enforced gate), just a type that permits shapes the decoder forbids. Cosmetic.

8. **[R1-L6] Rate limiter prune evicts still-active windows under extreme cardinality** — `host-context-controller/src/server.ts:66-76`. With >10,000 distinct `hostUidaction` keys active within one minute, the `while (size > 10_000)` loop evicts the oldest still-active window, resetting that host's counter early. Bounded-memory tradeoff on a DoS mitigation, not an auth boundary — noting for the record.

9. **[R1-L8] HTTP-boundary fixtures are hand-written** — `mcp-host/src/contextMapperClient.test.ts:27-45`. They stand in for HCC's output but are typed by hand rather than derived from the producer. I verified they match `AuthorizedMcpServerInfo`/`AuthorizedCredential` exactly today, so the risk is low — but the two shapes live in separate modules and can drift silently. A shared contract fixture would close it.

## What checked out clean

Spent most of the time trying to break the fence, and it held:
- **Cross-context is closed by the signed `host_uid` binding.** `resolveHostContext` requires `host.metadata.uid === principal.hostUid`; control-api signs `host_uid` from the live Host CRD UID and HCC re-verifies it. A compromised mcp-host in Context A can't resolve Context B without a token control-api only issues for the real Host B.
- **TOCTOU is fenced.** `getCredential` re-resolves the whole grant after reading the Secret and compares `sameGrant` (uids + resourceVersions + `credentialRevision` + principal expiry) before emitting the token.
- **v1 caller-selected routes are retired** at both the app (410 tombstone in `server.ts:184-187`, all methods incl. preflight, before the readiness gate and independent of the gateway) and the nginx gateway. No v1 handler serves credentials anymore. The residual `/api/v1/mcpservers` is metadata-only — `name/contextRef/transport/enabled/status`, no `auth`/`secretRef`/`secretKey`/token — and mcp-proxy (its only live consumer) reads only fields that survive the projection, so it isn't broken.
- **No cross-namespace Secret reads**; RBAC for Secrets is unchanged (no widening). No secret/token logging in the changed paths; credential responses carry `Cache-Control: no-store` and no wildcard CORS.
- **JWT hardening** is thorough: RS256-only (rejects none/HS256/private-key, requires RSA≥2048), exact issuer/audience, exact scope + capability, lifetime ≤600, bearer parsing that rejects duplicate/array `Authorization` headers. No escalation via refresh/reissue — `hccCredential` derives from verified claims, never the body.
- **Both contracts trace field-for-field** (control-api↔HCC JWT and HCC↔mcp-host v2 wire); the params that could drift go through shared ConfigMap keys.
- Suites green where I ran them: HCC full `1682 passed / 71 files` + `tsc` clean; mcp-host touched suites `38 + 59 passed` + `tsc` clean; control-api focused `47 + 28 passed`. I did not run the T1(pg)/T2(minikube) lanes — lean on branch CI + the manual T2 lane for those.

alfredolopez80:
## NP-08 PR1 closure and clerum-dev deployment blast-radius record

This comment records the final PR #422 evidence. PR #421 is closed and is not an active implementation path.

### Closed in PR1

- The credential path is bound as: valid Control API JWT -> live Host and UID -> live Context -> Context-owned McpServer allowlist -> Secret read -> post-read grant/principal fence.
- The caller cannot select an arbitrary Context or server. Unauthorized and cross-Context requests fail before any Secret metadata or value read.
- Host recreation, Context/McpServer/Secret deletion or rotation, disabled grants, credential revision changes, and stale authority revoke fail closed.
- Caller-selected v1 Context and per-server auth routes are tombstoned at the HCC application boundary and in the intended gateway configuration.
- mcp-host uses the Host-bound v2 contract and revokes its published fleet on caller rejection, inventory loss, or bounded authority staleness.
- mcp-proxy is unchanged and remains a documented PR2 compatibility residual; no new mcp-host -> mcp-proxy data-plane access was added.

### Validation at head ab87e7bb

- HCC full suite: 71 files, 1,683 tests passed.
- mcp-host full suite: 324 files, 3,530 passed, 1 todo.
- Control API full suite: 360 files, 4,463 passed, 103 skipped.
- HCC, mcp-host, and Control API builds passed; Prettier and git diff check passed.
- All GitHub check runs reported completed/success for ab87e7bb.
- No Secret values, tokens, GKE mutations, or production operations were performed.

Merge readiness is therefore: application security boundary PASS; clerum-dev functional rollout requires the gateway restart prerequisite above.

PR #423, opened 2026-08-20
fix(control-ui): bound concurrent pre-auth body reads in the control-api proxy

jozer-rami:
## Summary

The `/control-api/*` proxy route buffers a request body **before** control-api authenticates it. `CONTROL_UI_PROXY_MAX_BODY_BYTES` (24 MiB default) already caps a single request, but nothing capped the total buffered at once, so the real ceiling was `cap x concurrency`, not `cap`. control-ui runs `replicas: 1` under a 512Mi limit.

This bounds the **bytes** buffered pod-wide, charged from admission through the upstream response, via `CONTROL_UI_PROXY_MAX_INFLIGHT_BODY_BYTES` (192 MiB default).

Bytes rather than a reader count, for two reasons:

- **A reader count measures the wrong thing.** `body` stays referenced by `fetchInit` for the whole upstream round trip — the slow phase, at `CONTROL_API_PROXY_TIMEOUT_MS: "300000"` in `deploy/base/control-plane/control-ui.yaml`. Releasing when the *read* finishes bounds the rate buffers are created at, not how many are alive, leaving retained memory at `limit x bandwidth x latency`.
- **A slot prices a 4 KiB mutation like a 16 MiB upload part.** The GFS client runs `GFS_FILE_UPLOAD_DEFAULT_CONCURRENCY = 4` workers that hold their slot for the entire ingest of a part. Nine bulk actions fan out over `Promise.all` with no retry in `apiSend`, so a concurrent upload would have turned any of them into a partial write.

Charging bytes fixes both: memory is bounded directly, and a small mutation costs a few KiB so it interleaves freely with an upload.

Design notes:

- **Charged at 2x during the read** — `readBodyCapped` holds the chunk list and the contiguous copy it is flattened into at the same time — then settled to the resident copy for the fetch.
- **Charged as bytes arrive, not reserved from `content-length`.** A client that declares 24 MiB and then dribbles must not be able to hold budget it never uses. The declared length is used only as a pre-read forecast, to reject a request already known not to fit before buffering anything.
- **The read phase has an idle deadline** (`CONTROL_UI_PROXY_BODY_READ_IDLE_TIMEOUT_MS`, 30s). It previously had none: `AbortSignal.timeout` is constructed after the read and only ever covered the fetch. Idle rather than total, because a total deadline generous enough for a 16 MiB part on a domestic uplink is also long enough to be worth attacking.
- **Sheds rather than queues** — a queued caller still holds a connection and would buffer the moment budget frees.
- **Sheds with `429`, not `503`.** `503` is in `GFS_UPLOAD_AMBIGUOUS_STATUS` (`lib/gfsFileUpload.ts:16`), so a shed part triggers `reconcileAmbiguousPart` — up to 3 x 60s — to establish the fate of a request that provably never left the process. `429` is retryable but unambiguous, and matches the repo's prior art at `control-api/src/middleware/gfsUploadAdmission.ts`.
- **GET/HEAD are never gated**, so long-lived SSE notification streams cannot be shed. Bodyless POST/DELETE and the `OPTIONS` preflight charge 0 bytes.

Budget sizing, written down in the route so it can be re-derived: 512Mi limit, minus a ~150 MiB Next.js baseline, leaves ~360 MiB; 192 MiB charged keeps ~170 MiB headroom. It also clears a full upload wave (4 x 16 MiB x 2 = 128 MiB) with 64 MiB left for the mutations that must keep working while someone uploads.

Not an authorization issue: the operator ACL was never bypassed and control-api still returns 401. The gap was availability.

Found during the 2026-08 whitebox audit (finding C1).

## Also in this PR

From review feedback on the first revision:

- Log the shed. It was silent, so an operator could not distinguish a shedding proxy from a broken client.
- Render `proxy_busy` as a sentence rather than a raw code (`lib/api.ts`). `apiSend` never reads response headers, so the accompanying `Retry-After` is not reachable from the general client path and the wait is stated in words.
- Fold six copies of the env resolver into one.
- Correct two manifests that described a memory model this route does not have: v2 upload parts are buffered and flattened, not streamed through, and the read peak is 2x, not one copy.
- Document all four proxy knobs in `control-ui/README.md`. None were listed, and the "30 s upstream timeout" line contradicted the deployed 300s.

## Testing

Node 24 (`.nvmrc`), exit codes captured without a pipe.

- `control-ui` full suite: **1906 passed**, 172 files, exit 0
- proxy route file: **26 passed** (6 new)
- `next build` (typecheck + lint + compile): exit 0
- Merged `origin/dev` (189 commits) and re-ran all of the above.

Mutation-tested rather than trusting green — 9 mutants, all killed:

| mutation | killed by |
|---|---|
| release the charge when the read finishes (the bug this fixes) | `holds the charge across the upstream fetch` |
| no idle deadline on the read phase | `aborts a body that stalls mid-read` |
| price every request as one flat unit instead of by bytes | `lets a small mutation through while a large upload is buffering` |
| reserve the declared length instead of actual bytes | `does not let a dribbling client reserve the length it declared` |
| default budget widened (the `4 -> 1024` analogue) | `defaults the budget to 192MiB exactly` |
| env parse `Number.isInteger` -> `!Number.isNaN` | `falls back to the default when the env value is...` |
| shed with `503` again | `sheds with 429 once the budget is spent` |
| drop the outer `finally` release | 3 tests |
| upload branch forecasts from `content-length` | `sheds a part whose declared length cannot fit` |

## Known gaps, deliberately not in this diff

- `app/admin-invitations/[token]/complete/route.ts` and `app/admin-password-resets/[token]/complete/route.ts` both `await req.formData()` pre-auth with no byte cap, no counter and no timeout, in the same process under the same 512Mi. Pre-existing and its own change; the counter comment no longer claims to cover them.
- The budget is global, with no per-session or per-IP quota, so an unauthenticated caller can still consume budget an authenticated one wanted. Charging real bytes makes that expensive rather than free, but it is not partitioned.
- `Retry-After: 1` carries no jitter.

## Checklist

- [x] This is not a new third-party MCP server / WorkflowRecipe (those go to the registry)


alfredolopez80:
@jozer-rami — full multi-agent review of `4527537f` (six parallel reviewers: correctness/concurrency, security, tests, client-contract, config/deploy/docs, architecture), each verifying findings against the real code in a worktree at the PR HEAD. Independent mutation testing was run rather than trusting the green suite.

First, credit where it's due: the diagnosis is right, the layer is right, and the write-up is unusually honest. There is no `control-ui/middleware.ts`, the nginx limits in the repo are size limits not concurrency limits (`deploy/base/profiles/configmaps.yaml:36`), and control-api sits *downstream* of the buffering — so the route handler is genuinely the only place this control can live. The counter is also correctly per-pod: `next start -p 3000`, one Node process, `replicas: 1`, `runtime = 'nodejs'`. Both mutations you claim to have tested do die, and the `finally`-decrement one is **understated** — it fails 10 tests, not 1.

That said, I don't think this should merge as-is. Four issues, two of which are new regressions rather than incomplete mitigation.

---

## Blockers

### B1 — The slot is released before `fetch`, so buffered memory is still a multiple of load

`route.ts:266-268` vs `:283-291`

The `finally` releases the slot when the *read* finishes, but `body` stays referenced by `fetchInit` for the whole `await fetch()`. The counter bounds the **rate at which buffers are created**, not the **number of buffers alive**. Since the read is the fast phase and the upstream round trip is the slow one, buffers stack up.

Roughly: `retained ≈ limit × ingest_bandwidth × upstream_latency` — independent of the per-request cap.

The deployed config makes this 10× worse than the default would suggest: `deploy/base/control-plane/control-ui.yaml:39-40` sets `CONTROL_API_PROXY_TIMEOUT_MS: "300000"` (300s), not 30s. Reproducing the exact attack from your description (25 × 20 MiB) needs only ~1.7 MB/s sustained.

| | |
|---|---|
| Budget claimed at `route.ts:57-63` | 4 × 24 MiB = 96 MiB |
| Live bodies measured, 25 connections | 500 MiB |
| Pod limit (`control-ui.yaml:72`) | 512 MiB |
| Next.js baseline | ~100-150 MiB |

So the OOMKill this PR sets out to prevent is still reachable. The PR does roughly halve the peak (it serializes the phase where `chunks` and the contiguous copy coexist), which is real — but it doesn't get under the pod limit, so the outcome is unchanged.

The two comment blocks also contradict each other: `:62-63` promises a *fixed budget*, and `:65-66` explains exactly why there isn't one. Whatever you decide on the mechanism, `:62-63` can't keep asserting a bound the code doesn't provide.

Fix options: decrement after the `fetch` (throttles upstream — which is what the canonical middleware does via `res.once('finish'|'close')`), or track *live bytes* separately from *readers*.

### B2 — Slow-loris: 4 dribbling sockets shed every mutation, permanently

`route.ts:224-230` and `:273-274`

`readBodyCapped(req.body, cap)` takes no `AbortSignal`. `timeoutSignal` is constructed at line 273 — **after** the read block — so it only ever covers the upstream `fetch`. The phase that now holds a slot has no application deadline at all.

Attack: 4 sockets, `POST /control-api/...` with `Content-Length: 1000`, send one byte, go quiet. Each holds a slot until Node's `server.requestTimeout` (300s default, no override in the repo). That is **~0.013 connections/second** — 4 sockets refreshed every five minutes — to make every mutation on the pod return 503. Memory cost to the attacker: ~zero.

Before this PR that was a harmless trickle. After it, 4 sockets block every write in the product. The PR trades an expensive memory DoS for a nearly free availability DoS.

Your own test is the proof of concept: `parkedBodyRequest()` opens a stream that is never closed and the next POST gets 503. That test *is* the exploit, at budget 1.

### B3 — The default (4) is exactly the GFS upload client's concurrency (4)

`route.ts:69` + `control-ui/app/constants/gfsFileUpload.ts:6`

`GFS_FILE_UPLOAD_DEFAULT_CONCURRENCY = 4`. The four `runParts` workers (`lib/gfsFileUpload.ts:900`) `PUT` parts straight through this proxy, and the slot is held for the **entire ingest of that part from the browser** — not "microseconds", as `route.ts:65-67` says. On a domestic uplink that's ~13s per part, continuously, for the whole file. `replicas: 1`.

No attacker needed: user A uploads 1 GB → all four slots pinned for minutes → user B clicks anything that mutates → `503 proxy_busy`. And `control-ui/lib/api.ts:289-294` has no retry whatsoever (`grep -cE 'for \(let attempt|while \(attempt|retryCount'` → 0), so it surfaces as a hard error on the action.

The calibration is the worst available value, specifically. The client already self-corrects: after 3 consecutive retryable failures it halves concurrency from 4 to 2 (`lib/gfsFileUpload.ts:842-850`). But with the budget set at exactly 4, the four workers **fill the pool without ever shedding each other** — the gate only fires on a fifth. So the downgrade never triggers, the upload keeps all four slots for the whole file, and the 503 is paid by everyone else. Either value on the other side of 4 behaves better.

Worth knowing: that client-side 4 is a compile-time constant. `Math.min(GFS_FILE_UPLOAD_DEFAULT_CONCURRENCY, advertised)` at `lib/gfsFileUpload.ts:1380-1384` can only lower it, and no env var touches it. The server default is the only operational lever you have.

### B4 — Parallel mutation fan-out + no retry = partial writes

`control-ui/lib/api.ts:268-297` and 9 UI call sites

9 of the 10 bulk actions use `Promise.all` / `Promise.allSettled` over mutations, with N = whatever the user selected:

| file:line | action |
|---|---|
| `app/contexts/[name]/page.tsx:667` | link N hosts to a context (5-20) |
| `components/LlmDiscoveryPanel/index.tsx:313` | bulk-enable N models (dozens) |
| `components/HostAccessTab/index.tsx:136`, `:203` | grant access to N users / N teams |
| `app/mcp-servers/page.tsx:238-250` | add server to N contexts |
| `app/profile-admin/**` (4 sites) | bulk membership changes |

(Only `components/WorkflowAccessPanel/index.tsx:334-351` is sequential, deliberately.) Six of the nine do GET+PUT per item, which staggers the PUTs and *widens* the collision window rather than narrowing it.

`Promise.all` rejects on the first 503 while the other mutations have already been applied. The user doesn't get a clean failure — they get a **partial write** with a generic error and no way to tell what landed.

**B3 + B4 together are what makes this blocking.** For small JSON the collision is scheduling-dependent and may not happen. But while a GFS upload is running, all four slots are held for minutes, and then every bulk action fails deterministically, leaving state half-applied.

---

## High

**H1 — `503` triggers the upload client's reconciliation path; `429` wouldn't.** `lib/gfsFileUpload.ts:16` puts 503 in `GFS_UPLOAD_AMBIGUOUS_STATUS`, so each shed fires `reconcileAmbiguousPart` (`:796`) — up to 60s × 3 attempts (`:29-30`) — for a part that provably never reached upstream (your own test asserts `expect(fetchMock).not.toHaveBeenCalled()`). `429` is in `GFS_UPLOAD_RETRYABLE_STATUS` and *not* in the ambiguous set, so it would produce a cheap retry instead. It's also what the repo's canonical prior art already emits: `control-api/src/middleware/gfsUploadAdmission.ts:112-122` (`429` + `Retry-After` + `X-RateLimit-*`). That middleware is worth a look generally — it solves this same admission problem with per-user/per-IP buckets, Prometheus counters, and an idempotent release.

**H2 — The fix is partial: two sibling pre-auth handlers buffer with no cap at all.** `app/admin-invitations/[token]/complete/route.ts:46` and `app/admin-password-resets/[token]/complete/route.ts:18` both `await req.formData()` on public pre-auth endpoints — no byte cap, no counter, no timeout — in the same Node process under the same 512Mi. Pre-existing and out of this diff's scope, but it means `route.ts:79-80` overstates its reach: the counter is this *route's* in-flight count, not the pod's.

**H3 — The production configuration of the gate is untested.** `CONTROL_UI_PROXY_MAX_CONCURRENT_BODY_READS` appears in exactly two files in the repo — `route.ts` and the test — so the default of 4 *is* the production config, and nothing pins it. Surviving mutants:

| mutation | result |
|---|---|
| `DEFAULT_MAX_CONCURRENT_BODY_READS` 4 → 1024 | **survives** |
| ceiling 1024 → `MAX_SAFE_INTEGER` | **survives** |
| `Number.isInteger` → `!Number.isNaN` | **survives** |
| env absent/NaN → limit disappears entirely | **survives** |

**H4 — `Retry-After` is unreadable to the general client and `proxy_busy` renders raw.** `apiSend` never reads response headers (`lib/api.ts:93-141`), so the `Retry-After: 1` is inaccessible to the ~80 call sites that go through it. And `proxy_busy` isn't in any error catalogue — in `sfsUpload` (`:1372-1377`) the raw JSON reaches the user's screen.

---

## Medium

- **The shed emits no signal at all** (`route.ts:224-229`): no log, no metric, no counter. An operator can't distinguish "the proxy is shedding" from "the client is broken". There's precedent for `console.error` in this very function (`:306`), and the canonical middleware increments a Prometheus counter (`gfsUploadAdmission.ts:240`). Best cost/benefit item on this list.
- **Bodyless requests consume the budget** (`route.ts:219`): `hasBody` only inspects the method, never `req.body`. 56 of ~80 mutations (bodyless POST/DELETE, plus the `OPTIONS` preflight exported at `:327`) take a slot from an anti-buffering budget while buffering nothing.
- **The budget is global with no per-session or per-IP quota** (`route.ts:224-229`), so an unauthenticated caller denies service to authenticated ones.
- **`falls back to the default…` is vacuous in 4 of its 6 cases** (`test.ts:402-414`). With the counter at 0 and a single sequential request, `0 >= NaN`, `0 >= 1.5` and `0 >= 99999` are all `false` — those admit regardless of whether the fallback applied. Only `'0'` and `'-1'` prove anything, and only that the resolved value is `> 0`, not that it's 4. Suggested fix: park 4 reads and assert 503 on the fifth.
- **Module state has no reset between tests** (`test.ts:300-415`). A failure before `parked.release()` leaks the slot and poisons its neighbours in the same file — confirmed by the fact that the dropped-decrement mutant fails 10 tests.
- **The manifests contradict this code path**: `control-ui.yaml:36-38` and `:69-71` say v2 part requests "bypass that path and stream", but `route.ts:244` buffers them whole. This PR touches exactly that memory model and leaves the contradiction standing.
- **`resource-limits.yaml:44-45` models one copy** ("~22.4MB for a 16MB upload") while this PR correctly models 2×. The manifest now understates the footprint the PR itself describes.

## Low

`rawUpload || hasBody` ≡ `hasBody`, since `isGfsUploadRoute` requires `PUT` (confirmed equivalent mutant) · ceiling 1024 is unjustified and non-protective (1024 × 2 × 24 MiB ≈ 48 GiB, 96× the pod limit) · the arithmetic behind default 4 (4 × 2 × 24 MiB = 192 MiB vs 512Mi) isn't written down anywhere · fourth verbatim copy of the same 7-line env resolver — a shared helper is overdue · the 34-line `try` with 4 early returns adds two indentation levels, and extracting `readRequestBody()` would make the `finally` much harder to break later · `Retry-After: 1` with no jitter makes the four upload workers retry in lockstep · `control-ui/README.md:52-58` omits the new knob · no runbook documents `proxy_busy` anywhere · no test covers the shed on the `rawUpload` branch, which is precisely the one that holds slots for minutes · the PR body's numbers predate the two `dev` merges — actual on `4527537f` is **1918 passed / 1 skipped across 173 files**, not 1605/154 (green either way, just stale).

---

## Verified clean

Everything below was executed or read, not assumed.

| command | result | exit |
|---|---|---|
| `npx vitest run lib/__tests__/controlApiProxyRoute.test.ts` | **20 executed**, 20 passed | 0 |
| `npx vitest run` | **1919 executed**, 1918 passed / 1 skipped, 173 files | 0 |
| `npx tsc --noEmit` | no output | 0 |

Node v24.18.0, exit codes captured without a pipe. No run executed zero tests. CI: 37 checks green.

No counter leak — the `finally` covers all five early returns and throws. No TOCTOU — there is no `await` between the check at `:224` and the increment at `:230`. `+= 1` outside the `try` is correct and canonical, not fragile. `rawUpload` is never true for GET/HEAD. The `if/else` → `if (rawUpload || hasBody)` refactor preserves the inner semantics exactly. The env parser is fail-safe across `''`, `'  '`, `'0'`, `'-1'`, `'1.5'`, `'1025'`, `'abc'`, `'Infinity'`. Changing the env at runtime can't desync the counter. GET/HEAD are never gated, so SSE streams can't be shed — and it's tested. `retry-after` casing is correct and it *is* consumed at `gfsFileUpload.ts:343,411`. `desktop-app` and `profile-ui` don't consume this route at all. The shed logs nothing — no bodies, no tokens. `stripCrlf` covers url, message and cause. No new `console.*`, so the repo logging standard holds. Config parity with `CONTROL_UI_PROXY_MAX_BODY_BYTES` is strict — zero asymmetric omissions, since the sibling isn't in any manifest either; that's the repo's convention, not something this PR skipped. The manifest comment you cite exists and is accurate. And there's no overlap with the per-user rate limiting merged in #446 — different planes (pre-auth/memory vs post-auth/abuse).

---

## Suggested path

B2 and B3 are the actionable ones and need no redesign: put a deadline on the read phase, and separate the upload budget from the mutation budget (or simply raise the default above `GFS_FILE_UPLOAD_DEFAULT_CONCURRENCY`).

B1 is a genuine design call — hold the slot across the `fetch` and accept upstream throttling, or count live bytes instead of readers. Either way the claim at `:62-63` needs to match what the code delivers.

Two calls that are yours, not mine: whether to accept B1 as a partial improvement with an honest bound documented instead of redesigning, and whether `429` is worth switching to now that `503` is written.

---

https://claude.ai/code/session_017Zr7UQ3cRtBwNxWctkW2LX

alfredolopez80:
Review of **427cc8cb** against **e5afb0de**, including the earlier review of **4527537f**.

**Changes requested: two High findings remain.** The current suite and builds are green, but the new resource bound is not yet established.

1. **High — idle cleanup still depends on client progress.** At [route.ts:185–187](https://github.com/evenfire-ai/evenfire/blob/427cc8cb0a915d1573b83951c6bfe06c6ecd556d/control-ui/app/control-api/%5B...path%5D/route.ts#L185-L187), the timeout waits for `reader.cancel()` before releasing the charge. In the built Next server, with idle=100 ms and budget=128 bytes, a request declaring 64 bytes but sending 63 still held 126 bytes after 401 ms. A separate two-byte POST returned **200 on BASE and 429 on HEAD**. Sending the missing byte finally produced 408 and released capacity. The read's bounded termination must not require the client to finish it.

2. **High — the charge omits transport copies and ends before their lifetime does.** [Settlement](https://github.com/evenfire-ai/evenfire/blob/427cc8cb0a915d1573b83951c6bfe06c6ecd556d/control-ui/app/control-api/%5B...path%5D/route.ts#L413-L417) reduces the charge to one copy; [finally](https://github.com/evenfire-ai/evenfire/blob/427cc8cb0a915d1573b83951c6bfe06c6ecd556d/control-ui/app/control-api/%5B...path%5D/route.ts#L470-L471) releases it at response headers. Re-executed native Fetch/HTTP evidence: four fully uploaded 8 MiB bodies awaiting headers retained **96 MiB of ArrayBuffers while HEAD charged 32 MiB**, after GC. Independent active-response probes also retained body-derived memory after the counter reached zero. **Those copies already exist in BASE**; the defect is the new accounting guarantee, not a new increase in allocations. No pod OOM was demonstrated. Ownership must cover retained copies through actual transport disposal, including response EOF/error/cancellation.

3. **Medium — the aggregate-admission tests accept a per-request-only implementation.** At [test.ts:358–372](https://github.com/evenfire-ai/evenfire/blob/427cc8cb0a915d1573b83951c6bfe06c6ecd556d/control-ui/lib/__tests__/controlApiProxyRoute.test.ts#L358-L372), the contender costs 22 bytes against a 16-byte budget and fails even alone. Replacing the two aggregate checks with per-request checks leaves **26/26 tests passing**. Add an individually admissible control: budget 24, incumbent charge 8, contender cost 22; reject only the overlap, then admit after release. The current implementation passes that contrasting check; the missing regression protection is the finding.

Three **Low** follow-ups:

- [Fallback test, line 559](https://github.com/evenfire-ai/evenfire/blob/427cc8cb0a915d1573b83951c6bfe06c6ecd556d/control-ui/lib/__tests__/controlApiProxyRoute.test.ts#L559): the fractional-budget mutant survives this test alone. The full suite kills it through the pre-existing timeout test. Check both sides of the fallback boundary for each invalid value; the current parser itself is correct.
- [Fixture cleanup, line 324](https://github.com/evenfire-ai/evenfire/blob/427cc8cb0a915d1573b83951c6bfe06c6ecd556d/control-ui/lib/__tests__/controlApiProxyRoute.test.ts#L324): resetting before a parked request finishes allows `16 → 0 → -16` when its finally runs. Await fixture disposal before resetting. This is a failed-test isolation issue, not a production negative-counter claim.
- [README, lines 55–56](https://github.com/evenfire-ai/evenfire/blob/427cc8cb0a915d1573b83951c6bfe06c6ecd556d/control-ui/README.md#L55-L56): document raw GFS's fixed 16 MiB cap and separate `CONTROL_UI_GFS_UPLOAD_PROXY_TIMEOUT_MS` default of 600000 ms. Generic settings do not control those existing exceptions.

The previous four-reader collision, 503 retry classification, empty-body charging, duplicated resolver, missing friendly message, and missing raw-part shedding test are addressed. The earlier B1/B2 concerns change into the two lifecycle findings above. The two public form handlers remain a separate pre-existing risk. The CodeQL `fetchMock unused` comment does not apply to this head: the variable is used at lines 457 and 461.

**Validation:** 37 CI checks successful; proxy BASE 16/16 and HEAD 26/26; complete HEAD suite **1906/1906 across 172 files**, no skips; both builds exit 0 on Node 24.18.0 / Next 16.2.11. Targeted mutation and native HTTP results are separate from those green checks.

**Recommendation:** resolve the two High findings with explicit bounded resource ownership, preserve validation before forwarding and current 429/408/GFS compatibility, and add the aggregate counterexample before re-review. The Low items can be follow-ups.

**Coverage limits:** no browser E2E or full-deployment memory/OOM certification. The dedicated security reviewer did not complete after a platform safety interruption, so this review does not claim exhaustive independent security coverage.

jozer-rami:
Addressed in `7c83b44f4`. Point by point.

**H1 — idle cleanup depended on client progress.** Confirmed and fixed. The idle path did `await reader.cancel()`, and in the built server the request-body stream's `cancel()` does not settle until the peer sends more, so the charge stayed pinned exactly as long as the stalled client chose. Abort now fires `cancel()` and returns immediately; the caller's `finally` releases the charge at the deadline. The `parkedBodyRequest` fixture cancels instantly under jsdom, which is precisely why the original test passed over this — the new regression models a `cancel()` that never resolves and asserts the charge is reclaimed and a later request admitted anyway. Mutation-checked: restoring `await reader.cancel()` now fails that test.

**H2 — charge omitted transport copies and settled to 1×.** Fixed the magnitude: the 2× charge is held for the whole lifetime (admission → upstream response) instead of settled to one resident copy after the read, so the awaiting-headers window is charged ~2×, matching `body` + undici's send copy. The comment no longer over-claims — it states 2× is a conservative steady figure that transport can transiently exceed, which is why the budget sits well under the pod limit.

On the two calls you flagged as mine:
- **Release timing** stays at `fetch()` resolution rather than response EOF. By the time `await fetch` resolves the request body has been fully sent and its transport copies freed, so holding to EOF would not reduce request-body memory; and wrapping `upstream.body` to release on EOF would have to re-implement backpressure or it buffers the whole GFS *download* in memory — a worse regression than the one it closes. I took the honest-bound path you offered rather than the response-stream wrapper. Happy to revisit if you have a backpressure-safe hook in mind.
- On your 3× measurement vs my 2× charge: 2× covers `body` + one undici copy; the residual above that is short-lived transport internals I don't model exactly, which is what the ~170MiB of headroom under the 512Mi limit is for. Documented as such rather than pretending the charge is an exact bound.

**Medium — aggregate tests accepted a per-request implementation.** Added exactly your counterexample: budget 24, incumbent charged 8, contender 22 — individually admissible (22 ≤ 24), shed only because of the overlap (30 > 24), then admitted after release. Verified it kills the drop-the-incumbent-charge mutant, which the previous tests let survive.

**Lows:**
- Fallback test now asserts both sides of the 192MiB boundary for every invalid value, so the fractional-budget mutant no longer survives this test in isolation.
- Fixture disposal is awaited before the counter reset, the reset runs in a `finally`, and the raw-GFS block resets defensively — a failed test can no longer underflow the counter into a neighbouring block. (Found and fixed a real instance of this while adding it.)
- README documents raw GFS's fixed 16MiB part cap and the separate `CONTROL_UI_GFS_UPLOAD_PROXY_TIMEOUT_MS` (600000) default.

**Still not in scope, by design:** the two public `req.formData()` form handlers (pre-existing, separate change — happy to file an issue), and holding the charge past response EOF (backpressure risk above).

Validation on `7c83b44f4`, Node 24 / exit codes captured without a pipe: proxy file **28/28** (idle-never-cancels, aggregate counterexample, both-sides fallback new), full suite **1908/1908**, `tsc --noEmit` and `next build` both exit 0. Four mutants — restored `await` cancel, settle-to-1×, per-request admission, and a 2.1 multiplier — all die, each on the test written for it.

PR #428, opened 2026-08-21
fix(control-api): enforce live admin revocation on workflow routes (F4)

alfredolopez80:
## Summary

- Extract `authenticateAdminSession` as the canonical live administrator decision (JWT + JTI revocation + active account + matching `sessionVersion`).
- Delegate both `requireAuthForControlUI` (cookie-only) and `requireAdminWorkflowCaller` (bearer or cookie) to the shared service, closing the F4 bypass on `/admin/workflows/*` and defense-in-depth on `/admin/outputs`.
- Add PG + `express-rate-limit` IP edge backstops on admin workflow grants, read routes, and `/admin/outputs`; include `control_ui_admin_session` in `workflowTriggerRateLimit` buckets (cookie triggers were previously fail-open).
- IP-only edge backstop on grant/read/outputs routes so rotating unverified bearer tokens cannot evade the cap; PG limiter remains cross-replica source of truth per credential hash.
- Add focused regression tests across all 20 protected admin workflow handler paths plus unit/middleware coverage for revoked, stale, disabled, and missing administrator sessions.

## Follow-ups (pre-existing, out of scope for this PR)

Tracked as separate issues — not blockers for F4 merge:

| Issue | Topic | Why out of scope |
|-------|-------|------------------|
| [#431](https://github.com/evenfire-ai/evenfire/issues/431) | External trigger rate limit keyed by service bearer (shared 10/min for all external-rest-api users) | Pre-existing on `dev` (`bearer \|\| userSessionToken` order unchanged except cookie fallback); fairness/abuse ticket |
| [#432](https://github.com/evenfire-ai/evenfire/issues/432) | Bearer-only blocked on `/admin/outputs` and `/admin/workflow-recipes/*/allowed-teams` by global cookie middleware | Pre-existing `app.ts` mount mismatch; Control UI (cookie) unaffected |

Additional follow-up candidates (not yet filed): rate limits on `runs.routes.ts` (8 handlers), stronger integration tests for edge limiter 429 paths, `GET /admin/outputs` happy-path coverage.

## Test plan

- [x] `cd control-api && npm run test:run -- test/services.adminSessionAuth.test.ts test/routes.adminWorkflows.test.ts test/middleware.controlUIAuth.test.ts test/workflows.shared.rateLimit.test.ts` — 134 passed
- [x] `cd control-api && npx tsc --noEmit` — pass
- [x] `cd control-api && npm test` — 4512 passed, 21 skipped
- [x] `cd control-api && npm run test:run -- test/routes.externalWorkflows.test.ts test/routes.mcp-host.workflows.test.ts` — 84 passed (no regression on external/mcp-host lanes)
- [x] CI + CodeQL on head `a225d5e5` — green
- [ ] Minikube T0/T1/T2 — not run locally on this branch; pending post-merge or manual gate


control-api/src/routes/admin/workflows/grants.routes.ts
```diff
@@ -169,10 +180,10 @@ export function createAdminWorkflowGrantRoutes(gateway: K8sGateway): Router {
 
   router.get(
     `/admin/workflow-recipes/:ns/:name/allowed-teams`,
-    workflowGrantReadRateLimit(),
+    ...WORKFLOW_GRANT_READ_RATE_LIMITS,
+    bindAdminWorkflowAuth,
```
alfredolopez80, line 180: **CodeQL #81 — fixed in this PR (Capa 1; do not dismiss in Security tab)**

Commits `649e4e10` + `7a5a341e` on `fix/f4-admin-workflow-live-revocation`.

This admin workflow grant route uses dual-layer rate limiting: an `express-rate-limit` edge backstop (CodeQL-visible) plus PG `rateLimitMiddleware` (cross-replica source of truth), 60/min read or 20/min write per admin credential hash (bearer or `control_ui_admin_session` cookie).

`bindAdminWorkflowAuth` (`requireAdminWorkflowCallerMiddleware`) runs after the limiters and before the business handler, keeping live admin session checks (F4) on bounded traffic.

CodeQL marks the matching instance `fixed` on `refs/pull/428/merge`; it will close on `dev`/`main` after merge without a manual dismiss.

control-api/src/routes/admin/workflows/grants.routes.ts
```diff
@@ -193,10 +204,10 @@ export function createAdminWorkflowGrantRoutes(gateway: K8sGateway): Router {
 
   router.put(
     `/admin/workflow-recipes/:ns/:name/allowed-teams/:teamId`,
-    workflowGrantWriteRateLimit(),
+    ...WORKFLOW_GRANT_WRITE_RATE_LIMITS,
+    bindAdminWorkflowAuth,
```
alfredolopez80, line 204: **CodeQL #82 — fixed in this PR (Capa 1; do not dismiss in Security tab)**

Commits `649e4e10` + `7a5a341e` on `fix/f4-admin-workflow-live-revocation`.

This admin workflow grant route uses dual-layer rate limiting: an `express-rate-limit` edge backstop (CodeQL-visible) plus PG `rateLimitMiddleware` (cross-replica source of truth), 60/min read or 20/min write per admin credential hash (bearer or `control_ui_admin_session` cookie).

`bindAdminWorkflowAuth` (`requireAdminWorkflowCallerMiddleware`) runs after the limiters and before the business handler, keeping live admin session checks (F4) on bounded traffic.

CodeQL marks the matching instance `fixed` on `refs/pull/428/merge`; it will close on `dev`/`main` after merge without a manual dismiss.

control-api/src/routes/admin/workflows/grants.routes.ts
```diff
@@ -232,10 +243,10 @@ export function createAdminWorkflowGrantRoutes(gateway: K8sGateway): Router {
 
   router.delete(
     `/admin/workflow-recipes/:ns/:name/allowed-teams/:teamId`,
-    workflowGrantWriteRateLimit(),
+    ...WORKFLOW_GRANT_WRITE_RATE_LIMITS,
+    bindAdminWorkflowAuth,
```
alfredolopez80, line 244: **CodeQL #83 — fixed in this PR (Capa 1; do not dismiss in Security tab)**

Commits `649e4e10` + `7a5a341e` on `fix/f4-admin-workflow-live-revocation`.

This admin workflow grant route uses dual-layer rate limiting: an `express-rate-limit` edge backstop (CodeQL-visible) plus PG `rateLimitMiddleware` (cross-replica source of truth), 60/min read or 20/min write per admin credential hash (bearer or `control_ui_admin_session` cookie).

`bindAdminWorkflowAuth` (`requireAdminWorkflowCallerMiddleware`) runs after the limiters and before the business handler, keeping live admin session checks (F4) on bounded traffic.

CodeQL marks the matching instance `fixed` on `refs/pull/428/merge`; it will close on `dev`/`main` after merge without a manual dismiss.

control-api/src/routes/admin/workflows/grants.routes.ts
```diff
@@ -39,10 +50,10 @@
 
   router.get(
     `${BASE}/:ns/:name/grants`,
-    workflowGrantReadRateLimit(),
+    ...WORKFLOW_GRANT_READ_RATE_LIMITS,
+    bindAdminWorkflowAuth,
     asyncHandler(async (req: Request, res: Response) => {
-      const caller = await requireAdminWorkflowCaller(req, res)
-      if (!caller) return
+      const caller = adminCaller(req)!
```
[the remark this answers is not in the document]
alfredolopez80: CodeQL https://github.com/evenfire-ai/evenfire/pull/82 — fixed in this PR (Capa 1; do not dismiss in Security tab)

Commits 649e4e10 + 7a5a341e on fix/f4-admin-workflow-live-revocation.

This admin workflow grant route uses dual-layer rate limiting: an express-rate-limit edge backstop (CodeQL-visible) plus PG rateLimitMiddleware (cross-replica source of truth), 60/min read or 20/min write per admin credential hash (bearer or control_ui_admin_session cookie).

bindAdminWorkflowAuth (requireAdminWorkflowCallerMiddleware) runs after the limiters and before the business handler, keeping live admin session checks (F4) on bounded traffic.

CodeQL marks the matching instance fixed on refs/pull/428/merge; it will close on dev/main after merge without a manual dismiss.

control-api/src/routes/admin/workflows/grants.routes.ts
alfredolopez80, line 180: **CodeQL #81 — fixed in this PR (Capa 1; do not dismiss in Security tab)**

Commits `649e4e10` + `7a5a341e` on `fix/f4-admin-workflow-live-revocation`.

This admin workflow grant route uses dual-layer rate limiting: an `express-rate-limit` edge backstop (CodeQL-visible) plus PG `rateLimitMiddleware` (cross-replica source of truth), 60/min read or 20/min write per admin credential hash (bearer or `control_ui_admin_session` cookie).

`bindAdminWorkflowAuth` (`requireAdminWorkflowCallerMiddleware`) runs after the limiters and before the business handler, keeping live admin session checks (F4) on bounded traffic.

CodeQL marks the matching instance `fixed` on `refs/pull/428/merge`; it will close on `dev`/`main` after merge without a manual dismiss.

control-api/src/routes/admin/workflows/grants.routes.ts
alfredolopez80, line 204: **CodeQL #82 — fixed in this PR (Capa 1; do not dismiss in Security tab)**

Commits `649e4e10` + `7a5a341e` on `fix/f4-admin-workflow-live-revocation`.

This admin workflow grant route uses dual-layer rate limiting: an `express-rate-limit` edge backstop (CodeQL-visible) plus PG `rateLimitMiddleware` (cross-replica source of truth), 60/min read or 20/min write per admin credential hash (bearer or `control_ui_admin_session` cookie).

`bindAdminWorkflowAuth` (`requireAdminWorkflowCallerMiddleware`) runs after the limiters and before the business handler, keeping live admin session checks (F4) on bounded traffic.

CodeQL marks the matching instance `fixed` on `refs/pull/428/merge`; it will close on `dev`/`main` after merge without a manual dismiss.

control-api/src/routes/admin/workflows/grants.routes.ts
alfredolopez80, line 244: **CodeQL #83 — fixed in this PR (Capa 1; do not dismiss in Security tab)**

Commits `649e4e10` + `7a5a341e` on `fix/f4-admin-workflow-live-revocation`.

This admin workflow grant route uses dual-layer rate limiting: an `express-rate-limit` edge backstop (CodeQL-visible) plus PG `rateLimitMiddleware` (cross-replica source of truth), 60/min read or 20/min write per admin credential hash (bearer or `control_ui_admin_session` cookie).

`bindAdminWorkflowAuth` (`requireAdminWorkflowCallerMiddleware`) runs after the limiters and before the business handler, keeping live admin session checks (F4) on bounded traffic.

CodeQL marks the matching instance `fixed` on `refs/pull/428/merge`; it will close on `dev`/`main` after merge without a manual dismiss.

alfredolopez80:
@claude Please review PR #428 (F4 live admin revocation on workflow routes).

Scope: `authenticateAdminSession`, grant route rate limiting (PG + express backstop), auth middleware error propagation (`bb5ab65f`).

Focus: security regression on `/admin/workflows/*`, cookie vs bearer parity, rate-limit/auth ordering, and test coverage gaps.

CI may still be running on `bb5ab65f`; re-check checks before final verdict.

jozer-rami:
## Deep review — F4 live admin revocation

Independent review off a worktree at PR head `a225d5e5` (0 commits behind `dev`, all 33 checks green). Method: three independent lenses (security/authz, correctness, test-integrity), every finding verified against source before it survived, plus **mutation testing** of the load-bearing assertions.

**Short version: the auth half is good work and I'd land it as-is. The rate-limiting half is what I'd hold — it ships an unauthenticated admin lockout, and its own tests can't see it.**

---

### What is genuinely solid

- **The F4 chokepoint is real.** `verifyAdminToken` and `isAdminTokenRevoked` now have exactly one call site each in all of `control-api/src` — `services/adminSessionAuth.ts:19` and `:24`. No residual bypass anywhere in the service.
- **The 20-route matrix is honest, not decorative.** Mutation-proven: deleting the `sessionVersion` compare (`adminSessionAuth.ts:32`) fails **all 20** entries; deleting the revocation check fails 4; deleting the status check fails 3. The route tests never stub `adminSessionAuth.js`, so the real function executes — live revocation is genuinely exercised, not simulated.
- **The "all 20 protected admin workflow handler paths" claim checks out** — enumerated 3 (`read.routes`) + 7 (`grants.routes`) + 1 (`trigger.routes`) + 8 (`runs.routes`) + 1 (`outputs`) = 20, matched 1:1. No route missing.
- **Skipping `lockedUntil` is not just defensible, it's required.** `lockedUntil` is set only by `registerAdminFailedLogin` (`adminAuthService.ts:249-265`). Honoring it in session validation would hand anyone who knows an admin's username a session-kill primitive: fail 5 logins, victim's live session dies. Real deactivation bumps `session_version` anyway (`adminAuthService.ts:589`), and `db.ts:247` constrains status to `('active','disabled')`, so `status !== 'active'` is exhaustive. The comment in `adminSessionAuth.ts` is correct.
- `tsc --noEmit` failures are pre-existing and in files this PR doesn't touch.

---

### 🔴 Blocker — the pre-auth IP-keyed edge limiter lets an unauthenticated client lock every admin out of grant management

`routes/workflows/shared/rateLimit.ts:34-44` · `app.ts:73,130-141` · `grants.routes.ts:45-46` · `control-ui/app/control-api/[...path]/route.ts:143-151`

Three facts compose into a regression:

1. `app.ts:133-135` deliberately skips `requireAuthForControlUI` for `/admin/workflows` and `/admin/workflows/*` (documented at `routes/workflows/shared/auth.ts:17-21`). On those paths the new limiters are the **first** thing that runs.
2. The edge key is IP-only (`rateLimit.ts:38-40`), and every browser call is re-issued server-side by the control-ui pod, whose proxy copies headers verbatim and **never appends its own peer address** (`route.ts:143-151`). With `app.set('trust proxy', 1)`, `req.ip` is therefore the rightmost XFF entry the client supplied — or the control-ui pod IP when no XFF exists.
3. `deploy/` ships **no Ingress manifest at all**, and the documented self-hosted path is `kubectl port-forward` to control-ui. There is no XFF-attesting hop in the shipped topology.

**Two exploits, one root cause.**

*Omit XFF:* 20 anonymous write requests/min all land in `workflow_grants_write_edge:ip:<control-ui pod IP>` — the same bucket every real admin shares, because the control-ui pod is the socket peer for all of them. express-rate-limit counts on arrival and there's no `skipFailedRequests`, so the attacker's own 401s fill it. Every admin then gets 429 on grant writes for the rest of the window; their valid session is irrelevant because the limiter runs first.

*Rotate XFF:* send `X-Forwarded-For: <random>` per request. Fresh bucket every time, the backstop never fires, and each request still reaches the PG limiter's `INSERT … ON CONFLICT` **before** authentication (`rateLimit.ts:144-150` → `rateLimitMiddleware.ts:42` → `rateLimiterService.ts:82-88`). One anonymous HTTP request = one new `rate_limit_buckets` row.

This is newly introduced. On `dev`, `hashedBearerBucket` returned `null` for credential-less callers and `rateLimitMiddleware.ts:35-40` fails open, so anonymous traffic was never counted.

The comment at `rateLimit.ts:34-37` — *"Ingress-style edge key: IP-only so unverified bearer rotation cannot evade the backstop"* — does not hold in this topology. Worth contrasting with `routes/external/gfs.ts:404-407`, which uses the identical IP-key pattern but explicitly earns it: *"the authenticated external-rest-api boundary overwrites XFF with its trust-proxy-derived client address; externalGfsSourceIp accepts only that valid first IP, so a raw caller cannot select a bucket."* That boundary doesn't exist on this path.

**Fix:** put `bindAdminWorkflowAuth` ahead of the limiters and key on the credential — `adminWorkflowRateLimitCredential` is already written at `rateLimit.ts:17-23`. CodeQL visibility doesn't require an IP key. If a genuine pre-auth backstop is wanted, key it on a proxy-attested identity (`middleware/externalClientIdentity.ts:12-25` has that pattern).

> This is the one point where I'd differ from the earlier automated review, which concluded *"the edge limiter falls back to an IP key, so anonymous traffic is still bounded."* It is bounded — but the bound is shared with every legitimate admin, and the key is client-supplied.

---

### 🟠 Major — grant writes newly cap at 20/min on a path the UI fans out with `Promise.all`

`rateLimit.ts:25-32` · `control-ui/components/WorkflowAccessPanel/index.tsx:300-303`

The cookie fallback makes the `workflow_grants_write` bucket bind on browser sessions for the first time (previously `null` key → fail-open → browser writes entirely unmetered). `persistApprovalTeams` issues `added.length + removed.length` **concurrent** requests:

```ts
await Promise.all([
  ...added.map(teamId => allowWorkflowApprovalTeam(namespace, recipeName, teamId)),
  ...removed.map(teamId => revokeWorkflowApprovalTeam(namespace, recipeName, teamId)),
])
```

A single save changing >20 team memberships gets 429 on the tail, `Promise.all` rejects, and the writes that already succeeded stay applied — partial write behind a generic failure toast. Roughly 20 individual checkbox toggles in a minute does it too, since each is a write plus an echo read.

**Fix:** batch `allowed-teams` into one bulk endpoint (as `/grants` and `/team-grants` already are), or serialize with backoff the way `RecipeEditor.tsx:1702-1710` does — and/or raise `WORKFLOW_GRANT_WRITE_PER_MINUTE` now that it actually binds.

---

### 🟠 Major — the entire rate-limit deliverable is untested

`test/workflows.shared.rateLimit.test.ts:49-68`

The one test claiming to prove the edge backstop constructs its **own** `rateLimit({ limit: 2, standardHeaders: false, handler: … })` inline. It imports `workflowGrantEdgeRateLimitKey` but never `workflowGrantReadEdgeRateLimit()` or any `*RateLimits()` factory — so it tests express-rate-limit, not this PR.

Mutation-proven: stripping the edge limiter out of `workflowGrantReadRateLimits()`, `workflowAdminReadRateLimits()` and `adminOutputsReadRateLimits()` — the PR's entire CodeQL-motivated deliverable across 11 admin routes — leaves the suite **134/134 green**. No test anywhere asserts a 429 from a real limiter, and `workflowGrantEdgeRateLimitHandler`'s `retryAfterSeconds` derivation (`rateLimit.ts:46-55`) is never exercised.

The test also can't observe the blocker above: `test/workflows.shared.rateLimit.test.ts:61` builds its app without `app.set('trust proxy', 1)`, so XFF is never honored.

**Fix:** assert `workflowGrantReadRateLimits().length === 2` (and the other three factories), plus one 429 test against a router built by `createAdminWorkflowGrantRoutes`.

---

### 🟠 Major — `/admin/outputs` has no happy-path test, so a total outage would ship green

`test/routes.adminWorkflows.test.ts:2753`

`/admin/outputs` appears in the entire test suite twice: the import at `:4` and one 401 matrix entry. Because `requireBoundAdminWorkflowCaller` also returns 401, deleting `bindAdminWorkflowAuth` at `src/routes/admin/outputs.ts:36` leaves the suite **134/134 green** while the route denies every legitimate admin forever. The identical mutation on read/grants produced 28 failures — those routes have happy-path coverage; this one doesn't.

**Fix:** add a 200 test for `GET /admin/outputs` with a live admin session.

---

### 🟠 Major — a mock-queue shim silently couples 12 unrelated assertions to middleware count

`test/routes.adminWorkflows.test.ts:186-188`

`mockRateLimiterAllowed()` is queue alignment, not an assertion. `checkAndIncrement` fails open when rows are empty (`rateLimiterService.ts:110-122`), so the fake `{count: 1}` never influences a limiter decision — it only consumes a slot in the `mockPoolQuery.mockResolvedValueOnce` chain. Proof: removing the PG limiters broke 12 tests named things like *"lists approval target teams for admin callers"* and *"replaces grants, returns deduped canonical set"* — pure queue misalignment, nothing to do with rate limiting. Any future limiter add/remove on these routes silently corrupts a dozen unrelated assertions.

Related: `:537` **deletes** `expect(mockPoolQuery).not.toHaveBeenCalled()` from *"rejects mcp-host-control tokens on the admin workflows mount"*. That assertion was the only bound on pre-auth DB writes — precisely the property the blocker above violates. Nothing replaced it.

**Fix:** `vi.mock('../src/services/rateLimiterService.js')` here, the way `workflows.shared.rateLimit.test.ts:13` already does; and restore a `toHaveBeenCalledTimes` bound at `:537`.

---

### 🟡 Minor findings

**1. Each `*RateLimits()` call mints a fresh limiter, so the real quota is 3–4× the constant.** `rateLimit.ts:84-90,114-120` are invoked per route registration, producing 11 independent `rateLimit()` instances each with its own `MemoryStore`. Per client IP: grant reads 3×60 = 180/min, grant writes 4×20 = 80/min, admin reads 3×60 = 180/min — while the PG side shares one bucket per prefix at the nominal 60/20/60. The shared `workflow_grants_read_edge` prefix string reads like a shared bucket that doesn't exist. Both `routes/external/gfs.ts:331-399` and `routes/external/invitations.ts:37` hoist singletons; this diverges from that established pattern.

**2. The budget is on the wrong routes.** `read.routes.ts:26,39,70` gained 6 middlewares and 3 timers, but `GET /admin/workflows`, `/:ns/:name` and `/health` have **no shipped caller** in control-ui, desktop-app or profile-ui — the UI reads recipes via `GET /api/v1/admin/recipes` (`control-ui/lib/api.ts:2481`); the only callers are Playwright wait-loops. Meanwhile `runs.routes.ts` — 8 routes including run listing, artifact download and artifact delete — imports **no rate-limit middleware at all**, still holds authorization inline, and `/runs` is polled every 5s (`control-ui/app/workflow-recipes/[namespace]/[name]/page.tsx:67`), now paying a new `findAdminById` round-trip per poll.

**3. Four routes authenticate twice, and their bearer branch is unreachable.** `/admin/outputs` and the three `/admin/workflow-recipes/*/allowed-teams*` routes are **not** covered by the `app.ts:133-135` exemption (`'/workflow-recipes/x'.startsWith('/workflows/')` is false), so `requireAuthForControlUI` — cookie-only (`controlUIAuth.ts:10-12`) — already ran. `bindAdminWorkflowAuth` then re-runs `authenticateAdminSession`: 4 auth DB round-trips per request. The "bearer for automation" fallback at `auth.ts:24` can never fire on those four. Consider deriving the caller from `req.adminAuth` there.

**4. `workflowTriggerRateLimit` also flips from fail-open to enforced.** `rateLimit.ts:122-141` — the cookie fallback makes `POST /admin/workflows/:ns/:name/trigger` a hard 10/min per session (1 per 6s) for browser users, who previously sent neither credential the old key inspected. Correct in principle, but the PR body frames the cookie fallback as a grants-only change. Worth a release note and a sanity-check against real recipe-authoring loops.

**5. The new doc contradicts this PR.** `docs/how-to/codeql-js-missing-rate-limiting.md` lists alert #63 (`outputs.ts:34`) under **"Won't fix (accepted) — real gap, follow-up backstop PR"** with a *planned* 30/min tier — but this same PR shipped `adminOutputsReadRateLimits()` at `outputs.ts:35`. The tier table also omits the new admin-read (60) and outputs (30) tiers, and its "per admin credential hash unless noted" header is wrong for every edge limiter added here, which are per-IP.

**6. Four exports are module-private in practice.** `auth.ts:46` (`requireAdminWorkflowCallerMiddleware`), `auth.ts:61` (`adminWorkflowCaller`), `rateLimit.ts:61` and `:73` (`workflowGrantReadEdgeRateLimit` / `workflowGrantWriteEdgeRateLimit`) each have exactly one internal consumer.

---

### Checked and refuted (no action needed)

- **`Retry-After` parsing.** `standardHeaders: 'draft-7'` is truthy, so express-rate-limit sets the header before invoking a custom handler (`dist/index.mjs:1001-1006`). The string branch at `rateLimit.ts:51-52` is the live path; the `60` fallback isn't what runs. Only nit: the `typeof raw === 'number'` branch at `:49-50` is unreachable, since `setRetryAfterHeader` stringifies.
- **express-rate-limit v8 API.** Lockfile pins 8.6.1; `dist/index.d.ts:20` declares `ipKeyGenerator(ip: string, ipv6Subnet?)`, so `ipKeyGenerator(req.ip ?? 'unknown')` is the correct signature. The IPv6 validator only throws when the `keyGenerator` source contains `req.ip` without `ipKeyGenerator`; the compiled closure contains neither. No ValidationError, no warning.
- **No double-send in `bindAdminWorkflowAuth`.** `requireAdminWorkflowCallerMiddleware` (`auth.ts:46-59`) catches internally and calls `next(error)`, so the promise never rejects and the outer `.catch(next)` at `:79-81` is dead code rather than a second response. The 401 path returns without calling `next`.
- **No route can run without `bindAdminWorkflowAuth`.** Both routers are mounted exactly once (`app.ts:203`, `routes/admin/index.ts:33`), and `requireBoundAdminWorkflowCaller` fails closed with a `res.headersSent` guard.
- **No forced logout on deploy, no automation breakage.** `admin.sessionVersion` is coerced at `adminAuthService.ts:96` and the claim is normalized at `adminAuthToken.ts:56`, so `?? 0` vs `|| 0` is behaviourally identical here. Every bearer caller found (`scripts/smoke-workflow-grants.sh:81`, the e2e scripts, Playwright specs) mints its token via `POST /api/v1/admin/auth/login`, so the new `findAdminById` + status + version checks all pass. `signAdminToken` has no synthetic/service-account call site.
- **No 429-vs-401 oracle**, and the GET grant handlers discarding `caller` is not an IDOR — `workflowGrantManagementService.ts:32` validates the namespace itself.

---

### One closing observation

The only unchecked box in the test plan is *"Minikube T0/T1/T2 — not run locally on this branch."* That's precisely the gate that would have surfaced the blocker: the IP-bucket collapse is structurally invisible to the unit tests and only appears against a real control-ui → control-api hop.

Happy to split the rate-limiting changes onto their own branch so the F4 auth fix — which is solid and well covered — can land now.

alfredolopez80:
@claude **Closing re-review requested** — PR #428 at head `a225d5e5` (8 commits vs `dev`).

This is a **second review**, not a repeat of the 15:32 pass. Since your first review we landed **6 additional commits**, including rate-limit hardening and auth middleware fixes. Please validate the **full delta** and reconcile **all prior findings** in one response.

---

### Priority: validate the last two commits

| Commit | Summary | Files to read |
|--------|---------|-----------------|
| `91fcee84` | IP-only grant edge limiter; lazy limiter stacks per router; GET handler unused `caller` fix | `rateLimit.ts`, `grants.routes.ts`, `workflows.shared.rateLimit.test.ts` |
| `a225d5e5` | Rate limits on workflow read + `/admin/outputs`; cookie in `workflowTriggerRateLimit`; `bindAdminWorkflowAuth` helpers | `read.routes.ts`, `outputs.ts`, `auth.ts`, `rateLimit.ts`, `routes.adminWorkflows.test.ts` |

---

### Required deliverable: findings reconciliation table

For **every row below**, state: **Fixed in PR** / **Pre-existing (tracked)** / **Still open in PR** / **Won't fix (accepted)** — with file:line evidence and merge impact.

| ID | Source | Finding |
|----|--------|---------|
| F4-core | Original scope | Live admin revocation (`authenticateAdminSession`: JTI + active + sessionVersion) on all 20 workflow handlers + outputs |
| F4-err | `bb5ab65f` | Auth middleware `try/catch` + `.catch(next)` + DB failure propagation |
| F4-edge | `91fcee84` / adversarial review | Edge limiter IP-only (not bearer-keyed) so token rotation cannot evade backstop |
| F4-cookie-trigger | `a225d5e5` | Admin cookie included in `workflowTriggerRateLimit` (was fail-open for Control UI triggers) |
| F4-read-outputs | `a225d5e5` | PG + edge limits on read routes and `/admin/outputs` |
| CodeQL-grants | GHAS / prior review | `js/missing-rate-limiting` on grant routes — middleware order edge → PG → auth |
| CodeQL-tests | `4345d684`, `d8647ffc` | Test harness false positives resolved |
| SOL-1 | SOL closing audit | External trigger: service bearer wins over `x-user-session-token` → shared 10/min bucket → **[#431](https://github.com/evenfire-ai/evenfire/issues/431)** |
| SOL-2 | SOL closing audit | Bearer-only unreachable on outputs + allowed-teams (global cookie gate) → **[#432](https://github.com/evenfire-ai/evenfire/issues/432)** |
| JR-1 | @jozer-rami deep review | IP bucket collapse / trust proxy / pre-auth edge behavior |
| JR-2 | @jozer-rami | Rate-limit deliverable untested (edge factories, real 429) |
| JR-3 | @jozer-rami | `/admin/outputs` no happy-path 200 test |
| JR-4 | @jozer-rami | `mockRateLimiterAllowed` queue shim coupling unrelated tests |
| JR-5 | @jozer-rami | Multiple limiter instances per route (3–4× edge quota) |
| JR-6 | @jozer-rami | Limits on unused read routes vs hot `runs.routes.ts` unpaged |
| JR-7 | @jozer-rami | Double auth on outputs + allowed-teams |
| JR-8 | @jozer-rami | `docs/how-to/codeql-js-missing-rate-limiting.md` stale vs shipped outputs limiter |
| JR-9 | @jozer-rami | Admin trigger cookie limit 10/min — release-note / UX risk |
| CL-1 | Your first review (15:32) | Any items you flagged then — re-check against current head |

---

### Also confirm

1. **Blast radius** — control-api only; impact on high-volume paths (triggers, outputs, grant writes, polling).
2. **Merge verdict** — APPROVE / REQUEST CHANGES / COMMENT, with explicit list of **blockers vs follow-up issues**.
3. **CI** — head `a225d5e5` has green CI + CodeQL; do not re-run unless you see a gap.

Do **not** suggest splitting the PR unless you find a **security blocker** introduced by the last two commits.

zach88:
Took a deep pass through this one — auth/session core, the rate-limit rework, test quality, security, and the control-ui↔control-api contract. The F4 fix itself is solid: `authenticateAdminSession` is the single live-admin decision now (JWT sig + JTI revocation + active account + `sessionVersion`), all 20 admin workflow routes route through it, and `verifyAdminToken` is no longer used as an auth decision anywhere (bucket keying only). It's fail-closed on DB errors, and I reproduced T3 — the new F4 tests fail against the merge-base, so they're real enforcement, not golden-of-the-bug. What's left is UX/test polish on the control-ui side plus some rate-limit follow-ups.

## Verdict

🟡 **Approve with nits**

Nothing gating here — no Blocker/Critical/High, no regressions. This is my read from the review, not a formal GH approval.

None.

## Optional (non-blocking)

This PR can merge without these — follow-up or author's call.

1. **[R1-M1] `>20` approval-team edits produce a repeated, unactionable 429 message** — `control-ui/components/WorkflowAccessPanel/index.tsx:299-336` (+ `lib/api.ts:256-284` `apiSend`, `:88-133` `formatApiError`). Serializing allow/revoke against the shared 20/min bucket makes the 429 deterministic past team 21; the client treats 429 as a generic error (not 401), never reads `retryAfterSeconds`/`Retry-After`, and the panel concatenates one message per failed team → the user sees `"429 Too Many Requests - Too Many Requests; …"` repeated N−20 times with no hint to wait ~60s. Good news: integrity is fine — `loadApprovalTeams()` reloads from the server before the combined throw, so there's no silent partial write (which was the whole point of the change). It's message UX, not data. Suggest: in the per-team catch, detect `error.status === 429` and show a single "too many requests, retry in ~Ns" using `error.body.retryAfterSeconds`.

2. **[R1-M2] The declared invariant of the control-ui change has no test (T5)** — `control-ui/components/WorkflowAccessPanel/index.tsx:299-336`, test file `control-ui/components/__tests__/WorkflowAccessPanel.test.tsx`. The change serializes allow→revoke, collects errors, and throws after `loadApprovalTeams()` so a partial failure surfaces an error and doesn't fire the success toast — that's the "no partial write left behind a `Promise.all` rejection" invariant, and none of the 4 existing tests exercise a rejecting `allowWorkflowApprovalTeam`/`revokeWorkflowApprovalTeam`. I verified the flow is correct by hand (the throw precedes `showToast`, `mutating:false` clears in `finally`), so this is a missing test, not a bug. Add a case that rejects one write and asserts (a) no success toast, (b) the accumulated error message, (c) `loadApprovalTeams` was called.

3. **[R1-L1] `runs.routes.ts` (8 handlers) and `trigger.routes.ts:26` weren't migrated to `bindAdminWorkflowAuth`** — still `await requireAdminWorkflowCaller(req,res)` inside the handler. Not a security gap (the F4 fix lives inside that function, so they're covered), but it contradicts the stated point of the pattern ("keep authorization out of business handlers for SAST"). Either migrate them too, or note in the body why they're out of scope for this round.

4. **[R1-L2] The IP-fallback keying generalization (F2) didn't reach the trigger** — `control-api/src/routes/workflows/shared/rateLimit.ts:133-148` `workflowTriggerRateLimitCredential`. This PR fixed the "unverified credential → IP bucket" idiom in `hashedAdminWorkflowCredentialBucket`, but the sibling `workflowTriggerRateLimitCredential` (rewritten in this same PR) still returns the raw bearer for an unverified token. The underlying evasion reproduces identically on the base, so the defect itself is pre-existing — what's on this PR is the F2 discipline: the function was reshaped with that exact intent, three siblings got the fix, the trigger didn't, and there's no written class-check. Fair caveat: this function is shared with mcp-host/external callers whose valid non-admin bearers must key by raw token, so closing it cleanly means splitting keying by route family (issue-sized). At minimum: a bearer-rotation test on the trigger (there's a cookie one, none for bearer) and a one-line note on why the trigger keeps the raw bearer.

5. **[R1-L3] Same-subject bucket sharing can 429 an admin's live session** — `control-api/src/routes/workflows/shared/rateLimit.ts:27-33`. Keying by `sub` (stricter against rotation) means a valid-signature but revoked/stale token of admin X can consume the `workflow_*:<hash(X)>` bucket and 429 X's legit session. Needs prior possession of a signed JWT for X, and it only causes rate-limiting (F4 still rejects the action — no bypass). Tradeoff: it kills the self-token-rotation evasion. If you want to isolate sessions, mix `jti` into the key; otherwise fine to accept and document.

6. **[R1-L4] The serialized loop swallows `AuthExpiredError` and re-fires the redirect** — `control-ui/components/WorkflowAccessPanel/index.tsx:303-320`. If the session invalidates mid-series, the per-team try/catch catches the silent `AuthExpiredError`, pushes `'Session expired'`, and keeps going; each remaining iteration re-calls `handleUnauthorized()` (redirect + `clearAdminAuthToken`), and you end up with "Session expired; …" repeated. The global redirect does fire on the first 401 so the user lands on login — it's cosmetic + redundant calls. Introduced by the serialization (the old `Promise.all` propagated once).

7. **[R1-L5] New read-path 429s aren't handled by control-ui's fetchers** — `control-api/src/routes/admin/workflows/read.routes.ts` (list/detail/health) and `admin/outputs.ts:35` now emit 429 (60/min, 30/min); `apiGet` (`lib/api.ts:195-254`) shows a bare `"429 Too Many Requests"`. Limits are generous so unlikely to trip — same class as R1-M1 on the read side.

8. **[R1-L6] Edge MemoryStore is per-registration, so the edge limit isn't aggregated** — `control-api/src/routes/workflows/shared/rateLimit.ts:85-100`. `createWorkflowEdgeRateLimit` builds a fresh `rateLimit()` (own store) per route, so the N grant-write routes each get their own 20/min edge bucket → a verified subject can do ~20/min×N at the edge. Mitigated: the PG limiter (source of truth) does aggregate to a real 20/min. Matches the "Shared MemoryStore … follow-up (Low 1)" comment — fine as documented.

Minor nits, take or leave: `next()` sits inside the `try` in `requireAdminWorkflowCallerMiddleware` (`auth.ts:55`) — harmless today since the next middleware is throw-safe `asyncHandler`, but worth hoisting out; the `.catch(next)` in `bindAdminWorkflowAuth` (`auth.ts:79-81`) is dead defensive code (the middleware already does `next(error)`); the verified-subject key marker is inconsistent (PG `${prefix}:${hash}` vs edge `${prefix}:sub:${hash}` — `rateLimit.ts:39/65`); `verifyAdminToken` runs ~3× per authenticated request (edge keyer + PG keyer + `authenticateAdminSession`); and `middleware.controlUIAuth.test.ts` moving from real supertest/express to a hand-rolled `mockRes` drops some integration fidelity (no bug masked — the SUT only touches `req.header`/`res.status().json()`/`next()`).

## What checked out clean

- **F4 closed fail-closed across all 20 routes** (read 3, grants 7, trigger 1, runs 8, outputs 1) — enumerated on both the auth and security passes; `verifyAdminToken` no longer an auth decision anywhere.
- **T3 reproduced** — the new F4 tests fail against merge-base `9e20e575` (stale/disabled/missing → 401 with the fix, 500/non-401 without it). Real enforcement.
- **Fail-closed on DB errors** — a throwing `findAdminById`/`isAdminTokenRevoked` propagates → 500, never grants access or degrades to a permissive 401.
- **No new fail-open** — the cookie-trigger fail-open is closed, and the `null` credential path is strictly narrower than before.
- **Anti-poisoning between admins** — verified admins isolated in `:sub:<hash>`; only unverifiable credentials fall to the IP bucket, and an attacker can't sign a token with the victim's `sub`. `trust proxy=1` makes the IP fallback meaningful.
- **External / mcp-host trust roots untouched** — `requireExternalWorkflowCaller` and `requireMcpHostControlWorkflowCaller` unchanged.
- **429 body shape is identical** between the PG limiter and the edge limiter (`{error:'Too Many Requests', retryAfterSeconds}` + `Retry-After`), and `Retry-After` parsing handles number/string/missing/negative without crashing.
- **`/admin/outputs` 200 shape** `{workflowOutputs, chatArtifacts, warnings?}` matches `AdminOutputsOverview` and the consumer; 401/500 preserved.
- **Suites**: control-api HEAD `services.adminSessionAuth` + `middleware.controlUIAuth` + `routes.adminWorkflows` + `workflows.shared.rateLimit` = 147 passed; control-ui `WorkflowAccessPanel.test.tsx` = 4 passed.

zach88 (approved):

PR #429, opened 2026-08-21
fix: show deeplink confirm dialog above an open sandbox app

zach88:
## What

Fixes the "Open app link?" / "App link could not be opened" confirmation dialogs being drawn **behind** an already-open sandbox app, where the user cannot see or click them.

When a deeplink (`evenfire://app/<ns>/<name>`) arrives while another app is already mounted, the confirmation dialog now correctly appears on top: the native embed is hidden for the lifetime of the dialog and restored when it closes.

## Why

The confirmation dialogs render as renderer DOM at the App shell level, but a sandbox app is embedded in a native `WebContentsView`, which **always paints above renderer DOM regardless of CSS z-index**. With an app mounted, the dialog was painted underneath the embed and was unreachable — the user saw no confirmation and could not open the deeplinked app. As a side benefit this is fail-safe against a malicious embed: the untrusted content is hidden while the user makes the trust decision, so it cannot overlay or capture the confirmation.

## How

- Add a `deepLinkShellOverlayOpen` signal, bound to `sandboxUiDeepLinkDialog !== null` — the exact value that decides whether the dialog renders, so the signal cannot drift from the dialog it guards.
- Fold it into `SandboxUiPage`'s existing `shellOverlayOpen`, so the deeplink dialogs reuse the same hide/restore cycle already used by the header/sidebar/toast overlays (`capturePreview` + `setVisible(false)`, then `setVisible(true)` + re-apply bounds on close).

This adds no new decision logic and no new IPC surface — it routes through the existing `sandboxUi.setVisible` channel and the existing visibility `useLayoutEffect`. The deeplink state machine, the OS protocol router, and the pending-link queue are untouched. The new prop keeps the `*ShellOverlayOpen` family name so the full set stays greppable.

## Testing

- `SandboxUiPage.test.tsx` and `App.deepLinks.test.tsx`: 58/58 green.
- New regression tests assert the observable effect (jsdom does not model the native `WebContentsView`): the embed is hidden (`setVisible(false)`) while the deeplink dialog is up and restored (`setVisible(true)`) on close, and the `deepLinkShellOverlayOpen` prop tracks both the confirm and the failed/retry dialogs. The core production hunks fail against the prior head and pass with the fix.
- Coverage hardened after review: the Retry branch of the failure dialog (`onConfirm` → `handleRetryFailedSandboxUiDeepLink`) is now exercised and pinned to drop the signal in lockstep with the dialog; a multi-overlay case (deep-link + header both open, one closes) proves the embed stays hidden; overlay assertions use exact `setVisible` call counts and identity (`toBe(false)`) rather than `toBeFalsy`, and the deep-link preview `src` is asserted.
- `verify:electron` passes; typecheck introduces no new errors.

## Notes on the changed-file set

Two files beyond the core fix appear in the diff and are accounted for here:

- `useGfsBrowserController.test.tsx` — re-applies the `swallow()` helper for fire-and-forget probe calls. The original commit's content was silently discarded by an earlier merge that took dev's rewrite of the file wholesale; it is restored here so the PR ships the fix its commit list advertises.
- `config.onboardingPreview.test.ts` — an unrelated reformat that had leaked in via a merge has been dropped; the file now matches `origin/dev`.

The three-dot GitHub view (`gh pr diff`) additionally lists `control-ui/components/RecipeEditor.tsx`, `control-ui/components/SecretsTable.tsx` and `control-ui/lib/codexSubscription.ts`. These are pure merge-base artifacts: the two-dot diff against `origin/dev` is 0 lines for each, so they are byte-identical to dev and no-ops on merge.

## Risks / Follow-ups

- Low-severity cosmetic flicker: on the failed → Retry → fail-again path with another app mounted, the underlying embed is restored for the duration of each retry attempt (a full hide/restore cycle per attempt across the backoff ladder). Left as-is intentionally — keeping it hidden for the entire retry window (up to ~30s across the ladder) would be worse UX than the brief restore.
- Pre-existing, out of scope for this PR (candidates for separate issues): embed visibility has two owners — `GfsImagePreview` writes `setVisible` imperatively (`App.tsx:475`/`:494`), bypassing `shellOverlayOpen`, so a GFS preview close can re-show the embed over a live dialog; `capturePreview()` has no timeout, so a hung embed can defer the hide; and `DesktopUpdateRequiredDialog` (`App.tsx:1937`) is a hard-stop modal that raises no overlay signal. A single owner deriving "any shell-level modal is mounted" would subsume these.
- Follow-up (separate proposal, not in this PR): replace the blocking modal with a non-blocking, approval-style notification in the notification bell, which also inherits the embed-hide coordination.


alfredolopez80:
## Review — PR #429 · `fix: show deeplink confirm dialog above an open sandbox app`

Parallel multi-agent review (4 agents: correctness · security · tests · contract/architecture) with each finding verified against the code at PR head (`3e70c222`). Scope = the PR diff only (5 files, +127/-1).

**Recommendation: does not block merge.** Clean, well-scoped fix; the core mechanism is verified correct.

| Severity | Count |
|----------|-------|
| Blocker / Critical / High | 0 |
| Medium | 2 (1 test-coverage · 1 **pre-existing** security caveat) |
| Low | 5 |

### ✅ Verified clean (checked, not assumed)
- **Signal is byte-identical to the dialog condition** — `App.tsx:1881-1883` `Boolean(pendingSandboxUiConfirmation || failedSandboxUiDeepLink)` is exactly the condition that renders the dialog at `App.tsx:1378-1388`; both derive from the same two variables (`1327-1334`) in the same render, so signal and dialog cannot desync.
- **Restore path is sound** — `SandboxUiPage.tsx:503-550`: open → `capturePreview` + `setVisible(false)`; close → `setBounds` + `setVisible(true)`, with a `cancelled` guard against stale async runs. Deps `[launch.kind, shellOverlayOpen]` cover both transitions.
- **OR composition is race-free** — if header/sidebar/toast overlays are still open when the deeplink closes, the OR keeps the embed hidden; no flip-flop.
- **No new IPC, no fail-open** — same `sandboxUi:setVisible` channel (boolean-validated + `assertTrustedSender`); the new term can only hide *more*, never less.
- **Embed cannot bypass the hide** — `WebContentsView` with `sandbox:true` / `contextIsolation:true` / dedicated preload, loading an http(s) URL that `assertTrustedSender` rejects.
- **Contract is complete & symmetric** with the three sibling overlay props; `Boolean(...)` coercion is required (operands are `object | null | undefined`).

### 🧪 Test suite (executed)
- **PR head: 56/56 green, exit 0** (SandboxUiPage 24 + App.deepLinks 32) — matches the PR claim.
- **With the 3 production hunks reverted to base: exactly the 3 new tests fail**, nothing else regresses → the "fails against prior head" claim is honest, and the R1-L1 coverage-completeness nuance in the commit message is accurate.

---

### Medium

**M1 · tests — Retry branch of the failure dialog is untested.**
`App.tsx:1400` wires `confirmLabel="Retry"` → `handleRetryFailedSandboxUiDeepLink` (`App.tsx:1361`), but no test exercises Retry. That is exactly the flicker path (failed → Retry → in-flight → fail-again) where both predicates are null and `deepLinkOverlayOpen` drops to `false` for a frame. A test pinning "overlay stays raised / embed never left uncovered with a decision pending across Retry" would either prove no-flicker or expose it. Follow-up.

**M2 · security — *pre-existing, not introduced by this PR* — GFS-preview close can re-show the embed over a live deeplink dialog.**
`App.tsx:491-494` `closePluginGfsPreview` calls `setVisible(true)` **directly**, on a parallel path that does **not** go through `shellOverlayOpen`. Scenario: GFS image preview open (embed hidden via the direct `setVisible(false)` at `App.tsx:474`) → a deeplink dialog appears → the user closes the GFS preview → embed forced visible; since the visibility effect's deps `[launch.kind, shellOverlayOpen]` (`SandboxUiPage.tsx:550`) don't change, it does not re-hide, and the embed covers the still-pending dialog for the rest of its lifetime. This reintroduces the original bug on that path and qualifies the "hidden for the FULL lifetime of the trust decision" wording. `git blame` → `3559031b`, predates this PR. Suggested direction: centralize embed-visibility ownership in `shellOverlayOpen` instead of parallel direct `setVisible` callers. Not decision-capture (the embed cannot press the dialog's DOM buttons). Separate follow-up.

### Low
- **L1 · architecture — predicate duplication.** `App.tsx:1881` re-derives `Boolean(a||b)` while `sandboxUiDeepLinkDialog !== null` (`1378`) already means "a dialog exists". Equivalent today, but if the dialog ternary gains a third branch, line 1881 drifts silently. `deepLinkOverlayOpen={sandboxUiDeepLinkDialog !== null}` binds the signal to the actual dialog. (Flagged by 2 agents.)
- **L2 · security — async hide window.** The hide is async (1 `capturePreview` IPC + 1 rAF, `SandboxUiPage.tsx:503-520`), so for tens of ms the dialog is mounted while the embed still paints on top. Not decision-capture (the user can't reach the buttons in that window either). Inherent to async `setVisible`.
- **L3 · security (unverified live) — retained keyboard focus.** `setVisible(false)` stops painting but does not blur the embed's `webContents`; if it held focus, it may keep receiving keystrokes while hidden. Cannot cause a false confirm. Worth a live check.
- **L4 · tests — no multi-overlay interaction test.** `shellOverlayOpen` is now a 4-way OR; nothing pins "deeplink + header both open, one closes → embed stays hidden" against a future refactor.
- **L5 · naming — prop suffix.** Siblings are `*ShellOverlayOpen`; the new one is `deepLinkOverlayOpen` (no "Shell"). Defensible; cosmetic.

---

No open review comments or objections existed on the PR at review time. Nothing here is blocking — M1/M2 are follow-ups at the author's discretion; L1 is a cheap robustness improvement.

<sub>🤖 Multi-agent review (Fable 5 fan-out) via Claude Code. Findings verified against real code; no code was modified.</sub>

alfredolopez80:
## Review — PR #429 · `fix: show deeplink confirm dialog above an open sandbox app`

Re-review at head `7dae5b20` (the [previous review](https://github.com/evenfire-ai/evenfire/pull/429#issuecomment-5375324567) was at `3e70c222`). Five parallel agents — correctness · security · tests · history · contract/types. Scope = the PR diff only (6 files, +128/−5). Suites executed, not assumed.

**Recommendation: does not block merge.** The three production hunks are correct, monotone, and genuinely test-covered (RED→GREEN reproduced). Of 15 findings, **one is introduced by this PR** (M1, a lost commit — history hygiene, not a product defect); the rest are pre-existing defects of the same class, surfaced because this PR makes them newly relevant.

| Severity | Count | Introduced here |
|---|---|---|
| Blocker / Critical | 0 | 0 |
| High | 1 | 0 |
| Medium | 4 | 1 |
| Low | 10 | 4 |

**Nine findings are posted as inline comments** on their exact lines. The six below live in code this PR does not touch, so GitHub cannot anchor them.

---

## H1 · High · PRE-EXISTING · REPRODUCED · `App.tsx:475`, `:492-495`
**The GFS image preview writes embed visibility imperatively, and breaks in both directions.**

`GfsImagePreview` hides the embed with a direct `setVisible(false)` (`:475`) and restores it with a direct `setVisible(true)` (`:494`), never touching `SandboxUiPage`. With `SandboxUiPage.tsx:520`/`:539` these are the only four `setVisible` callers in the renderer.

**A — the embed covers a live deep-link dialog, permanently.** Embed mounted → a plugin calls `gfs.open()` on an image → `:475` hides it → a deep link arrives, `deepLinkOverlayOpen` goes true → the user closes the preview → `closePluginGfsPreview` fires `setVisible(true)` **while "Open app link?" is still up**, unconditionally, with an empty dep array. `SandboxUiPage`'s effect cannot correct it: `shellOverlayOpen` never changed, and it is in the deps (`SandboxUiPage.tsx:550`).

> **Reproduced.** Throwaway App-level test (since deleted): with `title === 'Open app link?'` and `deepLinkOverlayOpen === true`, recorded IPC was `setVisible calls = [[false],[true]]` — final call `true`.

**B — the embed covers the preview itself.** Because the hide at `:475` was imperative, `shellOverlayOpen` stays `false`, so `SandboxUiPage`'s effect believes nothing is overlaid. Any later run executes `setVisible(true)` (`:539`) over the open preview — reachable by a toast appearing and dismissing, or by `launch.kind` changing.

**Why it never self-corrects:** `preload.ts:582-614` exposes `setVisible` but no `isVisible`/`getState`. Neither owner can read back. Last writer wins, silently.

**Impact:** UI spoofing plus durable denial of the trust decision — *not* consent forgery. The covering embed swallows the click (a `WebContentsView` hit-tests above the DOM, not just paints above it), and the full-viewport `.confirm-dialog-backdrop` maps stray clicks outside the embed rect to **Dismiss**, never **Open**.

Blame `3559031b`, Andy Pavia, 2026-08-06. Ironically `App.tsx:473-474` already carries the exact reasoning on the *open* path; only the close path fails to reconcile.

---

## M1 · Medium · **INTRODUCED BY THIS PR** · REPRODUCED · commit `0369d90b`
**The PR advertises a fix whose content a merge silently discarded.**

`0369d90b` ("stop probe fire-and-forget calls leaking unhandled rejections") adds a `swallow()` helper to `useGfsBrowserController.test.tsx`. Merge `e0bd84dc` resolved a conflict by taking dev's side wholesale:

```
42c562f6    blob=0da60220  swallow=0     # before
0369d90b    blob=90c813f2  swallow=14    # the commit
e0bd84dc    blob=dc95e8c4  swallow=0     # <-- content GONE
7dae5b20    blob=dc95e8c4  swallow=0     # = HEAD
origin/dev  blob=dc95e8c4  swallow=0     # identical blob
```

`git merge-base --is-ancestor 0369d90b origin/dev` → exit 1 (PR-only commit). `git diff origin/dev...HEAD | grep -c swallow` → **0**. Dev had independently rewritten that file (`633ea500`, `2dc5ec90`, `93f211d0`).

The commit still sits in the PR's commit list with its full rationale, so anyone reading the history credits this PR with a fix it does not ship. No conflict marker survives; it is invisible in the diff. **Blast radius is exactly one file** — every added line of `3e4f2821` and `3e70c222` survives byte-for-byte.

**The lost commit was sound and worth re-applying.** The 13 sites are `onClick` handlers in throwaway probe components; assertions run on spies and rendered state, untouched by `.catch(() => {})`. The rejection was *already* discarded by `void` — `swallow()` only stops it escaping as a process-level unhandled rejection that fails an unrelated later test. Its FilesPage claim also checks out: `FilesPage.tsx:460-473` awaits `ctrl.createFolder` in `try/catch` and surfaces errors via `pushToast(..., 'error')`.

**Action:** drop the commit, or re-apply it onto dev's rewritten file.

---

## M2 · Medium · PRE-EXISTING · `App.tsx:1937`
**`DesktopUpdateRequiredDialog` is a hard stop with no escape, three lines above the dialog this PR fixes, and raises no overlay signal.**

The most severe instance of this defect class in the file — worse than the deep-link case, on three counts:

1. **No way out.** `App.tsx:183-235` exposes exactly one action, `<Button onClick={onDownload}>Download update</Button>`. No `onCancel`, no Dismiss, no backdrop handler — deliberate, since the workspace refuses to run on the old version. A dialog with Dismiss can be closed blind; this one cannot.
2. **Reachable mid-session.** `refreshDesktopReleaseStatus()` runs on every session-state application (`useAppController.ts:794`), not only at boot, so `updateRequired` can flip while an embed is mounted on the apps route.
3. **It lands inside the embed rect.** `.desktop-update-dialog-backdrop` is `position: fixed; inset: 0; place-items: center` (`styles.css:5944-5952`) — window-centered — while `.sandbox-ui-embed-slot` is `flex: 1` filling the content column (`styles.css:12464-12469`). Its `z-index: 2400` is irrelevant against a native view.

**Net effect: the app becomes unusable with no recovery path.** Blame `5d663a49`, Jose Ramirez, 2026-07-14.

The two `environmentSetup*` dialogs (`:1938-1939`) have the identical shape and also raise no signal, but I did **not** establish that either is reachable with an embed mounted — structurally exposed, reachability unproven.

---

## M3 · Medium · PRE-EXISTING · READ-ONLY · `SandboxUiPage.tsx:509-520`
**An unresponsive embed can keep the dialog unreachable indefinitely: `capturePreview` has no timeout.**

Hiding is gated on `await capturePreview()`; `setVisible(false)` is only reached once it settles. The gating is deliberate — the sibling test *"keeps the native view visible until the toast preview is ready"* pins it with `not.toHaveBeenCalled()` while the promise is pending, to avoid flashing an empty slot.

A hung embed `webContents` → `capturePage()` never settles → `setVisible(false)` never fires → the native view keeps painting over "Open app link?" and neither Open nor Dismiss is reachable. Same symptom this PR fixes, different cause.

For a toast the trade-off is fine (no input needed). For a modal it is a liveness hazard. Suggested fix: race the capture against ~150 ms and hide regardless.

**Not reproduced** — no hung `capturePreview` was constructed; reported from control flow plus the test that pins the gating.

---

## L9 · Low · PRE-EXISTING · `sandboxUiDriver.ts:855-859`
**Focus is pushed into a hidden embed while the trust dialog is up.**

`focusActiveSandboxUi()` checks only `!active` and `isDestroyed()` — no visibility guard. Reachable: open the command palette from inside the embed → a deeplink arrives and the dialog mounts → close the palette → `App.tsx:1137-1146` fires `focusActive()` into the still-hidden embed. `ConfirmDialog`'s Escape handler is a renderer `window` listener, so the user can be typing into an invisible untrusted app while looking at a visible dialog.

**Consent forgery is not reachable**: the embed is a separate `webContents` with no channel to synthesize key events into the renderer, and `ConfirmDialog` focuses `cancelButtonRef` on mount, so default activation is Dismiss.

**UNVERIFIED without headed Electron:** whether Chromium honors `focus()` on a view with `setVisible(false)` in effect. What *is* verified from source is that the app supplies no guard of its own — `setSandboxUiVisible` (`:740-743`) is two lines, no `blur()`, no focus transfer. Second caller, same shape: `SandboxUiPage.tsx:250-258`. Blame `2be6bcf68`, Josué, 2026-08-13.

---

## L10 · Low · PRE-EXISTING · `SandboxUiPage.tsx:519-523`
**A throwing `setVisible(false)` is silently swallowed — and the asymmetry matters.**

Four empty catches in this effect (`:510`, `:521`, `:532`, `:540`). A failed `setVisible(false)` (`:520`) leaves the embed painted over the trust dialog with the renderer never learning — **fail-open**. A failed `setVisible(true)` (`:539`) leaves it hidden — fail-closed and harmless. Only one of the four is dangerous, and it is the one this PR now exercises more often. Blame `106efe7c4`, Jose Ramirez, 2026-07-19.

---

## Verified clean — checked, not assumed

- **Signal ≡ dialog predicate, cannot desync.** `App.tsx:1884-1886` is exactly the predicate of the ternary at `:1379-1404`: same two consts (`:1328-1335`), same component body, no `useMemo`, no `React.memo` anywhere in the file, `SandboxUiPage` a plain function. Both commit in one pass.
- **Fail-closed, zero new IPC surface.** A 4th OR disjunct is arithmetically incapable of making the embed more visible. The `= false` default keeps every existing call site byte-identical. Same `sandboxUi:setVisible` channel, whose handler (`ipc.ts:1748-1754`) does `assertTrustedSender(event)` **and** `typeof payload?.visible !== 'boolean' → throw`.
- **Effect deps and async safety.** `[launch.kind, shellOverlayOpen]` is correct; the `cancelled` guard is checked after every `await` and before every mutating IPC (`:513`, `:517`, `:537`), so a cancelled run cannot issue a late `setVisible`. Every transition traced: no state leaves the embed permanently hidden.
- **Contract.** One production call site (`App.tsx:1874`); 45 occurrences total, 44 tests. `SandboxUiPageProps = {}` makes the whole object optional by design, and the page unmounts on route change, so the `= false` default is unreachable-with-an-embed-mounted rather than silently buggy.
- **`PluginConsentModal` is not affected** — the main process owns its hide (`pluginConsentGate.ts`).
- **No injection, no secrets.** `appRef` is a React text child (`:1382`, auto-escaped); `parseSandboxUiDeepLink` restricts `appRef`/`teamId` to ASCII, blocking homograph spoofing of the dialog text. Zero hits for console/log/token/secret/password/apikey/authorization/URLs across added lines.
- **Comment accuracy.** "The native WebContentsView always paints above renderer DOM" is correct for how this app mounts it (`sandboxUiDriver.ts:489` `addChildView`), and is already stated verbatim at `:735-741` and `pluginConsentGate.ts:13-17`.

## Gates executed

| Gate | Result |
|---|---|
| Node | v24.18.0 (contract requires 24) |
| `npm ci`, no `--ignore-scripts`, no `ELECTRON_SKIP_BINARY_DOWNLOAD` | EXIT=0 |
| `npm run verify:electron` | **EXIT=0** — real executable resolved |
| Targeted suites | **56 passed / 0 failed / total=56 > 0**, EXIT=0 |
| Full desktop-app suite | **213 files / 2101 tests passed**, EXIT=0 |
| `tsc --noEmit` | 14 errors at HEAD, 14 at base; normalized set-diff **empty both ways** → zero new |
| Prettier `--check`, all 6 files | EXIT=0 |
| vitest `retry:` | **none** (default 0) — no deterministic red can print green |

Exit codes captured with `; echo "EXIT=$?"`, never through a pipe; `total > 0` asserted on every run.

**All three author claims verified honest.** "56/56 green" reproduced. "The 3 new tests fail against the prior head" reproduced **exactly**: reverting only the production hunks yields `3 failed | 53 passed`, precisely the 3 new tests, zero collateral regressions, failing on the intended assertions. "No new typecheck errors" reproduced.

**Environment note, not a PR issue:** `npm ci` exits 0 while printing `allow-scripts ... electron@41.3.0 (postinstall: node install.js)`. That is the hazard `CLAUDE.md` § "Desktop Electron runtime" describes, arriving by a new route — the postinstall is now blocked by npm *policy*, not by a flag. `verify:electron` caught it, so the guard works, but the contract's premise ("run `npm ci` without `--ignore-scripts`") is no longer sufficient on this npm version.

---

## Recommendation

**In this PR** (cheap): M1 — drop or re-apply `0369d90b`; L7 — extract the unrelated reformat; L8 — the description mentions neither `config.onboardingPreview.test.ts` (1 of 6 changed files) nor the discarded commit; optionally L1 and L5, one line each.

**Follow-up issues, in priority order:** **M2 first** — a hard-stop dialog with no escape, painted behind the embed, is worse than the case this PR fixes. Then **H1 + M3** together, since they share one root: embed visibility has two owners, no read-back path, and no timeout. A single owner deriving "any shell-level modal is mounted" subsumes H1, M2 and L4. Then M4, L9, L10.

<sub>🤖 Multi-agent review (5-agent fan-out, Opus 5) via Claude Code. Findings verified against real code at `7dae5b20`; suites actually executed; every line reference re-checked. No code was modified.</sub>

alfredolopez80 (commented):
Inline anchors for the findings that live **inside this diff**, so each one is unambiguous at its exact line.

The remaining findings — H1 (`App.tsx:475`/`:494`), M1 (the commit whose content a merge discarded), M2 (`App.tsx:1937`), M3 (`SandboxUiPage.tsx:509-520`), L9 and L10 — sit in code this PR does not touch, so GitHub cannot anchor them. They stay in the [review comment](https://github.com/evenfire-ai/evenfire/pull/429#issuecomment-5482825246), with every line reference re-verified against the code at `7dae5b20`.

Posted as a neutral comment: **nothing here blocks the merge.**

desktop-app/ui/src/pages/SandboxUiPage.tsx
```diff
@@ -184,8 +185,14 @@ export function SandboxUiPage({
...
   const lastActionRequestIdRef = useRef(0)
+  // The deep-link confirm/error dialogs render as renderer DOM at the App
+  // shell level; like the other shell overlays they must hide the native
+  // WebContentsView, which always paints above the DOM, to be visible at all.
   const shellOverlayOpen =
-    headerShellOverlayOpen || sidebarShellOverlayOpen || toastShellOverlayOpen
+    headerShellOverlayOpen ||
+    sidebarShellOverlayOpen ||
+    toastShellOverlayOpen ||
+    deepLinkOverlayOpen
```
alfredolopez80: **Architecture — the mechanism is right, the ownership model is opt-in.**

This 4-term OR is correct and monotone: an added disjunct can only hide the embed *more* often, never less, so the change is fail-closed by construction.

The limitation is that every shell dialog must *remember* to raise its own signal, and the enumeration is already incomplete. Three findings hang off this line, all pre-existing and none introduced here:

- `App.tsx:1937` renders `desktopUpdateRequiredDialog` three lines above the deep-link dialog and feeds nothing in. It is a hard stop — `App.tsx:183-235` exposes only `Download update`, no `onCancel`, no Dismiss — and `refreshDesktopReleaseStatus()` runs on every session-state application (`useAppController.ts:794`), not just at boot. Painted behind the embed it leaves the app unusable with no way out.
- `App.tsx:475` / `:494` write the same native state imperatively, bypassing this aggregation entirely. Blame `3559031b`.
- `SandboxUiPage.tsx:509-520` gates the hide on an untimed `capturePreview()`, so a hung embed can defer `setVisible(false)` indefinitely.

A single owner deriving *"any shell-level modal is mounted"* would subsume all three. Not a change for this PR.

desktop-app/ui/src/App.tsx
```diff
@@ -1881,6 +1881,9 @@ export function App() {
                                   }
                                   sidebarShellOverlayOpen={sidebarSettingsMenuOpen}
                                   toastShellOverlayOpen={vm.toasts.length > 0}
+                                  deepLinkOverlayOpen={Boolean(
+                                    pendingSandboxUiConfirmation || failedSandboxUiDeepLink
```
alfredolopez80: **L5 · Low — two sources of truth for one question.**

`Boolean(pendingSandboxUiConfirmation || failedSandboxUiDeepLink)` re-derives what `sandboxUiDeepLinkDialog` (`App.tsx:1379-1404`) already computes from the same two consts. Equivalent today — verified, they cannot desync (same render pass, no `useMemo`, no `React.memo` in this file).

The risk is drift: if that ternary ever gains a third branch, this line keeps returning the old answer and a dialog renders **without** raising the signal — reintroducing exactly the bug this PR fixes, silently.

For the record the `Boolean()` coercion is *required*, not stylistic: `a || b` is `PendingSandboxUiDeepLink | undefined | null`, not assignable to `boolean | undefined`. Binding to the dialog value drops the coercion too:

```suggestion
                                  deepLinkOverlayOpen={sandboxUiDeepLinkDialog !== null}
```

desktop-app/ui/src/__tests__/App.deepLinks.test.tsx
```diff
@@ -612,6 +613,81 @@ describe('App deep-link orchestration', () => {
...
+          ready: true,
+          phase: 'active',
+        },
+      ],
+    })
+    render(<App />)
+    await waitFor(() => expect(sandboxUiPageHarness.props).not.toBeNull())
+    // The native WebContentsView paints above renderer DOM, so the overlay
+    // signal must stay off until a deep-link dialog actually needs to show.
+    expect(sandboxUiPageHarness.props?.deepLinkOverlayOpen).toBeFalsy()
```
alfredolopez80: **L1 · Low — this assertion is documentation, not proof.**

`toBeFalsy()` does not distinguish *"the signal is off"* from *"the prop was never passed"*. **Reproduced**: in the RED run (production hunks reverted) this prop was literally `undefined` and this line passed — the failure came 8 lines later. Delete `deepLinkOverlayOpen` from `App.tsx` entirely and this assertion still goes green.

The test as a whole does catch it, via `:644` and `:649` which demand identity. But if a later edit drops those, prop-absence becomes undetectable.

```suggestion
    expect(sandboxUiPageHarness.props?.deepLinkOverlayOpen).toBe(false)
```

desktop-app/ui/src/__tests__/App.deepLinks.test.tsx
```diff
@@ -612,6 +613,81 @@ describe('App deep-link orchestration', () => {
...
+      })
+    }
+
+    await waitFor(() =>
+      expect(confirmDialogHarness.props?.title).toBe('App link could not be opened')
+    )
+    expect(sandboxUiPageHarness.props?.deepLinkOverlayOpen).toBe(true)
+
+    await act(async () => {
+      confirmDialogHarness.props?.onCancel()
```
alfredolopez80, line 697: **M4 · Medium — this test covers Dismiss; the Retry branch is unreached.**

`onCancel()` is the Dismiss path. `App.tsx:1401` binds `onConfirm={handleRetryFailedSandboxUiDeepLink}` and nothing in the suite clicks it — the only two `onConfirm()` calls (`:511`, `:557`) are both gated on `title === 'Open app link?'`, i.e. the *confirm* dialog.

That is the path the description discloses as a flicker. I reproduced it against the real `lib/sandboxUiDeepLinkState.ts` and the disclosure is **honest as to severity** — both predicates do go null, but the dialog is driven by the identical expression (`App.tsx:1379`) so it unmounts in the same commit. There is no frame with a dialog on screen and the signal false.

One nuance the wording understates: it is not a frame, it is the whole retry window. With the ladder at `sandboxUiDeepLinkState.ts:87-93` (1+2+4+8+15 s) the embed stays fully restored for roughly 30 s — a complete hide/restore cycle per attempt. *"restored for the duration of each retry attempt"* would be more accurate than *"brief flicker"*.

desktop-app/ui/src/pages/__tests__/SandboxUiPage.test.tsx
```diff
@@ -851,6 +851,45 @@ describe('SandboxUiPage', () => {
...
+    sandboxUi.setVisible.mockClear()
+
+    // The "Open app link?" / "App link could not be opened" dialogs live in the
+    // renderer DOM; without hiding the WebContentsView they render behind the
+    // embedded app and the user can never reach the confirm/dismiss buttons.
+    rerender(<SandboxUiPage deepLinkOverlayOpen />)
+
+    await waitFor(() => {
+      expect(screen.getByTestId('sandbox-ui-embed-preview')).toBeTruthy()
+      expect(sandboxUi.setVisible).toHaveBeenLastCalledWith(false)
```
alfredolopez80: **L2 · Low — `toHaveBeenLastCalledWith` cannot see an intermediate wrong call.**

The `mockClear()` at `:875` correctly scopes the phase, so nothing leaks in from earlier phases — that part is sound. But *within* the phase, `false → true → false` also satisfies "last call was false". A transient `setVisible(true)` racing in from the bounds/resize effect would flash the native view over the dialog and this test would stay green.

This file already uses the stronger idiom 40 lines up (`:843` `toHaveBeenCalledTimes(1)`, `:844` `not.toHaveBeenCalled()`), so it is an inconsistency rather than a missing capability.

desktop-app/ui/src/pages/__tests__/SandboxUiPage.test.tsx
```diff
@@ -851,6 +851,45 @@ describe('SandboxUiPage', () => {
...
+          title: "Andy's Sales CRM",
+          defaultPath: '/',
+          ready: true,
+          phase: 'active',
+          updatedAt: null,
+        },
+      ],
+    })
+    sandboxUi.open.mockResolvedValueOnce(undefined)
+    sandboxUi.capturePreview.mockResolvedValueOnce('data:image/png;base64,deeplink')
```
alfredolopez80, line 868: **L3 · Low — the distinctive payload is never asserted.**

`'data:image/png;base64,deeplink'` reads as though the test pins *which* preview is rendered, but nothing asserts it — the check below only verifies that `sandbox-ui-embed-preview` exists. The test passes with any preview value, so a wrong or stale preview being mounted would go unnoticed. Either assert the `src`, or use a neutral literal so the test does not imply a check it is not making.

desktop-app/ui/src/pages/__tests__/SandboxUiPage.test.tsx
```diff
@@ -851,6 +851,45 @@ describe('SandboxUiPage', () => {
...
+
+    fireEvent.click(await screen.findByRole('button', { name: "Open Andy's Sales CRM" }))
+    await screen.findByRole('button', { name: 'Back to apps' })
+    await waitFor(() => expect(sandboxUi.setVisible).toHaveBeenLastCalledWith(true))
+    sandboxUi.setVisible.mockClear()
+
+    // The "Open app link?" / "App link could not be opened" dialogs live in the
+    // renderer DOM; without hiding the WebContentsView they render behind the
+    // embedded app and the user can never reach the confirm/dismiss buttons.
+    rerender(<SandboxUiPage deepLinkOverlayOpen />)
```
alfredolopez80: **L4 · Low — no multi-overlay interaction is pinned anywhere.**

Every overlay assertion in this file passes exactly one flag: `:779` header, `:841` toast, `:880` deep-link, `:897` header. Nothing covers two open at once.

The uncovered regression is concrete: with deep-link **and** header both open, if a future refactor turns the OR into an assignment or an early return, closing one would un-hide the embed while the other overlay is still up. The current code is correct — I traced every transition — but nothing would catch that change.

desktop-app/ui/src/pages/SandboxUiPage.types.ts
```diff
@@ -26,6 +26,7 @@ export type SandboxUiPageProps = {
   headerShellOverlayOpen?: boolean
   sidebarShellOverlayOpen?: boolean
   toastShellOverlayOpen?: boolean
+  deepLinkOverlayOpen?: boolean
```
alfredolopez80: **L6 · Low — naming breaks the family.**

Its three siblings on `:26-28` are `headerShellOverlayOpen` / `sidebarShellOverlayOpen` / `toastShellOverlayOpen`. `deepLinkOverlayOpen` drops the `Shell` infix, so `grep ShellOverlayOpen` no longer returns the full set — which matters precisely because the failure mode described above is *forgetting one of them*. `deepLinkShellOverlayOpen` would keep the family greppable. Cosmetic; the `deepLink` camelCase itself is correct and matches the codebase.

desktop-app/src/__tests__/config.onboardingPreview.test.ts
```diff
@@ -132,10 +132,7 @@ describe('EVENFIRE_ONBOARDING_PREVIEW', () => {
     // activeProfileId into the preview's own directory, so the NEXT launch
     // came up configured and went to sign-in. The switch showed onboarding
     // once and then quietly stopped working.
-    const previewDir = path.join(
-      testUserDataDir,
-      'runtime-configs-onboarding-preview'
-    )
+    const previewDir = path.join(testUserDataDir, 'runtime-configs-onboarding-preview')
```
alfredolopez80: **L7 · Low — unrelated scope, and it arrived invisibly.**

The reformat is *legitimate*: `prettier --check` exits 0 on this version and 1 on dev's (`printWidth: 100`; the collapsed line is 87 chars). Dev is unformatted here from CodeQL autofix `7aa4a31b`.

The problem is where it lives. It was introduced by merge commit `7dae5b20` **itself** — an evil merge, differing from *both* parents (`git diff 7dae5b20^1 7dae5b20` shows the file as new; `^2` shows the reformat). It has no commit message of its own and does not appear in the commit list, so it is invisible to anyone reviewing history rather than the diff. Worth extracting as `style: reformat config.onboardingPreview.test.ts` against dev, or dropping.

zach88:
Thanks for the two rounds @alfredolopez80 — every finding that lives inside this PR's diff is now addressed. Rundown below, plus what we consciously left out and why.

## Resolved in this PR (one commit per finding, ID-tagged)

| Finding | What changed | Commit |
|---|---|---|
| **M1** — lost `swallow()` commit | Re-applied the helper for fire-and-forget probe calls onto dev's rewrite of the file (16 sites live again) | `79895ef82` |
| **M4** — Retry branch untested | New test exercises `onConfirm` → `handleRetryFailedSandboxUiDeepLink` and pins the signal dropping in lockstep with the dialog | `7ea7abcc8` |
| **L5** — two sources of truth | Signal now binds to `sandboxUiDeepLinkDialog !== null` instead of re-deriving `Boolean(a\|\|b)` — can't drift from the dialog it guards | `90bff125f` |
| **L6** — naming breaks the family | Renamed `deepLinkOverlayOpen` → `deepLinkShellOverlayOpen`, so `grep ShellOverlayOpen` returns the full set | `df2f7832c` |
| **L1** — `toBeFalsy` proves nothing | `.toBe(false)` / `.toBe(true)` — prop-absence is now detectable | `9c7a18e26` |
| **L2** — `toHaveBeenLastCalledWith` blind to intermediate call | Exact `setVisible` call counts in the deep-link phase | `03534adc0` |
| **L3** — distinctive payload not asserted | Assert the preview `src` (`data:image/png;base64,deeplink`) | `333a5f267` |
| **L4** — no multi-overlay test | New test: deep-link + header both open, one closes → embed stays hidden | `f7f0bdc39` |
| **L7** — unrelated reformat via evil merge | Dropped; `config.onboardingPreview.test.ts` now matches `origin/dev` | `85cf5cc37` |
| **L8 / R4** — description drift | PR body updated: `deepLinkShellOverlayOpen`, 58/58, and a "changed-file set" note covering `useGfsBrowserController.test.tsx` and `config.onboardingPreview.test.ts` | body edit |

Targeted suites: **58/58 green** (up from 56 — the two new tests are M4 and L4). All ten fixes verified alive at HEAD after the latest `origin/dev` merge — worth double-checking given M1 was exactly a fix a merge had swallowed.

The inline threads for the applied changes are marked resolved.

## Left out on purpose — pre-existing, out of scope (R0)

The rest — **H1** (`App.tsx:475`/`:494`, `GfsImagePreview` writing `setVisible` imperatively), **M2** (same root, the GFS-close re-show), **M3** (`SandboxUiPage.tsx:509-520`, untimed `capturePreview`), **the architecture note** (`SandboxUiPage.tsx:195`, single-owner model), **L9** (`sandboxUiDriver.ts:855`, focus into a hidden embed), **L10** (`SandboxUiPage.tsx:519-523`, swallowed `setVisible(false)`) — all reproduce on the merge-base **without** this diff. They're pre-existing defects in code this PR doesn't touch or aggravate, so by our diff-scope rule they're not findings of this PR — and you flagged them as follow-ups yourself.

Their real fix is the shared root you called out: embed visibility has two owners with no read-back path. That's a design change (a single owner deriving "any shell-level modal is mounted"), not something to bolt onto this fix. Landing it here would just re-open the same review surface. We'll track H1/M2/M3/L9/L10 as separate issues so the ownership rework gets its own spec instead of a patch.

Left the architecture thread (`SandboxUiPage.tsx:195`) open since it's the anchor for that follow-up, not something addressed here.

alfredolopez80:
## Review — PR #429 · `fix: show deeplink confirm dialog above an open sandbox app`

Third round, at head `be35e096` (previous rounds: [`3e70c222`](https://github.com/evenfire-ai/evenfire/pull/429#issuecomment-5375324567), [`7dae5b20`](https://github.com/evenfire-ai/evenfire/pull/429#issuecomment-5482825246)). Eight parallel agents — correctness · contract/architecture · security · test-quality/mutation · adjudication · merge-hygiene · gates · docs. Suites **executed**, not assumed.

**Recommendation: does not block merge.** Zero defects introduced. Every one of the ten fixes claimed in the [author's response](https://github.com/evenfire-ai/evenfire/pull/429#issuecomment-5488750154) is verified alive at the current head. There is one **new** finding this round — High, but **pre-existing and recommended for deferral to a separate issue**; it extends the ownership diagnosis rather than contradicting it, and nothing about it is this PR's to fix.

| Severity | Count | Introduced here |
|---|---|---|
| Blocker / Critical | 0 | 0 |
| High | 1 | 0 |
| Medium | 2 | 0 |
| Low | 5 | 2 (cosmetic: one comment, one body note) |

---

### Gates — hard numbers, nothing inferred

Node v24.18.0 (matches `engines: 24.x`). `node_modules` provisioned by APFS copy-on-write clone after verifying `package-lock.json` is byte-identical to the main checkout.

| Gate | Exit | executed/passed/failed | Verdict |
|---|---|---|---|
| PR-touched suites (`App.deepLinks` + `SandboxUiPage`) | 0 | **58 / 58 / 0** | PASS — matches the PR's claim exactly |
| `useGfsBrowserController.test.tsx` | 0 | **21 / 21 / 0** | PASS — zero unhandled-rejection output in the verbose log |
| typecheck, PR head | 2 | 14 pre-existing errors | none in files the PR adds logic to |
| typecheck, `origin/dev` baseline | 2 | **identical 14 errors** | byte-exact diff, only line-number drift → "no new errors" **confirmed** |
| lint (`npm run style-rules`) | 0 | — | PASS |
| `prettier --check` (6 real files) | 0 | — | PASS |
| `npm run verify:electron` | 0 | — | PASS, Electron 41.3.0 resolved |

### Mutation testing on the production hunk

RED/GREEN reproduced: reverting **only** the three production lines yields **5 failed / 53 passed**, and the five failures are exactly the five new tests. Restored → 58/58.

| Mutant | Result |
|---|---|
| `!== null` → `!= null` (`App.tsx:1884`) | **SURVIVES — equivalent, not a gap.** `sandboxUiDeepLinkDialog` is `JSX.Element \| null` (the ternary at `App.tsx:1379-1405` ends in `: null`), never `undefined` |
| → `pendingSandboxUiConfirmation !== null` (drops the failure dialog) | CAUGHT (2 tests) |
| remove the `deepLinkShellOverlayOpen` disjunct | CAUGHT (2 tests) |
| whole OR → `deepLinkShellOverlayOpen` alone | CAUGHT (3 **pre-existing** tests) |

Zero actionable surviving mutants. The two suites are complementary by design — `App.deepLinks` mocks `SandboxUiPage` and pins the *signal*; `SandboxUiPage.test` pins the *consumption*. No mutant survived both.

### Verified clean

- **Signal/dialog cannot desync** — both emitted in the same `App` render pass; no `React.memo`, `Suspense`, `useDeferredValue` or transitions in either file.
- **Hide/restore is race-free** — the `cancelled` flag in `SandboxUiPage.tsx:503-550` is checked at *every* await boundary. A late `capturePreview` cannot leave the embed hidden after the overlay closes; a stale `setVisible(true)` cannot land over a live dialog.
- **Retry ladder** — the disclosed flicker is real and is the correct behaviour of a purely derived signal. No latch: a successful retry cannot strand the embed hidden.
- **Security** — clickjacking on the trust decision is not exploitable: `capturePreview` is `webContents.capturePage()` in the **main** process (`sandboxUiDriver.ts:864`), a compositor API that never runs the embed's JS, so the occlusion window (~2-4 frames) is not attacker-controllable. `ConfirmDialog` focuses **Cancel** on mount with Escape→cancel and no Enter→confirm binding, so blind input dismisses rather than confirms. The `capturePreview` catch still proceeds to `setVisible(false)` — fail-closed. Zero `console.*` in the diff.
- **Merge** — `merge-tree` is conflict-free, and verified *not* a false clean: `UpdateConnectorCredentials/index.tsx` is 719 lines in dev, 719 in the computed merge tree, 454 in the PR's stale tip — the 3-way merge keeps dev's newer refactor rather than reverting it. Dev has not touched any of the six PR files since the merge base.
- **The `useGfsBrowserController.test.tsx` history claim checks out exactly**: `e0bd84dc`'s first parent is `0369d90b`, so that merge is indeed what discarded the original commit by taking dev's rewrite wholesale; `79895ef8` re-applies it correctly on dev's newer content (15 sites, not 13 — dev added two, all covered). The restoration reverts nothing of dev's.
- **`swallow()` and fail-loud** — correct, does not mask real failures. The 15 sites are `onClick` handlers of throwaway probe components; assertions run on spies and rendered state, which `.catch(() => {})` does not touch. It only prevents an already-discarded rejection from surfacing as an unhandled rejection that would fell an *unrelated* test later.

---

## New this round

### H-A · High · PRE-EXISTING · READ-ONLY · `desktop-app/src/pluginConsentGate.ts:290`, `:311`

**There is a third writer of embed visibility, and it lives in the main process.**

The previous rounds identified two owners of the native visibility bit. There are **three**, uncoordinated, last-writer-wins, no refcount:

1. The declarative effect at `SandboxUiPage.tsx:503-550` — the **only** one that consults the `*ShellOverlayOpen` signals.
2. The imperative `GfsImagePreview` pair at `App.tsx:475` / `:494`.
3. **`pluginConsentGate.ts:290` (hide) / `:311` (unconditional restore)**, via `pluginSdkRuntime.ts:70` — autonomous, and the renderer never observes it.

The consent gate hides the embed precisely so the plugin "can neither fake the prompt nor paint over it" (`App.tsx:431-435`). But `pluginConsentPrompt` raises **no overlay signal**. So whenever any `shellOverlayOpen` term flips `true → false` while the prompt is open — most commonly **a toast expiring** (`toastShellOverlayOpen={vm.toasts.length > 0}`, `App.tsx:1883`) — the effect runs its restore branch (`SandboxUiPage.tsx:539`) and re-composites the plugin's `WebContentsView` **above the consent modal**, re-enabling the overlay-the-prompt attack the gate exists to block.

The converse holds too: answering consent while a deep-link dialog is open hits the unconditional restore at `pluginConsentGate.ts:311` and puts the embed over the new dialog.

**Recommendation: defer this to a separate issue. Do not fix it in this PR.** It reproduces on the merge base with this diff reverted, it lives entirely in files this PR does not touch, and its trigger (`toastShellOverlayOpen`) predates the change. Bolting a fix onto a three-line bug fix would re-open the whole review surface for a design change that deserves its own spec — the same reasoning the author already applied to H1/M2/M3, and we agree with it.

What it does change is the **scope of the follow-up already planned**: the single-owner rework must be specified against **three** writers, with a refcounted acquire/release owned by the main process — not against two writers reconciled through renderer props. A renderer-side aggregation cannot see `pluginConsentGate` at all, so a design drawn from the previous two rounds would ship incomplete.

---

## Medium — both pre-existing, both already tracked

**M-A · `App.tsx:494`** — closing a `GfsImagePreview` while the deep-link dialog is open defeats the fix. Interleaving: app mounted → `gfs.open(image)` hides the embed imperatively (`:475`) → deep link arrives, `deepLinkShellOverlayOpen=true`, the declarative hide is a no-op → user closes the preview → `:494` forces `setVisible(true)` → **the embed re-composites over the live dialog**, and the effect does not re-run because its deps `[launch.kind, shellOverlayOpen]` did not change. Buttons unreachable. This is H1/M2 from round 2, confirmed alive at the current head. The class pre-exists with toasts; this PR raises the stakes from cosmetic to *decision dialog unreachable*.

**M-B · `App.tsx:183-240`, render gate `:1771-1778`** — `DesktopUpdateRequiredDialog` verified **genuinely non-dismissable** (only "Download update" plus an info toggle; backdrop is `role="presentation"` with no click handler and no `onClose`) and raises no overlay signal. `refreshDesktopReleaseStatus()` runs inside `loadSession` (`useAppController.ts:794`), i.e. on *every* session-state application, not only at boot. With an app mounted, `updateRequired` flipping mid-session leaves the backdrop blocking all renderer chrome while the native embed keeps painting **and stays interactive** above an invisible hard stop. The same gap applies to `environmentSetupConfirmationDialog` / `environmentSetupSuccessDialog` (`App.tsx:1728-1749`), which the Risks section does not mention — we could not confirm whether those are reachable with the embed mounted.

---

## Low

- **L-A · `SandboxUiPage.test.tsx:900-945`** — the new multi-overlay test covers less than its name suggests. The mutant that collapses the whole OR is caught by **three pre-existing tests**, not by this one: `deepLink` stays `true` through all its phases, so it never discriminates. The suite as a whole protects; the individual test is not the safety net it appears to be. No action needed — noting it so a future edit does not delete the older tests believing this one subsumes them.
- **L-B · PR body** — `gh pr diff` (the three-dot view) also shows `control-ui/components/RecipeEditor.tsx`, `control-ui/components/SecretsTable.tsx` and `control-ui/lib/codexSubscription.ts`. All three are byte-identical to `origin/dev` (two-dot diff = 0 lines) — pure merge-base artifacts, no-ops on merge. The "Notes on the changed-file set" section covers only two extra files, so a reviewer sees three unexplained entries every round. One line closes it.
- **L-C · `App.deepLinks.test.tsx:753-759`** — the comment over-promises. It claims there is no frame with a dialog mounted and the embed shown; the assertions are final-state after `waitFor`, not frame-by-frame. The lockstep does hold, but **by construction** (a single expression), not because this test proves it. Soften the comment, not the test.
- **L-D · `SandboxUiPage.test.tsx:~930`** — `await waitFor(() => expect(capturePreview).toHaveBeenCalledTimes(1))` passes immediately (it was already 1 from the hide phase), so it is a no-op settle; a microtask-deferred second hide cycle could slip past the following `not.toHaveBeenCalled()`. Mitigated by the exact `toEqual([[true]])` in the final phase. Not reproducible as a real failure.
- **L-E · history only** — a second evil merge exists: `e0bd84dc` silently reformatted two `vi.fn().mockResolvedValue(...)` chains in `mcp-host/src/contextMapperClient.test.ts`, content present in neither parent, with no commit message of its own — alongside the already-known one in `7dae5b20`. **Both are formatting-only and both have already been overwritten**: the diff against `origin/dev` at the PR tip is 0 lines for each. Zero residual risk to what lands; noting it because the pattern recurred.

---

## Verdict on the open discussion

- **The architecture thread at `SandboxUiPage.tsx:195`, deliberately left open as the follow-up anchor — correct to keep open.** H-A adds the third writer that was missing from the diagnosis, and it moves the follow-up from "renderer prop hygiene" to "main-process ownership".
- **Deferring H1/M2/M3/L9/L10 under the diff-scope rule — valid and well argued.** Verified: the PR's hunks (`@@1881`, `@@156`, `@@185`) do not touch any of the cited regions.
- **Re-verifying the ten fixes after the latest dev merge was the right instinct** — M1 was precisely a fix a merge had swallowed. All ten are alive at `be35e096`.

---

## Suggested split — what belongs where

**Nothing here blocks the merge.** To be unambiguous about what we are and are not asking of this PR:

| Item | Where it belongs | Why |
|---|---|---|
| The three production lines + tests | **Merge as-is** | Zero defects introduced; 6/6 gates green; mutation-resistant |
| **H-A** (`pluginConsentGate` third writer) | **New issue — security, separate spec** | Pre-existing, reproduces on the merge base, lives in untouched main-process files. Should be folded into the single-owner rework, not patched |
| **M-A**, **M-B** | **Existing follow-ups the author already committed to tracking** | Same root cause as H-A; same rework absorbs them |
| **L-B** (body note on the three no-op `control-ui` files) | Optional, one line in this PR | Saves a reviewer question every round |
| **L-C** (soften the over-promising test comment) | Optional, one line in this PR | The test is right; only the comment overstates it |
| **L-A**, **L-D**, **L-E** | No action | Recorded so a future edit does not undo the protection by mistake |

Concretely, we suggest **one issue** covering H-A + M-A + M-B under a single title along the lines of *"Single owner for sandbox embed visibility (three writers, no read-back)"*, with H-A called out as the security-relevant instance and the acceptance criterion being that every shell-level modal — including the plugin consent prompt, `DesktopUpdateRequiredDialog` and the `environmentSetup*Dialog` pair — is covered by one refcounted owner rather than per-dialog signals. Filing it separately keeps this PR's scope honest and gives the rework the spec it needs.

<sub>🤖 Multi-agent review (8 agents: 5× Fable 5 · 3× Sonnet) in isolated worktrees via Claude Code. Every finding verified against real code; suites executed with recorded exit codes. No code was modified — all three worktrees ended pristine at `be35e096`.</sub>

zach88:
Thanks for the third round @alfredolopez80 — the read-through on the mutation testing and the three-writer diagnosis is really useful. Rundown of what we did with it.

## Applied in this PR (the two cosmetics you flagged as introduced here)

Both are comment-only, no behavior change — commit `059f8e7c`:

- **L-C** (`App.deepLinks.test.tsx`) — softened the retry comment. It no longer claims a frame-by-frame guarantee; it now says the dialog value and the overlay signal move in lockstep *by construction* (a single derived expression) and that the assertions check the settled final state after `waitFor`, not each render frame.
- **L-A** (`SandboxUiPage.test.tsx`) — added the note you suggested above the multi-overlay test: it does **not** discriminate the whole-OR-collapse mutant on its own (`deepLinkShellOverlayOpen` stays true through every phase), so that mutant is caught by the single-overlay tests, and this one does not subsume them — don't delete them in a future edit.

- **L-B** — updated the PR body. The "Notes on the changed-file set" section now accounts for the three extra `control-ui` files in the three-dot view (`RecipeEditor.tsx`, `SecretsTable.tsx`, `codexSubscription.ts`) — all byte-identical to `origin/dev` (two-dot diff = 0 lines), pure merge-base artifacts.

Suites still green after the comment edits: `App.deepLinks` + `SandboxUiPage` = 58/58.

## Filed as a follow-up, not touched here

**H-A / M-A / M-B → #538.** You're right that the third writer moves the follow-up from "renderer prop hygiene" to "main-process ownership", so I rolled the three into one issue framed as a single structural class: the embed visibility bit has three uncoordinated writers (the declarative effect, the `GfsImagePreview` imperative pair, and the consent gate in the main process), last-writer-wins, no refcount. Filed as security-leaning per H-A. The proposed direction is a single owner in the main process with refcounted acquire/release — which per our PR-discipline rules needs a mini-spec (decision table up front) before implementing, specified against all three writers. Deferred here under the diff-scope rule as you validated.

L-D and L-E we left as-is (not reproducible / history-only, zero residual risk), matching your read.

Thanks again — this closes the round from our side.

alfredolopez80:
## Round 3 — closing verification at `059f8e7c`

Everything from the [third-round review](https://github.com/evenfire-ai/evenfire/pull/429#issuecomment-5494600536) closes. Re-verified at the new head rather than taken on trust.

| Item | Status | Evidence |
|---|---|---|
| **L-C** — soften the over-promising retry comment | ✅ Closed | No longer claims a frame-by-frame guarantee; states lockstep *by construction* (a single derived expression) and that the assertions check settled final state |
| **L-A** — note that the multi-overlay test does not subsume the single-overlay ones | ✅ Closed | Goes further than asked: *"Keep those tests: this one does not subsume them"* — that is exactly the regression the note needed to prevent |
| **L-B** — account for the three `control-ui` files | ✅ Closed | Body now explains them as merge-base artifacts with the right argument (two-dot diff = 0 lines, no-ops on merge) |
| **H-A / M-A / M-B** — defer to a separate issue | ✅ Closed | [#538](https://github.com/evenfire-ai/evenfire/issues/538), rolled up as one structural class with all three writers at `file:line`, framed security-leaning per H-A, with a stated acceptance criterion (single refcounted owner in the main process) and a mini-spec gate before implementation |
| **L-D / L-E** — left as-is | ✅ Agreed | Not reproducible / history-only, zero residual risk |

### Gates re-run at `059f8e7c` — executed, not assumed

| Gate | Exit | executed/passed/failed | Verdict |
|---|---|---|---|
| `be35e096..059f8e7c` is comment-only | — | 2 files, +11/−3 | **CONFIRMED** — every changed line sits inside a `//` block; no assertion, test name or production line touched |
| PR suites | 0 | **58 / 58 / 0** | PASS (33 + 25) |
| **RED re-check** — 3 production lines reverted | 1 | 53 passed / **5 failed** | **The five new tests still fail, and only those five** |
| Restore → GREEN | 0 | **58 / 58 / 0** | PASS |
| typecheck | 2 | 14 pre-existing errors | none inside the PR's hunks |
| `prettier --check` on both changed files | 0 | — | CLEAN |

The RED re-check is the one that mattered here. The new commit edits lines 752-758 and 897-905 — the same blocks the `waitFor`/`expect` calls live in — so a softened comment that had swallowed an assertion would have left the suite green while silently losing its detection power. Reverting the three production lines still fails exactly the five new tests, which proves the assertions still bite.

### Verdict

Nothing outstanding. The three-line fix is correct, monotone, mutation-resistant and fully covered; the pre-existing structural findings are correctly deferred with a real spec behind them in #538.

One operational note, unrelated to the code: `mergeStateStatus` is `BEHIND` (`mergeable: MERGEABLE`, and `merge-tree` is conflict-free and verified to preserve dev's newer work rather than revert it). If branch protection requires branches to be up to date, an update will be needed before the merge button — it does not affect the approval.

Approving. Thanks for the three clean rounds, @zach88.

<sub>🤖 Closing verification via Claude Code — gates executed with recorded exit codes in an isolated worktree, which was removed afterwards. No code was modified.</sub>

alfredolopez80 (approved):
Approved at `059f8e7c`.

All five items from the third-round review close, verified at the new head rather than taken on trust — details and gate numbers in the [closing comment](https://github.com/evenfire-ai/evenfire/pull/429#issuecomment-5495054782).

Summary: the new commit is confirmed comment-only; suites 58/58 green; the RED re-check still fails exactly the five new tests with the three production lines reverted, so the softened comments did not weaken any assertion; typecheck introduces no new errors; prettier clean. The pre-existing structural findings (H-A/M-A/M-B) are correctly deferred to #538 with a real acceptance criterion behind them.

Note: `mergeStateStatus` is BEHIND — an update may be required before merging if branch protection demands up-to-date branches.

PR #430, opened 2026-08-21
Feat/ux improvements 29

miguelo-999:
## Summary

This intentionally broad PR delivers the Control UI agent and credential UX improvements, while preserving the existing backend model and transitional Context behavior.

## Control UI UX changes

- Replaced the user-facing Context step in agent creation with an Add Connectors step.
- Automatically creates the required private Context for new agents.
- Added connector add/remove actions and three-dot menus.
- Renamed agent connector routes to `/agents/:name/connectors`.
- Added compatibility for legacy `/agents/:name/contexts` links.
- Added inline agent renaming from the Overview tab.
- Shows agent descriptions with 100-character truncation and full text on hover.
- Refined LLM Secret selection with custom provider-aware UI and provider icons.
- Improved light-mode provider icon rendering and related navigation labels.
- Moved model configuration actions to the bottom of the editing experience.

## Model configuration and credentials

- Model provider, current model, allowed models, and fallback policy remain configured per Agent.
- LLM Secrets remain the credential store containing provider credential slots.
- Editing an Agent’s LLM Secret opens the existing credential editor modal.
- Stored credential values remain write-only; operators can replace or add credential slots without exposing existing values.

## Fallback credential protection

The control-api now protects active fallback credential slots at the authoritative API boundary.

If any active Host references a Secret key through:

`llmPolicy.fallbacks[].credentialSlot`

that key cannot be removed through:

- Merge updates using `removeKeys`.
- Full Secret replacement that omits the referenced key.
- Whole-Secret deletion.

The API returns a conflict response before performing the mutation. Producer-backed regression tests cover these paths.

## Transitional shared-Context behavior

Existing shared Context behavior remains intentionally transitional and is out of scope for further ownership migration or reassignment in this PR. The agent-facing UI abstracts Contexts from users while maintaining compatibility with existing Context-backed data.

## Scope

This is intentionally a broad UX and platform-safety update spanning agent creation, agent detail views, connectors, model configuration, LLM Secrets, route compatibility, API protection, and regression coverage.


miguelo-999:
@claude pls provide feedback or green light if everything is good

miguelo-999:
@claude pls provide feedback or green light if everything is good

miguelo-999:
@claude pls check again

josue-tech-94:
@miguelo-999 

The PR size / number of changed files is **intentional**, so please **do not split the PR or reduce scope just to satisfy the D1 size recommendation**. Treat the size-related finding (`R1-M5`) as accepted by the owner.

Please address the following:

### `R1-H1` — Shared Contexts can be edited as if they were agent-private

`control-ui/app/hosts/[name]/page.tsx`

Two agents can reference the same Context, but editing connectors through one agent currently updates that shared Context in place, which can silently change the other agent.

Please:

* Define/enforce the shared-Context ownership behavior.
* Do not allow an agent-local edit to silently mutate a Context shared by another agent.
* Either prevent the edit and link to the shared resource, or use an explicit clone/private-resource flow as appropriate.
* Add a **producer-backed test with two hosts using the same Context** proving an edit for host A cannot silently change host B.

### `R1-H2` — Model tab removed the existing host model configuration flow

`control-ui/app/hosts/[name]/page.tsx`

The new Model tab shows model/credential information but removes the previous editable provider/model/allowed-models/fallback-policy flow.

Please:

* Restore the editable host model configuration flow, including the existing model-policy fields, **or**
* If removing that functionality is intentional, get an explicit product-owner decision and update the related UX contract/docs accordingly.
* Add coverage for real host updates, including save, cancel, error, and reload behavior.

### `R1-M1` — Secret-key deletion can leave a dangling fallback credential

`control-ui/app/hosts/[name]/page.tsx`
`control-ui/components/LlmSecretUpdateModal/index.tsx`

A fallback policy can reference a custom `credentialSlot`, and the inline Secret editor can currently delete that key without updating the policy.

Please:

* Prevent deletion of keys currently referenced by an active fallback policy, or explicitly reconcile/revalidate the policy as part of the operation.
* Add a producer-valid test using a fallback with a custom credential slot.

### `R1-M2` — Optional Context loading can take down the entire hosts page

`control-ui/app/hosts/page.tsx`

`getHosts` and the Context lookup are currently coupled. If the optional Context request fails, the hosts list disappears even when `getHosts` succeeded.

Please:

* Load hosts independently.
* Treat Context information used for enrichment/hover UI as optional.
* If Context loading fails, keep the hosts table and primary actions available and degrade only that enrichment.
* Add a regression test for `getHosts` success + `getContexts` failure.

### `R1-M3` — Existing `/agents/:name/contexts` links are broken

`control-ui/app/hosts/[name]/[tab]/page.tsx`

The rename from Contexts to Connectors causes existing `/agents/:name/contexts` deep links/bookmarks to return not found.

Please:

* Preserve compatibility with a redirect or alias from `contexts` to the new `connectors` experience.
* Add a direct-route regression test for the legacy URL.

### `R1-M4` — Shared-Context tests need to use the real producer contract

`HostDetailsPage.connectors.test.tsx`
`HostWizard.test.tsx`

Some new fixtures are hand-built around the new UI assumptions and therefore do not prove the shared-Context behavior against the real producer contract.

Please:

* Add producer-backed shared-Context coverage.
* Include the two-host/shared-Context case needed for `R1-H1`.
* Keep explicit previous-head reproduction evidence for regression tests so we know the tests actually detect the old behavior.

### Before requesting the next review

The branch was also reported as **28 commits behind `dev`** at the reviewed head. Once the findings above are addressed, please sync the branch with the current `dev`, resolve any semantic conflicts, and rerun the affected validation.

At minimum rerun:

* `cd control-ui && npm test`
* `cd control-ui && npx tsc --noEmit`
* `cd control-ui && npm run build`
* Relevant focused tests for the findings above
* Relevant Playwright flows when the required Control API/Kubernetes environment is available

Please also update the PR description to describe the actual final behavior and validation performed. The large scope itself is intentional and does **not** need to be justified by splitting the PR.

One process note: `control-ui/components/HostOverviewTab/index.tsx` is already at the F3 three-fix threshold. If any of these fixes require another **behavioral** change in that file, stop and define the required mini-spec/decision table rather than applying a fourth corrective patch there.

Once these are resolved and the branch is synced, please request another full review.

miguelo-999:
@josue-tech-94 

R1-H1: This is temporary this is the first step moving towards a UI/UX with no contexts, so this will be evolving. 
R1-H2: Fixed 
R1-M3: Fixed 
R1-M4: Fixed 

branch up to date with dev, all checks in green

josue-tech-94:
@miguelo-999  R2 is complete. The PR is still **blocked**, but there are **no new R2 findings and no regressions introduced by the R1 fix round**. The remaining work is all carried from R1.

The PR size/scope is intentional and already accepted, so **do not split the PR because of the D1 file-count recommendation**.

Please address the following before requesting R3:

### `R1-H1` — Shared Context editing still silently changes other agents

`control-ui/app/hosts/[name]/page.tsx`

This is still the main blocker.

If two Hosts reference the same Context, editing connectors from Host A still updates that shared Context directly, which silently changes Host B as well.

Please change the UX/behavior so this cannot look like an agent-local edit while mutating a shared resource.

Acceptable directions include:

* block agent-local connector editing when the Context is shared and direct the user to the shared Context, or
* use an explicit clone/private-Context flow before making agent-specific changes.

The important invariant is:

**Editing connectors for Host A must not silently change Host B.**

Add a producer-backed two-Host/shared-Context regression test proving that invariant.

### `R1-M4` — The current shared-Context test is false-green

`control-ui/components/__tests__/HostDetailsPage.connectors.test.tsx`

The current test actually asserts that Host B sees Host A's edit, and it also passes on the previously reviewed buggy head.

Please rewrite this coverage around the corrected contract from `R1-H1`.

The regression test must:

* use producer-backed/shared-Context data,
* include two Hosts consuming the same Context,
* assert the observable behavior rather than the implementation detail,
* and **fail on `8c9e129a7ad4e7252bf8dae5a3820c08cb32ed5b` for the intended assertion**.

### `R1-H2` — Model editor behavior is restored; finish the required coverage

The actual model-editing behavior is mostly restored now, so this finding is narrowed to missing regression coverage.

Please add focused tests for:

* Cancel restoring the loaded/server state.
* Failed save preserving the user's draft and showing the error.
* Reload/server refresh after save or cancel using authoritative server state.

The tests should exercise the real Host update payload and observable UI state, including preservation of unrelated Host fields and `resourceVersion`.

### `R1-M1` — Prevent deleting credentials that active fallbacks still reference

`control-ui/components/LlmSecretUpdateModal.tsx`
`control-ui/app/hosts/[name]/page.tsx`

The Secret editor still allows removing a key without knowing that an active:

`llmPolicy.fallbacks[].credentialSlot`

references it.

Please either:

* pass the active referenced credential slots into the modal and block deletion, or
* reconcile the Host policy as part of one explicit confirmed operation.

Do not leave a Host fallback pointing at a missing Secret key.

Add a producer-valid regression test with a custom fallback `credentialSlot`.

### `R1-M2` — Optional Context enrichment must not take down the Hosts page

`control-ui/app/hosts/page.tsx`

`getHosts()` and `getContexts()` are still coupled through the same `Promise.all`.

Please make Hosts the primary request and Context data optional enrichment.

Required behavior:

* `getHosts()` succeeds,
* `getContexts()` fails,
* Hosts table still renders,
* primary actions remain available,
* only Context count/hover enrichment degrades.

Add a regression test for that exact case.

### `R1-M5` — Update the PR description

The **size/split part is already accepted** and does not need further action.

The remaining issue is that the PR body is still blank/generic.

Once the behavior above is finalized, update the description to accurately cover:

* final agent/connector UX,
* shared-Context semantics,
* model configuration behavior,
* fallback credential integrity,
* `/agents/:name/contexts` → Connectors compatibility,
* the intentionally broad scope,
* and validation actually run.

Do not claim Playwright passed unless it was actually run successfully.

### Already fixed

`R1-M3` is closed: legacy `/agents/:name/contexts` links now redirect correctly to `/agents/:name/connectors`.

Please preserve that behavior.

### Important process note before editing

Both of these files are now at the **3 corrective-commit F3 threshold**:

* `control-ui/components/HostOverviewTab/index.tsx`
* `control-ui/app/hosts/[name]/page.tsx`

If resolving the remaining findings requires a **fourth behavioral corrective patch to either file**, stop and define the required mini-spec/decision table first rather than layering another patch onto it.

Also, `R1-M1`, `R1-M2`, `R1-M4`, and the remaining PR-body portion of `R1-M5` have now survived two review rounds. Please **fix them in this round rather than carrying them unchanged into R3**.

After the fixes, rerun at minimum:

* `cd control-ui && npm test`
* `cd control-ui && npx tsc --noEmit`
* `cd control-ui && npm run build`
* focused regression tests for every item above
* `git diff --check`

Run the relevant Playwright flows if a verified branch-owned authenticated runtime is available; otherwise document that gap.

Please keep the fixes attributable by finding, then request R3 once the carried items are resolved.

miguelo-999:
@josue-tech-94 R1-H1: I already marked this as out of scope in the previous comment 

Included:

- Producer-backed shared-Context coverage with two Hosts.
- Model editor cancel, failed-save, refresh, payload, and resourceVersion tests.
- Protection against deleting fallback-referenced credential slots.
- Optional Context enrichment for the Hosts page.

josue-tech-94:
@miguelo-999  Thanks for clarifying. I’m accepting the scope decision for `R1-H1` here.

### `R1-H1` / `R1-M4` — no further change required in this PR

The shared-Context behavior is intentionally transitional as part of the move toward a UX without Contexts, so **changing shared-Context ownership semantics is out of scope for PR #430**.

Please treat:

* `R1-H1` as **closed-in-writing / owner accepted as out of scope**.
* `R1-M4` as **closed-in-writing as a consequence of that scope decision**.

The producer-backed two-Host shared-Context coverage you added is sufficient as **characterization coverage of the current transitional behavior**. We are not requiring it to fail on the previous head for the non-mutation invariant, because that invariant is not being implemented in this PR.

So please **do not make another behavioral change to `control-ui/app/hosts/[name]/page.tsx` for R1-H1/R1-M4**.

### Still required: `R1-M1` — protect fallback credential slots across every edit path

This is the remaining behavioral issue.

The current Host-detail protection is useful, but the review found that an active fallback slot can still be removed through the general LLM Secrets flow / server `removeKeys` path.

Please make sure that if any active Host has something like:

`llmPolicy.fallbacks[].credentialSlot = "some-key"`

that key cannot be removed from the referenced LLM Secret through **any** supported editing path without safely reconciling the Host policy.

The preferred solution is enforcement at the authoritative server/API boundary so this cannot be bypassed by another UI or caller.

Please include producer-valid coverage for the general Secret/API path with an active Host fallback reference, not only the Host-detail modal.

If this requires a `control-api` change, please run its relevant focused tests plus the normal `control-api` test/build gates as well.

### Still required: `R1-M5` — update the PR description

The PR size/scope itself is already accepted and does **not** need to be split or justified further.

Please replace the current one-line description with the actual final state of the PR, including:

* the Control UI UX changes;
* the transitional shared-Context behavior being intentionally out of scope for further ownership changes in this PR;
* model configuration behavior;
* fallback credential protection;
* legacy `/agents/:name/contexts` compatibility;
* the intentionally broad scope;
* validation actually run.

Do not claim Playwright unless it was actually run successfully.

### No further action needed for

* `R1-H2` — closed.
* `R1-M2` — closed.
* `R1-M3` — closed and re-verified.
* `R1-H1` — accepted out of scope above.
* `R1-M4` — closed in writing with `R1-H1` above.

After `R1-M1` is fully covered across all edit paths and the PR description is updated, rerun the affected validation and request the next review.

miguelo-999:
@josue-tech-94 done

josue-tech-94 (approved):

PR #436, opened 2026-08-22
fix(auth): reconcile JWT consumers by runtime contract

alfredolopez80:
## Summary

- discover every Deployment binding to `mcp-host-config.CLERUM_AUTH_JWT_PUBLIC_KEY` from the pod template instead of the broad HCC owner label
- resolve the effective binding with the real referenced ConfigMap/Secret key inventory, including ordered `envFrom`, Kubernetes-valid prefixes, explicit overrides, optional refs, and direct key refs
- derive the real selector, container, effective environment name, and desired replicas for each consumer
- prove all Ready pods and bindings, restart only stale or unprovable Deployments once and sequentially, then prove every binding again
- keep zero-replica consumers non-blocking and write the applied SHA marker only after the ConfigMap, consumer inventory, runtime bindings, and source hash agree
- preserve remote authorization, the exact context allowlist, source-rotation detection, and the shared-profile bootstrap prohibition

## Root cause

The public integration in #416 selected every Deployment labelled `clerum.io/managed-by=host-context-controller` and assumed this runtime contract:

- selector: `app=<deployment>`
- container: `mcp-host`
- environment: `CLERUM_AUTH_JWT_PUBLIC_KEY`

That is valid for the usual first-party MCP host, but not for `wfc-21352205ce`. WFC is the Workspace Files Controller generated by `sharedFileSystemFactory.ts`; it consumes the same ConfigMap key as `workspace-files-controller/WSF_JWT_PUBLIC_KEY` and uses the shared-filesystem selector.

In [infra run 32566719029](https://github.com/keyper-labs/evenfire-infra/actions/runs/32566719029), the WFC rollout completed successfully and the post-rollout proof failed because it used the MCP-host contract. The earlier [run 32563372922](https://github.com/keyper-labs/evenfire-infra/actions/runs/32563372922) / [job 97008434076](https://github.com/keyper-labs/evenfire-infra/actions/runs/32563372922/job/97008434076) failed at the remote lock before [infra #58](https://github.com/keyper-labs/evenfire-infra/pull/58) added the explicit one-step authorization.

## Reconciliation behavior

ConfigMap drift and process drift are independent:

1. repair the ConfigMap when needed
2. discover candidate containers from the exact target ConfigMap/key reference
3. read only referenced ConfigMaps/Secrets and reduce them immediately to sanitized identity, resourceVersion, and key-presence metadata
4. reproduce kubelet precedence to derive every effective target binding, using Kubernetes `IsEnvVarName` semantics for `envFrom` prefixes
5. read and prove every current Ready pod binding
6. perform zero rollouts when every process already has the source key
7. restart only a stale or unprovable Deployment once
8. wait for its rollout and re-prove all of its bindings
9. re-enumerate the binding inventory, rehash the target ConfigMap, and recheck the source before attestation

There are no WFC, recipe, first-party, or third-party cases in production code. A Workflow Recipes host referencing `clerum-wrc-public-key.CLERUM_WRC_SIGNING_PUBLIC_KEY` is ignored because it does not bind the target key, even if its effective environment name is similar.

## Review closure corrections

Zach88's [final review](https://github.com/evenfire-ai/evenfire/pull/436#issuecomment-5381272203) listed four Medium and six Low optional findings. On current head `7a319b5dc57f356bd986ba31092d17cd48c4944f`:

- **M1 fixed:** effective `envFrom` names mirror Kubernetes `IsEnvVarName` (`[-._a-zA-Z][-._a-zA-Z0-9]*`, with the `.`/`..*` chdir-prefix exclusions), so valid dot/hyphen prefixes cannot disappear into an empty discovery set.
- **M2 fixed:** this body and the validation evidence refer to the current exact head, not the superseded `7e2d9d37` review head.
- **M3 fixed:** present ConfigMaps and Secrets whose required target key is absent fail closed, with both regression cases covered.
- **M4 fixed:** a deterministic property suite imports and executes the production resolver across finite domains for prefixes, precedence/shadowing, provenance, optional/required references, expansion ordering, and snapshot determinism; it is registered in the T2 contract lane without adding dependencies.
- **Cheap Lows fixed:** replicas-zero selector hardening, duplicate reference traversal removal, dead fake-kubectl handlers, and HCC desired-spec preservation assertions.

L3 (sanitized resolver input still traverses the helper process) and L4 (hand-authored fixture composition) remain non-blocking defense-in-depth/test-maintenance observations; neither changes the fail-closed runtime contract.

## Regression matrix

Covered by the cluster-free fixtures and focused resolver/probe checks:

- ConfigMap drift with already-converged MCP and WFC processes: patch and attest with zero rollouts
- only MCP stale: restart only MCP
- only WFC stale: restart only WFC and prove its real selector/container/environment
- desired replicas zero: skip proof and rollout
- no Ready WFC after rollout: fail closed with no marker
- multiple containers and bindings: prove all
- one stale binding: one Deployment restart and complete re-proof
- ordered ConfigMap/Secret `envFrom`, later source shadowing, optional absent object/key, dot/hyphen prefixes, exact self-expansion, and direct key refs
- option-like effective environment name: stale value cannot be falsely attested and causes exactly one rollout
- required missing object/key or ambiguous expansion: fail closed
- source rotation: fail closed
- unrelated HCC Deployment: ignore
- WRC ConfigMap/key: ignore
- exact remote allowlist and no remote shared-profile bootstrap

## PR #416 blast radius

The merge of #416 changed 68 files, but the failed remote deploy reaches the public change through the assembled auth sync script. Both failed runs completed image mirroring, migration, strategy repair, render, and apply before stopping at JWT sync. The last deploy on the parent public SHA `9e20e575` completed NetworkPolicy verification, GFS provisioning, drift audit, scoped restarts, the rollout gate, and seed.

The HCC Deployment change in #416 is documentation-only. The fail-red HCC rollback from [infra #52](https://github.com/keyper-labs/evenfire-infra/pull/52) remains protected by the infra assembly and is not changed here. The GFS production flow is not redesigned; its explicit authorization, ordering, rerun, and fail-closed contracts pass. The blast review found one #416 test-only assertion that omitted the safe `$ROOT` path form; this PR corrects that assertion without changing GFS production logic.

Merged #422 (`9e20e575`) is an ancestor of this PR's exact base `e3b2c2d9`. Its HCC/MCP timing contract is unchanged: this PR neither restarts HCC nor changes staleness or rollout budgets, and consumer rollouts remain sequential with the existing 180-second timeout.

## Validation at current head

No cluster was accessed. `origin/dev` and the merge-base are `e3b2c2d9fc92d7ce831b0046b83f59077e4f5a7a`; the feature branch is clean at `7a319b5dc57f356bd986ba31092d17cd48c4944f`.

- `python3 scripts/tests/test-discover-configmap-key-consumers-properties.py` — 6/6 passed
- `bash -n` for the changed shell files — passed
- Python AST/bytecode parsing for the resolver and property suite — passed
- ShellCheck `--severity=error` for the changed shell files — passed
- `git diff --check origin/dev...HEAD` — passed
- The fan-out harness run passed `test-minikube-sync-auth-key.sh` and `test-minikube-sync-auth-key-durable.sh` at this head. In the primary Codex shell, a local PreToolUse policy blocked aggregate wrapper execution before the fake `kubectl` ran; no hook bypass or production/test workaround was introduced.
- The targeted HCC Vitest is not locally runnable in this worktree because `vitest` is not installed; the HCC build/test/typecheck check is included in GitHub CI.
- All 33 GitHub checks passed in [CI run 32584183497](https://github.com/evenfire-ai/evenfire/actions/runs/32584183497) and [Analyze run 32584183563](https://github.com/evenfire-ai/evenfire/actions/runs/32584183563).

The single focused GPT-5.6 Sol closure review returned **PASS** with no residual blocker, High, or Medium finding for exact head `7a319b5dc57f356bd986ba31092d17cd48c4944f`. It verified Kubernetes prefix semantics, the WFC binding, required-key fail-closed behavior, property-suite registration, HCC annotation/spec preservation, and the selected cheap-Low fixes.

This closure is public-repo/code-contract evidence only. It does not certify the private `evenfire-infra` workflow or a live cluster run; those require the post-merge dispatch gate below.

## Release gates

- do not merge without explicit approval
- after merge, use a new `mirror-images.yml` dispatch from `evenfire-infra/main` with the new public SHA
- do not rerun either old failed run because GitHub preserves its original workflow and SHA
- the follow-up is complete only after the new `clerum-dev` apply finishes fully green

Refs evenfire-ai/evenfire#416.


alfredolopez80:
@claude Please perform a complete review of this PR at the exact current head `7e2d9d37b027bf7f3189ce6b4d10096c1491a206`, against base `e3b2c2d9fc92d7ce831b0046b83f59077e4f5a7a` (`evenfire:dev`). Review the full diff for correctness, security, fail-closed behavior, shell safety, Kubernetes-runtime fidelity, race conditions, and regression risk. Please keep findings within the defect introduced by public PR #416 and the reconciliation path it changed; do not request unrelated refactors.

## Incident and repository boundary

This is the public-repository fix for the `clerum-dev` deployment failure introduced after [evenfire-ai/evenfire#416](https://github.com/evenfire-ai/evenfire/pull/416). The implementation belongs exclusively in `evenfire-ai/evenfire`; `keyper-labs/evenfire-infra` must not be changed to hide or bypass the public-script defect.

Relevant evidence:

- [infra PR #52](https://github.com/keyper-labs/evenfire-infra/pull/52)
- [public PR #416](https://github.com/evenfire-ai/evenfire/pull/416)
- [infra PR #58](https://github.com/keyper-labs/evenfire-infra/pull/58)
- [first failed run 32563372922](https://github.com/keyper-labs/evenfire-infra/actions/runs/32563372922)
- [original failed job 97008434076](https://github.com/keyper-labs/evenfire-infra/actions/runs/32563372922/job/97008434076)
- [second run after infra #58, 32566719029](https://github.com/keyper-labs/evenfire-infra/actions/runs/32566719029)
- public integration SHA: `e3b2c2d9fc92d7ce831b0046b83f59077e4f5a7a`
- infra main after #58: `c55642ad371a64ad9b2ade23b6420ea7f7999578`

## Confirmed diagnosis

1. Infra #52 did **not** introduce JWT synchronization. Its fail-red HCC rollback behavior must remain intact.
2. Infra #58 is correct. It explicitly authorizes only the remote reconciliation step with `GFS_REMOTE_RECONCILE_AUTHORIZED=true` and an exact `ALLOWED_CONTEXTS` list. This PR must not weaken that lock or the prohibition on remote `--shared-profile-bootstrap`.
3. The defect was in public `scripts/minikube/sync-auth-key.sh`. The #416 implementation selected Deployments by `clerum.io/managed-by=host-context-controller` and assumed every selected workload had:
   - selector `app=<deployment>`;
   - container `mcp-host`;
   - environment variable `CLERUM_AUTH_JWT_PUBLIC_KEY`.
4. `wfc-21352205ce` is a real first-party key consumer, not a third-party workload to exclude. It is the Workspace Files Controller generated by `host-context-controller/src/k8s/sharedFileSystemFactory.ts`. Its pod template consumes `mcp-host-config.CLERUM_AUTH_JWT_PUBLIC_KEY` as `WSF_JWT_PUBLIC_KEY` in container `workspace-files-controller`, using a different selector.
5. In the second infra run, the WFC rollout completed successfully. The job failed because the post-rollout proof used the MCP-host contract against WFC.
6. Proving only one MCP host is insufficient during key rotation because environment variables are fixed when a process starts; active consumers can temporarily hold different keys.
7. Every active real consumer must be proved, but only stale or unprovable Deployments should roll out. The normal converged path must be read-only and cause zero rollouts.
8. ConfigMap drift and process drift are independent. Repairing the ConfigMap must not automatically force rollout when existing processes already contain the source key.

Workflow Recipes/WRC is not excluded by a hard-coded category. It is ignored because its real binding references `clerum-wrc-public-key.CLERUM_WRC_SIGNING_PUBLIC_KEY`, not the target `mcp-host-config.CLERUM_AUTH_JWT_PUBLIC_KEY`. Production logic must not hard-code `wfc-*`, recipes, or first/third-party categories.

## Required reconciliation contract

- Discover consumers from the actual pod-template reference to ConfigMap `mcp-host-config`, key `CLERUM_AUTH_JWT_PUBLIC_KEY`.
- Support `envFrom.configMapRef`, including prefixes and effective overrides, plus `env[].valueFrom.configMapKeyRef`.
- Correctly account for referenced ConfigMap/Secret key presence when later `envFrom`, explicit refs, or optional refs determine the effective variable provenance.
- Derive the real selector, container, effective environment variable, and desired replicas.
- Ignore Deployments that do not effectively bind the target key, regardless of owner label.
- After repairing the ConfigMap, always inspect current Ready pods.
- Do not restart a Deployment when every effective binding already contains the source key.
- If stale or not provable, restart the Deployment once, wait sequentially for rollout, and prove every binding again.
- Desired replicas `0`: do not block and do not restart.
- Fail closed if an active consumer cannot be proved after rollout.
- Write `clerum.io/auth-key-applied-sha256` only when source hash, target ConfigMap, consumer inventory, and all active runtime bindings agree.
- Preserve source-change detection, exact remote authorization/allowlist, the shared-profile-bootstrap prohibition, and existing GFS behavior.
- No new dependencies and no parallel rollouts.

## What this PR implements

The PR replaces label/category discovery with a pure resolver plus a shell-orchestrated, sanitized object inventory:

- `references` enumerates only ConfigMap/Secret objects referenced by candidate containers.
- The shell performs exact, explicit-context `get` calls for those objects only.
- `sanitize-object` immediately reduces each object to kind, namespace, name, presence, UID, resourceVersion, and key names. Secret values are never printed, persisted, or passed to the resolver.
- `resolve` simulates effective kubelet environment provenance and emits a normalized consumer snapshot plus binding rows.

The resolver models:

- ordered `envFrom`; a later source shadows an earlier variable only when the later object actually contains that key;
- ConfigMap and Secret sources;
- `prefix`;
- explicit `env` after `envFrom`;
- optional absent object/key preserving the previous `envFrom` value;
- required absent object/key failing closed;
- direct target `configMapKeyRef` bindings;
- exact `$(NAME)` expansion preserving provenance;
- ambiguous composite expansion depending on the target failing closed.

The normalized snapshot includes the candidate pod-template contract, Deployment UID, replicas, selector, effective bindings, and referenced-object UID/resourceVersion/key sets. It deliberately excludes Deployment resourceVersion and rollout annotations so a legitimate rollout does not create false inventory drift. The final phase re-discovers and compares the full snapshot, re-proves active bindings, rechecks the target ConfigMap and source hash, and only then permits the applied-hash marker.

The runtime Node probe now invokes `node -e ... -- "$env_name"`, preventing option-like effective environment names such as `--version` from bypassing the script or creating false results. Tests execute the captured remote Node argv rather than merely comparing fixture bytes.

The two High findings corrected in the final head were:

1. PodTemplate-only precedence could not distinguish whether a later or optional referenced object really contained the overriding key.
2. The effective environment name reached `node -e` without Node's option terminator.

Two cheap Low findings were also corrected:

- valid empty values in `matchExpressions` are accepted;
- the T2 runbook now describes all active effective Deployment bindings, including WFC, rather than only managed MCP hosts.

## Scope and #422 compatibility

Merged public PR #422 commit `9e20e5754ac541fed3bb65571f2dd189223ca6ef` is an ancestor of the exact PR base. This PR does not change HCC startup/readiness behavior, MCP staleness configuration, or rollout budgets. Consumer rollouts remain sequential with the existing 180-second timeout. Please verify that no changed path accidentally undermines #422's timing protection.

The production changes are limited to auth-key consumer discovery/reconciliation. There are no GFS production changes, no HCC changes, no infra changes, no dependency additions, and no rollout parallelization. The small `test-provision-gfs-runtime.sh` diff fixes a #416 test-only path assertion; it must not alter GFS runtime behavior.

## Regression and validation evidence

The fixture matrix covers:

- ConfigMap drift while MCP and WFC processes are already converged: target repair and marker with zero rollouts;
- only MCP stale and only WFC stale;
- the real WFC selector/container/`WSF_JWT_PUBLIC_KEY` contract;
- multiple containers and multiple bindings;
- one stale binding causing exactly one Deployment restart and complete re-proof;
- no Ready pods after rollout failing closed with no marker;
- desired replicas zero;
- ordered ConfigMap/Secret `envFrom`, shadow/no-shadow, prefixes, optional absent object/key, direct refs, and exact self-expansion;
- required missing refs and ambiguous composite expansion failing closed;
- option-like environment names using the exact Node argv;
- source rotation and final inventory races;
- WRC and unrelated HCC-owned Deployments ignored by binding contract;
- exact remote allowlist, explicit remote authorization, and rejection of remote shared-profile bootstrap;
- existing GFS contracts.

All GitHub checks for exact head `7e2d9d37` are green, including [CI run 32577495080](https://github.com/evenfire-ai/evenfire/actions/runs/32577495080) and [CodeQL run 32577495078](https://github.com/evenfire-ai/evenfire/actions/runs/32577495078). Static validation also passed: `bash -n`, ShellCheck with the documented baseline exclusions, Python AST parsing, `git diff --check`, the public boundary target, a pure ConfigMap/Secret precedence matrix, and real Node option parsing.

Evidence limitation: the full fake-runtime shell wrappers passed on the initial PR head `e13168f1`. On corrected head `7e2d9d37`, the local Codex PreToolUse policy blocked fresh aggregate-wrapper execution before their fake `kubectl` binaries started. The hook was not bypassed and the implementation was not changed to satisfy it. GitHub's shell job validates syntax but does not execute the full fake-runtime matrix, so do not treat green CI alone as proof of those behavioral cases; please inspect the final tests and implementation together.

## Review questions

Please review the entire diff and specifically try to find:

1. any path that can write the marker while an active binding is stale, absent, unprovable, or discovered from an unstable inventory;
2. any false binding that causes unnecessary mass rollout, or any missed real consumer such as WFC;
3. mismatches with kubelet precedence for ordered `envFrom`, prefixes, overrides, optional refs, explicit refs, or exact expansion;
4. unsafe Secret handling, excessive object reads, missing explicit-context use, or shell/argv injection issues;
5. races across source rotation, target ConfigMap repair, Deployment rollout, inventory re-discovery, and final JSON Patch/CAS;
6. selectors that can target the wrong pods, zero-replica mistakes, duplicate/parallel rollout, or more than one rollout per Deployment;
7. regressions to the exact remote lock, `ALLOWED_CONTEXTS`, shared-profile prohibition, HCC fail-red behavior, GFS flow, or #422 timing;
8. missing or misleading regression coverage relevant to the final implementation.

Report every actionable finding with severity and precise file/line evidence. Distinguish merge blockers from cheap non-blocking corrections. If no actionable findings remain, explicitly state that the PR is ready for approval. Do not merge or deploy; those require separate explicit authorization.

alfredolopez80:
Adversarial closure for a2ac25139464cb6b4e0be634ac57957b14f22e93: the consolidated review was revalidated against the current diff with GPT-5.6 Sol Max. Fixed findings: (1) fake kubectl exec now propagates the probe status via exec, (2) the escaped-expansion regression is resolved independently instead of being asserted in the effective fixture, (3) durable assertions are fail-loud and helpers are defined before first use, (4) kubelet $$ escaping and invalid envFrom effective names are handled, and (5) HCC preserves live pod-template restart annotations so a stale WFC restart is not followed by a second Recreate rollout. Added a focused SharedFileSystemReconciler regression for the HCC/WFC interaction. GitHub reports 30/30 checks green on this SHA, including host-context-controller build/test/typecheck, shell syntax, leak guards, manifests, and CodeQL. No evenfire-infra changes, merge, old-run rerun, or clerum-dev dispatch was performed. The local wrapper execution remains blocked before start by the repository command-safety hook; no product code was changed for that hook.

zach88:
Took a deep pass through this one, both areas: the new Python resolver + shell orchestration, and the HCC reconciler change. Overall this is solid work. The kubelet-precedence resolver holds up end-to-end, the `-- "$env_name"` node guard genuinely neutralizes option-like names, the one-rollout / skip-replicas=0 / fail-closed-attestation invariants all hold, and the HCC `preserveDeploymentAnnotations` wiring is correct (it reuses the existing helper and its new test goes red against the parent). Nothing here blocks a merge. The two things I'd actually want you to look at are a latent fail-open in the resolver and a body/head mismatch.

## Verdict

🟡 **Approve with nits**
- **Optional / non-blocking: 10** (4 Medium, 6 Low)
- **Pre-existing (out of scope, not introduced here): 2** (2 Low)

## Optional (non-blocking)

The PR can merge without these. That said, I'd genuinely prioritize the first two before merge given the fail-open direction and the stale attestation claim.

1. **[R1-M1] `C_IDENTIFIER` skip is stricter than kubelet → fail-open false negative** (`scripts/minikube/discover-configmap-key-consumers.py:317`). The `a2ac2513` commit added `if not C_IDENTIFIER.fullmatch(effective_name): continue` with a comment saying kubelet skips non-C-identifier envFrom keys. Kubelet actually validates envFrom-derived names with `IsEnvVarName` (`[-._a-zA-Z][-._a-zA-Z0-9]*`), not C-identifier (`[A-Za-z_][A-Za-z0-9_]*`). Names with `.`/`-` do materialize as pod env vars but this code drops them.
   - Repro: a Deployment with `envFrom: [{prefix: "AUTH.", configMapRef: {name: mcp-host-config}}]` and the CM holding `CLERUM_AUTH_JWT_PUBLIC_KEY` gives effective name `AUTH.CLERUM_AUTH_JWT_PUBLIC_KEY`. Kubelet sets it; the script emits zero TARGET records → that consumer is skipped → never restarted on rotation → keeps validating with the old key. That's the fail-open the contract is meant to prevent.
   - Narrow trigger (target key is a fixed C-identifier; only bites when an `envFrom.prefix` contains `.`/`-`), so it's latent, not reachable in today's mcp-host/WFC layouts.
   - Suggested fix: swap `C_IDENTIFIER` for `^[-._a-zA-Z][-._a-zA-Z0-9]*$` (plus reject bare `.`/`..`) and fix the comment. This is a hypothesis, worth confirming against the kubelet version on the dev cluster before applying.

2. **[R1-M2] Body's Validation/Closure-review certify a stale head.** The description cites `7e2d9d37` three times as the "corrected head" and says the closure review returned PASS with no actionable findings for exact head `7e2d9d37`. But the head is `a2ac2513` (58 min later), which added production logic not covered by that review or the validation matrix: the `C_IDENTIFIER` skip + the `$$` escape in the resolver, and `preserveDeploymentAnnotations` in the HCC reconciler. That new skip line is exactly what item 1 flags, so the "no findings" closure demonstrably doesn't cover what's about to merge. Worth re-running validation/closure on `a2ac2513` and syncing the body, or spelling out which head each claim covers.

3. **[R1-M3] Fail-closed coverage gap: required key absent from a present object isn't tested** (`scripts/tests/test-minikube-sync-auth-key-durable.sh:117-118,206-209`). The "required missing" matrix only covers the absent-object case, which short-circuits at the `inventory_object` guard before reaching the `raw_key not in obj["keys"]` + required → `raise` branch in the resolver (`discover-configmap-key-consumers.py:365-371`). Confirmed by mutation: turning that `raise` into `continue` leaves the suite green. The code is correct; the test that would guard it is missing. Add a fixture with a required (non-optional) `configMapKeyRef` to a present object whose `keys` lacks the target, asserting it fails.

4. **[R1-M4] No property-based tests on the pure precedence resolver** (`discover-configmap-key-consumers.py`, `effective_bindings`/`resolve_records`). It's a pure precedence/order function (envFrom ordering, shadowing, expansion, dedup, total order) with example tests only; the gap in item 3 is exactly what a fuzz would have caught. `fast-check`/hypothesis over total order, no dup ids, nothing representable lost, sort/dedup idempotence.

5. **[R1-L1] Empty selector emitted when `replicas == 0`** (`discover-configmap-key-consumers.py:492`). A Deployment scaled to 0 that binds the target emits records with an empty `selector`. The shell skips it before the selector is used (verified), so it's not wrong today, but not emitting records (or flagging no-pods) at `replicas == 0` would harden the output contract.

6. **[R1-L2] Duplicated container iteration** (`discover-configmap-key-consumers.py:453-486` vs `176-198`). `resolve_records` reimplements the `items→deployment→pod_spec→containers` + candidate filter that already lives in `candidate_containers`. Consistent today; future drift risk if the criterion changes in one spot.

7. **[R1-L3] Full Secret JSON transits the resolver's stdin** (`scripts/minikube/sync-auth-key.sh:302-307`). The sanitize step pipes `kubectl get secret … -o json` (base64 values included) into the resolver. Not a leak — `sanitize_object` provably emits only metadata (verified against a real Secret, 0 value hits) — but it moves Secret material through a subprocess unnecessarily, against the docstring's "never emits or persists values" at the input boundary. Tightening the fetch to metadata/`.data` keys only is purely defensive.

8. **[R1-L4] Hand-written `DISCOVERY_INVENTORY` fixture** (`test-minikube-sync-auth-key-durable.sh:44-52`). Represents `sanitize-object` output. The shape is faithful and the `sanitize-object → resolve` composition is exercised e2e in the main flow, so risk is mitigated, but the resolver unit tests depend entirely on the hand fixture; deriving it by calling `sanitize-object` on temp objects would close the T1 gap.

9. **[R1-L5] Dead handlers in the kubectl fake** (`test-minikube-sync-auth-key-durable.sh:511-521,557-562`). `get deployment -l …` and `get deployment gfsc-reader` aren't exercised by the new script (it enumerates with `get deployments -o json` / `get pods -l`; line 540 catches `gfsc-reader` first). Cosmetic.

10. **[R1-L6] Test doesn't assert the desired `spec.template` survives the merge** (`host-context-controller/src/__tests__/sharedFileSystemReconciler.test.ts:335`). It checks annotations are preserved but not that the desired template's labels/containers/initContainer survive. The code does preserve them; a future helper refactor that broke it wouldn't be caught here. Optional.

## Pre-existing (out of scope)

Not introduced by this PR — both reproduce at the merge-base, non-blocking. Listing them only because they sit in a touched file; ping me if you want these opened as issues.

- **Low ×2** — `SC2034` (unused var) at `scripts/minikube/sync-auth-key.sh:121` (`T2_WORKTREE_ID`, in the untouched lease block) and `:341` (`ignored_b64`, in the untouched `assert_source_unchanged`). Both benign.

## What checked out clean

- **Security: SAFE.** No secret leakage (`sanitize_object` emits metadata only — reproduced with a real Secret, 0 hits); injection hardened with `--` on the node probe (reproduced: `--eval`/`-e` aren't parsed as flags); restart set is namespace-pinned to `mcp-host` and rejects out-of-namespace objects (reproduced); the WRC signing key is correctly ignored; attestation is fail-closed (marker only after proof + contract-stability + hash-match).
- **`$$` escape** (reproduced): `value.replace("$$","\x00")` + `findall` is faithful to kubelet; never fabricates or hides a `$(`.
- **Precedence / shadowing / exact `$(VAR)` expansion / direct keyRef / optional-vs-required**: faithful to kubelet.
- **HCC TS fix**: `preserveDeploymentAnnotations` is a pre-existing pattern (`utils.ts:121`, already used by hostReconciler/llmHookReconciler/reconciler). The SFS authors no annotations of its own, so the plain variant (no `Except`) is the right call — nothing to invert. New test goes red against the parent; 38 tests green.
- **shell↔python↔node contract**: field counts and 0x1f delimiters match exactly across layers; herestrings (not pipes) make `return 1` abort correctly; no dangling refs to removed vars.
- **Suites**: `test-minikube-sync-auth-key.sh`, `test-minikube-sync-auth-key-durable.sh`, `test-provision-gfs-runtime.sh` → PASS. `py_compile` OK. `shellcheck -S warning` → only the 2 pre-existing SC2034. Runbook (`minikube-t2-runbook.md:166-172`) is accurate against this head.

zach88 (approved):

PR #440, opened 2026-08-24
Fix desktop chat tab selection contrast

josue-tech-94:
## Summary

Active and inactive chat tabs were visually ambiguous, and the selected tab could appear disconnected from the chat surface.

Inactive tabs now have no bottom border and use 25% of the existing border token, keeping inactive tabs visually quieter in both light and dark themes. The selected tab uses the same chat-surface border token for its top and side borders, omits its lower border, and paints an opaque selected surface over the panel token.

All chat tabs now use a consistent browser-style width, so short titles leave empty tab space and long titles truncate inside the fixed tab body.

The chat surface no longer paints its own top border. Instead, the tab list paints one full-width 1px top boundary for the chat surface, and the selected tab stacks above that boundary so the line remains visible to the left and right but is covered under the selected tab. This keeps the selected tab and content visually connected without a doubled line or gap.

The native horizontal overflow control remains intentionally above the tabs so a narrow window does not insert a scrollbar between the selected tab and its content surface. The tab overflow geometry is covered by tests; the exact scrollbar-above-tabs placement is a manual visual verification because browser scrollbar rendering is not asserted reliably by the automated harness.

## Validation

Final review-fix round validation run on 2026-08-25:

- `cd desktop-app && node --version` (`v24.3.0`)
- `cd desktop-app && npm run verify:electron`
- `cd desktop-app && npm test -- ui/src/components/ChatTabs/ChatTabs.test.tsx ui/src/components/ChatViewWorkspace/ChatViewWorkspace.test.tsx` (2 files, 8 tests)
- `cd desktop-app && npm run test:native:chat-workspace-geometry`
- `cd desktop-app && npm test` (193 files, 1,924 tests)
- `cd desktop-app && npm run build`
- `npm run style-rules -- --strict`
- `git diff --check`

Skipped: packaged desktop UI/E2E was not run in this review-fix round because the signed-in backend/runtime prerequisites were not available locally.

## UI

Before:
image
image

After:
image
image


josue-tech-94:
@claude Please do a full review of this PR against `dev`, focusing on the complete current diff rather than only the intent described in the PR body.

Please review it as a merge-readiness pass and **do not make code changes**.

Pay particular attention to:

* Correctness and regressions in the desktop chat-tab behavior.
* Whether the selected tab is clearly distinguished from inactive tabs in both light and dark themes.
* Whether inactive tabs truly have no bottom border and use the intended reduced-opacity border treatment without introducing theme inconsistencies.
* Whether the active tab correctly masks the content surface's top border only beneath the selected tab, without gaps, doubled borders, or seams.
* The `rotateX(180deg)` approach used to place the horizontal scrollbar above the tabs: check carefully for unintended effects on rendering, pointer/hover behavior, focus outlines, text/icon orientation, scrolling direction, accessibility, layout geometry, or browser/Electron edge cases.
* Tab overflow and horizontal scrolling at narrow widths.
* Hover, focus-visible, close-button, and tab-switching behavior.
* Whether the tests actually validate observable UI behavior rather than merely certifying the implementation.
* Whether the native geometry test's fixture/reference for the 25% border treatment is sufficiently independent of the production CSS to catch a regression instead of duplicating the same expression on both sides of the assertion.
* Any missing edge cases involving one tab, many tabs, changing the active tab, resizing, and theme switching.
* Reuse/simplification opportunities only where they materially reduce risk in this changed surface.
* Any security implications introduced by the diff, if any.

Please distinguish findings introduced by this PR from pre-existing issues. Do not block this PR for unrelated existing code.

For each actionable finding, please include:

* severity (`Blocker`, `High`, `Medium`, or `Low`);
* exact `file:line`;
* what is wrong;
* a concrete failure scenario;
* user or maintenance impact;
* the required fix;
* the required regression/validation test;
* confidence (`reproduced`, `code-verified`, `inferred`, or `needs confirmation`).

Also verify the PR description against the implementation and call out any claim that is not actually supported by the code/tests.

If you find no merge-blocking problems, please say that plainly rather than manufacturing findings, and summarize the areas you verified.

zach88:
Took a deep pass through this one, tracing the actual compositing (z-index paint order, the `--panel` opacity layer, the double `rotateX` flip) against the real tokens rather than just the intent. The mechanism is sound: no correctness bugs, no security concerns, and the `paintedSeam` pixel test is genuinely validating the observable result in both themes. Everything open is test-quality or body/code sync, nothing structural.

**🟡 Approve with nits**
- Optional / non-blocking: 5

None. Nothing blocks the merge.

## Optional (non-blocking)

1. **[R1-M1] Native fixture hand-writes `Button`'s class output** (`desktop-app/test/native/chat-workspace-geometry.fixture.cjs:7-10`). The tab markup hardcodes the classes `Button` emits (`ui-button--ghost ui-button--neutral ui-button--sm ui-button--align-start`, and `--xs --align-center` for close). I checked these class-by-class against `components/Common/Button/index.tsx` and they match exactly today, so nothing is broken right now. But the fixture is decoupled from the producer: if `Button` ever renames its class scheme without touching `ChatTabs`, the native harness keeps measuring geometry over stale markup and stays green with no warning. Worth deriving the markup from a real `ChatTabs` render (snapshot its `outerHTML`), or a small unit test asserting the class set against `Button`'s output.

2. **[R1-M2] Inactive-border reference is circular** (`desktop-app/test/native/chat-workspace-geometry.fixture.cjs:173-177`, asserted in `assertTabTreatment`). The `[data-inactive-tab-border-reference]` node hardcodes `color-mix(in srgb, var(--border-soft) 25%, transparent)`, which is a literal copy of the production `.chat-view-tab:not(.is-active)` rule, then asserts equality against it. That only proves the CSS equals a hand-copied duplicate of itself: if the 25% design value were wrong, or if both sides change in lockstep, the assert still passes. Stronger version: derive the reference from base `--border-soft` independently and assert the relationship (alpha ~25% of the normal tab border) instead of equality with a copy.

3. **[R1-L1] PR description describes the old mechanism.** The body says "A one-pixel inherited-background mask covers the content border beneath the selected tab." That's the removed approach (the old `.is-active::after { background: inherit }` plus the surface top border). The final design dropped both: the surface has no top border anymore, and masking is done by the active tab's opaque gradient over a full-width `list::after` seam. Worth updating the paragraph so it matches what shipped.

4. **[R1-L2] "Scrollbar above the tabs" invariant has no test** (body claim). That's the whole reason for the double `rotateX`, but nothing asserts the scrollbar position, so it rests on manual verification. It's hard to test reliably, so at minimum a one-line CSS comment on the `rotateX` trick would help the next reader, and a note in the body that it's manually verified.

5. **[R1-L3] `paintedSeam` never asserts tab-fill == surface-fill** (`desktop-app/test/native/chat-workspace-geometry.fixture.cjs`, `paintedSeam`). It checks the seam is masked (`selectedSeam ~ selectedFill`) but not that `selectedFill ~ surfaceFill`. They match today only because both composite `--surface-strong` over `--panel`; if a future change gives `.chat-view-workspace` or an intermediate a non-transparent background, a visible tab/surface color step would appear with no failing test. A `colorDistance(selectedFill, surfaceFill)` check sampling `surface.top + 3` at the active tab's center-X, in both themes, would close that gap.

One process note, not a finding: `styles.css` and the fixture each carry 5 `fix:` commits in this branch. They're all iterations of the same seam change rather than independent fixes, and the final state is coherent, so no concern about the result. But it'd be worth squashing that history before merge so the log reads as one change.

## What checked out clean

- **Security:** nothing. Renderer-only change, no auth/secret/JWT boundary, no `dangerouslySetInnerHTML`, `tab.title` is React-escaped, the fixture only interpolates literals, and `capturePage().toBitmap()` indices are clamped (no OOB). `webPreferences`/preload/`contextIsolation` untouched.
- **Double `rotateX(180deg)`:** hit-testing, visual order, badge, hover, and horizontal scroll all compose to identity for the leaf content. No functional defect.
- **Seam geometry:** paint order is correct, the active tab covers the 1px row even with `margin-bottom:-1px` removed, no gap or doubled line. The new `Math.abs(surfaceTop - activeBottom) <= 1` is stricter than the old `<= +1`.
- **`--panel` layer:** correct and deliberate, gives full opacity and an exact color match with the surface.
- **`paintedSeam` thresholds:** comfortably loose, not flaky (channel distance ~46-56 vs a floor of 5), and BGRA to RGBA conversion is right.
- **Regression coverage (T3):** the new asserts would fail against the merge-base (old CSS had `margin-bottom:-1px`, the `::after` on the active tab, surface `border-width:1px`). No tests that would pass against the previous head while claiming to cover a fix.
- **Suites:** `vitest ChatTabs.test.tsx ChatViewWorkspace.test.tsx` is 8/8 green. The native `test:native:chat-workspace-geometry` needs a real display so it was reviewed statically, not run here.

josue-tech-94:
@claude please review again and check Zach's comment also

zach88:
Took another pass through this one now that the fixed-width tab change and the dev merge landed. The chat-tabs side is solid: the 220px fixed width doesn't break the seam geometry or the horizontal overflow, and the native geometry harness actually ran and passed this round. Nothing here blocks the merge. The two things worth a look are a scope leak that snuck into the width commit and a test assert that doesn't quite bite, plus a few test-quality items still carried from round 1.

## Verdict

🟡 **Approve with nits**

## Optional (non-blocking)

The PR can merge without these. My one real ask is the profile-ui leak, since it's cheap and leaves a false claim in history; the rest is test quality.

1. **[R2-M1] profile-ui scope leak in the width commit.** `b581f58e` ("keep chat tabs equal width", renderer-only CSS) also carries a `profile-ui` version bump `0.1.123 → 0.1.124` plus lockfile churn (`profile-ui/package.json:3` + `profile-ui/package-lock.json`). Reproduced: dev and the merge-base are both on `0.1.123`, and `git log <base>..HEAD -- profile-ui/package.json` shows only `b581f58e`. It pulls a second service into an otherwise single-service PR, and the commit body says *"Existing unrelated profile-ui/package-lock.json change remains untouched"* which isn't true, the change is in the commit. No security angle (the lockfile adds no runtime deps, just a `fast-check` devDep and peer-flag reshuffles). Suggest pulling `profile-ui/*` out into its own PR.

2. **[R2-L1] `width: 220px` assert doesn't fail against the old head.** `desktop-app/ui/src/components/ChatTabs/ChatTabs.test.tsx:65`. The regex `/\.chat-view-tab\s*\{[^}]*width:\s*220px;/s` also matches the old `max-width: 220px;` (it's a substring, and `[^}]*` swallows the `max-`). Reproduced by running both regexes against the old `styles.css`: `flex: 0 0 220px` correctly fails, but `width: 220px` passes. So that assert wouldn't catch a revert of `width:220px` back to `max-width:220px`. Low because the companion `flex: 0 0 220px` assert does cover the change. Fix: anchor the prefix, e.g. `/\.chat-view-tab\s*\{[^}]*[\n;]\s*width:\s*220px;/s`.

3. **[R1-M1] Native fixture hand-writes `Button`'s class output (still open since round 1, re-triage).** `desktop-app/test/native/chat-workspace-geometry.fixture.cjs:7-10`. The `tab()` helper hardcodes `ui-button ui-button--ghost ui-button--neutral ...` instead of deriving it from the real `Button`. Classes match class-for-class today, so no live bug, but if `Button` renames its class scheme without touching `ChatTabs` the harness measures stale markup and stays green. This has now been open two rounds, so it either gets fixed (snapshot the real `ChatTabs` render, or a unit test asserting the class set) or closed with a written note. Your call.

4. **[R1-M2] Inactive-border reference is tautological (still open since round 1, re-triage).** `desktop-app/test/native/chat-workspace-geometry.fixture.cjs:173-177`. The reference div hardcodes the same `color-mix(in srgb, var(--border-soft) 25%, transparent)` the assert compares against, so it proves "CSS == a hand copy of the CSS". If the 25% design value were wrong the test stays green. Same re-triage situation as above: fix it to assert the relationship (alpha ≈ 25% of `--border-soft`) rather than equality with a copy, or close with a note.

5. **[R1-L3] `paintedSeam` never asserts `selectedFill ≈ surfaceFill`.** `desktop-app/test/native/chat-workspace-geometry.fixture.cjs:117-122`. It returns `selectedFill` but nothing compares it to `surfaceFill`, so the "active tab fill matches the surface fill" invariant isn't verified. They match today because both compose `--surface-strong` over `--panel`; a future non-transparent intermediate background would introduce a color step with no failing test. Add a small `colorDistance(selectedFill, surfaceFill)` check.

## What checked out clean

- Security: nothing. Cosmetic renderer-only change, `tab.title` escaped by React (children + `aria-label`), no `dangerouslySetInnerHTML`/`eval`, preload/`contextIsolation`/`webPreferences` untouched, and the profile-ui lockfile adds no runtime dependencies.
- Fixed width vs geometry: no fixture assertion depends on an individual tab's width; wide/narrow blocks still line up, `activeCenterX`/`selectedFillY` land inside the active tab, and overflow still triggers (`N×220 + gaps > viewport`).
- Truncation intact (`__select flex:1 1 auto; min-width:0` + `__label` ellipsis). The `min-width:0` on `.chat-view-tab` is now redundant but harmless.
- No regressions: the width fix was audited in isolation from the dev merge, no lost behavior beyond the intended design change, no JS consumer of tab width, no contract change.
- Suites: `ChatTabs.test.tsx` + `ChatViewWorkspace.test.tsx` → 8/8 passed. Native `test:native:chat-workspace-geometry` ran and passed this round.
- The good assert (`flex: 0 0 220px`) does fail against the previous head, so it's a valid guard.

zach88 (approved):

PR #441, opened 2026-08-24
feat(control-ui): give agent hooks their own advanced tab and table

andypavia:
## Summary

The Hooks UI on the agent Advanced tab (`/agents/<name>/advanced`) hung below the sub-tab bar as an always-visible section with its own edit mode. This promotes it to a sub-tab of its own — **first and default**, alongside Per-tool approval and Env vars — and rebuilds it on the same paradigm the Context and Access tables already use.

## What changed

**Hooks is now a sub-tab.** `HostAdvancedTab` gains a `hooks` entry, placed first and set as the default. `HostGuardrailsSection` moved inside it; the separate render in `app/hosts/[name]/page.tsx` is gone. `guardrailsData` / `persistGuardrails` pass through unchanged, so save and optimistic-concurrency behaviour are identical.

**The table follows Context/Access.** Section header with the primary `Add hook` control on the right, `cu-table cu-table--header-band` in a `cu-table-wrap`, `cu-empty` empty state, and a per-row danger `IconX` behind a confirm dialog. No edit mode and no Save/Cancel — removing saves immediately, the same way revoking access does.

**Hook names link to their detail page.** `/guardrails/<hook-name>`, via the same `cu-link` button pattern Access uses for members and teams. Navigation is deliberately *not* gated on `canWrite`: a read-only operator loses Add/Remove but can still click through.

**`Add hook` goes to `/marketplace/org/entries?type=llm-hook`** — the org marketplace entries list, narrowed to guardrail hooks. The unfiltered list mixes in every connector and plugin the org has published, which is not what someone adding a hook is looking for.

**Org marketplace entries can be filtered by type.** `?type=` on `/marketplace/org/entries`, which is what makes the above work. An unrecognised value falls back to the full list rather than rendering an empty one, and "no hooks yet" reads differently from "you have published nothing yet".

**Built-in guardrail editing is dropped for now.** `builtins` still rides along untouched on every save, so removing the editor does not wipe it from the Host spec — there is a test pinning exactly that. The CSS that styled the built-in control row and the old edit-mode hook chips is removed with it.

## Notes for review

- **A Phase column was added.** A hook that runs at two lifecycle points is genuinely two rows — that is how it is stored — so a 2-column table could not tell them apart. Access and Context do not need this because a member is a member.
- **`AccessGrantModal` moved to `components/SelectionModal`.** It had nothing access-specific in it. An intermediate revision of this branch used it for an installed-hook picker on the Hooks tab; that was replaced by plain navigation, so today the only caller is `HostAccessTab` (Add member / Add team). The extraction is kept because it is the right home for a shared control, but reviewers should know it is currently a one-consumer component.
- **E2E fix:** `host-approval-tools-e2e.spec.ts` waited on `.cu-host-approval-section` right after landing on `/advanced`. The guardrails panel reuses that class, so with Hooks defaulting it would have matched the wrong section — the spec now selects the Per-tool approval sub-tab explicitly.

## Testing

- `tsc --noEmit` clean for every file touched.
- `vitest run`: **1576 passed**, 1 skipped. Nine tests on the hooks section, three on the entries filter, one on the default sub-tab.
- Six test files fail to load on missing deps (`fast-check`, `@clerum/display-field`). Confirmed pre-existing by stashing — they fail identically on the base.
- Swept the touched files for unused exports and orphaned CSS classes; everything left has a caller.
- Exercised manually against minikube: `/agents/chatllm/advanced`, `/marketplace/org/entries?type=llm-hook`, and `/guardrails/hook-acme1-token-compactor-v0-2-1-6da759e1` all serve 200 with real hook data.

### DIFF
--- a/control-ui/app/constants/marketplaceEntryTypes.ts
+++ b/control-ui/app/constants/marketplaceEntryTypes.ts
@@ -5,8 +5,15 @@ export const GUARDRAIL_ENTRY_TYPE = 'llm-hook'
 
 // Human labels for the active-filter notice on the entries list. Only types
 // that something links to with `?type=` need an entry here.
-export const MARKETPLACE_ENTRY_TYPE_LABELS: Record<string, string> = {
-  'llm-hook': 'Guardrail hooks',
-  'mcp-server': 'Connectors',
-  recipe: 'Plugins',
-}
+//
+// Null-prototype on purpose: the keys are indexed with an untrusted `?type=`
+// value off the URL, so an inherited key (`toString`, `constructor`, …) would
+// otherwise resolve to a function and be treated as a real label.
+export const MARKETPLACE_ENTRY_TYPE_LABELS: Record<string, string> = Object.assign(
+  Object.create(null) as Record<string, string>,
+  {
+    'llm-hook': 'Guardrail hooks',
+    'mcp-server': 'Connectors',
+    recipe: 'Plugins',
+  }
+)
--- a/control-ui/app/constants/marketplaceEntryTypes.ts
+++ b/control-ui/app/constants/marketplaceEntryTypes.ts
@@ -1,0 +1,12 @@
+// Registry `entry_type` values, as sent on the wire. Shared by the surfaces
+// that link into a type-filtered entries list and the list that reads the
+// filter back off the URL, so the two never drift apart.
+export const GUARDRAIL_ENTRY_TYPE = 'llm-hook'
+
+// Human labels for the active-filter notice on the entries list. Only types
+// that something links to with `?type=` need an entry here.
+export const MARKETPLACE_ENTRY_TYPE_LABELS: Record<string, string> = {
+  'llm-hook': 'Guardrail hooks',
+  'mcp-server': 'Connectors',
+  recipe: 'Plugins',
+}
--- a/control-ui/app/constants/routes.ts
+++ b/control-ui/app/constants/routes.ts
@@ -79,6 +79,10 @@ export const CONTROL_ROUTES = {
     // The org-named tab (design spec §4): everything this deployment owns.
     org: '/marketplace/org',
     orgEntries: '/marketplace/org/entries',
+    // Same section, pre-filtered by entry type (`?type=llm-hook`). The type is
+    // a filter over one list, not a section of its own, so it stays a query
+    // param rather than becoming a route segment.
+    orgEntriesFiltered: (query?: ControlRouteQuery) => withQuery('/marketplace/org/entries', query),
     orgImages: '/marketplace/org/images',
     orgCredentials: '/marketplace/org/credentials',
     orgConnection: '/marketplace/org/connection',
--- a/control-ui/app/globals.css
+++ b/control-ui/app/globals.css
@@ -1968,20 +1968,6 @@ textarea.cu-input {
   justify-content: flex-end;
 }
 
-/* Per-agent Guardrails tab: add-built-in control row + removable hook chip. */
-.cu-guardrails-add-row {
-  display: flex;
-  flex-wrap: wrap;
-  gap: var(--cu-space-2);
-  align-items: center;
-}
-
-.cu-guardrails-ref-chip {
-  display: inline-flex;
-  align-items: center;
-  gap: var(--cu-space-1);
-}
-
 .cu-expandable-row__name,
 .cu-installed-plugin__name {
   color: var(--cu-text);
--- a/control-ui/app/globals.css
+++ b/control-ui/app/globals.css
@@ -544,6 +544,17 @@ body.cu-body {
   background: var(--cu-bg-elevated);
 }
 
+/* Nested inside a container that already draws a bordered card (a form
+   section), the list's own border, radius, and elevated background repaint the
+   same box in the same tokens — the inner corners then read as a stray rounded
+   edge under the header. Flush keeps the row rhythm and drops the chrome. */
+.cu-summary-list--flush {
+  padding: 0;
+  border: 0;
+  border-radius: 0;
+  background: none;
+}
+
 .cu-summary-list__row {
   display: flex;
   justify-content: space-between;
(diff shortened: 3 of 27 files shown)


zach88:
Took a solid pass through this one across the hooks tab rebuild, the `?type=` marketplace filter, the `SelectionModal` extraction, and a security sweep. Really clean PR overall: the risky bit (the `?type=llm-hook` filter contract that "Add hook" relies on) checks out end to end, the built-in passthrough is preserved, and everything is green on `tsc --noEmit` + the touched test suites. Nothing here blocks a merge, there's just one behavioral nit I'd fix and a small pile of cleanups.

## Verdict

🟡 **Approve with nits**
- **Optional / non-blocking: 7**

## Optional (non-blocking)

This PR is mergeable as-is. #1 is the one I'd actually fix before merging; the rest are your call as follow-ups.

1. **[R1-M1] Double success toast on every hook remove.** `components/HostGuardrailsSection/index.tsx:59` + `app/hosts/[name]/page.tsx:633`. Removing a hook fires two overlapping success toasts: the parent `persistGuardrails` shows `'Guardrails saved.'` after the PUT + reload, and the section then shows its own `'<id> removed from this agent.'` on resolve. On base there was a single toast (the old section never called `showToast`). The paradigm you're copying from (`HostAccessTab`) is self-contained, it does its own save and its own toast without an `onSave` round-trip to the parent, which is why it doesn't double up. Repro: traced through the save path. Fix: keep one toast, either drop `'Guardrails saved.'` from `persistGuardrails` (the section already emits the per-hook message) or don't toast in the section.

2. **[R1-L2] `?type=` render crash on inherited object keys.** `components/PublisherView/OwnedEntries.tsx` (~L248). `typeFilter` from the URL is used as an index into the plain-object `MARKETPLACE_ENTRY_TYPE_LABELS`, so `?type=toString` (or `__proto__`, `hasOwnProperty`, `constructor`) returns a truthy function instead of `undefined`, `activeTypeLabel` goes truthy, and the banner branch calls `activeTypeLabel.toLowerCase()` which throws. The entries view crashes for whoever opens that link (self-DoS, no data exposure). Repro: by reading the render path. Fix: `Object.prototype.hasOwnProperty.call(LABELS, typeFilter)` before using the label, or build the map with `Object.create(null)`.

3. **[R1-L1] Stale mock + false comment.** `components/__tests__/HostDetailsPage.identity.test.tsx:50`. The added `getLlmHooks` mock carries the comment "The Hooks sub-tab loads the installed LlmHooks to populate its picker," but the new section navigates plain with no picker and never calls `getLlmHooks` on that path (grep-confirmed, the only real caller is `app/guardrails/page.tsx`). Dead mock + a false claim about the contract. Fix: drop the mock entry and its comment.

4. **[R1-L3] Contradictory empty state under a filter.** `components/PublisherView/OwnedEntries.tsx` (~L247). When the org has published nothing but a recognized `?type=` is present, you render both "Showing guardrail hooks only. Show all entries" and "You haven't published any registry entries yet." Gate the banner on `entries.length > 0`, or the "haven't published" line on `!activeTypeLabel`.

5. **[R1-L4] `hasPrivateEntries` now computed over `visibleEntries`.** `components/PublisherView/OwnedEntries.tsx:223`. It feeds the `sharingUnavailable` subtitle, so under `?type=llm-hook` with only public hooks visible, the "private entries stay visible only to your org" notice can disappear even when the org does have private entries of another type. Defensible (the notice describes the filtered view), flagging in case the intent was to describe org state.

6. **[R1-L5] Body invariant only half-covered (T5).** `components/__tests__/HostGuardrailsSection.test.tsx`. The body and the type comment promise `builtins`, rules, ceiling, trustLevel *and* `limits` ride through untouched on save. The remove test pins `saved.builtins` but not `saved.limits`, even though the fixture sets `limits: { maxHooksPerPhase: 8 }` and `withHooks` carries it via spread. Add `expect(saved.limits).toEqual(GUARDRAILS.limits)`.

7. **[R1-L6] Dead class + stale comment.** `components/HostGuardrailsSection/index.tsx:89` wraps the section in `cu-guardrails-tab`, which has no rule in `globals.css` (harmless, layout comes from `cu-access-section__header`). And `app/hosts/[name]/page.tsx:644` still comments "stays in edit mode," which no longer applies now that the section has no edit mode. Housekeeping.

## What checked out clean

- **The `?type=llm-hook` filter contract holds end to end.** `ownedEntryKind` returns `'llm-hook'`, which matches `GUARDRAIL_ENTRY_TYPE` and the `MARKETPLACE_ENTRY_TYPE_LABELS` keys, and `OwnedRegistryEntry` documents `entryType` camelCase with `getOwnedRegistryEntries` forwarding the body verbatim. That dependency predates this PR (`entryTypeLabel` already read it), so "Add hook" lands on a real, populated list. Test fixture matches the real producer shape (T1 ok).
- **Guardrails passthrough is intact.** `withHooks = {...source, hooks}` preserves `builtins`/`rules`/`limits`/`ceiling`/`trustLevel`; covered by test and `tsc`.
- **No stale-state race on rapid successive removes.** `persistGuardrails` does `setBusy(true)` + `await loadData('none')` before `setBusy(false)`, and `disabled = busy || saving` keeps the buttons disabled through the whole cycle.
- **`guardrails.detail(id)` exists and encodes** (`encodeURIComponent`), no path traversal / open redirect.
- **E2E selector is valid**, `.cu-advanced-tabs` is emitted and sub-tabs render as `<button>`, so the `.click()` matches.
- **`SelectionModal` is a 1:1 extraction** of the old `AccessGrantModal`, no dead imports left in `HostAccessTab`, removed CSS classes and `GuardrailBuiltinType`/`GUARDRAIL_BUILTIN_OPTIONS` have no remaining refs.
- **Security: SAFE.** URL-derived values are encoded via `URLSearchParams`/`encodeURIComponent`, React escapes everything rendered, no `dangerouslySetInnerHTML` or unsanitized href, and real write authz lives on the `PUT /api/v1/admin/hosts` admin endpoint.
- **Suites:** `tsc --noEmit` clean, 39/39 tests green across the touched suites.

PR #444, opened 2026-08-25
feat(control-ui): pre-fill the guardrail install form from the author's declarations

andypavia:
## What

The registry already publishes what a guardrail hook needs to run — `hook_meta.requiredCapabilities` and `hook_meta.authorDefaults` — but the install form ignored both and opened with every capability unticked.

That default is not merely unhelpful. `enforceCapabilities` (`mcp-host/src/core/guardrails/capabilities.ts`) **discards** ungranted actions rather than erroring:

```
deny       without may_deny               → downgraded to 'no_decision'  ← the call PROCEEDS
rewrite    without may_rewrite            → dropped
substitute without may_substitute_result  → dropped
```

So clicking through the wizard installed a hook that reports `Ready` and silently does nothing. Evenfire's own PII-redaction reference hook is reachable this way — it records its needs in a source comment because there was nowhere else to put them.

## Changes

**`191c7504` — a collapsible variant for `FormSection`**

- `collapsible` renders the title as the disclosure toggle; `defaultOpen` seeds its state. Both optional, so all ~40 existing callers are untouched.
- Uses the rotating `IconChevronRight` + `.is-expanded` idiom that `LlmProviderConfig` and the expandable tables already use, rather than the hardcoded `▸`/`▾` glyphs in `HostApprovalSection`.
- Adds `cu-form-section__group` / `__subheading` — a fieldset + legend for a labelled group of controls inside one section, so grouped checkboxes keep their name for assistive tech without nesting a second bordered section inside the first.

**`ed16e5fc` — read the author's declarations in the install form**

- `requiredCapabilities` seeds the capability selection, so the default install is the *working* one.
- `authorDefaults.failMode` seeds the fail-mode select.
- `authorDefaults.orderHint` is shown as text, not converted — see Open questions.
- Step 1 is restructured so the hook's own facts, the capability grant, and the ordering knobs sit behind one **Advanced details** disclosure, leaving the target agent and credentials in front.

## Declaring is not granting

`hook_meta` is publisher-supplied, so this is the part worth reviewing closely. The declaration only pre-ticks an already-permitted box:

- Values are filtered to the four known capabilities — a malformed or hostile entry degrades to "declared nothing" and can **never widen** what the operator may grant.
- The operator's selection remains the grant, still narrowed client-side by the agent's `capabilityCeiling` and still rejected server-side (`capability_exceeds_ceiling`) when it exceeds it. An absent ceiling is the empty set server-side, so it fails closed.
- A declared `may_deny` forces fail-closed exactly as a manual tick does, matching the CRD's CEL rule.
- A declared capability outside the ceiling stays unticked and disabled, and is reported as needed-but-unavailable.
- The final review step still states the effective grant before installing.

The label map for `orderHint` is null-prototype: the key is publisher-supplied, so an inherited key (`toString`, `constructor`) would otherwise resolve to a function and render as a real hint. Same remedy already applied to `MARKETPLACE_ENTRY_TYPE_LABELS` in #441.

## Deliberately not included

`authorDefaults.timeoutMs` and `onUnavailable` are **not** surfaced. The CRD supports both and the author publishes them, but the install route (`control-api/src/routes/admin/registry.ts`) never forwards them to the LlmHook spec — the CRD defaults would apply instead. Showing them would display settings that are silently never applied. Forwarding them needs a control-api change first.

## Testing

- `HookInstallForm` had **no tests**; this adds 9, covering the seeding, malformed and inherited-key inputs, and the ceiling invariants above.
- `FormSection` adds 3, covering the non-collapsible path being unchanged and the disclosure's `aria-expanded` / `aria-controls` wiring.
- Full suite: **158 files, 1633 tests green**. `tsc --noEmit` clean.
- Verified manually against `@acme1/token-compactor@0.2.3` on `clerum-test`, which is the first version carrying these fields (0.2.0–0.2.2 have neither, and correctly render the old empty form).

## Open questions

1. **`orderHint` → `order` has no published mapping.** The author sends `"early"`; the CRD wants an integer. Inventing a mapping would silently write a value the author never chose, so Order stays at 100 and the hint is shown as text. If there's an intended mapping, I'll wire it.
2. **Should the collapsed header summarise the grant?** Capabilities now sit inside the disclosure, so an operator can reach the review step without having seen the list. The review step does state the effective grant, but a summary on the collapsed header (e.g. *Advanced details — no capabilities granted*) would surface it earlier.

## Adjacent drift found, not fixed here

`lifecyclePoints` is **six** values in the CRD (the four LLM-lane points plus `preToolUse`/`postToolUse`) but four in `lib/api.ts` and in `GuardrailHooksTable.types.ts`, while `app/guardrails/[name]/page.tsx` already uses all six. Left for a separate change.


jozer-rami:
## Review

Checked out at `7f77932e` and reviewed against `origin/dev`, plus the control-api enforcement path, the Host CRD, and the mcp-host capability enforcer. Registry-side facts below come from reviewing registry PR #73 alongside this.

The design is right and the security reasoning is careful. Three things block, and one of them means the feature currently fails on its own happy path.

### Blockers

**1. On an agent with no `capabilityCeiling`, the default install now 403s.**

`hostCapabilityCeiling` returns `null` when `spec.guardrails.capabilityCeiling` is absent (`HookInstallForm/index.tsx:42`). Three things key off that null and all of them no-op:

* the narrowing effect early-returns, `if (!ceiling) return` (`index.tsx:149`)
* `outsideCeiling = !!ceiling && ...` is `false`, so nothing is disabled
* the seeded `setCapabilities(new Set(declaredCapabilities))` (`index.tsx:114`) survives intact

control-api does the opposite. `const ceiling = guardrails.capabilityCeiling ?? []` (`control-api/src/routes/admin/registry.ts:1907`), so every requested capability lands in `outOfCeiling` and the request is rejected with `capability_exceeds_ceiling` (`:1924`).

This mismatch predates the PR, but it was unreachable by default: capabilities started empty, so only an operator who deliberately ticked a box hit it. Seeding puts it on the straight-through path.

It is not an edge case either. `capabilityCeiling` has no `default:` in the CRD (`charts/clerum-crds/crds/host.yaml:480`), and control-ui never writes one. The field exists only as a type (`HostGuardrailsSection/types.ts:29`) with no editor anywhere. Every agent created through the UI has no ceiling, so any hook declaring `requiredCapabilities` installed onto one fails. That is exactly the population this PR targets.

The fix is to mirror the server and treat an absent ceiling as the empty set client-side. Watch the loading window, where `selectedHost` is `undefined` and currently produces the same `null`.

The new tests encode this rather than catch it. The shared fixture is `host('agent-a', null)` (`__tests__/HookInstallForm.test.tsx:60`), and both `pre-ticks the capabilities the author declares` and `a declared may_deny still forces fail closed` assert boxes are ticked against a ceiling-less host.

**2. The warning copy is false for `may_add_context`.**

`enforceCapabilities` branches on `deny`, `rewrite`, and `substitute` only (`mcp-host/src/core/guardrails/capabilities.ts:34-41`). `may_add_context` appears in the enum lists and in tests, and nowhere else in `mcp-host/src`. A hook that is not granted it still adds context normally.

So for that one capability the banner ("Without it, it installs and reports Ready, but the actions it takes are discarded") tells the operator something untrue, and pushes them to widen a grant that changes nothing. It also costs the banner credibility for the three capabilities where it is accurate. Either scope the message to the three enforced capabilities, or say something different for `may_add_context`.

**3. Merge order: this cannot land before registry #73 merges and deploys.**

Neither `requiredCapabilities` nor `authorDefaults` exists on any entry today. The registry deploys from `main`, and two independent gates reject the unknown keys at publish, so the fields cannot be seeded ahead of the release. Every path this PR adds is currently dead, and the feature is not testable end to end until #73 ships.

### Should fix

**`orderHint: "normal"` renders no hint.** The registry enum is three values, `early | normal | late`, but `ORDER_HINT_LABELS` defines only two (`constants.ts:47-48`) and the comment above it says "a word ('early', 'late')". `normal` is legal and publishes clean today. One line.

**The degraded-install warning is invisible on the default path.** The `missingDeclared` banner sits inside `<FormSection title="Advanced details" collapsible>` with `defaultOpen = false` (`ui/index.tsx:62`), and collapsed content is unmounted rather than hidden. Step 1 reports the effective grant but never says what was dropped (`index.tsx:474-483`). In the case this PR exists to fix, a ceiling silently narrowing the grant, the operator sees nothing at either step. The dropped capabilities belong in the review line.

Related: `role="status"` will not announce here, because the live region enters the DOM already populated.

**Back re-collapses "Advanced details".** Step-0 content is `{step === 0 ? (...) : null}` (`index.tsx:275`), so Continue unmounts the subtree and destroys `FormSection`'s local `open` (`ui/index.tsx:64`). Back re-mounts at `useState(defaultOpen)`, which is `false`. The capabilities the operator just reviewed, and any warning their edit produced, disappear. Lifting `open` into `HookInstallForm` fixes it.

**No test asserts the install payload.** `installHookFromRegistry` is mocked and never asserted; all nine new tests stop at DOM state. One test that walks to step 1 with the default ceiling-less host and asserts the `capabilities` array would have caught blocker 1 outright. The payload is what the PR's central claim is about, and it is the one thing untested.

**No contract backstop.** The `Registry contract (consumer → registry)` job checks out `keyper-labs/evenfire-registry` with no `ref:`, so it runs against registry `main`, which lacks both fields, and `clerumRegistryClient.test.ts` mentions neither. It also short-circuits whenever `CROSS_REPO_READ_TOKEN` is absent. Worth adding both fields to that test once #73 is on `main`.

### On "declaring is not granting"

The conclusion holds, but the stated reasoning leans on the wrong thing. Per #73, the registry is a shape boundary, not a truth boundary. It validates the enum, bounds and dedupes the arrays, rebuilds field by field with no body spread, and freezes `hook_meta` per `(name, version)`. It forms no opinion on whether a declaration is truthful or minimal, and never reconciles it against a ceiling it cannot see. A hostile author can legally publish all four capabilities.

So every bit of narrowing has to come from this component. Checking each:

* **Enum filter**: correct. `index.tsx:91-97` is a strict allowlist intersection, so a malformed or hostile entry degrades to "declared nothing".
* **Ceiling intersection**: correct when a ceiling exists, absent when it does not (blocker 1). That failure is fail-closed at the server, so it is a UX defect and not a privilege escalation.
* **Disclosure before the click**: this is what actually carries the argument. Step 1 states the effective grant, so an all-four declaration is surfaced before Install. Worth saying so in the PR body directly, rather than resting the case on the registry.
* **Headless inheritance**: not reachable. The pre-fill is client-side React state, so no non-UI path inherits it.
* **Escalation by upgrade**: not reachable yet, since there is no hook-upgrade UI in control-ui. It goes live the moment someone builds one, and re-acceptance of a changed `requiredCapabilities` should be designed in then rather than retrofitted.

One qualifier on the consent flow. Moving the grant behind a collapsed disclosure in the same commit that makes it pre-ticked means consent degrades from a deliberate tick to a prose acknowledgment of raw enum ids at step 1. That is a real weakening, and it should be an explicit decision in the PR rather than a side effect of the layout change.

### Nits

* The ceiling explanation is dropped whenever a declaration exists (`index.tsx:394`). Both facts can apply at once: a declared capability outside the ceiling shows disabled checkboxes with no sentence explaining why.
* `missingDeclared.join(', ')` prints raw enum ids three lines below checkboxes labelled "May deny" and "May rewrite". `HOOK_CAPABILITY_OPTIONS` is already imported and has the labels.
* `HOOK_INSTALL_ERROR_LABELS` (`constants.ts:54`) is a plain object literal indexed by a server-supplied `code` in `describeHookInstallError` (`index.tsx:53`), three lines below the new null-prototype map. Same defect class, low severity since the key comes from control-api, but odd to fix one and not the other in the commit that names the pattern.
* `defaultOpen` has no production caller and only seeds `useState`, so a future caller computing it from async data gets a section stuck closed with no error.
* This is now the third hand-rolled disclosure, alongside `HostApprovalSection` and `LlmProviderConfig`. The next fix to the pattern has to be made in three places.
* The Order and Fail-mode `description` ternaries repeat their base sentence in every branch and look up `ORDER_HINT_LABELS[authorOrderHint]` twice (`index.tsx:433-457`).
* `timeoutMs` and `onUnavailable` are added to the type (`lib/api.ts:3039`) but never read. Either wire them or leave them out, so the type does not advertise support that is not there. If they do get wired, note `timeoutMs` is unbounded on #73 and gets capped by #74.
* `if (!Array.isArray(raw)) return []` (`index.tsx:94`) collapses "author said nothing" and "author explicitly said none needed", which the registry deliberately keeps distinct. No visible difference today, since both take the same branch.
* `aria-controls` points at an id that is not in the DOM while collapsed (`ui/index.tsx:79`). Either render the div with `hidden`, or drop the attribute when closed. `FormSection.test.tsx` only covers the `defaultOpen` case, so the collapsed state is unverified.

### What holds up well

The null-prototype `ORDER_HINT_LABELS` genuinely closes the inherited-key path: `toString`, `constructor`, and `__proto__` all resolve to `undefined`. Refusing to convert `orderHint` into a number is the right call, and matches the registry's reasoning that absolute order is a property of the set of hooks on a Host that no author can see. Making `collapsible` and `defaultOpen` optional keeps all existing `FormSection` callers untouched. The `may_deny` implies fail-closed handling is correct and mirrors the CRD rule, and the registry rejects that pair at publish anyway, so it is sound defense in depth.

The problem this PR identifies is real and worth fixing. It is blocker 1 and the `may_add_context` copy that need another pass.

andypavia:
Thanks — this was a genuinely useful review. Both blockers were real and are fixed, along with four of the five should-fixes. Six commits on top of `7f77932e`.

## Blockers

**1. Ceiling-less agent → 403 on the default path** — `8a08bfd4`

Confirmed on both sides and fixed by mirroring the server: an absent `capabilityCeiling` is now the empty set client-side.

One thing that fell out of doing it: `null` had to be kept, but re-pointed at *"no agent resolved yet"*. Returning `[]` for the undefined `selectedHost` during the load window is the case you flagged, and it is worse than it looks — the narrowing effect would run on mount and clear the author's seed *before* the hosts arrived, so even an agent with a real ceiling would lose its pre-tick. So `null` = unknown, narrows nothing; `[]` = known and empty, narrows everything.

The fixture is inverted too: the default host now carries an explicit all-permitting ceiling, and a new test pins the ceiling-less behaviour rather than encoding it.

**2. `may_add_context` copy is false** — `6bbbb991`

Correct, and the gap is bigger than the missing branch. `Contributor` has fields for `rewrite`, `substitute`, and `audit` and **none for added context**, so `enforceCapabilities`' three branches are already the complete set of gateable fields. The LLM lane's three response actions never add context, and the tool lane advertises `additionalContext` in its protocol comment (`remoteToolHook.ts:9`) but never reads it from `body` — so a hook author can write against a documented field and have it dropped silently, granted or not.

`may_add_context` is a capability with no corresponding action anywhere in the runtime. Filed as #459 with four directions; it needs a spec answer given the Phase 3 TODO, so I have not touched it here.

In this PR the warning now splits: the three enforced capabilities keep the sharp wording, anything else gets an informational line saying the runtime does not act on it today. `ENFORCED_HOOK_CAPABILITIES` in `constants.ts` points at both `enforceCapabilities` and #459 so the coupling is findable when that lands.

**3. Merge order** — agreed, this waits on registry #73.

One clarification: the self-hosted registry in `clerum-test` *does* carry both fields — `@acme1/token-compactor@0.2.3` has `requiredCapabilities: ["may_rewrite"]` and the full `authorDefaults`. That is what the manual verification ran against. Your point stands for the hosted registry on `main`, which is what the contract job and any real install would hit.

## Should fix

- **`orderHint: "normal"`** — `6ca42103`. All three registry values are labelled now. `normal` says the author expressed no preference rather than claiming a position they did not choose; rendering nothing made it indistinguishable from no hint at all. The stale `"early", "late"` comment is corrected.

- **Degraded install invisible** — `6a489240`. Three parts: the disclosure now opens itself the first time the grant falls short (once per entry, so closing it sticks), the review step repeats the notices, and both steps render one shared component so they cannot drift. Whether the grant falls short is only knowable after the ceiling loads, which is why this could not be a static default.

  Your `role="status"` catch was right, and the fix surfaced a second bug: the live region is now always mounted with the notices inside it, so ticking a box announces. The first version of the auto-open also fired for a *fully granted* hook, because the seed is applied by an effect and the first render therefore sees every declared capability as missing — gated on `loading`, and pinned by a test.

- **Back re-collapses** — `8f9fb5cd`. Took your suggestion and lifted it. `defaultOpen` is replaced by an optional controlled `open`/`onOpenChange` pair rather than sitting alongside it: it had no production caller, and one mechanism covers both the unmount case and the async-data case you raised in the nits. Uncontrolled sections still start closed.

- **No test asserts the payload** — `6d5db28f`. Three now do: no ceiling requests nothing, a permitted `may_deny` carries its forced fail-closed through to the request, and a capability outside the ceiling never leaves the form. Your claim checks out — I reverted the blocker 1 fix and re-ran, and `requests no capability from an agent with no ceiling` fails, along with six others. Previously the whole suite went green on the broken behaviour.

- **Contract backstop** — not done, agreed on the reasoning. It needs #73 on `main` before the job would test anything.

Full suite: 158 files, 1696 tests green. `tsc --noEmit` clean.

## On the consent point

This is the part of the review I think is most worth settling, and you are right that it should be an explicit decision rather than a side effect.

Where it stands after these changes: the grant is still behind the disclosure, but the disclosure opens itself whenever the install would be degraded, and the review step now states what was dropped rather than only what was granted. So the operator lands on the checkboxes in the case that matters instead of reading a prose summary of enum ids.

That mitigates it; it does not answer it. If you still think a pre-ticked grant belongs in the main flow regardless, say so and I will move it back out — it is a small change now that the state is lifted.

I also take the point that the "declaring is not granting" argument should rest on the step 1 disclosure rather than on registry validation. The PR body predates all of this and needs rewriting; I will do that separately.

## Nits

Two are closed as side effects: the ceiling explanation and the declaration copy are now composed rather than chosen between, so both facts can appear at once; and `defaultOpen` is gone. The rest — raw enum ids in the notices, `HOOK_INSTALL_ERROR_LABELS`, the third hand-rolled disclosure, `aria-controls` while collapsed, the repeated ternaries, `timeoutMs`/`onUnavailable` typed but unread, and the `[]`-vs-absent collapse — are not addressed yet.

jozer-rami (approved):

PR #446, opened 2026-08-25
fix(control-api): key workflow trigger rate limits per user session

### LINKED ISSUE #431: fix(control-api): per-user workflow trigger rate limit for external-rest-api callers
## Summary

`workflowTriggerRateLimit()` keys the PostgreSQL bucket from **service bearer first**, then `x-user-session-token`, then admin cookie. Every call from `external-rest-api` sends **both** headers:

- `Authorization: Bearer <control-api service token>` (`external-rest-api/src/controlApiClient.ts`)
- `x-user-session-token: <end-user session JWT>`

Because bearer is checked first, **all external users share one global 10 triggers/minute bucket** for the entire service identity, instead of per-user isolation.

## Pre-existing

This behavior predates PR #428 (`fix/f4-admin-workflow-live-revocation`). On `dev`, the limiter already used `bearer || userSessionToken`; PR #428 only added admin cookie as a third fallback without changing priority.

```typescript
// control-api/src/routes/workflows/shared/rateLimit.ts (current)
function workflowTriggerRateLimitCredential(req: Request): string | null {
  const bearer = extractBearerToken(req)
  if (bearer) return bearer  // ← wins for external-rest-api
  const userSessionToken = String(req.header('x-user-session-token') || '').trim()
  if (userSessionToken) return userSessionToken
  // ...
}
```

## Impact

- One noisy or abusive external user can exhaust trigger quota for **all** Profile/external users until the window resets.
- Legitimate high-volume automation (multiple users behind external-rest-api) appears as a single caller to rate limiting.
- Auth and authorization are unaffected; this is a **fairness / abuse-containment** bug, not an authz bypass.

## Affected route

- `POST /external/workflows/:ns/:name/trigger` (`control-api/src/routes/external/workflows/trigger.routes.ts`)

## Suggested fix

Prefer end-user identity for bucket key when present:

1. `x-user-session-token` (external user session)
2. admin cookie (Control UI trigger)
3. verified admin/mcp bearer where applicable
4. service bearer only as last resort (or exclude service bearer entirely on `/external/*` trigger paths)

Add regression tests in `control-api/test/routes.externalWorkflows.test.ts` proving two different `x-user-session-token` values get independent buckets even when the same service bearer is present.

## Discovered during

Closing audit of PR #428 (SOL High auditor, 2026-08-21).

alfredolopez80:
## Summary
- Prefer the end-user session over the service bearer when keying `workflowTriggerRateLimit`, so Profile users no longer share one 10/min PostgreSQL bucket.
- HEAD keys the external lane on the verified `user:<userId>` (unverified session header falls back to IP) and gives mcp-host a bearer-only limiter.
- Add unit and route regressions that send both production headers and assert independent `bucket_key` values.

Fixes #431

## Fireworks tech graph

Style 1 Flat Icon comparison of the broken `origin/dev` key order versus this PR:

Issue 431 per-user workflow trigger rate limit

Source: [`docs/architecture/issue-431-workflow-trigger-rate-limit.svg`](https://github.com/evenfire-ai/evenfire/blob/fix/control-api-per-user-workflow-trigger-rate-limit/docs/architecture/issue-431-workflow-trigger-rate-limit.svg)

## Test plan
- [x] `cd control-api && ./node_modules/.bin/tsc --noEmit`
- [x] `cd control-api && npm run test:run -- test/workflows.shared.rateLimit.test.ts test/routes.externalWorkflows.test.ts` (58 passed)
- [x] CI on `8ead7b21` (control-api + Real Postgres 16 + full matrix)
- [ ] Optional: two distinct Profile sessions triggering the same recipe through `external-rest-api` do not share 429s


alfredolopez80:
@claude closing review requested for https://github.com/evenfire-ai/evenfire/pull/446 (`8ead7b21`).

Please treat this as the close-out review of issue #431 (per-user workflow trigger rate limits). CI on this exact HEAD is fully green, including `control-api` and Real Postgres 16. Do not merge; this is review-only.

## Intent

`external-rest-api` always forwards both `Authorization: Bearer <service token>` and `x-user-session-token` on `POST /external/workflows/:ns/:name/trigger`. The original limiter keyed the shared 10/min PG bucket from the service bearer first, so every Profile user shared one quota.

## Commits

1. `173e3541` — prefer `x-user-session-token` over the service bearer (raw token as key).
2. `8ead7b21` — close review findings with a KISS split:
   - External: verified `user:<userId>` via `verifyExternalSessionToken`. Present-but-unverified session header falls back to IP (same rotation-evasion pattern as the admin cookie / #428).
   - mcp-host: new `mcpHostWorkflowTriggerRateLimit()` that keys only on the bearer. Injecting `x-user-session-token` must not mint a new bucket.
   - Shared hash helper; no auth/429 reorder, no extra `/external` IP edge limiter, no metric-prefix split.

## Files

- `control-api/src/routes/workflows/shared/rateLimit.ts`
- `control-api/src/routes/mcp-host/workflows/trigger.routes.ts`
- `control-api/test/workflows.shared.rateLimit.test.ts`
- `control-api/test/routes.externalWorkflows.test.ts`
- `control-api/package.json` / lock (0.1.497)

## Please verify

- Two verified users behind one service bearer get distinct `workflow_trigger:<hash>` buckets.
- Two session tokens for the same `userId` reuse one bucket.
- Forged/rotated unverified `x-user-session-token` + bearer collapse to one IP bucket (not the service identity).
- mcp-host trigger stays on the bearer when a session header is injected.
- Dual-header success path (`201`) still attributes the run to the session user.
- `requireExternalWorkflowCaller` / `requireMcpHostControlWorkflowCaller` authorization is unchanged.
- Cookie-vs-bearer order is only relevant for this helper; neither real caller sends the admin cookie.

## Evidence (AGENTS.md lanes)

- **T0 PASS (local + CI):** `control-api` `tsc --noEmit` and `vitest` `test/workflows.shared.rateLimit.test.ts` + `test/routes.externalWorkflows.test.ts` = 58/58. CI `Build / test / typecheck (control-api)` SUCCESS on `8ead7b21`. Full CI matrix SUCCESS (33 checks), including CodeQL and leak guards.
- **T1 PASS (CI Real Postgres):** `Real Postgres integration (Postgres 16)` SUCCESS on this HEAD. Suite is HTTP/middleware credential selection; no new SQL/role-reset path.
- **T2 NOT_RUN (AGENTS.md):** this worktree has no branch-owned `MINIKUBE_PROFILE` / `.local-notes/minikube-profiles/branch.mk`. Creating or borrowing another worktree’s profile is forbidden. The change does not touch deploy manifests, images, GFS, or cluster runtime, so a Minikube `MINIKUBE_T2_PASS` is not required to close this HTTP bucket-key fix.

Please post findings (or a clean close-out) on this PR. Merge stays blocked on a human Approve plus explicit maintainer authorization.

zach88:
Took a careful pass through this one. It's a tight, well-scoped change: the credential-precedence reorder cleanly fixes the shared-bucket problem from #431 without breaking either of the two consumers, and the bearer-only mcp-host lane preserves the exact base behavior. Nothing here blocks a merge.

**🟡 Approve with nits**
- Optional / non-blocking: 1

## Optional (non-blocking)

The PR can merge without this — it's follow-up / author's call.

1. **[R1-L1] Keying tests mock the session verifier, so the sign→verify→key seam isn't exercised end-to-end** — `test/workflows.shared.rateLimit.test.ts:~139` and `test/routes.externalWorkflows.test.ts:~336` both fully mock `verifyExternalSessionToken` with hand-written claims, so the "unverified session header → falls back to `ip:`" property is asserted against the stub rather than the real RS256 verifier. Not blocking: the mock is faithful to the real shape (`{userId}` / `null`), it follows the file's existing idiom, and real rejection of forged/expired/bad-audience tokens is already covered in `middleware.externalSessionAuth.legacyToken.test.ts` and `.lifecycle.test.ts`, which sign with the real producer `signExternalSessionToken`. If you want to close the seam in one place: add a case that signs a valid token with `signExternalSessionToken` and a forged one, without mocking the verifier.

## What checked out clean

- **Precedence reorder does what the body says.** With both a bearer and a session token present, base keyed on the bearer (the #431 bug); head keys on the verified `user:<userId>`. Verified by reproducing: the route regression test `keys independent trigger buckets…` fails against the merge-base `42cd1b9` (both buckets identical) and passes on head — genuine T3 coverage.
- **No caller regressions.** The only two consumers are `external/workflows/trigger.routes.ts` (`workflowTriggerRateLimit()`) and `mcp-host/workflows/trigger.routes.ts` (`mcpHostWorkflowTriggerRateLimit()`). The mcp-host bearer-only lane produces the *same* key as before — base already keyed on the raw bearer since the mcp-host control-JWT doesn't verify as an admin token, so no dimension is lost. mcp-host never sends `x-user-session-token` or an admin cookie.
- **Token trust is sound.** `user:<userId>` only comes from `verifyExternalSessionToken(token)?.userId` — RS256 pinned (no alg:none / HS confusion), issuer/audience/exp/role/authGeneration checked, public key, length capped at 4096. Unverifiable token → `null` → `ip:` fallback, so token rotation collapses to one IP bucket instead of minting fresh ones.
- **No secret leakage.** The only rate-limit log emits the already-hashed `bucketKey` (`workflow_trigger:<sha256>`), never the raw token or bearer.
- **Suites green:** `test/workflows.shared.rateLimit.test.ts` + `test/routes.externalWorkflows.test.ts` → 58 passed; `test/routes.mcp-host.workflows.test.ts` + `test/routes.adminWorkflows.test.ts` → 171 passed.

zach88 (approved):

PR #449, opened 2026-08-25
fix(hcc): skip no-op token restarts and name /ready 503 clauses

### LINKED ISSUE #447: fix(hcc): recover readiness after watch closure and avoid unnecessary token restarts
## Incident

The 2026-08-25 dev deploy failed three times in the HCC rollout gate:

- [Run #88 / job 97875558473](https://github.com/keyper-labs/evenfire-infra/actions/runs/32870261204/job/97875558473), runner `1000134242`
- [Run #90 attempt 1 / job 97959060287](https://github.com/keyper-labs/evenfire-infra/actions/runs/32896043800/job/97959060287), runner `1000134270`
- [Run #90 attempt 2 / job 97963953737](https://github.com/keyper-labs/evenfire-infra/actions/runs/32896043800/job/97963953737), runner `1000134274`

All setup, GCP authentication, image mirroring, image pinning, migration, and apply steps succeeded. The only failing step was `Verify rollout status across namespaces`.

The failing deployment was `control-plane/deployment.apps/host-context-controller`. It remained `0/1` with readiness HTTP 503 until the 90-second rollout gate expired. The workflow rolled back HCC, but correctly kept the deploy failed.

## Evidence and timeline

The token step reported:

`internal-control-jwt-secrets patched (no change)`

Despite that, HCC was restarted. The unconditional restart path is in:

https://github.com/evenfire-ai/evenfire/blob/d90f564daba7283d12961e0c1a764874c7adb12b/deploy/scripts/apply-inter-service-tokens.sh#L298-L317

After rollback, the HCC pod started successfully and logged `Provider started` and `Host Context Controller running`.

The invalid `mcp-web-research` egress error was already being logged, but HCC still started. Later, the controller logged repeated:

- `Premature close` for multiple watches
- `HostInventoryAuthorityUnavailableError`

Readiness then failed with HTTP 503. The cluster also had 97–99% CPU requests reserved and produced `Insufficient cpu` scheduling events, which amplified rollout latency.

Deleting `mcp-web-research` did not restore HCC: the McpServer was deleted, but HCC remained `Running`, `0/1`, with readiness HTTP 503.

## Root cause

The deploy unnecessarily restarts an unchanged HCC. After the restart, HCC does not reliably recover readiness when its watches or inventory authority are interrupted. The rollout gate then fails and rolls back.

The GitHub Actions runners are not the root cause; three different runners reproduced the same HCC rollout failure.

## Separate McpServer defect

`mcp-web-research` had an invalid `public-web` egress binding containing `protocol: TCP`. That configuration should be rejected only for that McpServer and must not make HCC globally unready. The resource has since been deleted, but deleting it did not resolve the HCC readiness failure.

## Required fixes

1. Restart HCC only when the exact HCC-consumed secret changes.
2. Make HCC re-list and re-establish watches without remaining unready.
3. Expose a structured readiness failure reason and include HCC container logs before rollback.
4. Keep sufficient schedulable CPU capacity for control-plane rollouts.
5. Ensure an invalid McpServer only affects that McpServer and never makes HCC globally unready.

## Acceptance criteria

- A control-api-only deploy does not change the HCC pod or ReplicaSet when the HCC secret is unchanged.
- HCC recovers readiness after watch interruption without a pod restart.
- A bad McpServer is isolated with `ExternalEgressReady=False`; HCC and other MCPs remain operational.
- The dev deploy completes with a green HCC rollout gate.

alfredolopez80:
## Summary

Clerum-dev deploys were Recreating `host-context-controller` on every token apply even when `internal-control-jwt-secrets` reported `patched (no change)`. HCC is `Recreate` + `replicas: 1` and stays `/ready` 503 until inventory is recertified, so that gratuitous restart raced the infra rollout gate (evenfire-infra runs 88, 90×3, 91). The first observed failure (run 88 / PR #440) did not change HCC code or the HMAC; the token script was the only rollout trigger. Issue #447 tracks the incident; this PR stops the recurrence and makes the next 503 diagnosable.

- Skip `kubectl rollout restart` for `control-plane/host-context-controller` unless the in-memory `INTERNAL_CONTROL_JWT_HCC_HMAC_SECRET` actually changed (empty/unreadable fails open; `FORCE_CONSUMER_RESTART=true` still restarts). WRC-only rotations do not Recreate HCC. Image/spec applies still roll HCC themselves.
- `/ready` 503 now includes closed-gate `reasons` (`mcp_watch_unsynced`, `context_watch_unsynced`, `host_watch_unsynced`, `safety_pass_uncertified`, `revocation_revision_mismatch`). The 200 body stays `{status:'ready', ready:true}`.
- Wire `scripts/tests/test-inter-service-tokens.sh` into `ci-public.yml`.

Fixes #447 (P0 + P1). Serialized safety-pass hardening remains deferred until `reasons` name a persistent false clause.

## Test plan

- [x] `bash scripts/tests/test-inter-service-tokens.sh` (unchanged HMAC skip, HMAC change, WRC-only, empty-before, FORCE, no secret material in logs)
- [x] `npm --prefix host-context-controller test -- src/readinessGate.test.ts src/server.test.ts src/k8sClient.test.ts src/readinessSafetyFence.integration.test.ts` (303 passed)
- [ ] CI `ci-public` on this PR
- [ ] After merge: a control-api-only clerum-dev deploy with a byte-identical HCC HMAC must not create a new HCC pod/ReplicaSet

### DIFF
--- a/.github/workflows/ci-public.yml
+++ b/.github/workflows/ci-public.yml
@@ -412,6 +412,9 @@ jobs:
       - name: Test WRC-to-HCC contracts
         run: bash scripts/tests/test-wrc-hcc-contracts-gate.sh
 
+      - name: Test inter-service token HCC restart skip
+        run: bash scripts/tests/test-inter-service-tokens.sh
+
       - name: Test WRC internal dependency NetworkPolicy contracts
         run: bash scripts/tests/test-wrc-internal-dependency-networkpolicy-gate.sh
 
--- a/deploy/base/control-plane/host-context-controller.yaml
+++ b/deploy/base/control-plane/host-context-controller.yaml
@@ -36,10 +36,12 @@ spec:
   # reconciliation finishes, so time-to-Ready scales with the McpServer/Context
   # fleet — clerum-dev took 651s at 108 McpServers / 22 Contexts. The unset default
   # (600s) governed the deploy because `kubectl rollout status` aborts the instant
-  # the deadline is exceeded, and the clock starts at apply-inter-service-tokens.sh's
-  # `rollout restart`, so the wait inherits only the remaining budget (600-160=440s
-  # vs a 480s timeout on 2026-08-05: #202's timeout never bound, deploy failed 50s
-  # short). 1200s covers the 900s wait plus a 300s head-start. Mitigation, not the
+  # the deadline is exceeded. apply-inter-service-tokens.sh used to start that
+  # clock on every token apply via `rollout restart`, even when
+  # INTERNAL_CONTROL_JWT_HCC_HMAC_SECRET was unchanged; no-op token deploys no
+  # longer Recreate this Deployment, so they no longer consume this budget.
+  # A real HMAC change (or FORCE_CONSUMER_RESTART) still restarts HCC — then
+  # 1200s covers the 900s wait plus a 300s head-start. Mitigation, not the
   # fix — #205 decouples readiness from fleet convergence, after which this should
   # come back down.
   progressDeadlineSeconds: 1200
--- a/deploy/scripts/apply-inter-service-tokens.sh
+++ b/deploy/scripts/apply-inter-service-tokens.sh
@@ -43,6 +43,8 @@ set -euo pipefail
 #                                        control-api Secret already has a value.
 #   INTERNAL_CONTROL_JWT_WRC_HMAC_SECRET override WRC HMAC (default: preserve-or-generate)
 #   INTERNAL_CONTROL_JWT_HCC_HMAC_SECRET override HCC HMAC (default: preserve-or-generate)
+#   FORCE_CONSUMER_RESTART               if true, restart every consumer including
+#                                        HCC even when the HCC HMAC is unchanged
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
 # shellcheck source=lib/clerum-minikube-context.sh
@@ -229,6 +231,17 @@ kctl -n channels patch secret workflow-approval-request-reader-credentials \
   -p='[{"op":"remove","path":"/data/service-token"}]' >/dev/null 2>&1 || true
 
 # --- 2. internal-control-jwt-secrets (control-plane) ---
+# Compare the HCC-consumed key only. The merge patch writes WRC + HCC together,
+# so a whole-Secret resourceVersion (or parsing `patched (no change)`) would
+# Recreate HCC on a WRC-only rotation. Values stay in shell locals — never logged.
+HCC_HMAC_BEFORE="$(read_secret_key control-plane internal-control-jwt-secrets INTERNAL_CONTROL_JWT_HCC_HMAC_SECRET)"
+if [ "${FORCE_CONSUMER_RESTART:-}" = "true" ] || \
+   [ -z "$HCC_HMAC_BEFORE" ] || \
+   [ "$HCC_HMAC_BEFORE" != "$INTERNAL_CONTROL_HCC_HMAC" ]; then
+  HCC_SECRET_CHANGED=true
+else
+  HCC_SECRET_CHANGED=false
+fi
 log "Patching Secret internal-control-jwt-secrets (control-plane)"
 ensure_secret control-plane internal-control-jwt-secrets
 INTERNAL_CONTROL_PATCH="$(jq -cn \
@@ -296,7 +309,9 @@ WA_READER_PATCH="$(jq -cn \
 kctl -n channels patch secret workflow-approval-request-reader-credentials --type=merge -p "$WA_READER_PATCH"
 
 # Roll the consumers so they pick up the fresh tokens on first deploy. Safe
-# no-ops if the Deployments do not yet exist.
+# no-ops if the Deployments do not yet exist. HCC is Recreate + replicas:1
+# and reloads HMAC only at process start, so skip that extra rollout when the
+# in-memory HCC key is unchanged (image/spec applies still roll HCC themselves).
 for pair in "control-plane:control-api" \
             "control-plane:workflow-recipes" \
             "control-plane:host-context-controller" \
@@ -306,6 +321,11 @@ for pair in "control-plane:control-api" \
             "channels:clerum-workflow-approval-request-reader"; do
   ns="${pair%%:*}"
   dep="${pair#*:}"
+  if [ "$ns" = "control-plane" ] && [ "$dep" = "host-context-controller" ] && \
+     [ "$HCC_SECRET_CHANGED" != "true" ]; then
+    log "Skipping rollout of $ns/$dep: hcc-hmac unchanged"
+    continue
+  fi
   if kctl -n "$ns" get deploy "$dep" >/dev/null 2>&1; then
     log "Rolling deployment $ns/$dep to pick up fresh Secret values"
     kctl -n "$ns" rollout restart deploy "$dep" >/dev/null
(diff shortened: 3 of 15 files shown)


alfredolopez80:
@claude Please perform a complete re-review of this PR at HEAD `b11b9663f3989578be17baec6b1a92e24c6cd706` (base `dev`).

Review the full delta versus `dev`, not only the latest commit. The latest commit maps the `stopped` inventory clause to `/ready` 503 reason `controller_stopped`, flattens HCC logger `Error` fields so `{ err }` is not empty, migrates the provider-authority catch in `getReadinessState` to the structured readiness logger, and adds HTTP tests that use the same `resolve*Fn` wiring as `main.ts`.

Focus:
- skip HCC `rollout restart` only when `INTERNAL_CONTROL_JWT_HCC_HMAC_SECRET` is unchanged
- WRC-only rotation must not Recreate HCC
- empty/unreadable before-value and `FORCE_CONSUMER_RESTART=true` still restart HCC
- `/ready` 200 body stays `{status:'ready', ready:true}`
- 503 reasons are a closed vocabulary with no secrets, CR names, or inventory contents
- HMAC values must never be logged

Ticket: #447

zach88:
Took a deep pass through this one. Tight, well-scoped PR and the tests actually earn their keep (the new bash case fails against the base, so it's a real regression guard, not a golden). The core refactor of `isReadinessInventoryAuthoritative()` is logically identical to the old version, the HMAC compare is apples-to-apples plaintext on both sides, and nothing sensitive leaks into logs or the `/ready` body. Only one thing I'd nudge you on, and it's a sibling-cleanup, not a bug.

**🟡 Approve with nits**
- Optional / non-blocking: **7** (1 Medium, 6 Low)

## Optional (non-blocking)

This PR is safe to merge without these; they're follow-up or your call.

1. **[R1-M1] Sibling `console.error` left un-migrated** — `host-context-controller/src/server.ts:144`. Intent #3 migrated `console.error('[Server] Provider authority readiness check failed:', err)` to `readinessLog.error(...)`, but its identical twin in `safeHostAuthoritative()` is still raw `console.error('[Server] Host authority check failed:', err)` — and its own comment even says *"same shape as the provider check in getReadinessState()"* (`server.ts:138`). Same class, same file, right next door. If `hostAuthoritativeFn()` throws on `/api/v1/desktop/*`, that error bypasses the `sanitize`/structured shape you just standardized. Suggest `readinessLog.error('host authority check failed', { err })`. Flagging it here (not as pre-existing) because it's the sibling instance of the exact idiom this diff fixed.

2. **[R1-L1] Skip closes on "Secret changed this run", not "running pod matches Secret"** — `deploy/scripts/apply-inter-service-tokens.sh:238-244,324-328`. If someone patches the HCC key out-of-band (direct kubectl) without restarting HCC, then reruns the script *without* setting `INTERNAL_CONTROL_JWT_HCC_HMAC_SECRET`, both sides read the new Secret value → `HCC_SECRET_CHANGED=false` → skip, and the pod keeps signing/validating with the stale in-memory HMAC. Low because the sanctioned rotation path (env var) always restarts, and a real rotation always restarts control-api too, so there's no signer/verifier skew — the scenario needs an unsanctioned flow where you'd probably just use `FORCE_CONSUMER_RESTART=true` anyway. Optional hardening: note the assumption in the block comment, or compare against the HMAC actually mounted in the running pod.

3. **[R1-L2] Returned shape declared as an inline literal** — `host-context-controller/src/k8sClient.ts:1016`. `getReadinessInventoryDetail()` spells out its 7-key return literal instead of annotating `: ReadinessInventoryDetail`. Type-safe today (drift gets caught at the `resolveReadinessDetailFn` boundary in `main.ts:152`, tsc passes), but the shape lives in two places. Annotating with the imported type makes the coupling explicit at the producer.

4. **[R1-L3] `sanitize(Error)` drops `cause`** — `host-context-controller/src/logger.ts:23`. It flattens to `{name, message}` and discards `cause` / `AggregateError.errors`. Not a regression (base logged `{}`), but worth revisiting if you start chaining errors with `cause`.

5. **[R1-L4] Test name oversells** — `host-context-controller/src/server.test.ts:621`. 'emits closed reasons through the same resolvers main.ts uses' runs the real resolvers but over a watcher stub whose two methods re-implement the formula by hand, not the real `McpServerWatcher`. It validates resolver wiring, not the real alignment of the two watcher methods (that's covered in `k8sClient.test.ts`). Coverage is complete across both files — just the name is misleading.

6. **[R1-L5] No pure preserve-or-generate case** — `scripts/tests/test-inter-service-tokens.sh:157-195`. All 5 cases pass `INTERNAL_CONTROL_JWT_HCC_HMAC_SECRET` as an env var, so `INTERNAL_CONTROL_HCC_HMAC` always comes from `resolve_token`'s env branch, never its `read_secret_key` branch — which is the dominant CI redeploy path. The decode equivalence is still covered via `HCC_HMAC_BEFORE`, so risk is low; a 6th case with no HCC override (only `KUBE_SECRET_HCC` for both reads) would mirror the real path faithfully.

7. **[R1-L6] Version bump skips 0.1.233** — `host-context-controller/package.json:3` (0.1.232 → 0.1.234). Probably consumed on another branch; just confirm it's intentional and not a local double-bump.

## What checked out clean

- **Refactor equivalence** (reproduced): `getReadinessInventoryDetail()` exposes the same 7 terms `isReadinessInventoryAuthoritative()` ANDs, same order. `hasCertifiedSafetyInventory()` is a pure read, so short-circuit vs full-eval yields the same boolean.
- **detail ↔ reasons ↔ authoritative** (reproduced): each of the 7 `false`s maps to exactly one reason; the two revision booleans collapse to `revocation_revision_mismatch` intentionally. A 503 with `reasons:[]` from misalignment is impossible — `getReadinessState()` is synchronous, no yield point between the authority check and the detail read.
- **Fail-safe** (reproduced): a throw in `readinessDetailFn` is caught, logged, returns `undefined`; the 503 still emits with statusCode intact.
- **HMAC contract** (reproduced): both compare sides are plaintext, read happens before the patch (compares against the old value), and compared value == written value (round-trip stringData→.data→decode is consistent).
- **Constructor wiring** (reproduced): 9 positional params line up position-by-position between the signature (`server.ts:103-119`) and the call (`main.ts:186-196`); `readinessDetailFn` lands at position 9, no shift.
- **`/ready` body** (reproduced): 200 → `{status:'ready', ready:true}` with no `reasons`; 503 adds `reasons` only when non-empty. `reasons` are fixed enums — no CR names, counts, or revision numbers.
- **Suites**: 302 TS tests pass, `tsc --noEmit` EXIT 0, `test-inter-service-tokens.sh` passes and fails against the base (T3 verified).

zach88 (approved):

alfredolopez80:
@zach88 thanks — validated the 7 nits. None were merge blockers.

Cheap ones landed in `21453c28`:

- **R1-M1** — `safeHostAuthoritative()` now uses `readinessLog.error('host authority check failed', { err })` plus a desktop 503 test.
- **R1-L1 (comment only)** — script comment now states the skip is Secret-to-write, not running-pod-env. Hardening to compare the live pod is #452.
- **R1-L2** — `getReadinessInventoryDetail()` returns `ReadinessInventoryDetail`.
- **R1-L4** — test renamed to `emits closed reasons through the main.ts resolve*Fn wiring`.
- **R1-L5** — added a preserve-or-generate case (`env -u` both HMAC overrides; both reads come from `KUBE_SECRET_HCC`).
- **R1-L6** — intentional. First commit hook was `0.1.232 → 0.1.233`; the review-fix commit bumped `0.1.233 → 0.1.234`; this nit commit is `0.1.235`.

Left as issues (not cheap / new design):

- **R1-L1 remaining** — #452 (skip vs running pod HMAC)
- **R1-L3** — #453 (`Error.cause` / `AggregateError.errors`)

PR #458, opened 2026-08-26
feat(desktop-app): first-run onboarding questionnaire

andypavia:
## Why

The desktop app is a client that cannot function without a server, and it had no way to acquire one. A cold install showed a single email field labelled "Continue setup" — which only worked if that address already had an invitation registered with `member-registration-service`. Everyone else hit a dead end with no explanation of what an "Environment" was or where to get one.

## What

A first-run questionnaire that asks at most two questions and routes to one of five paths. Implements PR 1 of `specs/desktop-app-onboarding-spec.md`, plus the hosted link-out.

**Q1 — "Do you already have an Evenfire server?"**
- *My team already uses Evenfire* → invitation flow (existing handlers, new home)
- *I have a server address* → manual environment form (also the one-click skip)
- *No, I'm just getting started* → Q2

**Q2 — "How do you want to run Evenfire?"**
- *Evenfire hosts it for me* → links to the Evenfire site, falls back to the manual form
- *I'll run it myself* → links to the deployment guide, falls back to the manual form
- *I have no idea* → a comparison of the two, framed around the requirements that force the decision

`AuthPage` becomes sign-in-only. Its environment dock moves to a shared `RuntimeConfigDock` rendered by both pages, because a cold install no longer renders `AuthPage` at all and the Localhost escape hatch lives there.

## Notable decisions

**Self-hosting means any cluster the user controls — remote or local.** The spec originally scripted a minikube install with a `127.0.0.1` default and a 10GB Docker prerequisite, which quietly redefined self-hosting as "on this laptop". `docker-compose.yml` settled it: 43 lines, `mcp-host` + `channel-reader` only, `CLERUM_ENABLE_AUTH: "false"` — no external-rest-api, so it cannot produce a `DesktopRuntimeConfig` target at all. Path B now links to documentation covering both shapes and shows no prerequisites, no install command, and no default address.

**No cost claims.** Hosted signup does not exist yet — no tenant provisioning, no `evenfire://desktop-environment` handoff — so the hosted path is a link-out, not an in-app signup, and the copy promises no trial or price. Two tests assert the absence of `free|trial|price|$|per seat|subscription`, so shipping hosted signup fails them and puts the copy decision in front of a human.

**Every compared option carries a downside.** A comparison listing only upsides is an advert; a test pins the count of downside marks.

## New IPC — three calls, all no-argument

`probeLocalhostReachable`, `openDeploymentDocs`, `openHostedSignup`. All call `assertTrustedSender`, and none take a parameter: the renderer cannot name a URL for the main process to fetch or open. The probe only ever checks the built-in Localhost option.

## Dev switch

`EVENFIRE_ONBOARDING_PREVIEW=true` (default false, only the literal `"true"`) makes the app start genuinely cold so the flow can be reviewed on a configured machine — separate config directory, env runtime config ignored, real environments never read or written. `make local-app-onboarding` wraps it. Verified in both directions: on → `screen: onboarding` on the preview directory; off → `screen: sign-in` on the real one.

## Verification

- Full desktop-app suite: **198 files, 1997 tests** passing
- `npm run style-rules` clean; no new CSS files, no raw hex or font sizes
- Driven by hand in the running app across every step, both themes of the brand lockup, and with the preview flag on and off
- Profile UI handoff confirmed unaffected: both deep-link listeners live in `useAuthController` (mounted app-wide), and the confirmation/success dialogs are siblings of the page switch, so they render over onboarding as they did over sign-in

## Not in this PR

Path A proper (the `state` nonce and waiting screen), path C's `invitation_config_not_found` copy, and the trial pill — PRs 2–4 in the spec's sequencing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -783,8 +783,12 @@ local-web: ## Run Control UI locally against minikube control-api port-forward
 	wait_for_port control-api 8090 "$$api_pid" http://127.0.0.1:8090/health; \
 	CONTROL_API_INTERNAL_URL=http://localhost:8090 npm --prefix control-ui run dev
 
+.PHONY: local-app-onboarding
+local-app-onboarding: ## Run Desktop App showing the first-run onboarding flow (isolated; real environments untouched)
+	@$(MAKE) --no-print-directory EVENFIRE_ONBOARDING_PREVIEW=true local-app
+
 .PHONY: local-app
-local-app: ## Run Desktop App locally against minikube API port-forwards
+local-app: ## Run Desktop App locally against minikube API port-forwards (EVENFIRE_ONBOARDING_PREVIEW=true previews onboarding)
 	@set -euo pipefail; \
 	cleanup() { \
 		local pids; \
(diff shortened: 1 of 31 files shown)


josue-tech-94:
@andypavia  **Approve with follow-ups**. No merge-gating correctness or security issues were found.

There is one non-blocking follow-up I recommend addressing:

### R1-L1 — Add direct IPC boundary coverage

Please add table-driven coverage in `desktop-app/test/ipc.test.ts` for these three new onboarding IPC channels:

* `auth:probeLocalhostReachable`
* `auth:openDeploymentDocs`
* `auth:openHostedSignup`

For each channel, directly verify that:

1. An untrusted sender is rejected through `assertTrustedSender`.
2. A trusted sender dispatches to the expected AppService method.
3. The AppService method is invoked with **no renderer-controlled arguments**, including if an unexpected extra IPC argument is supplied.

The implementation is currently safe; this is to pin the security-sensitive Electron boundary so a future change cannot accidentally start forwarding renderer input or remove the trusted-sender check while the higher-level onboarding tests continue passing.

One operational item as well: GitHub reported the reviewed branch as **behind `dev`**. Please bring it up to date before merge using the normal branch-sync process and make sure CI/CodeQL remain green on the resulting head.

Everything else from R1 looked good.

josue-tech-94 (approved):

Issue #460, opened 2026-08-26
bug(wrc,hcc): known no-op write-storm gate (mcpDelegation `specHashUnchanged`) never propagated to the shared-Context path — cross-controller write loop pins `/ready` at 503


alfredolopez80:
## Summary

This is not an unknown defect. **The codebase documents this exact failure mode in three separate
files, already fixed it for one path, and never propagated the fix to the adjacent one — in the same
file, 200 lines below, with the gate already imported at the top.**

The module docstring of `workflow-recipes/src/reconciler/specHash.ts` (lines 4–18) states the whole
mechanism:

> ```
> * Why this exists: WRC's apply path does an unconditional full PUT
> * (replaceNamespaced*) every reconcile. Because builders emit a partial spec, the
> * API server re-defaults server-managed fields (strategy, revisionHistoryLimit,
> * progressDeadlineSeconds, …) on every replace, bumping `metadata.generation`
> * with no real change. That perpetual churn drove a degraded↔active status flap
> * (observedGeneration always one behind) and a downstream HCC NetworkPolicy
> * no-op write storm. Hashing the *desired* manifest (never the server's defaulted
> * view) and gating the write makes reconciles idempotent.
> ```

`mcpDelegation.ts:466–472` repeats it at the call site:

> ```
> // Idempotency gate: skip the replace when the existing recipe-owned object
> // already carries the spec-hash we are about to write. […] WRC
> // re-reconciles every McpServer each resync; an unconditional replace bumps
> // resourceVersion, which HCC watches → it recomputes context NetworkPolicies
> // every loop (a no-op write storm). Skipping unchanged writes stops that.
> ```

and `workflowRecipeReconciler.ts` (~3628–3634) repeats it a third time for `createOrReplace`.

**The decisive detail is the locality.** In `mcpDelegation.ts`:

| Line | What |
|---:|---|
| `33` | `import { specHashUnchanged, stampSpecHash } from './specHash'` |
| `474` | `specHashUnchanged(…)` — guards the **McpServer** path |
| `696–750` | `replaceExistingContext` — the **Context** path, **never calls it** |

Same file. Symbol already in scope. Two hundred lines apart. The predicted consequence —
`unconditional replace` → `resourceVersion` bump → HCC watch → `recompute context NetworkPolicies
every loop` → *no-op write storm* — is exactly what `clerum-dev` is doing right now, continuously,
and it pins `/ready` at 503.

[...]

PR #467, opened 2026-08-26
fix(hcc): re-arm silent NetworkPolicy convergence sinks and name pass outcomes (#447)

alfredolopez80:
## Summary

- Completes the leftover on `runInitialNetworkPolicyConvergenceCore`: the comment above `:3046` already said LIST→WATCH recovery schedules a fresh pass, but unsynced (`:3046`) and post-pass authority-lost (`:3108`) were bare `return`s. Only `catch` reached `scheduleInitialConvergenceRetry`. `/ready` stayed 503 until a random watch event (20–35 min measured).
- Re-arm those sinks on the existing ladder `[5, 15, 30, 60, 300]` s (does not exhaust). Count dropped mid-pass effects (`:3072` / `:3080`) without a second re-arm. Name each pass result: `deferred-unsynced`, `aborted-authority`, `aborted-bump` (inventory-changed throw only), `failed` (any other `catch`), `certified`.
- This is **#447 P2a only**. It does not include #460 write-amp, Change D (content-identity on the in-flight pass), or P2b scoped abort. It is not a revert of #382 readiness content-identity.

Closes nothing on #447 by itself (token-restart and other items remain). Proof for this leftover: Vitest T12–T14 + the existing HCC suite.

## Test plan

- [x] T14 (`:3046`): unsynced swallow warns, arms the ladder, then one reconcile after sync and the ladder clears
- [x] T13 (`:3108`): still no positive effects; asserts `pass ended without certifying: inventory authority lost`; retry armed; later pass applies current inventory
- [x] T12: silent-sink miss, no later watch events, 5 s timer runs another pass
- [x] Change C: generic `catch` is `failed`, inventory-changed throw is `aborted-bump`
- [x] Keep-green: `k8sClient.test.ts`, `readinessGate.test.ts`, `server.test.ts` (305) and full HCC suite (1700)
- [x] `npm run build` in `host-context-controller`
- [ ] Existing HCC e2e remain keep-green (no new GKE-recycle e2e; watch-churn already says the #382 livelock RED lives in Vitest)
- [ ] Post-merge (not merge-blocking): next natural `context1` bump → ladder logs + Ready 200 ≤300 s after the cause is gone


zach88:
Took a deep pass through this one — core convergence logic, the metrics, and the test suite, plus a security angle on the isolation boundary. This is a tight, well-scoped liveness fix and it holds up: the two re-arm paths, the lifecycle guards, the error classification, and the four new metrics all behave exactly as the body claims. The only things I'd flag are test-robustness and one log level; nothing in the production path breaks.

🟡 **Approve with nits**
- **Optional / non-blocking: 4** (1 Medium, 3 Low)

## Optional (non-blocking)

The PR is mergeable without these — follow-ups or your call.

1. **[R1-M1] `certified > 0` assertion proves nothing** — `src/k8sClient.test.ts:4450-4457`. prom-client counters are never reset across the file, and earlier tests already emit `result:'certified'` (the "fences" pass at :4019 and the "defers" pass at :4398), so this counter is already `> 0` before the silent-sink test starts. If the fake-timer retry fired but the pass ended `aborted-authority`/`failed` instead of `certified`, the assertion stays green. The test isn't actually a hole today because `toHaveBeenCalledOnce()` + the `timestamp > successTimestampBefore` delta carry it — but the intent-revealing assertion is inert. Suggest capturing `certifiedBefore` at the top and asserting `toBe(certifiedBefore + 1)`, same as the `failed`/`aborted-bump` tests already do.

2. **[R1-L1] `console.error` on the authority-lost path is alert noise** — `src/k8sClient.ts:3159`. This logs at `error` level on every re-arm, but authority-lost is an expected transient during watch-generation/revision churn, and by that point the safety certificate may already be recorded (so `/ready` is 200 — the gate is `hasCertifiedSafetyInventory()` + revision alignment, not this timestamp). So a healthy controller that churns inventory each pass emits an error-level line + a climbing retry ladder every pass. The `deferred-unsynced` sibling one line up correctly uses `console.warn`; suggest matching it here.

3. **[R1-L2] Loose `> 0` metric asserts on global counters** — `src/k8sClient.test.ts:3985-4004` (fences) and `:4383-4396` (defers). The `swallowed{sink}`, `pass_results{result}`, and `effects_dropped{kind}` checks use `.toBeGreaterThan(0)` on process-global counters, so they pass today only because each test is the first emitter of its label — order-fragile. Each test is independently carried by a real delta (`retryAttempts` 4→5 / 2→3, the log-line spy, the pinned-timestamp check), so this is consistency, not a hole. Same `before + 1` treatment the sibling tests use would make them order-proof.

4. **[R1-L3] Ladder non-exhaustion isn't walked for the new paths** — `src/k8sClient.test.ts:4422`. The body promises the re-armed paths recover ≤300s and the ladder doesn't exhaust; the silent-sink test advances only the first rung (5000ms). Non-exhaustion is covered transitively by the existing shared-scheduler ladder tests (:4355, :4647) that the new arming reuses, so it's covered structurally — just no new test drives the newly-armed path across multiple rungs.

## What checked out clean

Verified, not assumed:

- **Ladder / double-arm.** `scheduleInitialConvergenceRetry`'s `initialConvergenceRetryTimers.has(lane)` guard prevents two live timers per lane; both new paths respect it. The attempt counter climbs `Math.min(attempt-1, 4)` → pinned at 300s, retries forever, never exhausts — matches the "does not exhaust" claim. Ladder constant is `[5000,15000,30000,60000,300000]`.
- **Timer leak after stop.** `stop()` sets `stopped = true` before clearing timers, and both `scheduleInitialConvergenceRetry` and the timer callback re-check it, so a schedule reached after stop is a no-op.
- **`aborted-bump` vs `failed`.** The inventory-changed error is thrown unwrapped at `networkPolicyReconciler.ts:1582`, before the additive try/catch, so the `error.message === DESIRED_NETWORKPOLICY_INVENTORY_CHANGED_MESSAGE` match holds; it's the sole throw of that constant, and the additive path throws a distinct message → correctly `failed`. No mislabel.
- **`observeInitialNetworkPolicyPass`.** `startedAtMs` is a fresh local per call; every path observes exactly once; `certified` doesn't fall through to the catch; `deferred-unsynced` ~0s is intended with `Math.max(0, …)` guarding clock skew.
- **effects-dropped closures.** Byte-for-byte the old early-return — `await work()` stays behind the same guard, unreached — with a bounded-cardinality counter (`kind ∈ {context, server}`) added ahead of it.
- **Security.** Both re-arm paths hard-stop before any cluster mutation; neither writes certificate/revision state, so `/ready` can't flip to 200 from `deferred-unsynced`/`aborted-authority`. All new metric labels are fixed string literals (no tenant-derived cardinality), and no secret-bearing data lands in the new log lines.
- **Suites/typecheck at head:** `k8sClient.test.ts` 268 passed, `networkPolicyReconciler.test.ts` 160 passed, `tsc --noEmit` clean. The five new re-arm tests were confirmed to fail on assertions (not import/typecheck) against the parent commit, so they genuinely cover the bug.

zach88:
Took another pass over round 2 (head `2ef4c34d`) — production path, the metrics, the tightened test suite, and the new e2e script, plus a collateral audit of everything that landed since round 1. This round closed all four nits from last time, and it holds up cleanly: the only production change in the range is the `console.error`→`console.warn` you flagged, and it's backed by a repro test that fails against the previous head. Nothing left is blocking.

🟡 **Approve with nits**
- **Optional / non-blocking: 6** (6 Low)

## Round 1 follow-ups — all closed
- **[R1-M1]** `certified > 0` is now a `certifiedBefore + 1 / + 2` delta (`src/k8sClient.test.ts:4547,4571`).
- **[R1-L1]** `src/k8sClient.ts:3159` is `console.warn`. Verified the repro: the `warnSpy … 'inventory authority lost'` assert (`:4018`) fails against `d5cb470d` and passes at head.
- **[R1-L2]** the loose `>0` asserts on global prom-client counters are all deltas now (`:4019-4034`, `:4470-4489`).
- **[R1-L3]** the new paths route through `scheduleInitialConvergenceRetry`; the test walks the ladder with real timers (attempts 1→2), and non-exhaustion is covered by the existing clamp table (`:7009-7016`).

## Optional (non-blocking)

The PR is mergeable without these — follow-ups or your call.

1. **[R2-L1] authority-lost path doesn't assert the `/ready` 503→200 recovery** — `src/k8sClient.test.ts:3937`. For the `aborted-authority` branch the test only asserts the intermediate effect (timer armed, `retryAttempts==5`, `/ready`==503), not that `/ready` actually recovers to 200 when the ladder fires. The `deferred-unsynced` branch has the full 503→200 HTTP test (`:4527`) and the ladder is shared code, so it's covered transitively — but a direct `advanceTimers` re-running the pass with authority restored and asserting `/ready`==200 would close the body invariant head-on.
2. **[R2-L2] `readPromTextMetric` matches labels by substring** — `src/k8sClient.test.ts:4457`. Uses `labelPart.includes(\`${key}="${value}"\`)`. Fine for today's values (no `result`/`sink` is a substring of another), but fragile the day a label value is added that's a prefix of an existing one. The fixture itself is derived from the real `/metrics` of `ContextMapperServer`, so this is just the parser being loose.
3. **[R2-L3] e2e `trap cleanup EXIT` is missing `INT TERM`** — `scripts/e2e/e2e-hcc-np-convergence-ladder.sh:334`. The script waits on long windows (`SWALLOW_OBSERVE_SEC=90s`, `READY_BUDGET_SEC=300s`). A Ctrl-C / SIGTERM during those waits can exit bash without running `cleanup`, orphaning the watch-gate lock (shared cluster resource — can block later runs). One-liner: `trap cleanup EXIT INT TERM`.
4. **[R2-L4] tautological assertion in the e2e** — `scripts/e2e/e2e-hcc-np-convergence-ladder.sh:190`. `[ "${swallowed:-0}" = "$deferred" ]` compares two counters the producer bumps in the same branch unconditionally (`src/k8sClient.ts:3080-3081`), so the equality can never fail on a given pod. Doesn't cause a false green (the real gate is `deferred >= 1` at `:186`), but it adds no coverage — drop it or swap for a cross-check that can actually diverge.
5. **[R2-L5] dead function in the e2e** — `scripts/e2e/e2e-hcc-np-convergence-ladder.sh:110`. `hcc_ready_now()` is defined and never called. The `2ef4c34d` commit touched this same region and left it; safe to delete.
6. **[R2-L6] asymmetric probe parameterization** — `scripts/e2e/e2e-hcc-np-convergence-ladder.sh:130`. `2ef4c34d` moves the `/metrics` scrape onto `HCC_PROBE_HOST`/`HCC_PROBE_PORT` (identical defaults), but `/ready` goes through the `ready_status` lib helper, which doesn't read those vars. Override one and the two probes point at different hosts/ports. Latent (same pod/port in practice), but worth keeping them consistent.

## What checked out clean
- **Security (SAFE, 0 findings).** Both re-arm branches only bump counters, emit a static `console.warn`, and reschedule — neither writes the safety certificate or revision, so `/ready` can't flip to 200 from them. `isReadinessInventoryAuthoritative()` re-checks cache-sync + revision alignment on top, so a stale certificate can't pass either. Metric labels are fixed literals with bounded cardinality (no tenant/CRD-derived label), no secrets in the new logs, and the `.has(lane)` guard prevents duplicate timers / retry storms. This turns fail-closed-stuck into fail-closed-then-retry — strictly safer.
- **The e2e can't go falsely green.** Anti-vacuity is real: green requires a caches-unsynced swallow + retry log, `/ready` 200 within a budget measured from the observation, and `deferred >= 1` on the wire; a lucky `/ready` 200 without a swallow exits `VACUOUS` (3), and `print_results` fails on `e2e_total==0`. Context gating is thorough (branch-scoped, rejects `*prod*`, owned minikube profile, node label, fault-injection ack). Only mutation is `rollout restart`, not `delete`. Metric names, labels, log strings, port 8081, and the `Recreate` precondition all verified against the real producer.
- **No regressions.** Audited `d5cb470d..2ef4c34d`: the log-level change has a single consumer (the test, migrated to `warnSpy` in the same commit) — no alert/dashboard/Prometheus rule depends on `error` there. The `>0`→delta move is correct against monotonic prom-client counters (suite doesn't reset the registry), and each test captures its own `*Before`. Suite stable in source order.
- **Suites.** `vitest run src/k8sClient.test.ts src/networkPolicyReconciler.test.ts` at head → 429 passed. `bash -n` + `shellcheck -x` on the e2e → clean (only benign SC1091/SC2015).

zach88 (approved):

PR #469, opened 2026-08-26
fix(hcc): skip no-op Service replaces at all four writers (#460)

alfredolopez80:
## Summary
- Skip no-op Service replaces at all four HCC writers (`ensureService`, channel-reader Service, McpServer Service, LlmHook Service) via `serviceMatchesDesired` after `preserveServiceAssignedFields`.
- Drop the duplicate Host `Updated Service` logs so only the helper logs a real write.
- Closes the #460 Service slice only. NetworkPolicy, Role, WRC timer, and fan-out stay out.

## Test plan
- [x] HCC Vitest: new NOOP/CREATE/LOG/CMP/RETRY/ORDER/GATE cases plus full suite (1719 passed)
- [x] `tsc -p tsconfig.build.json`
- [x] CI on this PR
- [ ] After deploy: `apiserver_request_total{verb="PUT",resource="services"}` attributable to HCC drops; do not use a frozen `resourceVersion`

### DIFF
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.240",
+  "version": "0.1.241",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.240",
+      "version": "0.1.241",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.241",
+  "version": "0.1.242",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.241",
+      "version": "0.1.242",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.242",
+  "version": "0.1.243",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.242",
+      "version": "0.1.243",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.242",
+  "version": "0.1.243",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.241",
+  "version": "0.1.242",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.240",
+  "version": "0.1.241",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/src/__tests__/asApiserverService.ts
+++ b/host-context-controller/src/__tests__/asApiserverService.ts
@@ -1,0 +1,87 @@
+import { vi } from 'vitest'
+import type * as k8s from '@kubernetes/client-node'
+
+/**
+ * Recorded kube-apiserver GET of a ClusterIP Service (`kubectl get svc -o json`
+ * shape), not a field list copied from `normalizeServiceForComparison`.
+ * Stamp controller-owned name/selector/ports via `asApiserverService`.
+ * When a newer apiserver default-fills a field, refresh this blob so the
+ * suite goes red until the comparator learns that default.
+ */
+export const RECORDED_CLUSTERIP_SERVICE: k8s.V1Service = {
+  apiVersion: 'v1',
+  kind: 'Service',
+  metadata: {
+    annotations: {
+      'kubectl.kubernetes.io/last-applied-configuration': '{}',
+    },
+    creationTimestamp: new Date('2026-04-01T00:00:00.000Z'),
+    generation: 1,
+    managedFields: [
+      {
+        apiVersion: 'v1',
+        fieldsType: 'FieldsV1',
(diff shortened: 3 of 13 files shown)


zach88:
Took a deep pass through this one. The core logic is solid: I tried hard to break the no-op gate into a false positive (skip a replace that was actually needed, which would leave a Service drifting) and couldn't, because the canonicalization only ever drops a field when its value equals the apiserver default, on both sides. Security angle is clean too. Everything that's left is test effectiveness, not code correctness, so nothing here blocks the merge.

**🟡 Approve with nits**
- **Optional / non-blocking: 4** (2 Medium, 2 Low)

## Optional (non-blocking)

The PR is mergeable as-is. These are follow-ups, either a quick test-only round or your call.

1. **[R1-M1] `existing` fixtures are hand-built, not derived from a real apiserver object.** `asApiserverService(desired)` in the four `*NoopGate.test.ts` (`hostServiceNoopGate.test.ts:37`, `channelReaderServiceNoopGate.test.ts:37`, `llmHookServiceNoopGate.test.ts:33`, `mcpServerServiceNoopGate.test.ts:28`) and the inline `existing` in `serviceMatchesDesired.test.ts:31` enumerate by hand exactly the fields `normalizeServiceForComparison` knows how to canonicalize. So NOOP-*-1 ends up proving the normalizer strips what the author injected, not that a real apiserver Service compares equal. The failure mode this hides: if the apiserver fills a default the normalizer doesn't model (`spec.ports[].appProtocol`, `allocateLoadBalancerNodePorts`, `finalizers`, or something a future version adds, like it once did with `internalTrafficPolicy`), then in prod `serviceMatchesDesired` returns false and a `replace` fires on every reconcile (the exact thing this PR removes) while the suite stays green forever. It's non-blocking because the fail-open direction means the worst case is an extra write, never drift. Suggested fix: derive one canonical `existing` from a recorded real `kubectl get svc -o json` (or an envtest round-trip) and build the fixtures off that, so a new apiserver-filled field shows up as a red instead of hiding. (Verified: ran the tests, compared the fixture builder against `buildService`.)

2. **[R1-M2] The retry test asserts the retry happened but not the fresh resourceVersion it exists to thread.** In `replaceWithConflictRetryOrder.test.ts:58`, `read` returns the same object (resourceVersion `'9'`) on every attempt, and the assertions only check that `replace` ran twice plus the "after 2 attempts" log. The body handed to the second `replace` is never inspected, so a regression that froze the stale resourceVersion across attempts (precisely what conflict-retry prevents) would stay green. Related: there's no case where the post-409 re-read flips `isUpToDate` to true, so the merge to gate to retry path on a fresh read is untested, and the body's "do not use a frozen resourceVersion" invariant is only half covered. Suggested fix: have `read` return rv `'9'` then `'10'` and assert the second `replace` body carries `'10'`; add a case where the retry's re-read makes the gate skip. (Verified: read the test against the `replaceWithConflictRetry` loop.)

3. **[R1-L1] No property test for the pure normalization functions.** `normalizeServiceForComparison` / `serviceMatchesDesired` are pure, so idempotency (`normalize(normalize(x)) === normalize(x)`) and reflexivity (`serviceMatchesDesired(x, x) === true`) are natural properties. They hold by construction today and `fast-check` isn't a devDep yet, so this is genuinely optional, just suite hardening.

4. **[R1-L2] `normalizeServiceForComparison` and `normalizeServiceValue` are exported without need** (`utils.ts:198,222`). Only `serviceMatchesDesired` is consumed in prod; the helpers are test-only. Trim the public surface if you feel like it. Cosmetic.

## What checked out clean

- Worst-case correctness reproduced adversarially: `targetPort` omitted vs different, selector change, dropped port, `publishNotReadyAddresses` all force a write; realistic steady state skips. No false-positive skip.
- Security: canonicalization is symmetric and value-gated, so selector, ports (value and order), type, externalIPs, loadBalancerSourceRanges, externalName, ownerReferences all stay in the comparison and any deviation forces a revert. Fail-open (throw to false to write) is the safe direction; `validateExisting` still runs before the gate. The helper logs only the Service name, no secrets.
- All four writers wired consistently with `preserveServiceAssignedFields` + `serviceMatchesDesired`, and there's no fifth writer left ungated: `sharedFileSystemReconciler` is create-only by design, so "all four writers" is accurate.
- The two removed `console.log` lines match the helper's `Updated ${description}` output character for character (minus the additive `(after N attempts)` suffix), so no log grep breaks. The `console.error` failure paths are kept.
- Node `>=24` so `structuredClone` is available, no crash risk. No CRDs touched.
- Full HCC suite: 79 files, 1724/1724 passing. New tests isolated: 17/17. `tsc` clean.

alfredolopez80:
Retriggering CI: pushes `ee6ecd87` / `95b924de` did not start Actions; reopening the PR to fire `pull_request`.

alfredolopez80:
Temporary close to retrigger CI on 95b924de.

zach88 (approved):

zach88:
Took another pass on this now that the round-1 test-quality nits got addressed. This round is test-only plus a one-line un-export, and it lands well: both round-1 Mediums are closed, the refactor kept every case, and nothing regressed. What's left is three tiny test nits that are pure author's-call.

## 🟡 Approve with nits

## Optional (non-blocking)

The PR is mergeable as-is. These are cosmetic test cleanups, take them or leave them.

1. **[R2-L1] CMP-6 reflexivity is nearly tautological** (`serviceMatchesDesired.test.ts:102`). `serviceMatchesDesired(desired, desired)` and `(existing, existing)` compare an object against itself, so `JSON.stringify(normalize(x)) === JSON.stringify(normalize(x))` is always true unless `normalize` throws. The only real signal it carries is "normalize doesn't blow up on the full recorded blob (fail-open)". If you want it to have teeth, compare `desired` against `asApiserverService(desired)` with one preserve-field stripped, so its absence actually breaks equality.

2. **[R2-L2] Port-drift helper doesn't carry `targetPort` along** (`asApiserverService.ts:53`). In `asApiserverService(desired, {port})`, `port` uses the drifted value but `targetPort: port.targetPort ?? port.port` uses the original, so the drifted object ends up `{port: 9090, targetPort: 8080}`. It still forces the replace (any diff does), so the test is fine, but it models a slightly unreal drift.

3. **[R2-L3] Dead `spec.ports` in the recorded fixture** (`asApiserverService.ts:42`). `RECORDED_CLUSTERIP_SERVICE.spec.ports` is never read because `asApiserverService` overwrites `spec.ports` wholesale with the ports mapped from `desired`. Harmless, just a dead field in the blob.

## What checked out clean

- **Round-1 R1-M2 (fresh resourceVersion on retry): closed.** RETRY-SVC-1 now asserts the 2nd `replace` body carries the fresh rv (`'9'` then `'10'`), and a frozen-rv mutant turns that assertion red, so it actually discriminates. RETRY-SVC-2 is new and covers the post-409 re-read that converges and skips the second replace, using the real `serviceMatchesDesired`.
- **Round-1 R1-M1 (hand-written fixture): meaningfully improved.** The recorded blob now carries apiserver-assigned fields the comparator doesn't strip (`clusterIP/clusterIPs/ipFamilies/ipFamilyPolicy` + `last-applied-configuration`), so the NOOP passing genuinely proves `preserveServiceAssignedFields` neutralizes them. The one residual is that the blob is hand-authored, so a brand-new apiserver default not in it would go green-in-error until the blob is refreshed (the file comment already flags this). Reasonable to leave as-is for this PR.
- **No lost coverage.** The un-export of `normalizeServiceForComparison` breaks no consumers (it was only used inside `utils.ts`; typecheck green on both tsconfigs). Every `it()` from round 1 survived; `serviceMatchesDesired` gained CMP-6 and `replaceWithConflictRetryOrder` gained RETRY-SVC-2. The ~44 lines each NoopGate shed are just the inline helpers moved to the shared `asApiserverService` module.
- **Security clean.** The no-op gate is value-gated and symmetric (a field is only dropped when its value equals the apiserver default); selector, ports, externalIPs, loadBalancerSourceRanges, externalName, publishNotReadyAddresses, ownerReferences and labels stay in the comparison. No fail-closed-to-skip path, `validateExisting` runs before the gate, normalizers log nothing.
- **Suites/typecheck:** full HCC suite 1721/1721 green, touched files 19/19, `tsc` green on `tsconfig.json` and `tsconfig.build.json`.

Issue #477, opened 2026-08-27
bug(hcc): phase-2 certification clauses on serving surfaces pay availability without closing the invoked risk


alfredolopez80:
## Status (2026-08-27) — the original "one indivisible diff" is superseded

Review this body, not the comments that first proposed the split. Delivery was
partitioned after measuring that **no runtime alerting exists today**
(Prometheus/Grafana inactive, Loki 0 HCC streams, Cloud Monitoring **0
notification channels**, HCC `/metrics` exposed and **not scraped**).

| Slice | What ships | Where | Alert required? |
|---|---|---|---|
| **A1** | Cut phase 2 from the **readinessProbe only**. Per-request stays fail-closed on all six clauses. | PR **#479** (Oleada 1, merge **after #480**) | **No.** The per-request 503 and the #467 logs remain the detector. |
| **A2a** | Additive only: metric `uncertified_window_seconds`, `degraded` field/header, scrape annotations. **Cuts nothing.** | Later wave (Oleada 4). Not #479. | No. Useful to any operator who already scrapes. |
| **A2b** | Cut phase 2 from the **per-request** predicate. | Same later wave. **Blocked** until a detector exists in the *deployment* environment. | **Yes.** This is the cut that would create silence. |

Merge order for Oleada 1: **#480 (B) → #479 (A1) → #481 (C)**. Do not merge A1
before #462 / #480.

The original "steps 1–3 then cut both surfaces in one diff" rule **still
applies to A2b**. It does **not** apply to A1. A1 does not create a new silent
failure mode: discovery keeps returning 503 while uncertified.

---

## What happens

`/ready` and the per-request gate both withhold service when the NetworkPolicy
safety inventory is uncertified. Two of the six readiness clauses —
`safety_pass_uncertified` and `revocation_revision_mismatch`
(`host-context-controller/src/readinessGate.ts:17-18,26-28`) — are wired to
**serving surfaces**: the kubelet readinessProbe *and* the per-request predicate
(`server.ts:297-310`).

Measured on `clerum-dev` over 31 continuous hours (2026-08-26/27):

| | |
|---|---|
| Time not Ready | **31.3 %** |
| Longest single outage | **49.2 min** |
| Outages > 30 s | 1.25 / h |
| Dominant cause | `pass ended without certifying: inventory authority lost` — **1.8 / h** |

That **1.8 / h** figure is **pre-#471 / pre-#467**. Recalibrate against the
[...]

Issue #462, opened 2026-08-26
bug(hcc): doomed NetworkPolicy pass retries forever when duration exceeds GKE ~300s watch recycle (Change D; follow-up to #447)


alfredolopez80:
## What happened?

#447 P2a (maestro PR 3) re-arms silent NetworkPolicy certification sinks so `/ready` recovers in ≤300 s **after the cause is gone**. That PR must stay A/B/B′/C only.

A pass whose duration exceeds a GKE watch recycle (~300 s, `Premature close`) is still killed by **channel identity** (`contextWatchGeneration` / `mcpWatchGeneration`). Live: the 12:48 pass ran ~24 min and died at authority-lost; shorter passes (11 s / 27 s / 55 s) certified.

After P2a that death is **loud and scheduled**: the ladder retries forever (5 / 15 / 30 / 60 / 300 s, then 300 s indefinitely). Ready stays 503; CPU and LIST load go **up**. That is better than today’s silent unbounded 503, but on a large fleet it is a treadmill, not recovery.

Readiness already uses content identity (#382). The in-flight pass still pins channel generation (`k8sClient.ts` ~3051–3052, `hasContextInventoryAuthority` ~1175). Change D would use the revisions the pass already captures (`contextRevision` / `serverRevision` ~3056–3057) without weakening fail-closed on a **real** desired-state change.

## What this issue is not

- Not the P2a watchdog diff. Do not implement Change D in the #447 P2a PR.
- Not #460 write-amp (PRs 4–8 shrink pass duration; they do not change the authority contract).
- Not #447-P2b scoped abort (conditional on the same histograms).

## Version / commit

`origin/dev` @ `92be6464` (HCC cited files byte-identical to deployed `host-context-controller:sha-3d89978`).

## Environment

clerum-dev (`gke_eventfire-491421_us-central1-a_clerum-dev`) only. GKE watch recycle ~300 s. Do not Recreate HCC to “test” this.

## Decide from PR 3 Change C (required)

Named results must distinguish `aborted-authority` vs `aborted-bump` vs `failed` vs `deferred-unsynced` vs `certified`.

- If p50/p95 pass duration **without** storm is already >300 s → this issue becomes the next HCC PR after P2a.
- If `aborted-authority` is storm-only and p50 <300 s after #460 PRs 4–8 → keep this parked.

## Acceptance

- A recycle-only generation bump does not kill a pass whose desired revision is unchanged.
[...]

Issue #478, opened 2026-08-27
bug(hcc): NetworkPolicy orphan sweep runs only at startup — subtractive drift is unbounded until Pod restart


alfredolopez80:
## What happens

The NetworkPolicy orphan sweep runs **only at startup**. Subtractive drift — a lost `DELETED` event
— is therefore **unbounded in time**: it survives until the next Pod restart, which in production
means weeks.

The sweep lives inside `fullReconcile` (`networkPolicyReconciler.ts:757-791`), invoked once during
bootstrap. Nothing re-runs it on a schedule.

## Why neither branch covers this today

**`main` / prod — an accidental (but functional) resync that never covers deletions.**

Context and McpServer watches are reopened **without `resourceVersion`**
(`k8sClient.ts:2623`, `:2442`), so the apiserver replays **synthetic ADDED events for the whole live
inventory**; `applyNetworkPolicy` then does an **unconditional replace** (`utils.ts:169-192`). Every
watch recycle (~3.4 min) re-reconciles the entire live fleet. Those are prod's measured 109.8
`Updated policy`/min — **the write storm *is* its resync**.

But the replay only re-adds **what exists**. Nobody diffs the replay against the cache to detect
**absences**. A `DELETED` lost to a watch gap is never recovered.

**`dev` — the sweep exists but the pass carrying it dies first.**

The sweep is inside the authoritative pass, which dies **1.8/h** on `inventory authority lost`
before completing (measured over 31 h). Result: **dev has the same unbounded subtractive drift as
main, plus 31 % downtime.**

Also relevant: **no managed policy carries `ownerReferences`** (0 occurrences in
`networkPolicyReconciler.ts` on either branch). Kubernetes GC never collects anything — the
reconciler is the only cleanup path.

## Drift taxonomy

| Class | Corrected by the event path? | Bounded? |
|---|---|---|
| Stale allow (binding revoked, MODIFIED lost) | yes — next Context event or watch replay | ✅ minutes |
| Spec drift in L2/L3 | yes — unconditional replace; plus the #299 TTL/2 timer | ✅ minutes |
| Failed write | yes — next event/replay | ✅ minutes |
| **A — orphan from lost DELETE** | **no** | ❌ **until Pod restart** |
| **D — L0/L1 defaults edited** (deny-all, infra egress) | **no** — only `ensureDefaultPolicies` at startup | ❌ **until Pod restart** |
[...]

Issue #464, opened 2026-08-26
bug(hcc): McpServer initial-convergence core silent return (k8sClient.ts:2988; same class as #447, not on /ready)


alfredolopez80:
## What happened?

#447 P2a re-arms **NetworkPolicy** silent sinks (`:3046`, `:3108`). The McpServer initial-convergence core has the same swallow class and is **out of that PR**.

`runInitialMcpServerConvergenceCore` (`host-context-controller/src/k8sClient.ts` ~2988):

```ts
if (!inventoryAuthoritative()) return
initialConvergenceLastSuccessTimestampSeconds.set({ lane: 'McpServer' }, Date.now() / 1000)
this.clearInitialConvergenceRetry('McpServer')
```

A bare `return` after a lost-authority / unsynced check never reaches `scheduleInitialConvergenceRetry('McpServer')`. The catch (~2991) is the only re-arm, same bug #447 fixes for NetworkPolicy.

`/ready` does **not** currently gate on the McpServer lane timestamp (safety certificate is content-identity from #382). So this does not explain the 18–35 min Ready 503 windows. An 11+ MCP e2e or a “McpServer initial convergence never completed” log can look like a **regression of the #447 P2a PR**. It is not.

## What this issue is not

- Not the #447 P2a diff. Do not expand `:2988` in that PR.
- Not a `/ready` bug unless readiness later starts gating this timestamp (that would make this blocking).

## Version / commit

`origin/dev` @ `92be6464`.

## Environment

HCC McpServer lane. Fleet 11+ MCP is the likely false-alarm surface after P2a.

## Acceptance

- Same A/B pattern as #447, **lane `McpServer` only**: warn/error + swallow counter + existing ladder. Do not expand the McpServer core.
- Tests for this lane must not reuse T12–T14 IDs from #447 P2a.
- `/ready` contract unchanged unless a separate issue moves the gate.

Related: #447 (sibling class, other lane). Maestro §7 R8 / PR 12.

alfredolopez80:
## Priority raised — this is the only silent sink left without a schedule

An adversarial audit of the #460 delivery plan (2026-08-27) flagged this issue as the one gap that
belongs in the near-term wave rather than the backlog.

**Confirmed still open**: the McpServer lane keeps the bare sink at `k8sClient.ts:3011` — `return`
with no log, no metric, no re-arm. The `scheduleInitialConvergenceRetry('McpServer')` at `:3016`
[...]

PR #479, opened 2026-08-27
fix(hcc): keep phase-2 certification off the readiness probe (#477)

alfredolopez80:
## Summary
- Split kubelet `/ready` from the per-request authority gate: the probe now uses watch freshness only (clauses 1–4), so an uncertified NetworkPolicy safety pass no longer evicts the Pod from Service endpoints.
- The per-request predicate still uses all six clauses (`isReadinessInventoryAuthoritative`). Omitting the new probe function fails closed to that same 6-clause gate.
- Guardian merge gate is Vitest HTTP spies (A1-T01–T12, T14). Playwright, Minikube “Pod Ready”, and T2 are destination-only and are not this PR’s evidence.

## Test plan
- [x] `host-context-controller` `npm test` (1759) + `npx tsc --noEmit`
- [x] Review fan-out + fix findings
- [ ] Do not merge before #462 (Change D). Sequence: B → A1 → C.

### DIFF
--- a/docs/architecture/hcc-readiness-flows-diagram.svg
+++ b/docs/architecture/hcc-readiness-flows-diagram.svg
@@ -29,9 +29,9 @@
 <text x="640.0" y="178" text-anchor="middle" font-size="11" fill="#475569">· anchored to desired revision</text>
 <rect x="779.5" y="104" width="214" height="92" rx="11" fill="#dbeafe" stroke="#1d4ed8" stroke-width="1.8" filter="url(#sh)"/>
 <text x="886.5" y="127" text-anchor="middle" font-size="14" font-weight="700" fill="#0f172a">/ready = 200</text>
-<text x="886.5" y="146" text-anchor="middle" font-size="11" fill="#475569">only if inventory certified</text>
-<text x="886.5" y="162" text-anchor="middle" font-size="11" fill="#475569">and no stale allow</text>
-<text x="886.5" y="178" text-anchor="middle" font-size="11" fill="#475569">survives</text>
+<text x="886.5" y="146" text-anchor="middle" font-size="11" fill="#475569">watch freshness only</text>
+<text x="886.5" y="162" text-anchor="middle" font-size="11" fill="#475569">phase-2 stays on the</text>
+<text x="886.5" y="178" text-anchor="middle" font-size="11" fill="#475569">per-request data path</text>
 <rect x="1026.0" y="104" width="214" height="92" rx="11" fill="#eff6ff" stroke="#2563eb" stroke-width="1.8" filter="url(#sh)"/>
 <text x="1133.0" y="127" text-anchor="middle" font-size="14" font-weight="700" fill="#0f172a">Additive convergence</text>
 <text x="1133.0" y="146" text-anchor="middle" font-size="11" fill="#475569">fleet converges AFTER</text>
@@ -52,7 +52,7 @@
 <text x="40" y="275" font-size="13" font-weight="700" fill="#b91c1c">2 · Fail-closed fences (R3 · R4 · J3)</text>
 <rect x="40.0" y="284" width="380.0" height="76" rx="11" fill="#fef2f2" stroke="#dc2626" stroke-width="1.8" filter="url(#sh)"/>
 <text x="230.0" y="307" text-anchor="middle" font-size="14" font-weight="700" fill="#0f172a">Readiness gate — default () =&gt; false</text>
-<text x="230.0" y="326" text-anchor="middle" font-size="11" fill="#475569">/ready = ready &amp;&amp; providerAuthoritativeFn()</text>
+<text x="230.0" y="326" text-anchor="middle" font-size="11" fill="#475569">/ready = ready &amp;&amp; probeAuthoritativeFn()</text>
 <text x="230.0" y="342" text-anchor="middle" font-size="11" fill="#475569">throw → 503 · never a wrong 200</text>
 <rect x="450.0" y="284" width="380.0" height="76" rx="11" fill="#fef2f2" stroke="#dc2626" stroke-width="1.8" filter="url(#sh)"/>
 <text x="640.0" y="307" text-anchor="middle" font-size="14" font-weight="700" fill="#0f172a">G1 — deny without inventory-authority</text>
@@ -76,7 +76,7 @@
 <text x="693.0" y="456" text-anchor="middle" font-size="11" fill="#475569">same protocol/port/ownership/cidrs</text>
 <text x="693.0" y="472" text-anchor="middle" font-size="11" fill="#475569">· HCC stays Ready (modulo cidr)</text>
 <rect x="540" y="496" width="306" height="60" rx="11" fill="#fef2f2" stroke="#dc2626" stroke-width="1.8" filter="url(#sh)"/>
-<text x="693.0" y="519" text-anchor="middle" font-size="14" font-weight="700" fill="#0f172a">REVOKE before readiness</text>
+<text x="693.0" y="519" text-anchor="middle" font-size="14" font-weight="700" fill="#0f172a">REVOKE before data-path certify</text>
 <text x="693.0" y="538" text-anchor="middle" font-size="11" fill="#475569">drift TCP→UDP · port · ownership · zero-cidr</text>
 <text x="693.0" y="554" text-anchor="middle" font-size="11" fill="#475569">→ unprovable → cleanup fail-closed</text>
 <path d="M 472,470 H 506.0 V 444 H 540" fill="none" stroke="#059669" stroke-width="2.6" marker-end="url(#aGreen)" stroke-linejoin="round"/>
--- a/docs/architecture/hcc-readiness-flows.md
+++ b/docs/architecture/hcc-readiness-flows.md
@@ -4,7 +4,7 @@
 > `host-context-controller` (HCC). Addresses the "no design/decision doc for the
 > 500+-line external-egress coordinator and the certification barrier" review gap.
 
-![HCC readiness flows resolved by #382](./hcc-readiness-flows-diagram.png)
+![HCC readiness flows — probe vs data-path lanes after #477 A1](./hcc-readiness-flows-diagram.svg)
 
 ## 1. Readiness lifecycle — decoupled from initial fleet convergence
 
@@ -12,8 +12,9 @@ Before this work, `McpServerWatcher.start()` awaited full initial fleet
 reconciliation, so a large MCP/Context/Host fleet could keep `provider.start()`
 and `/ready` unavailable until every individual reconcile finished.
 
-Now readiness is gated on an **authoritative, fail-closed safety boundary**, not on
-full convergence:
+Readiness is now two lanes. Watch freshness (clauses 1–4) stays fail-closed on
+both. Phase-2 certification stays fail-closed on the **data path** and is off
+the kubelet probe (#477 A1):
 
 1. **HCC start** — four watch lanes (Context ingress, mcp-host egress, rpc-proxy
    egress, external egress) start. A transient LIST/WATCH failure on any lane enters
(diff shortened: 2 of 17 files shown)


alfredolopez80:
@claude please re-review and audit this PR for close.

This is **only** #479 / #477 A1 (probe phase-2 cut). Do not expand the review to #462 or #478.

Head now includes the review-fix commit `46c350fd` on top of `d653c91e`:

- `/ready` stays watch freshness (clauses 1–4). Per-request keeps the 6-clause gate.
- Fence integration tests now adjudicate recovery on the API lane (`apiStatus` 200 after a clean pass), not a tautological `/ready` 200.
- Readiness logs include `lane: probe | request`. `GET /` and the readiness design note/SVG match the two-lane contract.
- `e2e-hcc-mcp-context-readiness.sh` no longer treats `/ready` 200 + stale policy as fatal `rc=2`; that window is retryable. Success still requires `/ready` 200 **and** the stale allow gone. Hermetic pins in `test-hcc-watch-recovery-gate.sh` were updated.
- T0 on this worktree: HCC `tsc --noEmit` + `npm test` (1763) green. Hermetic gate script green. T1/T2 do not apply.

Please audit and close the review if the invariant holds. Do not merge — oleada order is B → A1 → C.

jozer-rami:
## Review: 2 blockers, 1 major

Reviewed at `46c350fd` in a clean worktree. CI 33/33 green, `tsc --noEmit` clean, 1763 tests pass.

The code is carefully fail-closed and I could not break it: 6/6 mutations were caught, including the two that matter most. Omitting the 10th constructor argument falls back to the 6-clause gate rather than open (27 tests fail if that is inverted), and pointing the data path at the probe lane fails 49 tests. `A1-T02` asserting that `isReadinessInventoryAuthoritative` is *never invoked* from the probe path is the right structural guard. What follows is about what is missing from the diff, not what is in it.

---

### Blocker 1: the declared prerequisite is not merged

The test plan carries an unchecked box: *"Do not merge before #462 (Change D). Sequence: B -> A1 -> C."*

#462 is OPEN, and its implementation (#480) is also OPEN.

This is not bookkeeping. #462 is what makes certification actually succeed. Until it lands, dev decertifies ~1.8/h, and this PR removes the loudest detector of that while the underlying failure is still live.

### Blocker 2: #477 steps 1-3 are absent, and #477 forbids exactly this shape

#477 scopes the work as *"ONE indivisible diff, in this order"*:

1. metric `uncertified_window_seconds`
2. `degraded` field/header on the response
3. alert on the metric, routed to a channel with a human owner
4. cut phase 2 out of the probe **and** out of the per-request predicate

Then, explicitly: *"**The order is mandatory.** [...] A deployed version with step 4 but without steps 1-3 must never exist. Atomic revert."*

Verified against this branch:

- `host-context-controller/src/metrics.ts` is untouched by the diff
- `grep -rn "uncertified_window"` across the repo returns nothing
- no `degraded` field or header is added to any data-path response

This ships step 4, and only half of step 4, with none of 1-3.

Concrete consequence: an HCC that never certifies now presents as a healthy Ready pod. The 31.3%-not-Ready signal that made #477 diagnosable in the first place disappears with nothing in its place. Per the issue itself: *"today the downtime is the detector."*

### Major: the half-cut breaks mcp-host's cold-start retry ladder

`mcp-host/src/mcp/authoritativeFleet.ts:24-36`. `listServersForHost()` sits inside `if (ready)` with no `try`/`catch`, so a 503 from the inventory route exits the loop on the first pass.

Measured against the real code:

| | `healthCheck()` calls | `listServersForHost()` | sleeps | outcome |
|---|---|---|---|---|
| before this PR (`/ready` 503) | 5 | never called | 4 x 2s | throws after the ladder |
| after this PR (`/ready` 200) | 1 | 1, returns 503 | **0** | throws on attempt 1 |

The throw is caught upstream (`mcpBackgroundInit.ts:28` at boot, `main.ts:992` in the poll loop), so nothing crashes. The cost is that cold-start fleet publication during a transient uncertified window falls from <=8s of tight retry to one `CLERUM_CONTEXT_MAPPER_POLL_INTERVAL`, default 30000ms (`mcp-host/src/config.ts:579`).

This follows directly from splitting step 4. In #477's design the sequence is `healthCheck()` true -> inventory answers 200 with a degraded marker -> fleet publishes. Probe-only gives `healthCheck()` true -> inventory 503 -> nothing publishes. The PR redefines a contract mcp-host keys on, touches zero mcp-host files, and adds no cross-service test.

Two ways to close it:

- land the per-request half in this PR, as #477 specifies, or
- make the ladder treat a 503 from `listServersForHost()` the same way it treats `ready === false`, so the retry still happens

---

### Suggested order

Merge #480 first, then either fold #477 steps 1-3 into this PR or land them immediately before it. If the probe-only split is the preferred shape, the mcp-host ladder fix should ship with it, otherwise the change trades kubelet eviction for a 30s cold-start hole.

alfredolopez80:
@jozer-rami thank you — the measurements are right. The disagreement is scope, not the code.

## What this PR is

**#479** implements **one surface** of [#477](https://github.com/evenfire-ai/evenfire/issues/477): take the phase-2 clauses (`safety_pass_uncertified`, `revocation_revision_mismatch`) off **`/ready` only**. The per-request gate stays fail-closed on all six clauses. HCC `readinessGate.ts` / `server.ts` / `main.ts` only.

[#477](https://github.com/evenfire-ai/evenfire/issues/477)'s body is updated to match. The original "one indivisible diff" paragraph is superseded **for the probe** and **kept** for the later per-request cut on that same issue.

## Why #477 ships as two changes, not one

The original #477 text asked for one indivisible diff: metric → `degraded` → alert → then cut phase 2 from **both** `/ready` **and** the per-request gate. That rule existed for one reason: *today the downtime is the detector*, so a deployed cut without a replacement detector would turn a loud failure into a silent one.

That reason applies to **one** of the two surfaces, not both.

- The **probe** is not a detector anyone should keep. No data route depends on it. The 503 revokes nothing (`server.ts` has no mutation path; mcp-host keeps a published fleet on 503 and only drops it on 401/403). Removing phase 2 from `/ready` opens nothing. After this PR, discovery still answers **503** while uncertified, and the #467 logs (`pass ended without certifying`) stay. **No new silence.** The 31 % NotReady and the external rollout-undo (#391) go away.
- The **per-request** gate *is* the remaining loud detector. Removing phase 2 there *would* create silence. That cut still needs the metric, the `degraded` marker, and an alert routed to a human **in the same change**, in that order. That remaining work stays on **#477** as a **later PR**. It is blocked on observability that does not exist today: nothing scrapes HCC `/metrics`, and the project has **zero** notification channels. Folding it into #479 would either ship the silent cut without a detector, or block the probe cut on an alert we cannot fire.

So: **this PR = probe only.** Still on #477, later: `uncertified_window_seconds` + `degraded` + alert, then the per-request cut — one change, that order. No metric, no `degraded`, no alert, no mcp-host change in #479.

## Related PRs — what each issue is, and the merge order

These three PRs are slices of the HCC mitigation whose principal issue is [**#460**](https://github.com/evenfire-ai/evenfire/issues/460) (no-op write-storm pinning `/ready` at 503). Merge and deploy in this order. Each line is "this PR closes / implements that issue."

1. **[#480](https://github.com/evenfire-ai/evenfire/pull/480)** → [**#462**](https://github.com/evenfire-ai/evenfire/issues/462)  
   Abort the in-flight NetworkPolicy safety pass on **content** identity, not watch-channel generation. GKE recycles watches about every 300 s; a pass longer than that dies even when desired state did not move. This is why certification fails today. **First**, because #479 only changes what a failed certify *costs*, and #481's sweep dies with a pass that cannot finish.

2. **[#479](https://github.com/evenfire-ai/evenfire/pull/479)** (this PR) → probe half of [**#477**](https://github.com/evenfire-ai/evenfire/issues/477)  
   Phase-2 clauses off `/ready` so kubelet stops evicting the Pod when the pass is uncertified. Discovery / credential stay 503. Does **not** merge before #480.

3. **[#481](https://github.com/evenfire-ai/evenfire/pull/481)** → [**#478**](https://github.com/evenfire-ai/evenfire/issues/478)  
   Coordinated NetworkPolicy resync timer (default **off**) plus the orphan sweep, so subtractive drift is not "until the next Pod restart." **Last**, so it inherits a pass that can survive a recycle.

Not in this merge set:

- The **rest of #477** (metric, `degraded`, alert, then the per-request cut) — later PR on the same issue, after a detector exists in the environment that will run it.
- [**#464**](https://github.com/evenfire-ai/evenfire/issues/464) — McpServer initial-convergence silent `return` (no log, no metric, no retry). Sibling of #447; not a #477/#479 file.

## Blocker 1

Agreed. This PR does not merge before #462 / #480. #480 is what makes certification actually succeed; this PR only changes what a failed certify costs.

## Major (mcp-host ladder)

Also measured correctly. After this PR, `healthCheck()` sees `/ready` 200 during an uncertified window, then `listServersForHost()` still 503s and leaves the 2 s ladder on attempt 1. The next try is the poll interval.

That is the contract of the probe-only split, not a missing file in this PR:

- `healthCheck()` true / inventory 503 is what "probe stops lying, discovery stays loud" *is*.
- Boot already catches the throw (`mcpBackgroundInit`) and starts the poll. The 503 never revoked a fleet. The cost is cold-start latency, not a safety hole.
- Teaching mcp-host to retry an inventory 503 like `/ready` 503 would paper over the split. We are not changing mcp-host in #479.

The **close** of that hole is a **later phase** of the same HCC mitigation whose principal issue is [**#460**](https://github.com/evenfire-ai/evenfire/issues/460) (the write-storm that pinned `/ready` at 503). #462 / #477 / #478 are slices of that mitigation. The later #477 change — inventory answers 200 with a `degraded` marker, then the per-request cut — is what restores `healthCheck()` true → fleet publishes, without a mcp-host workaround. It does not ship in this PR, and it does not ship before the detector that #477 still requires for that cut.

#479 stays probe-only. Please re-review against the updated #477 body.

jozer-rami (approved):
Approving at `46c350fd`. CI 33/33, `tsc --noEmit` clean, 1763 tests.

Reconciling my earlier review, since I raised two blockers.

## Blocker 2 — withdrawn

I blocked on #477's *"ONE indivisible diff [...] a deployed version with step 4 but without steps 1-3 must never exist."* That rule has been superseded: #477's body was updated at 12:46Z and now opens with *"the original 'one indivisible diff' is superseded."*

I did not take the supersession on trust, because an issue body amended mid-review is not independent authority. I verified the premise myself: `grep -rn "ServiceMonitor\|PodMonitor\|prometheus.io/scrape" deploy/` returns nothing. HCC's `/metrics` is exposed and scraped by nobody.

That settles it. Step 3 is *"alert on the metric, routed to a channel with a human owner."* With no alerting pipeline in the project, step 3 is not implementable, and gating step 4 on it would block the work indefinitely rather than protect anything. The argument that the rationale (*"today the downtime is the detector"*) applies to the per-request surface and not the probe is correct: no data route depends on the probe, the 503 revokes nothing, and discovery still answers 503 while uncertified.

## Blocker 1 — stands as a merge gate, not a code objection

Agreed on both sides: do not merge before #480. #480 is OPEN, so **this approval is not a green light to merge**. Order remains #480 → #479 → #481.

## Major (mcp-host ladder) — accepted consequence, on record

Restating so it is traceable rather than reopening it. After this PR, during an uncertified window `healthCheck()` sees `/ready` 200, then `listServersForHost()` still 503s and escapes `runAuthoritativeMcpInitialization` on attempt 1, bypassing the 5x2s ladder. Boot catches it (`mcpBackgroundInit.ts:28`) and the poll loop retries, so the cost is cold-start latency of up to one `CLERUM_CONTEXT_MAPPER_POLL_INTERVAL` (default 30000ms), not a safety hole.

Accepting that as the contract of the probe-only split. The later #477 phase (inventory answers 200 with a `degraded` marker, then the per-request cut) is what closes it without a mcp-host workaround. Worth carrying into that PR's acceptance criteria so it does not get lost between the two halves.

## What I verified in the code

6 of 6 mutations caught. The two that matter: omitting the 10th constructor argument falls back to the 6-clause gate rather than open (27 tests fail if inverted), and pointing the data path at the probe lane fails 49 tests. `A1-T02` asserting `isReadinessInventoryAuthoritative` is never invoked from the probe path is a structural guard, not a value check, which is the right shape here. Clauses 1-4 stay fail-closed on both lanes.

The e2e downgrade from fatal `rc=2` to retryable `rc=1` is correct for the new `/ready` semantics. Flagging one gap for later rather than for this PR: the e2e never probes `/api/v2/hosts/self/mcpservers`, so no e2e assertion now binds any serving surface to revocation. #477's regression test specifies both halves; only the `/ready` half is covered. The Vitest coverage is strong, so this is about e2e depth.

PR #480, opened 2026-08-27
fix(hcc): abort the NetworkPolicy pass on content identity, not watch generation (#462)

alfredolopez80:
## Summary
- Abort the in-flight NetworkPolicy safety pass on **content** identity (`contextDesiredRevision` / `mcpServerDesiredRevision`), not watch-channel generation. GKE watch recycles no longer kill a pass whose desired state is unchanged.
- Leave `hasContextInventoryAuthority` / `hasMcpServerInventoryAuthority` untouched (delete-confirm / watch effects). Both Context and McpServer sides of the in-flight closures change together. `cacheSynced` stays required.
- Tests G1–G9 pin recycle-only survival, real desired-state abort, and the existing lease/`cacheSynced` fail-closed path (G5).

Closes #462.

## Merge order
Oleada 1: **merge/deploy B (this PR) → A1 #479 → C (netpol resync)**. Do not merge this after A1. Do not merge C first.

## Test plan
- [x] HCC Vitest covering G1–G9 / content-identity pass abort
- [x] T0: `host-context-controller` `npx tsc --noEmit` + `npm test`
- [ ] T1/T2: not applicable (no Real PG; not a runtime/Minikube gate)

PR #445, opened 2026-08-25
fix(upload): clarify post-401 Upload v2 guidance

josue-tech-94:
## Summary

- Follow-up to #401: https://github.com/evenfire-ai/evenfire/pull/401
- Improve Upload v2 mixed-version Desktop guidance and writer-resolved Control UI guidance.
- Harden deferred Upload v2 E2E fixture cleanup and test-maintainability coverage.
- Preserve the policy and protocol ownership established by PR #311 and PR #401.

## Context

PR #401 completed runtime-configurable Upload v2 product policy and intentionally deferred these
non-blocking follow-ups. This PR does not change GFSC product-policy ownership or the established
Upload v2 boundaries: 200 MiB is the default and missing-field compatibility product policy,
1 GiB is the hard whole-file protocol/client safety ceiling, and 16 MiB is the part/request and
legacy-upload boundary.

## Changes

- Desktop distinguishes a writer value beyond its fixed 1 GiB protocol ceiling from generic malformed
  capability data, without enabling legacy fallback.
- Desktop and Control UI consistently present resolved writer, compatibility, and legacy limits in
  binary units.
- Control UI pre-selection copy identifies 1 GiB as a local safety ceiling and explains that the
  writer resolves the product policy during upload.
- Upload v2 disk fixtures use a descriptor-bound lease lifecycle. The lease owns the opened
  FileHandle immediately after open succeeds, before test hooks or wrapper factories can throw.
  Cleanup is single-flight per fixture, neutralizes only the retained file handle, closes it once,
  and performs no pathname-based file or directory deletion. Replacement paths, symlinks, and
  fabricated fixture objects cannot redirect cleanup. Small zero-byte OS-temp artifacts may remain.
- The descriptor-bound fixture regression suite runs in the canonical public Node 24 CI gate with a
  non-zero executed-test guard.
- Product-policy tests use below-protocol limits. The hard 1 GiB protocol boundary is proven at the
  protocol validator authority rather than through runtime product-policy admission.

### Passed locally

- `npm test --prefix tests/e2e -- gfsUploadV2Fixtures.test.ts` — 22 passed.
- `E2E_VITEST_SUITE_GROUP=node-unit bash scripts/e2e/run-vitest-e2e.sh` — 22 passed through the
  canonical runner and non-zero guard.
- `bash -n scripts/e2e/run-vitest-e2e.sh scripts/tests/test-e2e-vitest-node-unit-gate.sh`.
[...]

Issue #465, opened 2026-08-26
bug(hcc): SharedFileSystem and GlobalFileSystem initial convergence have no retry ladder


alfredolopez80:
## What happened?

NetworkPolicy (and, after #447 P2a, its silent sinks) use `scheduleInitialConvergenceRetry` (`INITIAL_CONVERGENCE_RETRY_DELAYS_MS` = 5 / 15 / 30 / 60 / 300 s, then 300 s forever).

SFS and GFS initial cores only log on failure and return:

- `runInitialSharedFileSystemConvergenceCore` (`k8sClient.ts` ~3167–3175): catch → `console.error`, **no** `scheduleInitialConvergenceRetry('SharedFileSystem')`.
- `runInitialGlobalFileSystemConvergenceCore` (~3184–3192): same for GFS.

Context watch recovery (~1486–1488) does kick `runInitialSharedFileSystemConvergence` again, so SFS is not totally orphaned — but there is no bounded ladder if that recovery does not run or the core fails for another reason. Third silent lane after NetworkPolicy (#447) and McpServer (sibling issue).

## What this issue is not

- Not #447 P2a. Do not add SFS/GFS re-arm to the NetworkPolicy watchdog PR.
- Not assuming SFS/GFS share the NetworkPolicy core or `/ready` certificate.

## Version / commit

`origin/dev` @ `92be6464`.

## Environment

HCC SFS/GFS lanes. Inventory call sites at implement time; do not assume they match the NetPol core.

## Acceptance

- Failed or swallowed SFS/GFS initial convergence schedules the existing ladder (or an equivalent noisy retry) per lane.
- Success clears the lane timer (same contract as NetPol `clearInitialConvergenceRetry`).
- No change to #447 P2a files beyond what that PR already owns, unless this lands much later on the same functions.

Related: #447 (do not mix), McpServer sibling (maestro R8). Maestro §7 R9 / PR 13.

PR #481, opened 2026-08-27
fix(hcc): resync NetworkPolicies through coordinated convergence (#478)

### LINKED ISSUE #483: test(hcc): mutation-survivor gaps — cap-tripped external-egress lane and ensureDefaults reset
Two test gaps in PR #481, both **proven by mutation** during an independent multi-agent review: the suite stays fully green with the guard removed. Neither blocks the merge — the production code is correct today — but nothing stops a future refactor from breaking either one silently.

## Gap A — no test exercises a tripped orphan-delete cap on the `external-egress` lane

**Mutation E**: remove `!capReason` from the external-egress loop gate (`host-context-controller/src/networkPolicyReconciler.ts:~1565`) ⇒ **188/188 tests still green**.

In the three reconciler test files, `ext-egress` appears exactly once — `networkPolicyReconciler.orphanSweep.test.ts:353`, inside S3, which runs **under** the cap and therefore deletes. The cap-tripped tests (M4 `:251`, its sibling `:291`, S2 `:311`) have no external orphans at all. The `cap × external-egress` gate is never exercised.

Why it matters: the cap is the guard against mass deletion of NetworkPolicies when the cache is in a bad state. A regression that re-enabled external-egress deletes under a tripped cap — exactly the failure the cap exists to prevent — **would pass the suite silently**.

Verified as a real gap, not indirect coverage: the mutation result is structurally consistent with the fixtures.

**Acceptance**: a test with a tripped absolute cap (N+1 context orphans) **plus** at least one `ext-egress-*` orphan for a non-desired server asserts that (a) `deleteNamespacedNetworkPolicy` is never called for the server namespace, (b) `clerum_hcc_netpol_orphan_sweep_capped_total{reason="absolute"}` increments, (c) the pass still certifies. Removing `!capReason` must turn the suite **red**.

## Gap B — the consume-and-reset of `netPolConvergenceEnsureDefaults` is unpinned

**Mutation D**: remove the reset at `host-context-controller/src/k8sClient.ts:3146` ⇒ **280/280 tests still green**.

The flag is **instance state**, not run state: the periodic tick sets it (`:3054`), and the core reads-and-resets it ~90 lines later (`:3145-3146`). The documented contract is that startup and event-driven entries stay `false` — only the timer path gets `ensureDefaults: true`.

`k8sClient.test.ts:11331-11353` (M1) asserts `call[0]` is `false` and the tick is `true`, but **never asserts a third call**. With a sticky flag, every subsequent event-driven pass would inherit `ensureDefaults: true`.

The failure direction is benign — it would repair more often than intended, and `ensureDefaultPolicies` is idempotent — hence Medium rather than High. But the mechanism is precisely the kind that a refactor breaks without anyone noticing: a new early-return added between the read (`:3145`) and the `await` (`:3147`) would leak the flag.

Independently confirmed correct **today** by two reviewers who reproduced the interleavings; the gap is in the test harness, not the code.

**Acceptance**: after a timer tick completes, trigger an event-driven pass and assert `ensureDefaults: false` on the third `fullReconcile` call. Removing the reset at `:3146` must turn the suite **red**.

## Notes

- Both gaps were found by mutation testing, so the acceptance criteria above are stated as "the mutation must turn the suite red" — a test that passes with and without the guard proves nothing.
- Verified not to overlap with deferred work: `SafetyInventoryLane` (the sweep's `external-egress`) and `InitialConvergenceLane` (the `McpServer` sink tracked in #464) are distinct types. #481's diff does not touch `k8sClient.ts:3011`.

Related: #481, #478.

alfredolopez80:
## Summary
- Add a coordinated NetworkPolicy resync timer whose body is `void this.runInitialNetworkPolicyConvergence({ ensureDefaults: true })` (single-flight, defer-unsynced, authority fences). Startup/event entry stays `ensureDefaults: false` because `start()` already ran `ensureDefaultPolicies`. Naked `netPolReconciler.fullReconcile` is forbidden on this timer. The tick covers #478 class A (orphan sweep) and class D (L0/L1 defaults).
- A timer tick while a NetworkPolicy pass is in flight is **skipped** (`clerum_hcc_netpol_resync_ticks_skipped_total{reason="pass-in-flight"}`) and does not set `trailingRequested`. Event-driven coalescing is unchanged. Without this guard, pass duration ≥ period becomes a treadmill on the shared `contextReconciliationQueues` (worst measured pass 17.1 min / 1026 s; infra rollout budget 1500 s).
- `CONTEXT_MAPPER_NETPOL_RESYNC_SEC` defaults to **0** (off). Do not set it in any public or infra manifest in this PR. A non-canonical value fails loud at load.
- **Enable (issue #478, after merge + deploy):** with the skip guard the chosen value is **`1500` s**. Without the guard the floor is **`3600` s**. `300` is not a valid enable while passes last minutes. Criterion: treadmill ⇔ pass duration ≥ period. Preconditions: image deployed, orphan census read (cap would not trip on the first sweep), several days of `clerum_hcc_initial_convergence_pass_duration_seconds` p99/max ≪ T, then `evenfire-infra` **dev first**. Class D of #478 HARD REQUIREMENT 1 is covered by that enable, not by this merge — the tick is the only call site that passes `ensureDefaults: true`.
- Mass-delete cap: if orphans exceed 10 or 20% of the managed fleet, **do not delete** and increment `clerum_hcc_netpol_orphan_sweep_capped_total`; the pass **still certifies** (#478: alert and do not delete). Absolute is checked first, so at a large fleet the percent rule does not bind. Successful deletes increment `clerum_hcc_netpol_orphans_deleted_total`.
- Job 4.1-D: `scripts/ops/hcc-netpol-orphan-census.sh` filters `policy-type ∈ {context-allow, rpc-proxy-egress, external-egress}` plus the reserved-name `repairable` class. `live_cap_would_trip=UNSET` when both cap keys are omitted. Zero listed managed policies is `INCONCLUSIVE_EMPTY`, never `CLEAN`.
- Auditor v2: explicit high-tail buckets `[1, 5, 15, 60, 120, 300, 600, 1200, 1800, 3600]` on the two pass-duration histograms. McpServer lane left on channel identity (#464).

Refs #478.
Closes #483.

## Merge order
**#480, #479, and #445 are on `dev`.** This PR is next. Do not merge without a human Approve on the post-sync HEAD.

## Test plan
- [x] HCC Vitest for timer wiring, default-off, in-flight tick skip, cap, orphan sweep, and 300s histogram buckets
- [x] T0 after the in-flight tick skip: `host-context-controller` `npx tsc --noEmit` + `npm test` (1804)
- [x] Follow-up T0 on `00b0d3e8`: `tsc --noEmit` + 318 tests in the five touched suites; cap×external and ensureDefaults-reset mutations go red
- [ ] T1/T2: not applicable (timer default 0; do not enable sweep as a merge gate)


alfredolopez80:
@claude please re-review and audit this PR for close.

This is **only** #481 / #478 C (resync timer + orphan sweep). Do not expand to #462 or #477.

Head now includes the review-fix on top of `65d51472`:

- Timer body stays `void this.runInitialNetworkPolicyConvergence()`. Default `CONTEXT_MAPPER_NETPOL_RESYNC_SEC=0`. Merge does not enable the tick.
- Mass-delete cap and census `policy-type` filter are unchanged.
- Auditor v2: the two pass-duration histograms now pass explicit buckets including the 300s GKE recycle cut. No new Prometheus labels. McpServer convergence lane is untouched (#464 / oleada 2).
- Enable (`interval > 0`) is written as a separate ops decision on issue #478: freshness gate **or** a controller-runtime 10h-scale period before any non-zero value. 300s is not the enable value.
- T0 on this worktree: HCC `tsc --noEmit` + `npm test` (1759) green. T1/T2 do not apply.

Please audit and close the review if the invariant holds. Do not merge — oleada order is B #480 → A1 #479 → C #481.

alfredolopez80:
The three non-blocking ops notes from the re-review are documented; no product behavior change.

1. **Census vs compiled defaults.** `scripts/ops/hcc-netpol-orphan-census.sh` still prints `live_cap_would_trip` from Deployment env only (`UNSET` → `none`). It now also prints `compiled_default_absolute=10` / `compiled_default_percent=20` and `controller_cap_would_trip` (live env, else those compiled defaults) so `live_cap_would_trip=none` cannot be read as "the running controller will sweep."

2. **Cap on every `fullReconcile`.** README + `evaluateNetPolOrphanSweepCap` JSDoc now say the cap governs startup, event-driven, and timer passes. A bulk offboarding above 10 / 20% withholds certification until an operator raises the env.

3. **Default-0 boot warn.** Left as `console.warn` (same pattern as Host/SFS/GFS/LlmHook). README notes it is expected in the merge state and names the self-heal gap.

Invariant unchanged: timer default 0, coordinated body, mass-delete cap, `policy-type` census filter. Do not merge before #480 and #479.

jozer-rami:
## Review: 1 major, 3 minors

Reviewed at `31f06b95`. Re-ran T0: `npx tsc --noEmit` clean, **1759/1759** pass, CI 33/33.

**HARD REQUIREMENT 2 is honoured and actually pinned.** The timer body is `void this.runInitialNetworkPolicyConvergence()` (`k8sClient.ts:2858`), and swapping it for a naked `netPolReconciler.fullReconcile` fails 3 tests. That was #478's top non-negotiable, and it is enforced by the suite rather than only asserted in the summary. 4 of 4 cap mutations were caught: cap never fires (3 tests), `>` becomes `>=` (1), removing the `percentThreshold >= 1` inertness guard (14), naked fullReconcile (3). The inertness guard is a genuine subtlety and S4 pins it. Timer clears on stop (`:4539`) and warns honestly when disabled (`:2863`). The 3600s histogram tail is exactly what #462's duration criterion needs.

---

### Major: the cap ships enabled by default, and a trip is a permanent certification wedge

The summary says `CONTEXT_MAPPER_NETPOL_RESYNC_SEC` defaults to 0 and that enabling is *"an ops step [...] not part of merge"*. That is true of the **timer**. It is not true of the **cap**, which is the part that carries risk.

`evaluateNetPolOrphanSweepCap` runs unconditionally inside `fullReconcile` (`networkPolicyReconciler.ts:1477`), and `fullReconcile` runs on the startup convergence pass (`k8sClient.ts:3122`) whether or not the timer is on. The cap ships on, in every cluster, at merge.

The throw at `:1477` exits before `onAuthoritativeRevocationComplete` at `:1666`, so the pass never certifies. Because the same throw also prevents the deletes, the orphan count never falls, so every retry the ladder schedules recounts the identical set and refuses identically.

Verified: three consecutive `fullReconcile` calls with 21 orphans all reject, `deleteNamespacedNetworkPolicy` is never called, `complete` is never called. Not self-clearing.

**Blast radius, measured read-only on clerum-dev.** 377 managed policies, 25 Contexts, 19 of them carrying policies. **0 orphans today**, so this would not wedge dev on merge. But each Context contributes 3x its server count across the three context lanes, and the distribution is:

| policies from one Context | how many Contexts | exceeds cap 10? |
|---:|---:|---|
| 90 | 1 | yes |
| 31 | 1 | yes |
| 21 (`context1`) | 1 | yes |
| 15 to 18 | 4 | yes |
| 12 | 1 | yes |
| 9 or fewer | 11 | no |

**8 of the 19 individually exceed the absolute cap of 10.** So the exact scenario #478 exists to fix, one lost Context `DELETED`, trips the cap for 8 of them. The outcome is that the orphan allows stay live (no better than today) **and** certification stops (worse than today). Stacked on #479, `/ready` stays 200 while the data path 503s indefinitely.

This also cuts against #478's own reasoning, which quotes controller-runtime approvingly: *"Kubernetes accepts drift between resyncs [...] Certification is a property that converges, not a precondition for serving."* And the issue asked for *"alert and do not delete"*, not "and do not certify". The certification withholding is this PR's addition, and it is what turns a bounded alert into an unbounded outage.

This is a deliberate, tested choice (`M4` asserts `expect(complete).not.toHaveBeenCalled()`), so it wants a decision rather than a patch. Options, best first:

1. Refuse the **deletes** and record the metric, but let the pass certify the rest. Matches the issue's wording and keeps convergence a converging property.
2. Keep the withholding, but raise the default absolute cap above a realistic single-Context blast radius (largest here is 90) or make it percent-only, and give the trip its own named pass result so the wedge is diagnosable.
3. Gate the cap behind the same default-off flag as the timer, so "not part of merge" is true of the whole diff.

### Minor 1: a cap trip is labelled `failed`

`k8sClient.ts:3199` only distinguishes `DESIRED_NETWORKPOLICY_INVENTORY_CHANGED_MESSAGE`; everything else falls to `'failed'`. #462's Change C built that taxonomy to be decision-grade. A cap trip is a new **permanent-condition** outcome sitting in the transient-failure bucket, and the retry ladder treats it as transient. `netPolOrphanSweepCappedTotal` makes it observable, but an `aborted-cap` result would make the histogram tell the truth.

### Minor 2: the percent cap is dead at production scale

`networkPolicyReconciler.ts:92` checks absolute first, so with defaults (10, 20%) the percent rule only binds when the listed fleet is under 50. clerum-dev's fleet is 377, so the percent threshold is 75.4 and the absolute cap of 10 always fires first. Fine as small-fleet defense in depth, but the summary's "exceed 10 or 20% of the managed fleet" reads as though both are live at any scale.

### Minor 3: the cap counts pre-recheck candidates

`isContextLaneOrphan` (`:1450`) tests only `!desiredContextIds.has(contextId)`, while the actual delete path re-checks liveness inside `runContextEffect` (`:1353`, `:1357`) and skips policies whose Context turns out to be alive. The cap therefore counts a superset of what would actually be deleted, and can refuse a pass that would have deleted fewer than the cap. The direction is conservative, but an operator cannot reconcile the capped number against the deletes they would have seen.

---

Merge order in the summary (B #480 -> A1 #479 -> C) matches the other two PRs. This is C and goes last.

alfredolopez80:
@jozer-rami option 1 is in. A cap trip still refuses every orphan delete and increments `clerum_hcc_netpol_orphan_sweep_capped_total`, then the pass certifies live inventory. No throw, so the retry ladder cannot wedge on the same count.

M4 now runs three consecutive `fullReconcile`s with 11 orphans: `deleteNamespacedNetworkPolicy` stays 0, `complete` is called three times, the absolute metric increments three times. Same for the percent sibling and S2's N+1 row.

That matches [#478](https://github.com/evenfire-ai/evenfire/issues/478) ("alert and do not delete") and keeps certification a converging property. Stacked on [#479](https://github.com/evenfire-ai/evenfire/pull/479), a cap trip no longer 503s the data path forever.

Minors:

1. `aborted-cap` is not added. The pass result is `certified`; the cap counter is the signal. A named failure bucket would re-describe a path that no longer fails.
2. README now says absolute is checked first, so at a fleet ≥ 50 with defaults the percent rule never binds.
3. Pre-recheck overcount is unchanged (conservative). It no longer withholds certification.

Timer default stays 0; base/overlays still do not set `CONTEXT_MAPPER_NETPOL_RESYNC_SEC`. Merge order unchanged: #480 → #479 → #481.

alfredolopez80:
This branch is now synced with `origin/dev` after [#480](https://github.com/evenfire-ai/evenfire/pull/480) and [#479](https://github.com/evenfire-ai/evenfire/pull/479) merged (`6177014d`). Product files auto-merged; only `package.json` / lock versions conflicted.

T0 on the post-sync tree: `npx tsc --noEmit` clean, HCC Vitest **1801/1801**. Timer default stays 0.

Finding status vs the reviews already on this PR:

**Jozer (reviewed `31f06b95`)**

- Major (cap throw wedges certification): **option 1 is in**. Cap trip refuses deletes, increments `clerum_hcc_netpol_orphan_sweep_capped_total`, does not throw, and still certifies. M4: three consecutive `fullReconcile`s with 11 orphans → 0 deletes, `complete` ×3, metric ×3.
- Minor 1 (`aborted-cap`): **not added**. Result is `certified`; the cap counter is the signal.
- Minor 2 (percent dead at fleet ≥ 50): **documented**. Absolute is checked first.
- Minor 3 (pre-recheck overcount): **unchanged**, conservative, and no longer withholds certification.

**Claude (reviewed `31f06b9` / `65d5147`)**

- Coordinated timer body, default-off, `policy-type` census, 300s histogram buckets: **still hold**.
- Cap-before-deletes: **holds**. The throw/withhold-certification wording from that review is stale vs option 1.
- Non-blocking census under-report: **`controller_cap_would_trip`** + compiled defaults.
- Cap on every `fullReconcile` and default-0 boot warn: **documented / kept**.

Composition with the landed stack: content-identity abort still runs before certify (G5/G9 present). Probe stays watch-freshness; data path stays 6-clause (`apiStatus` 503 + `/ready` 200 on a lost delete).

@jozer-rami please re-review at `6177014d`. Do not merge until that Approve is on this HEAD.

alfredolopez80:
@claude please re-review and audit this PR for close on HEAD `16df382c`.

This is **only** #481 / #478 (coordinated NetworkPolicy resync timer + orphan sweep). Do not expand into a new McpServer-lane change (#464).

## Why this HEAD

`origin/dev` is an ancestor. The branch absorbed, in order:

- [#480](https://github.com/evenfire-ai/evenfire/pull/480) content-identity abort (`e4c97605`)
- [#479](https://github.com/evenfire-ai/evenfire/pull/479) probe vs data-path split (`03b69691`)
- [#445](https://github.com/evenfire-ai/evenfire/pull/445) Upload v2 follow-up (`b2b7c9a4`) — no HCC product overlap

Product files from #480/#479 auto-merged. Version/lock was the only conflict on the first sync. #445 merged clean.

## Closing-audit result (adversarial, Grok 4.6)

Verdict: **keep**. No new code blocker. Option 1 is implemented and is the accepted #478 residual (alert and do not delete — not withhold certification). Do not merge from this comment; human Approve is still required.

### Invariants to re-check

1. Timer body is exactly `void this.runInitialNetworkPolicyConvergence()` (`k8sClient.ts` setInterval). Naked `netPolReconciler.fullReconcile` is forbidden on this timer.
2. `CONTEXT_MAPPER_NETPOL_RESYNC_SEC` defaults to **0**. No deploy overlay sets it. Merge does not enable the tick. Enable remains an ops decision on #478 (freshness gate **or** a controller-runtime 10h-scale period; `300` is not the enable value).
3. Mass-delete cap runs on every `fullReconcile` (startup / event / timer). Candidates are counted **before** any delete. A trip increments `clerum_hcc_netpol_orphan_sweep_capped_total`, **refuses deletes**, does **not** throw, and **still certifies** live inventory (`onAuthoritativeRevocationComplete`). M4: three consecutive `fullReconcile`s with 11 orphans → 0 deletes, `complete` ×3, metric ×3.
4. `#480` identity abort still fires **before** certify (`DESIRED_NETWORKPOLICY_INVENTORY_CHANGED`). G5/G9 must still pass.
5. `#479` probe stays watch-freshness only (`resolveProbeAuthoritativeFn`). Data path stays 6-clause. Lost-delete fence: `/api/v1/mcpservers` 503 and `/ready` 200.
6. Census `scripts/ops/hcc-netpol-orphan-census.sh` filters `policy-type ∈ {context-allow, rpc-proxy-egress, external-egress}`. `live_cap_would_trip` is Deployment env only; `controller_cap_would_trip` applies compiled defaults (10 / 20) when unset.
7. Histogram buckets stay `[1, 5, 15, 60, 120, 300, 600, 1200, 1800, 3600]` on both pass-duration histograms. No new Prometheus labels.

### Prior review findings — expected status

**Jozer** (reviewed `31f06b95`):

- Major (cap throw wedges certification): **resolved by option 1**.
- Minor 1 (`aborted-cap`): **declined**. Pass result is `certified`; the cap counter is the signal.
- Minor 2 (percent dead at fleet ≥ 50): **documented**. Absolute is checked first.
- Minor 3 (pre-recheck overcount): **accepted residual**. Conservative; no longer withholds certification.

**Claude** (reviewed `65d5147` / `31f06b9`):

- Coordinated timer, default-off, `policy-type` census, 300s buckets: **still hold**.
- Cap-before-deletes: **holds**. The older “throw / withhold certification” wording is stale vs option 1.
- Non-blocking census under-report: **`controller_cap_would_trip`**.
- Cap on every `fullReconcile` and default-0 boot warn: **documented / kept**.

Accepted residual (not a merge gate): cap=10 stays on at startup with timer 0, so a single large Context `DELETED` can leave orphans live while the data path opens. That is #478 option 1, not a new throw-wedge.

## Evidence already on this HEAD

- Local T0 after the #480/#479 sync and again after #445: HCC `npx tsc --noEmit` + `npm test` **1801/1801**.
- GitHub CI on `16df382c`: **34/34 pass**, including `host-context-controller`, CodeQL, Real Postgres, and E2E Vitest node-unit. `mergeable=MERGEABLE`. `reviewDecision` is still `REVIEW_REQUIRED`.

Please audit this HEAD and close the review if the invariant holds. Do **not** merge.

zach88:
Took a deep pass through this one. The controller side is genuinely solid: the timer wiring, the single-flight convergence, and the cap logic (counts candidates before any delete, still certifies on a trip) all hold up under reading and repro. Where the problems concentrate is the ops census script, plus a couple of "make the caps harder to misconfigure" gaps and one design call on isolation that's yours to make.

## Verdict

🔴 **Not ready to approve**
- **Must resolve before approval: 1** (1 Blocker)
- **Optional / non-blocking: 6** (4 Medium, 2 Low) + 3 nits

1. **[R1-B1] Census script exits 2 against the real HCC deployment** (`scripts/ops/hcc-netpol-orphan-census.sh:69-72`)
   The guard aborts if `MAPPER_NS|HOST_NS|RPC_NS == "UNSET"`, and `deployment_env` returns `UNSET` whenever the Deployment omits the key. But `deploy/base/control-plane/host-context-controller.yaml` declares `CONTEXT_MAPPER_NAMESPACE` and `CONTEXT_MAPPER_HOST_NAMESPACE` only, never `CONTEXT_MAPPER_RPC_PROXY_NAMESPACE` (it's nowhere in `deploy/`). The controller resolves that one via its compiled default `getEnv('CONTEXT_MAPPER_RPC_PROXY_NAMESPACE','rpc-proxy')` (`host-context-controller/src/config.ts:552`) and reconciles fine.
   **Repro:** run against the manifest's exact env shape and the script prints `FATAL: refusing to invent defaults` and exits 2 before sampling anything. So the census is inoperable on the very deployment it's meant to audit, out of the box. It's also internally inconsistent: for the cap the script already falls back to compiled defaults (`controller_cap_would_trip`), but for namespaces it refuses.
   **Fix:** fall back to the same compiled namespace defaults the controller uses (`mcp-server`/`mcp-host`/`rpc-proxy`) with an explicit "assuming compiled default" note, same pattern you already use for the cap. If aborting is really intended, document the "you must export `CONTEXT_MAPPER_RPC_PROXY_NAMESPACE`" requirement in the usage header (today it only shows `CONTEXT=`).

## Optional (non-blocking)

The PR can merge without these; they're follow-up or your call. One of them (R1-M2) is a design decision I'd like a yes/no on before it ships.

1. **[R1-M2] The cap leaves stale allows in place across every reconcile, and couples lanes** (`host-context-controller/src/networkPolicyReconciler.ts:1478-1491,1560`)
   Before this PR the orphan sweep always revoked the *allow* NetworkPolicies of deleted contexts/servers. Now, above the absolute (10) or percent (20%) cap, the sweep refuses to delete and just alerts, so allows that should be revoked persist until someone raises the cap. Real exposure only if a pod that still matches the selector is alive (deleted context/server label, or a reused namespace); if the pods are already gone the allow is inert, which is why it's Medium and not higher. It also couples independent lanes: 11 orphans in the context lane plus 1 in external-egress refuses all four lanes, including that safe isolated external revocation. This is the documented #478 tradeoff ("alert and don't delete"), so it's a call, not a bug: I just want you to confirm the persistent-stale-allow window (no time bound today) is acceptable. If you want to narrow it, the move is a **per-lane cap** instead of one over the total.

2. **[R1-M1] Orphan-delete caps aren't validated for negative/zero** (`host-context-controller/src/config.ts:153-154`)
   `netPolOrphanDeleteCap`/`Percent` parse through the silent `getEnvInt`, which accepts `-1` and `0`. `=-1` makes `orphanCount > -1` true even for 0 orphans, so every `fullReconcile` trips `reason='absolute'`, spams `clerum_hcc_netpol_orphan_sweep_capped_total`, and silently disables orphan deletion forever. `=0` refuses all deletes on a single orphan. The sibling external-egress knobs in the same file already got a loud validating parser (`getExternalEgressEnvInt`); these should reject non-positive values the same way.

3. **[R1-M3] Census policy-type-only filter misses the controller's `repairable` policies** (`scripts/ops/hcc-netpol-orphan-census.sh:91-110`)
   The controller counts, in both `listedManagedFleet` and `orphanCandidates`, policies classified `repairable` (`managed-by` present, `policy-type` absent, corroborated by reserved name + owner labels). The census filters strictly on `policy-type ∈ {context-allow, rpc-proxy-egress, external-egress}`, so in a policy-type-strip corruption it under-reports (possibly `VERDICT=CLEAN`) while the controller counts them and may trip, which is exactly the corruption class the cap guards. The header documents the policy-type-only choice on purpose, so either confirm it's acceptable against the spec or mirror the `managed-by + reserved-name` rule.

4. **[R1-M4] `evaluateNetPolOrphanSweepCap` has no direct test** (`host-context-controller/src/networkPolicyReconciler.ts:940`)
   It's a pure `(orphanCount, listedManaged) -> reason|null` with absolute-first precedence plus an inertia rule, exercised only through `fullReconcile`. No test asserts that when both rules would fire it reports `'absolute'`, nor the exact `=== cap` boundary or the 0/negative cases. A small table or `fast-check` test would pin it (T2).

5. **[R1-L1] Percent denominator can inflate under a mid-pass authority flip** (`host-context-controller/src/networkPolicyReconciler.ts:1465-1477`)
   `listedManagedFleet` always adds `allExternalPolicies.length`, while the external slice of `orphanCandidates` is gated on `serverCleanupAuthoritative()` re-evaluated later. They agree normally; they diverge only if `serverInventoryAuthoritative()` flips false between population and counting, which inflates the denominator and weakens the percent cap for the remaining context orphans. Narrow, and it only ever weakens the cap (never over-deletes). Computing the fleet from the same gated slices closes it.

6. **[R1-L2] Malformed numeric env in the census becomes 0 and trips spuriously** (`scripts/ops/hcc-netpol-orphan-census.sh:182,186`)
   A non-numeric `CONTEXT_MAPPER_NETPOL_ORPHAN_DELETE_CAP` makes `[[ ORPHAN_COUNT -gt ORPHAN_CAP ]]` treat it as a variable name, so it reads as 0 and reports a spurious `absolute` trip. Low likelihood (needs Deployment-env control), a `^[0-9]+$` guard hardens it.

**Nits:** [R1-N1] the `orphanContextAllow` fixture uses `ingress: [{ _from: [...] }]` where the real apiserver field is `from` (`networkPolicyReconciler.orphanSweep.test.ts:602`), harmless since classification reads labels only but no producer emits that shape. [R1-N2] the first bucket test asserts the constant against its own literal, tests 2/3 already cover the real observation (`metrics.test.ts:23-27`). [R1-N3] a one-line comment in the census noting it reflects on-cluster orphans, not the controller's current authority gate, would prevent misreads.

## What checked out clean

- **Timer:** `runInitialNetworkPolicyConvergence` routes through `runInitialConvergence('NetworkPolicy')`, which is single-flight for real (coalesces a concurrent tick via `trailingRequested`) and defers when caches are unsynced, re-arming the retry without registering duplicate timers. `stop()` clears the interval with no leak. Default-off matches the Host/SFS/GFS/LlmHook pattern.
- **Cap:** candidates are counted before any delete, certification is not gated by a trip, and `recordOrphanDelete` preserves the old `deleted===true -> recordSafetyPassRevocation()` semantics exactly. The counted set is a superset of the delete set in the safe (over-refuse) direction.
- **Cap parity census vs controller:** brute-forced 6480 `(orphan, listed, pct, abs)` cases, 0 mismatches. Absolute-first, the `listed*pct >= 100` inertia, and `orphan*100 > listed*pct` are all exactly equivalent to the TS float formula under integer arithmetic.
- **Census is genuinely read-only:** every `kubectl` is `get`, jsonpath uses fixed literals, namespaces pass as argv, jq uses `--arg`, nothing is eval'd. `bash -n` and `shellcheck` clean.
- **Metrics:** the histogram helper is idempotent per registry, buckets are uniform across both histograms, and the new counters declare `lane`/`reason` matching what the producer passes.
- **Suites:** 466/466 green across the five touched suites, `tsc --noEmit` clean. The new timer/metrics/cap tests fail against the base, so they cover real new behavior.

alfredolopez80:
@alfredolopez80 — one finding from an independent pass: a gap against this PR's own stated requirement.

## 🔴 The periodic pass does not cover drift class D

`HARD REQUIREMENT 1` of #478 names **A and D** as the unbounded classes the sweep must cover. Class D is defined in the issue body as:

> | **D — L0/L1 defaults edited** (deny-all, infra egress) | **no** — only `ensureDefaultPolicies` at startup | ❌ **until Pod restart** |

But the pass the timer re-enters passes `ensureDefaults: false`:

```ts
// host-context-controller/src/k8sClient.ts:3107-3111
await this.netPolReconciler.fullReconcile(initialContexts, initialServers, {
  serverInventoryComplete,
  ensureDefaults: false,          // ← class D never re-checked
  contextInventoryAuthoritative,
  serverInventoryAuthoritative,
```

And `networkPolicyReconciler.ts:1272` gates the default-policy work on `options.ensureDefaults !== false`.

**Verified against this PR's diff**: `ensureDefaults` appears 9 times, **all of them in `networkPolicyReconciler.orphanSweep.test.ts`, all as `false`**. No production line changes it. So the timer re-runs the pass every period and **class D still self-heals only on Pod restart** — exactly the "weeks" window the issue set out to close.

Net effect as it stands: the timer covers **A** (orphan from a lost DELETE) and **F** (failed initial LIST, retried each pass), but **not D**. That is half of `HARD REQUIREMENT 1`.

**Options** (your call, both defensible):

1. Pass `ensureDefaults: true` on the periodic path only — keeps startup semantics untouched, closes D. Needs a test asserting a hand-edited L0/L1 policy is repaired within one period.
2. Declare D explicitly out of scope in the PR body and open a follow-up. Legitimate, but then `HARD REQUIREMENT 1` in #478 should be amended so it does not read as satisfied.

What is not fine is leaving it as-is silently: the issue states the requirement, and a reader will assume it shipped.

alfredolopez80:
Validated that pass and landed it on `6e9db752`.

**Class D / HARD REQUIREMENT 1:** agreed. The only production `netPolReconciler.fullReconcile` is the coordinated core, and it passed `ensureDefaults: false`. `start()` still runs `ensureDefaultPolicies()` as the bootstrap barrier (class D until Pod restart). The timer now calls `runInitialNetworkPolicyConvergence({ ensureDefaults: true })`; startup and event entry stay false. M1/M2 pin the bag: first pass `ensureDefaults: false`, timer tick `true`. Did not amend #478.

**R1-B1:** agreed — two-tier fallback in this PR, not a manifest change. Live line is now:

`CONTEXT_MAPPER_RPC_PROXY_NAMESPACE=UNSET (controller applies compiled default: rpc-proxy)`

Declaring all five omitted namespaces belongs in a follow-up, not here.

Also on this HEAD from Diego: negative orphan-delete caps fail loud; `0` remains a never-delete kill-switch. Per-lane cap still declined (option 1).

alfredolopez80:
Landed on `36b832c0`:

- Timer ticks while a NetworkPolicy pass is in flight are skipped (`clerum_hcc_netpol_resync_ticks_skipped_total{reason="pass-in-flight"}`) and do not set `trailingRequested`. Event-driven coalescing is unchanged. M2 now pins the skip (removing the `if` fails the first assert).
- Class D stays closed on the periodic path only (`ensureDefaults: true` on the timer; startup/event stay `false`). HARD REQUIREMENT 1 is not being amended.
- Enable stays off in this PR. Chosen post-merge value is **1500 s** with the skip guard (floor **3600 s** without it). `300` is not a valid enable. Preconditions are on #478.

T0 on this HEAD: `npx tsc --noEmit` + `npx vitest run --fileParallelism=false` — 84 files, 1804 tests.

zach88:
Took another deep pass now that the two fix commits landed. Good news: the round-1 blocker (the census `exit 2`) is properly closed, the negative-cap validation is in, and nothing regressed. What's left is concentrated in two spots: the ops census script (a couple of parity/robustness gaps) and a still-untested pure cap function. All non-blocking.

## Verdict

**🟡 Approve with nits**
- **Optional / non-blocking: 6**

No open blockers and no regressions from the fixes (checked the `6177014d..36b832c0` range with the fix commits isolated from the #445 sync). This is my read from the review, not a formal GH approval.

Nothing blocking this round. The one round-1 blocker is closed (see below).

## Optional (non-blocking)

PR can merge without these; they're follow-up or author's call. Flagging the three Mediums that have now been open two rounds so they don't just float into a third.

1. **[R1-M2] Cap leaves stale allows in place + couples lanes** (`host-context-controller/src/networkPolicyReconciler.ts:1478-1491,1560`) — *re-triage: needs a call, up or written-off.* When the sweep trips the cap (10 absolute / 20% percent) it refuses to delete and only alerts, so stale `context-allow`/`rpc-proxy-egress`/`external-egress` allows for deleted contexts/servers persist until a human raises the cap or deletes by hand. It also couples independent lanes (11 context orphans + 1 external refuses all four lanes, including the safe isolated external revoke). Real exposure only if a pod still matching the selector is alive. This is the documented #478 "alert and do not delete" decision, so it's your call: accept the (currently unbounded) window and add an alert on `clerum_hcc_netpol_orphan_sweep_capped_total > 0`, or narrow to a per-lane cap as a follow-up. Either way it shouldn't quietly ride into another round.

2. **[R1-M3] Census misses `repairable` policies the controller counts** (`scripts/ops/hcc-netpol-orphan-census.sh:113-132`) — *re-triage.* The script filters strictly on `policy-type ∈ {context-allow, rpc-proxy-egress, external-egress}`, but the controller also counts `repairable` policies (`managed-by` present, `policy-type` absent, corroborated by reserved name + owner labels; they land in both `listedManagedFleet` and `orphanCandidates`). In a "policy-type stripped" corruption the census under-reports (can show `VERDICT=CLEAN` / `controller_cap_would_trip=none`) while the controller counts them and might trip. Fix: mirror the `managed-by + reserved-name` rule, or document explicitly that `controller_cap_would_trip` doesn't cover that case.

3. **[R1-M4] Pure cap function has no direct test** (`host-context-controller/src/networkPolicyReconciler.ts:97`) — *re-triage.* `evaluateNetPolOrphanSweepCap` is a precedence function (absolute-first, inertia `percentThreshold >= 1`), exported but only exercised via `fullReconcile`. The case that matters most, both rules firing at once so `'absolute'` must win, is never asserted; nor is `orphanCount === absoluteCap` exact, nor cap `0`/negative at the function level. Cheap to close with a direct table test.

4. **[R2-L1] Caps still coerce non-canonical numeric input** (`host-context-controller/src/config.ts:283,775-782`) — the negative-cap fix validates sign but not shape: `getEnvInt` runs through `parseInt`, which silently coerces `'2e1'→2`, `'10.9'→10`, `'0x1F'→0`, `'12abc'→12` (reproduced). `requireNonNegativeEnvInt` gets the already-truncated int, so it can't catch these. The sibling `getExternalEgressEnvInt`, built for exactly this class, rejects them with `/^\d+$/`. Scenario: someone sets `..._CAP_PERCENT='2e1'` meaning "20" and gets cap `2`. Direction is safe (coerces to ≤ intended, so more delete-refusals, never an accidental mass-delete), hence Low. Reusing `getExternalEgressEnvInt` for these two knobs closes the class.

5. **[R1-L2] Census can spurious-trip on a malformed cap value** (`scripts/ops/hcc-netpol-orphan-census.sh:204,208`) — `[[ "${ORPHAN_CAP}" -gt ... ]]` doesn't guard that the value is numeric. Reproduced: `ORPHAN_CAP=abc` + `ORPHAN_COUNT=3` treats `abc` as a var name → `0` → spurious `absolute` trip (exit 0, no error). Needs a malformed/negative value on the Deployment (cluster admin), so Low, but there's a live repro, so it wants either a `^[0-9]+$` guard or a written-off note. The "fail loud" from the fix lives in `config.ts` (the controller), not in the script.

6. **[R1-L1] Percent denominator can inflate under mid-pass authority flip** (`host-context-controller/src/networkPolicyReconciler.ts:1474-1477`) — `listedManagedFleet` always adds `allExternalPolicies.length` while the external slice of `orphanCandidates` is gated by `serverCleanupAuthoritative()`. If server authority flips to false between population and count, the denominator inflates and weakens the percent cap. Safe direction (only ever under-deletes), not reproduced. Optional: gate the external term of the denominator with the same condition as the candidates.

**Nits** (no action required):
- `[R2-N1]` `k8sClient.ts:3105-3118`: the `ensureDefaults` flag isn't cleared on the `deferred-unsynced` early-return, so a later event-driven pass can consume `ensureDefaults:true`, contradicting the "event entry stays false" comment. Benign (idempotent, fail-closed, resync is default-off). Cleanest fix is to also clear the flag in that early-return.
- `[R2-N2]` `config.ts:780`: percent comment says "(0-100)" but there's no upper bound; `'150'` is accepted. Harmless (percent > 100 never fires).
- `[R1-N1]` `networkPolicyReconciler.orphanSweep.test.ts:602`: fixture uses `_from` (real field is `from`). Harmless, classification reads labels only.
- `[R1-N2]` `metrics.test.ts:22-26`: first test is tautological (constant vs its own literal). The two `observe` tests already cover the real thing.

## What checked out clean

- **[R1-B1] closed** (`6e9db752`): census now falls to the compiled `rpc-proxy` default when the env is unset (matches `config.ts:561`), only mapper/host stay fatal. Verified it doesn't sample a wrong namespace, and no regression.
- **[R1-M1] closed** (`6e9db752`): `requireNonNegativeEnvInt` rejects negatives loud, `0` stays a valid documented never-delete kill-switch, with tests for both. (Shape residue tracked as R2-L1 above.)
- **[R1-N3] closed** (`6e9db752`): census now carries the "reflects on-cluster orphans by policy-type, not the controller's authority gate" note.
- **Skip-in-flight guard** does what the body says: a tick while a pass runs increments `clerum_hcc_netpol_resync_ticks_skipped_total{reason="pass-in-flight"}` and does not set `trailingRequested`; event-driven coalescing is untouched. The `M2` test is a genuine regression test (goes red against the parent on the behavior assertion, not on setup).
- **Sibling timers unaffected**: the four existing resync timers use separate mechanisms; the guard only filters the NetworkPolicy timer trigger. New counter doesn't collide in the registry. `stop()` clears the timer and the flag.
- **Security: nothing new.** The cap only omits deletes, never creates or widens a NetworkPolicy or relaxes deny-all; `ensureDefaults:true` only tightens. No `deploy/**/networkpolicies.yaml`, token, JWT, or redact surface touched. Census is read-only (`kubectl get` only), `shellcheck`/`bash -n` clean, no `eval`, jq via `--arg`.
- **Suites**: 469/469 green across the five touched suites, 280 in `k8sClient`, `tsc --noEmit` clean.

zach88 (approved):

alfredolopez80:
Second-pass findings (`#issuecomment-5442787233`), landed on `b7218bd4`.

**Written off (product call, not a follow-up ticket):**
- **[R1-M2]** One cap over the **sum of all lanes** stays. A per-lane cap would let each lane delete up to N and contradict #478 option 1 (alert and do not delete; still certify). Documented on the function and in the README: a trip leaves those orphans in place until an operator raises the env; the operator signal is `clerum_hcc_netpol_orphan_sweep_capped_total`. Infra alert is later, not this PR.

**Fixed:**
- **[R1-M3]** Census now includes the controller `repairable` class (managed-by + reserved `ctx-` / `rpc-egress-` / `ext-egress-` name, policy-type absent) in both listed and orphan sets, and prints `repairable_untyped`. Approximation only — reserved names without owner labels can over-count vs the controller (safer than a false `CLEAN`). Did not copy the full four-lane classifier into bash.
- **[R1-M4]** Direct table test on `evaluateNetPolOrphanSweepCap`: exact cap still deletes, both-rules → `absolute`, percent under abs, inert tiny fleet, cap `0`, negative abs, percent > 100 never fires.
- **[R2-L1]** Caps use `requireCanonicalNonNegativeEnvInt` (`/^\d+$/`). `2e1` / `10.9` / `12abc` / `0x1F` fail loud. Did not reuse `getExternalEgressEnvInt` (its error text says seconds). `0` remains a valid kill-switch.
- **[R1-L2]** Census `is_uint` / `^[0-9]+$`; malformed live or compiled cap prints `malformed` and never does arithmetic.
- **[R1-L1]** `listedManagedFleet` uses the same `countExternalLane` gate as `orphanCandidates`.
- **[R2-N1]** `deferred-unsynced` now clears `netPolConvergenceEnsureDefaults` so a later event/retry stays `ensureDefaults: false`.
- **[R2-N2]** Percent comment/README no longer imply a 0–100 bound; values > 100 never fire.

**Skipped (your nits, no action):**
- **[R1-N1]** `_from` fixture.
- **[R1-N2]** tautological metrics buckets.

Local T0 on this commit: `npx tsc --noEmit` clean; 310 tests green in `config.netPolResync`, `networkPolicyReconciler.orphanSweepCap`, `networkPolicyReconciler.orphanSweep`, `k8sClient`, `metrics` (`--fileParallelism=false`). Census `bash -n` clean; jq fixture covers typed + repairable-untyped.

alfredolopez80:
@claude please re-review and audit this PR for close on HEAD `00b0d3e8`.

This is **only** #481 / #478 (coordinated NetworkPolicy resync timer + orphan sweep + the census this PR introduces). Do not expand into #464, #465, or a new McpServer-lane / Host-role change.

GitHub CI on this HEAD is **34/34 SUCCESS**, including `host-context-controller`, `Shell script syntax (bash -n)`, CodeQL `Analyze (javascript-typescript)`, Real Postgres, and E2E Vitest node-unit. `mergeable=MERGEABLE`. Do **not** merge from this comment.

## Why this HEAD

After the previous close-audit at `16df382c` / `36b832c0`, two follow-up commits landed:

- `b7218bd4` — zach88 second-pass Medium/Low (canonical cap parse, census `repairable` class, direct cap table, percent-denom gate, `ensureDefaults` unsynced reset)
- `00b0d3e8` — fail-loud resync interval; mutation-survivor tests (cap × `external-egress`, event-driven `ensureDefaults` after a tick); census fail-closed kubectl / empty-fleet / `live_cap=UNSET` / uid dedupe

Body now **Refs #478** and **Closes #483**. #484 stays **open** (TS↔bash cap parity + `managed-by` filter — needs a product call, not a drive-by in this PR).

## Closing-audit request

Two jobs, both required:

1. **Closure ledger** — walk every finding below against `00b0d3e8`. For each: closed / written-off with a recorded call / declined / still open. Quote the file:line that proves it. If any claimed-closed item is still open, that is a blocker for this review.
2. **Your own complete audit** of this HEAD vs `origin/dev`. Same bar as the earlier close-reviews: timer wiring, skip-if-in-flight, cap-before-deletes + still-certify, class D call site, census fail-closed, no new NetworkPolicy widen, no deploy enable.

## Invariants to re-check

1. Timer body is exactly `void this.runInitialNetworkPolicyConvergence({ ensureDefaults: true })`. Naked `netPolReconciler.fullReconcile` is forbidden on this timer. A tick while a pass is in flight is **skipped** (`clerum_hcc_netpol_resync_ticks_skipped_total{reason="pass-in-flight"}`) and does **not** set `trailingRequested`.
2. `CONTEXT_MAPPER_NETPOL_RESYNC_SEC` defaults to **0**. No deploy overlay sets it. Non-canonical values (`25m`, `1e3`, `2e1`) **fail loud at load**. Merge does not enable the tick.
3. **Class D of #478 HARD REQUIREMENT 1 is not covered by this merge.** The tick is the only call site that passes `ensureDefaults: true`. Class D waits on the later enable (chosen **1500 s** with the skip guard). That is why the body Refs #478 instead of closing it.
4. Mass-delete cap: one cap over the **sum of all lanes**. Candidates counted before any delete. A trip refuses **all** orphan deletes (context **and** `external-egress`), increments `clerum_hcc_netpol_orphan_sweep_capped_total`, does not throw, **still certifies**.
5. `#480` identity abort still fires before certify. `#479` probe vs data-path split unchanged.
6. Census: `policy-type` plus reserved-name `repairable` class. kubectl failure is FATAL, not `UNSET`. Zero listed managed policies is `INCONCLUSIVE_EMPTY`, never `CLEAN`. `live_cap_would_trip=UNSET` when both cap keys are omitted. Do **not** “fix” the context-lane jq to read `clerum.io/mcpserver` — that HIGH-2 is **refuted**; context orphans are keyed only on `clerum.io/context` (parity with `networkPolicyReconciler.ts`).
7. Histogram buckets stay `[1, 5, 15, 60, 120, 300, 600, 1200, 1800, 3600]`. No new Prometheus labels.

## Prior findings — expected status on `00b0d3e8`

**jozer-rami** (`31f06b95`, option 1 reply on `08ef8c04`):

- Major (cap throw wedges certification): **closed** — refuse deletes, increment cap metric, still certify. M4 ×3 consecutive passes.
- Minor 1 (`aborted-cap` bucket): **declined** — pass result is `certified`; the cap counter is the signal.
- Minor 2 (percent dead at fleet ≥ 50): **documented** — absolute is checked first.
- Minor 3 (pre-recheck overcount): **accepted residual** — conservative; no longer withholds certification.

**zach88** round 1 (`#issuecomment-5442234005`) and round 2 (`#issuecomment-5442787233`):

- **[R1-B1]** census `exit 2` on omitted rpc-proxy ns: **closed** (`6e9db752`) — compiled `rpc-proxy` fallback, mapper/host stay fatal.
- **[R1-M1]** negative cap: **closed** — fail loud; `0` remains the kill-switch.
- **[R1-M2]** cap leaves stale allows + couples lanes: **written off** — product call, one all-lane cap, option 1. Signal is `clerum_hcc_netpol_orphan_sweep_capped_total`. Not a per-lane cap.
- **[R1-M3]** census misses `repairable`: **closed** (`b7218bd4`) — approximation, not the full four-lane classifier.
- **[R1-M4]** no direct cap-function test: **closed** (`b7218bd4`) — `evaluateNetPolOrphanSweepCap` table.
- **[R2-L1]** cap knobs coerce via `parseInt`: **closed** (`b7218bd4`) — `requireCanonicalNonNegativeEnvInt`.
- **[R1-L2]** census `ORPHAN_CAP=abc` → spurious trip: **closed** — `is_uint` / `malformed`.
- **[R1-L1]** percent denominator inflates if server authority flips: **closed** — same `countExternalLane` gate on listed fleet.
- **[R2-N1]** `ensureDefaults` uncleared on deferred-unsynced: **closed**.
- **[R2-N2]** percent comment implied 0–100: **closed**.
- **[R1-N1]** `_from` fixture / **[R1-N2]** tautological metrics buckets: **no action** (your nits).

**alfredolopez80** (self-pass + implementer r2):

- Class D gap vs HARD REQUIREMENT 1: **closed in code, not in production** — tick now passes `ensureDefaults: true`; enable still off. See invariant 3.
- Skip-if-in-flight treadmill: **closed** (`36b832c0`).
- Interval still on `getEnvInt`: **closed** (`00b0d3e8`).
- Cap × `external-egress` untested / `ensureDefaults` reset untested: **closed** (`00b0d3e8`, #483). Mutation: drop `!capReason` on the external loop, or drop the `:3146` reset → those tests go red.
- Census kubectl `|| true` / empty `CLEAN` / overloaded `none` / missing uid dedupe / `grep -c || true`: **closed** (`00b0d3e8`).
- README factor-100 (`percent * listed < 1`): **closed**.

**claude** (close-audits at `65d5147` / `31f06b9` / `16df382c`):

- Coordinated timer, default-off, 300s buckets, cap-before-deletes: **must still hold**.
- Older “throw / withhold certification” wording is stale vs option 1 — do not re-open that.
- Non-blocking census under-report of compiled defaults: **`controller_cap_would_trip`** still present; `live_cap` is now `UNSET` when both keys are omitted (no longer overloaded `none`).

Accepted residuals that must **not** be re-litigated as merge blockers unless you find a new hole:

- One all-lane cap (R1-M2 / #478 option 1).
- #484 (TS↔bash cap parity + `managed-by` filter).
- Class D production coverage waits on the enable, not this merge.

Please audit this HEAD and say whether the review can close. Do **not** merge.

zach88:
Took a full round-3 pass through this one, plus a collateral audit of the two new `fix:` commits against the previously-reviewed head (`36b832c0..00b0d3e8`). Short version: every open finding from round 2 is genuinely closed — I verified each by reproducing against the real code, not by trusting the commit messages — and the collateral audit came back clean, zero regressions. What's left is all cosmetic.

## 🟡 Approve with nits

Per the body, this still needs a human Approve on the post-sync HEAD — and that's the only gate I see left. This is my read from the review, not a formal GH approval.

Nothing. No blockers, criticals, highs, or open regressions. The three behavior changes this round (canonical-int fail-loud, the percent-cap denominator, the census exit codes) are all intentional and monotonic toward the safe side — no valid canonical config value changes behavior.

## Optional (non-blocking)
This PR can merge without any of these; they're follow-up or author's call.

1. **[R1-M2] One all-lane orphan-delete cap leaves stale allows in place — design decision, needs a sign-off** (`host-context-controller/src/networkPolicyReconciler.ts:1478-1491`). This is the #478 "alert and do not delete" behavior, now actually implemented ("one cap over the sum of all lanes") and documented in the README, with `clerum_hcc_netpol_orphan_sweep_capped_total` as the signal. Security reads it Low by-design (deleting an allow is fail-closed; stale allows only bite if a pod still matching the selector is alive). It's been open 3 rounds, so it shouldn't float another one: either accept the window and wire an alert on `..._capped_total > 0`, or schedule the per-lane narrowing as a follow-up issue. No code change required to merge.

2. **[R3-N1] The one real behavior change lacks a test asserting the new behavior** (`host-context-controller/src/networkPolicyReconciler.ts:1478`). The percent-cap denominator now excludes external policies when `countExternalLane` is false. It's the safe direction (only ever refuses *more* deletes, never fewer), so not a regression — but three of the new tests (`orphanSweepCap` suite, the k8sClient "M1 sibling", the orphanSweep "M4 sibling") pass against the parent head, which makes them characterization coverage rather than a repro for this line. Worth one test that goes red without the denominator change.

3. **[R1-N1] Hand-written fixture uses `_from` instead of the real `from` field** (`host-context-controller/src/networkPolicyReconciler.orphanSweep.test.ts:127`). The ingress rule ends up silently empty. Harmless to the sweep (it keys on labels/name/spec-presence only), but it's exactly the fixture-drift T1 warns about. Rename `_from` → `from`, and ideally run these fixtures through the same `asApiserverNetworkPolicy` normalizer the sibling suite uses.

4. **[R1-N2] First metrics test is tautological** (`host-context-controller/src/metrics.test.ts:22-26`) — asserts a constant against its own literal. The two `observe`-over-buckets tests below it do carry weight, so this is purely cosmetic.

5. **`setInterval(sec * 1000)` has no 2³¹-ms upper bound** (`host-context-controller/src/k8sClient.ts`, the `netPolResyncTimer` wiring). An absurd interval (> ~24.8 days) would get clamped by Node to 1ms. It's the established idiom across all four sibling timers in the file, none of which cap either, and the resync is default-off with a planned enable of 1500s — so fixing it here would just diverge from the siblings. Better as a whole-class follow-up than a change in this PR.

## What checked out clean
- **Security: SAFE.** The sweep only *skips* deletes under the cap; deleting an allow is fail-closed (traffic gets more denied, deny-all persists), never an opening. `ensureDefaults: true` only re-hardens. Env parsing fails closed at load. Census is read-only, `shellcheck` clean, no `eval`, jq via `--arg`, arithmetic `is_uint`-guarded.
- **The cap can't over-delete.** The pre-count predicates are the same ones the delete loops use, and `serverCleanupAuthoritative()` is monotonic, so a mid-pass authority flip can only delete *fewer* than counted. Absolute cap is checked first, so >10 orphans always refuses.
- **Canonical-int validation is class-complete.** All three new knobs go through `requireCanonicalNonNegativeEnvInt` (`/^\d+$/`); grepped the diff, no new knob left on raw `parseInt`. `2e1`/`10.9`/`12abc`/`25m`/`1e3`/negatives all throw at load. Reverting to `getEnvInt` turns 13 reject-assertions red.
- **Census parity with the controller.** It now counts the `repairable` untyped class; reproduced end-to-end with a mock kubectl — census set ⊇ controller set (over-counts, safe direction), so it can't print `CLEAN` while the controller would trip. Empty fleet → `INCONCLUSIVE_EMPTY` (exit 4), never `CLEAN`.
- **Suites/types green:** `tsc --noEmit` clean; config/metrics 22, k8sClient resync 30, orphanSweep+cap+reconciler 189; `shellcheck` + `bash -n` clean on the census.

alfredolopez80:
**[R1-M2] signed off** — this is the last open design call from the round-3 pass.

**Accept the window.** One cap over the sum of all lanes stays. A trip refuses every orphan delete, increments `clerum_hcc_netpol_orphan_sweep_capped_total`, and still certifies (#478 option 1: alert and do not delete). Stale allows remain until an operator raises the env. Security’s Low-by-design read is accepted: deleting an allow is fail-closed; a leftover allow only bites if a pod still matching the selector is alive.

**Per-lane narrowing is declined**, not deferred. Splitting the cap would let each lane delete up to N (e.g. 9+9+9+9) and re-open the mass-delete #478 refused.

**Alert is the follow-up, not this PR.** The controller already emits the signal. Wiring `clerum_hcc_netpol_orphan_sweep_capped_total > 0` lives in [#485](https://github.com/evenfire-ai/evenfire/issues/485) (`evenfire-infra`, dev first, after this image is scraped). No HCC code change for R1-M2.

No other optional from that comment is in this reply. Do not merge from here.

PR #486, opened 2026-08-27
fix(hcc): treat a 404 on conflict-retry read as gone, not fatal (#472)

### LINKED ISSUE #472: bug(hcc): applyNetworkPolicy raises an uncaught 404 when the object is deleted between the 409 and the read, losing the whole certification pass
`applyNetworkPolicy` follows a create-then-409-then-replace pattern. If the object is deleted in the
window between the `409` and the subsequent `read`, the `read` raises a `404` that nothing catches,
the additive reconciliation fails, and the whole certification pass is lost — taking readiness down
with it.

Observed on `clerum-dev` on 2026-08-26, image `sha-0ece1b8` (`origin/dev` @ `0ece1b8e`).

## Observed

```
23:32:16  [NetPol] Additive Context reconciliation failed for "b2b": ApiException HTTP-Code: 404
          body: networkpolicies.networking.k8s.io "ctx-…" not found
23:44:59  [K8s] Initial NetworkPolicy post-certification additive reconciliation failed: AggregateError
23:44:59  [K8s] Scheduling initial NetworkPolicy background convergence retry 1 in 5000ms
23:45:24  Pod .status.conditions[Ready] -> False
   …
23:58:07  readiness became ready
```

**12 min 43 s of unreadiness**, triggered by a single missing object out of ~861 NetworkPolicies.
The `Context/b2b` itself exists and is healthy (11 McpServers); the 404 was on one of its
`ctx-b2b-*` policies, and 11 such policies are present now — so the object was either recreated
afterwards or the 404 was a genuine race.

## Mechanism

`host-context-controller/src/utils.ts`:

```ts
// applyNetworkPolicy :249-266
try {
  await api.createNamespacedNetworkPolicy({ namespace, body: policy })
  return
} catch (error) {
  if (getErrorCode(error) !== 409) throw error     // only 409 is tolerated
}
await replaceWithConflictRetry<k8s.V1NetworkPolicy>({
  read: () => api.readNamespacedNetworkPolicy({ name, namespace }),
  ...
})

// replaceWithConflictRetry :58-60 — the read sits OUTSIDE the try
for (let attempt = 1; attempt <= maxAttempts; attempt++) {
  const existing = await read()          // <- a 404 here escapes uncaught
  validateExisting?.(existing)
  ...
  try { await replace(next) } catch (err) {
    if (getErrorCode(err) === 409 && attempt < maxAttempts) { /* backoff, retry */ }
    throw err                            // only 409 is retried
  }
}
```

The sequence is a textbook TOCTOU:

1. `create` returns **409** — the policy exists.
2. Something deletes the policy in the gap.
3. `read` returns **404** — an exception no handler expects.

The retry ladder does not help: it only re-attempts on `409`, and the failing call is the `read`,
which is not inside the guarded block.

## Impact

The failure is **recoverable but expensive**. The `AggregateError` propagates to the `catch` at
`k8sClient.ts:3181` and the convergence ladder reschedules, so the controller does recover on its
own. But the cost is disproportionate: one transient missing object aborts an entire certification
pass and removes the controller from its Service endpoints for minutes.

The correct response to that 404 is unambiguous — the object is gone, so create it — and it is
currently absent.

## Proposed fix

Preferred, narrow: in `applyNetworkPolicy`, catch a `404` from the `read` path and retry the
`create` once (with a loop guard so a create/delete flapper cannot spin).

Broader alternative: move the `read` inside the guarded block of `replaceWithConflictRetry` and
treat `404` as "object is gone — recreate". This is cleaner but changes shared behaviour for
**7 files** that use the helper (`hostReconciler`, `llmHookReconciler`, `sharedFileSystemReconciler`,
`networkPolicyReconciler`, `reconciler`, `k8s/gfsK8sApi`, `utils`), so it deserves its own review.

Either way the change is small. `applyNetworkPolicy` alone has **22 call sites across 7 files**, so
whichever path is taken should land as an isolated, revertible PR.

## Sequencing

Should land **after** #471, which is currently modifying `applyNetworkPolicy` to add the `isUpToDate`
equality gate. Landing both at once would entangle two unrelated changes in the same function and
make either one harder to revert.

## Acceptance criteria

- A unit test that deletes the object between the `409` and the `read` (fake/mocked client) proves
  the policy ends up created rather than raising. **The test must fail against current `main`** —
  otherwise it is not covering the race.
- No change to fail-closed semantics: `validateExisting` still runs before any write, and ownership
  checks are untouched.
- A create/delete flapper cannot cause an unbounded create↔404 loop.

## Notes

Found while diagnosing an unrelated readiness outage; the recovery path itself worked as designed
(the #467 watchdog logged and re-armed the interrupted pass). This issue is about removing the
trigger, not the recovery.

alfredolopez80:
## Summary
- `replaceWithConflictRetry` treated a 404 from `read()` as fatal because that call sat outside the replace `try`. After `create` returns 409, a delete in the gap aborted the whole NetPol certification pass (`AggregateError`).
- Wrap only `read()`: a 404 returns cleanly (object is gone; the next pass / #467 watchdog recreates). 403, 500, and network errors without a status still throw — including on the post-409 re-read (TOCTOU-NP-6/7).
- Does not retry `create` in the same call (anti-flapper). Does not change the #471 `isUpToDate` gate or the #480 409 retry ladder.

**Read-404 on deny-all (R1-L1):** if `deny-all-<ns>` is deleted in the create-409→read window, this pass can certify readiness while that deny-all is still absent. The *network* state matches base (the object is gone either way). Only the fail-closed certification signal is suppressed; the next resync / #467 watchdog recreates it. Conscious call: ≤1-pass self-heal. This PR does not special-case the deny path.

Closes #472 except same-call recreate; next pass / #467 watchdog recreates.

## Test plan
- [x] TOCTOU-NP-1: create 409 + read 404 (`ApiException` and `response.statusCode`) → no throw; create ×1; read ×1; replace ×0
- [x] TOCTOU-NP-2: helper read 403/500 still throw
- [x] TOCTOU-NP-2-APPLY: `applyNetworkPolicy` create 409 + read 403/500 still throw
- [x] TOCTOU-NP-3: replace 409 then second read 404 → return; no third replace
- [x] TOCTOU-NP-4: network `Error` without `code` is rethrown
- [x] TOCTOU-NP-5: helper replace 404 still throws
- [x] TOCTOU-NP-6/7: replace 409 then second-read 403/500/network still throw
- [x] Existing NOOP-NP-1/2 and ORDER-1 stay green
- [x] `npx tsc --noEmit` and `npm test` in `host-context-controller`

### DIFF
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.255",
+  "version": "0.1.256",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.255",
+      "version": "0.1.256",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.256",
+  "version": "0.1.257",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.256",
+      "version": "0.1.257",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.257",
+  "version": "0.1.258",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.257",
+      "version": "0.1.258",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.258",
+  "version": "0.1.259",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.258",
+      "version": "0.1.259",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.258",
+  "version": "0.1.259",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.257",
+  "version": "0.1.258",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.255",
+  "version": "0.1.256",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.256",
+  "version": "0.1.257",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/src/__tests__/applyNetworkPolicyNoopGate.test.ts
+++ b/host-context-controller/src/__tests__/applyNetworkPolicyNoopGate.test.ts
@@ -1,7 +1,12 @@
 import { describe, expect, it, vi } from 'vitest'
+import { ApiException } from '@kubernetes/client-node'
 import type * as k8s from '@kubernetes/client-node'
 import { applyNetworkPolicy, networkPolicyMatchesDesired } from '../utils'
 import { asApiserverNetworkPolicy, updatedPolicyLogs } from './asApiserverNetworkPolicy'
+
+function apiException(code: number): ApiException<unknown> {
+  return new ApiException(code, 'test', {}, {})
+}
 
 function desiredPolicy(): k8s.V1NetworkPolicy {
   return {
@@ -99,7 +104,7 @@ describe('applyNetworkPolicy no-op gate', () => {
   })
 
   it.each([
-    { form: 'code', error: { code: 404 } },
+    { form: 'ApiException', error: apiException(404) },
     { form: 'statusCode', error: { response: { statusCode: 404 } } },
   ])('TOCTOU-NP-1: create 409 + read 404 ($form) returns without replace', async ({ error }) => {
     const api = fakeNetworkingApi()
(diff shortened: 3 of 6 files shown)


alfredolopez80:
@claude please perform a complete re-review and audit this PR for close on HEAD `fdaec664095a1acbd14bffb8eb6ab6a03cd13507` (base `dev`).

This is **only** #486 / #472 (NetworkPolicy conflict-retry read 404 = gone). Do **not** expand the review to #487, Host Role skip, `isUpToDate` (#471), the 409 ladder (#480), `hcc_writes_total`, `k8sClient.ts`, `metrics.ts`, `config.ts`, or `networkPolicyReconciler.ts`. Do **not** merge from this comment.

GitHub CI on this HEAD is **34/34 SUCCESS**, including `host-context-controller`, CodeQL `Analyze (javascript-typescript)`, Real Postgres, E2E Vitest node-unit, and `Shell script syntax (bash -n)`.

## Product (must stay — Option A)

- Scope is `replaceWithConflictRetry` in `host-context-controller/src/utils.ts` only.
- Wrap **only** the conflict-retry `read()`. `getErrorCode(err) === 404` → return (object is gone; do not treat as fatal).
- 403 / 500 / network `Error` without a code must still throw.
- **No** same-call recreate after a gone read.
- A 404 on the **write** (`replace`) remains fatal — do not swallow it.

This helper is shared (including `createSafetyPolicy`). Option A is intentional: a gone object on retry-read is success-enough; recreating in the helper would change every call site.

`hcc_writes_total` is intentionally **not** in this PR. That metric waits until #472 merges.

## Review-fix commit on this HEAD

`fdaec664` — **TOCTOU-NP-5:** a `replace` 404 (`{ code: 404 }` and `{ response: { statusCode: 404 }`) **still throws**. Pins Option A so a future `catch { return }` on the write path cannot stay green.

**TOCTOU-NP-3** now covers both 404 shapes on the second read.

## Pins to verify

- **TOCTOU-NP-1 / NP-2 (apply):** `applyNetworkPolicyNoopGate.test.ts` — apply path unchanged; this PR does not alter `isUpToDate`.
- **TOCTOU-NP-2 (helper) / NP-3 / NP-4 / NP-5:** `replaceWithConflictRetryOrder.test.ts`
  - NP-3: second-read 404 (`code` and `statusCode`) → return, not throw
  - NP-4: non-404 read failures still throw
  - NP-5: write-path 404 still throws

Please review product + tests + the write-404 pin as one change, and report any remaining merge blockers.

zach88:
Took a careful pass through this one. The actual change is tiny and well-scoped — wrapping only `read()` so a 404 returns cleanly instead of aborting the whole pass — and it holds up: I walked all 14 call sites of the shared `replaceWithConflictRetry`, and 13 of them are create-then-409 where a read-404 genuinely means "gone in the gap" and the next pass recreates. No caller loses its object permanently, and 403/500/no-code errors still throw. Two small semantic wrinkles worth a look, but nothing that blocks a merge.

**🟡 Approve with nits**

None — no blockers.

## Optional (non-blocking)

The PR is mergeable as-is; these are follow-up / author's call.

1. **[R1-L1] A swallowed read-404 lets a pass certify readiness while the deny-all is briefly absent.** `src/utils.ts:60-65` interacting with `src/networkPolicyReconciler.ts:1306` (`ensureDefaultPolicies`, deny-all L0) and `:1685` (the safety-inventory certification). `ensureDefaultPolicies()` runs *before* certification. If `deny-all-<ns>` gets deleted in the create-409→read window: on base the read-404 threw → the pass stayed `uncertified` (fail-closed, readiness held); with this change it returns cleanly → the pass **certifies readiness while the deny-all is still missing**, until the next resync recreates it. Reproduced by reading (two paths over the same object now handle 404 oppositely; only reproduces with the diff). Calibrated Low because the *network state* is identical to base (the deny-all is absent either way once it's deleted) — the diff only suppresses the fail-closed signal on the certification, and it self-heals within one resync pass. Suggestion: either make a read-404 on the *deny* path (defaults) keep the pass uncertified instead of swallowing it, leaving the silent return for the additive/allow lanes — or just document in the PR body that the ≤1-pass self-heal is a conscious call.

2. **[R1-L2] Two contiguous reads of the same object now handle 404 oppositely.** `src/hostReconciler.ts:2270-2297` (`reconcileChannelReaderDeployment`). The outer read at 2272 deliberately treats a 404 as a transient failure — `console.warn("…disappeared after create conflict; treating the reconcile as a transient failure")` + `throw` (requeue → prompt recreate). But the helper read at 2297 now returns silently on the same 404 in the same function, so the reconcile reports success without recreating this pass (it recreates on the host's next natural reconcile). Narrow window, self-heals. Worth a decision: either the helper read-404 also warns/requeues for parity with 2274, or drop a comment accepting the divergence — just don't leave two adjacent reads with contradictory 404 semantics and no note.

3. **[R1-N1] Error fixtures are plain objects, not `ApiException`.** In the new tests, `{ code: 404 }` / `{ code: 403 }` aren't `ApiException` instances (what the real v1.4.0 client emits: `extends Error` with a numeric `.code`). Real-path coverage is fine since everything flows through `getErrorCode`, which reads `.code` identically — but the `{ response: { statusCode } }` shape is a legacy form the current client no longer emits (harmless over-coverage). Strict fidelity would use `new ApiException(404, …)`.

4. **[R1-N2] `TOCTOU-NP-2` ID reused across two files.** `applyNetworkPolicyNoopGate.test.ts` and `replaceWithConflictRetryOrder.test.ts` both use `TOCTOU-NP-2` for different assertions — hurts traceability, worth renaming one.

## What checked out clean

- `npx tsc --noEmit` → exit 0; full suite → 85 files, **1842 passed**.
- T3 verified against the merge-base: `TOCTOU-NP-1` and `TOCTOU-NP-3` **fail** on base (they cover the actual bug), while `TOCTOU-NP-2/4/5` **pass** on base — those are pins of unchanged behavior (write-404 stays fatal, 403/500 still throw), not regression tests. Correct that they're there.
- All 14 callers of `replaceWithConflictRetry` audited: 13 are create-then-409 where the silent return is benign; the one that isn't (`scaleDeployment`) comes out *more* consistent with the fix.
- Recreation is actually guaranteed: `netPolResyncTimer` → `runInitialNetworkPolicyConvergence` (`k8sClient.ts:2864-2871`) plus event-driven convergence, so the NetPol fail-open is transient (≤1 resync), not persistent — the network state matches base throughout.
- PR body claims check out: the resync-recreates story is real, and tsc + tests pass as described.

alfredolopez80:
Cheap nits from https://github.com/evenfire-ai/evenfire/pull/486#issuecomment-5445059874 landed on `35f304c1`.

- **[R1-L1]** Conscious call, documented on the PR body. Network state matches base; certification can go green for ≤1 resync while `deny-all-<ns>` is gone. No deny-path special case in this PR.
- **[R1-L2]** Comment on the channel-reader helper call: outer read 404 still requeues this pass; helper 404 is the #472 return. Accepted split.
- **[R1-N1]** `{ code }` fixtures are now `new ApiException(code, …)`. `response.statusCode` stays as the extra `getErrorCode` shape.
- **[R1-N2]** Apply-path case renamed `TOCTOU-NP-2-APPLY`. Helper keeps `TOCTOU-NP-2`.

No product change to certification or to the helper 404 contract.

zach88 (approved):

Issue #307, opened 2026-08-08
WRC: unconditional WorkflowRecipe status patch churns apiserver/DB on every forced reconcile (amplified by #299 refresh loop)


alfredolopez80:
## Summary

The WorkflowRecipe reconciler patches CRD status **unconditionally** on every full reconcile — `statusPatch` sets top-level `lastTransitionTime: now` with no dirty-check against the current status (`workflow-recipes/src/reconciler/workflowRecipeReconciler.ts`, the `statusPatch` object ~L6098-6130). Every forced reconcile therefore writes status and emits a `MODIFIED` watch event even when phase/message/conditions are unchanged.

This is **pre-existing** behavior (identical in `dev`), already triggered by the pre-existing periodic timers (`refreshWorkloadStatusesForSteadyRecipes` ~30s, `refreshRuntimeCredentialsForInProgressRecipes` 5–30s). It is being tracked now because PR #306 (issue #299) **amplifies** it for a new population.

## What PR #306 changed (the amplification)

PR #306 added a TTL-aware external-egress refresh loop (`startExternalEgressRefreshLoop` / `refreshExternalEgressForActiveRecipes`) that force-reconciles every recipe with exact-host egress. Consequences:

- Recipes with external egress but **no** named Secrets and all-ready workloads previously had **no** periodic forced reconcile (0 status writes/min). They now get one per refresh tick.
- The refresh cadence is TTL-accelerated: `min(configured, minObservedTtl/2)` clamped at a 5s floor. Under GitHub's ~15s TTL that is ~7.5s → **~8 CRD status writes + ~8 `MODIFIED` watch events per minute per egress recipe** (vs ~2/min for the pre-existing worst case).
- Each status-only `MODIFIED` also runs `dbRunProcessor.syncFromRecipeExecution` (`workflow-recipes/src/k8sClient.ts` status-only path), so each tick is a DB touch too, not only an etcd write.

Note: this is **churn (apiserver/etcd/watch/DB load), not a correctness bug** — there is no feedback loop (the resulting `MODIFIED` is generation-unchanged, classified status-only, and short-circuits before reconcile). Egress enforcement is unaffected.

## Why it was deferred out of PR #306

[...]

PR #487, opened 2026-08-27
fix(hcc): skip no-op Host Role replaces

alfredolopez80:
## Summary
- `ensureHostRole` already read the live Role and then always called `replaceNamespacedRole`, logging `Updated Role` on every reconcile (~10–15/min in the field).
- Skip the PUT when authored `rbacLabels` and `rules` already match (label keys canonicalized; rules / verbs / resourceNames stay in author order). Server metadata and annotations are ignored so last-applied noise does not force a write.
- Does **not** migrate onto `replaceWithConflictRetry`. Migrating and unswallowing the fail-quiet update catch stay a follow-up (G5). Does not touch `utils.ts`.

Related to #460 (G5 Role skip).

## Test plan
- [x] NOOP-ROLE-1: create 409 + apiserver-shaped Role (uid/RV/managedFields, shuffled label keys, same rules) → replace ×0 and zero `Updated Role`. Authored rule keys differ from live `attributeTypeMap` order (canonicalize must run).
- [x] NOOP-ROLE-2: existing equivalent, desired `secretRef` rotated → replace ×1 and one `Updated Role`
- [x] NOOP-ROLE-3/4: extra live rule or extra secrets `update` verb → replace ×1
- [x] NOOP-ROLE-5: reordered live verbs → replace ×1 (converge, do not oscillate)
- [x] NOOP-ROLE-6: tampered `clerum.io/host` label value → replace ×1
- [x] NOOP-ROLE-7/8: removed live rule or removed secret resourceName → replace ×1
- [x] FIXTURE-ROLE-1: `asApiserverPolicyRule` key order matches `V1PolicyRule.attributeTypeMap`
- [x] KEEP `test/hostReconciler.test.ts` (`rewrites Role resourceNames when spec.secretRef changes`): read without `rules` still replaces
- [x] KEEP `hostDeploymentAnnotationMerge.test.ts` unchanged
- [x] `npx tsc --noEmit` and `npm test` in `host-context-controller` (1857 tests)

### DIFF
--- a/host-context-controller/src/__tests__/asApiserverRole.ts
+++ b/host-context-controller/src/__tests__/asApiserverRole.ts
@@ -5,9 +5,9 @@ import type * as k8s from '@kubernetes/client-node'
  * Recorded kube-apiserver GET of a namespaced Role (`kubectl get role -o json`
  * shape), not a field list copied from the comparator. Stamp
  * controller-owned name/labels/rules via `asApiserverRole`. Label keys are
- * inserted in reverse author order so a byte-identical fixture cannot hide a
- * missing canonicalize. Do not reconstruct rules from desired alone when
- * asserting the read carried `rules` — the skip must see them or it fail-opens.
+ * inserted in reverse author order, and each rule is rebuilt in
+ * V1PolicyRule.attributeTypeMap order, so a byte-identical fixture cannot hide
+ * a missing canonicalize. The skip must still see `rules` or it fail-opens.
  */
 export const RECORDED_ROLE: k8s.V1Role = {
   apiVersion: 'rbac.authorization.k8s.io/v1',
@@ -49,6 +49,20 @@ function shuffleLabelKeys(labels: Record<string, string>): Record<string, string
   return shuffled
 }
 
+/**
+ * Rebuild a rule in client-node ObjectSerializer / V1PolicyRule.attributeTypeMap
+ * order so a missing key-canonicalize cannot hide behind a cloned builder object.
+ */
+export function asApiserverPolicyRule(rule: k8s.V1PolicyRule): k8s.V1PolicyRule {
+  const live: Record<string, unknown> = {}
+  if (rule.apiGroups !== undefined) live.apiGroups = rule.apiGroups
+  if (rule.nonResourceURLs !== undefined) live.nonResourceURLs = rule.nonResourceURLs
+  if (rule.resourceNames !== undefined) live.resourceNames = rule.resourceNames
+  if (rule.resources !== undefined) live.resources = rule.resources
+  if (rule.verbs !== undefined) live.verbs = rule.verbs
+  return live as unknown as k8s.V1PolicyRule
+}
+
 export function asApiserverRole(desired: k8s.V1Role): k8s.V1Role {
   const recorded = structuredClone(RECORDED_ROLE)
   const labels = desired.metadata?.labels ?? {}
@@ -66,7 +80,7 @@ export function asApiserverRole(desired: k8s.V1Role): k8s.V1Role {
       ownerReferences: desired.metadata?.ownerReferences,
       selfLink: `/apis/rbac.authorization.k8s.io/v1/namespaces/${desired.metadata?.namespace}/roles/${desired.metadata?.name}`,
     },
-    rules: structuredClone(desired.rules ?? []),
+    rules: (desired.rules ?? []).map(asApiserverPolicyRule),
   }
 }
 
--- a/host-context-controller/src/__tests__/asApiserverRole.ts
+++ b/host-context-controller/src/__tests__/asApiserverRole.ts
@@ -52,6 +52,7 @@ function shuffleLabelKeys(labels: Record<string, string>): Record<string, string
 /**
  * Rebuild a rule in client-node ObjectSerializer / V1PolicyRule.attributeTypeMap
  * order so a missing key-canonicalize cannot hide behind a cloned builder object.
+ * FIXTURE-ROLE-1 pins this order against the live client-node class.
  */
 export function asApiserverPolicyRule(rule: k8s.V1PolicyRule): k8s.V1PolicyRule {
   const live: Record<string, unknown> = {}
--- a/host-context-controller/src/__tests__/asApiserverRole.ts
+++ b/host-context-controller/src/__tests__/asApiserverRole.ts
@@ -1,0 +1,77 @@
+import { vi } from 'vitest'
+import type * as k8s from '@kubernetes/client-node'
+
+/**
+ * Recorded kube-apiserver GET of a namespaced Role (`kubectl get role -o json`
+ * shape), not a field list copied from the comparator. Stamp
+ * controller-owned name/labels/rules via `asApiserverRole`. Label keys are
+ * inserted in reverse author order so a byte-identical fixture cannot hide a
+ * missing canonicalize. Do not reconstruct rules from desired alone when
+ * asserting the read carried `rules` — the skip must see them or it fail-opens.
+ */
+export const RECORDED_ROLE: k8s.V1Role = {
+  apiVersion: 'rbac.authorization.k8s.io/v1',
+  kind: 'Role',
+  metadata: {
+    annotations: {
+      'kubectl.kubernetes.io/last-applied-configuration': '{}',
+    },
+    creationTimestamp: new Date('2026-04-01T00:00:00.000Z'),
+    generation: 1,
+    managedFields: [
+      {
+        apiVersion: 'rbac.authorization.k8s.io/v1',
+        fieldsType: 'FieldsV1',
+        fieldsV1: { 'f:rules': {} },
+        manager: 'kube-apiserver',
+        operation: 'Update',
+        time: new Date('2026-04-01T00:00:00.000Z'),
+      },
+    ],
+    name: 'recorded-role',
+    namespace: 'recorded-ns',
+    resourceVersion: '1784167',
+    uid: 'aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee',
+    selfLink: '/apis/rbac.authorization.k8s.io/v1/namespaces/recorded-ns/roles/recorded-role',
+    labels: {
+      'clerum.io/host': 'recorded',
+      'clerum.io/managed-by': 'host-context-controller',
+    },
+  },
+  rules: [],
+}
+
+function shuffleLabelKeys(labels: Record<string, string>): Record<string, string> {
+  const shuffled: Record<string, string> = {}
+  for (const key of Object.keys(labels).reverse()) {
+    shuffled[key] = labels[key]
+  }
+  return shuffled
+}
+
+export function asApiserverRole(desired: k8s.V1Role): k8s.V1Role {
+  const recorded = structuredClone(RECORDED_ROLE)
+  const labels = desired.metadata?.labels ?? {}
+  return {
+    ...recorded,
+    metadata: {
+      ...recorded.metadata,
+      name: desired.metadata?.name,
+      namespace: desired.metadata?.namespace,
+      labels: shuffleLabelKeys({ ...labels }),
+      annotations: {
+        ...recorded.metadata?.annotations,
+        ...desired.metadata?.annotations,
+      },
+      ownerReferences: desired.metadata?.ownerReferences,
+      selfLink: `/apis/rbac.authorization.k8s.io/v1/namespaces/${desired.metadata?.namespace}/roles/${desired.metadata?.name}`,
+    },
+    rules: structuredClone(desired.rules ?? []),
+  }
+}
+
+export function updatedRoleLogs(log: ReturnType<typeof vi.spyOn>, needle: string): string[] {
+  return log.mock.calls
+    .map((call: unknown[]) => String(call[0]))
+    .filter((line: string) => line.includes('Updated Role') && line.includes(needle))
+}
--- a/host-context-controller/src/__tests__/hostRoleNoopGate.test.ts
+++ b/host-context-controller/src/__tests__/hostRoleNoopGate.test.ts
@@ -165,19 +165,36 @@ describe('Host ensureHostRole no-op gate', () => {
     }
   })
 
-  it('NOOP-ROLE-5: reordered live verbs still replaces once', async () => {
+  it('NOOP-ROLE-5: reordered live verbs replace once then converge', async () => {
     rbacApi.createNamespacedRole.mockRejectedValue({ code: 409 })
     let existing: k8s.V1Role | undefined
+    let readPass = 0
     rbacApi.readNamespacedRole.mockImplementation(() => {
-      const live = structuredClone(desiredFromCreate(rbacApi))
-      secretGetRule(live).verbs = ['watch', 'get', 'list']
-      existing = asApiserverRole(live)
+      readPass += 1
+      if (readPass === 1) {
+        const live = structuredClone(desiredFromCreate(rbacApi))
+        secretGetRule(live).verbs = ['watch', 'get', 'list']
+        existing = asApiserverRole(live)
+        return Promise.resolve(existing)
+      }
+      const written = rbacApi.replaceNamespacedRole.mock.calls[0]?.[0].body as
+        | k8s.V1Role
+        | undefined
+      if (!written) throw new Error('NOOP-ROLE-5: second read expected a written Role body')
+      existing = asApiserverRole(structuredClone(written))
       return Promise.resolve(existing)
     })
     const log = vi.spyOn(console, 'log')
     try {
       await (reconciler as any).ensureHostRole(host)
       expect(secretGetRule(existing!).verbs).toEqual(['watch', 'get', 'list'])
+      expect(rbacApi.replaceNamespacedRole).toHaveBeenCalledOnce()
+      expect(updatedRoleLogs(log, '"host-chatllm-config-reader"')).toEqual([
+        '[HostReconciler] Updated Role "host-chatllm-config-reader"',
+      ])
+
+      await (reconciler as any).ensureHostRole(host)
+      expect(secretGetRule(existing!).verbs).toEqual(['get', 'watch', 'list'])
       expect(rbacApi.replaceNamespacedRole).toHaveBeenCalledOnce()
       expect(updatedRoleLogs(log, '"host-chatllm-config-reader"')).toEqual([
         '[HostReconciler] Updated Role "host-chatllm-config-reader"',
--- a/host-context-controller/src/__tests__/hostRoleNoopGate.test.ts
+++ b/host-context-controller/src/__tests__/hostRoleNoopGate.test.ts
@@ -36,9 +36,7 @@ function makeHost(secretRef = 'secret'): HostCRD {
 }
 
 function desiredFromCreate(rbacApi: MockRbacApi): k8s.V1Role {
-  const body = rbacApi.createNamespacedRole.mock.calls[0][0].body as k8s.V1Role
-  expect(body.rules?.length).toBeGreaterThan(0)
-  return body
+  return rbacApi.createNamespacedRole.mock.calls[0][0].body as k8s.V1Role
 }
 
 describe('Host ensureHostRole no-op gate', () => {
@@ -66,14 +64,23 @@ describe('Host ensureHostRole no-op gate', () => {
 
   it('NOOP-ROLE-1: apiserver-shaped Role with shuffled label keys skips replace', async () => {
     rbacApi.createNamespacedRole.mockRejectedValue({ code: 409 })
+    let existing: k8s.V1Role | undefined
     rbacApi.readNamespacedRole.mockImplementation(() => {
-      const existing = asApiserverRole(desiredFromCreate(rbacApi))
-      expect(existing.rules?.length).toBeGreaterThan(0)
+      existing = asApiserverRole(desiredFromCreate(rbacApi))
       return Promise.resolve(existing)
     })
     const log = vi.spyOn(console, 'log')
     try {
       await (reconciler as any).ensureHostRole(host)
+      expect(rbacApi.readNamespacedRole).toHaveBeenCalledOnce()
+      expect(existing?.rules?.length).toBeGreaterThan(0)
+      expect(Object.keys(existing!.rules![0] ?? {})).toEqual([
+        'apiGroups',
+        'resourceNames',
+        'resources',
+        'verbs',
+      ])
+      expect(desiredFromCreate(rbacApi).rules?.length).toBeGreaterThan(0)
       expect(rbacApi.replaceNamespacedRole).not.toHaveBeenCalled()
       expect(updatedRoleLogs(log, '"host-chatllm-config-reader"')).toEqual([])
     } finally {
@@ -83,20 +90,20 @@ describe('Host ensureHostRole no-op gate', () => {
 
   it('NOOP-ROLE-2: rotated secretRef still replaces once', async () => {
     rbacApi.createNamespacedRole.mockRejectedValue({ code: 409 })
+    let existing: k8s.V1Role | undefined
     rbacApi.readNamespacedRole.mockImplementation(() => {
       const live = structuredClone(desiredFromCreate(rbacApi))
       const secretRule = live.rules?.find(
         (rule: k8s.V1PolicyRule) => rule.resources?.[0] === 'secrets' && rule.verbs?.includes('get')
       )
-      expect(secretRule?.resourceNames?.[0]).toBe('rotated-secret')
       secretRule!.resourceNames![0] = host.spec.secretRef
-      const existing = asApiserverRole(live)
-      expect(existing.rules?.length).toBeGreaterThan(0)
+      existing = asApiserverRole(live)
       return Promise.resolve(existing)
     })
     const log = vi.spyOn(console, 'log')
     try {
       await (reconciler as any).ensureHostRole(makeHost('rotated-secret'))
+      expect(existing?.rules?.length).toBeGreaterThan(0)
       expect(rbacApi.replaceNamespacedRole).toHaveBeenCalledOnce()
       expect(updatedRoleLogs(log, '"host-chatllm-config-reader"')).toEqual([
         '[HostReconciler] Updated Role "host-chatllm-config-reader"',
--- a/host-context-controller/src/__tests__/hostRoleNoopGate.test.ts
+++ b/host-context-controller/src/__tests__/hostRoleNoopGate.test.ts
@@ -1,0 +1,108 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import type * as k8s from '@kubernetes/client-node'
+import {
+  type MockRbacApi,
+  asAppsApi,
+  asCoreApi,
+  asCustomApi,
+  asNetworkingApi,
+  asRbacApi,
+  createMockAppsApi,
+  createMockCoreApi,
+  createMockCustomApi,
+  createMockNetworkingApi,
+  createMockRbacApi,
+} from '../../test/__fixtures__/testMocks'
+import { HostReconciler } from '../hostReconciler'
+import type { HostCRD } from '../types'
+import { asApiserverRole, updatedRoleLogs } from './asApiserverRole'
+
+function makeStubKc(): k8s.KubeConfig {
+  const stub = new Proxy({}, { get: () => vi.fn() })
+  return { makeApiClient: () => stub } as unknown as k8s.KubeConfig
+}
+
+function makeHost(secretRef = 'secret'): HostCRD {
+  return {
+    name: 'chatllm',
+    namespace: 'mcp-host',
+    uid: 'chatllm-uid',
+    spec: {
+      host: 'chatllm',
+      contextRef: 'ctx',
+      secretRef,
+    },
+  }
+}
+
+function desiredFromCreate(rbacApi: MockRbacApi): k8s.V1Role {
+  const body = rbacApi.createNamespacedRole.mock.calls[0][0].body as k8s.V1Role
+  expect(body.rules?.length).toBeGreaterThan(0)
+  return body
+}
+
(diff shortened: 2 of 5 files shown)


alfredolopez80:
@claude please perform a complete re-review and audit this PR for close on HEAD `f6a2a1576dd7b6472eccf1af80089ab36ddab4f0` (base `dev`).

This is **only** #487 / #460 G5 (Host Role no-op skip). Do **not** expand the review to #472, #486, `replaceWithConflictRetry`, `utils.ts`, NetworkPolicy apply, or `hcc_writes_total`. Do **not** merge from this comment.

GitHub CI on this HEAD is **34/34 SUCCESS**, including `host-context-controller`, CodeQL `Analyze (javascript-typescript)`, Real Postgres, E2E Vitest node-unit, and `Shell script syntax (bash -n)`.

## Product (must stay)

- `ensureHostRole` already read the live Role, then always `replaceNamespacedRole` and logged `Updated Role` on every reconcile.
- Skip the PUT when authored `rbacLabels` and `rules` already match. Server metadata and **annotations are ignored** so last-applied noise does not force a write.
- **Do not** migrate this writer onto `replaceWithConflictRetry`. Migrating and unswallowing the fail-quiet update catch stay a follow-up so this PR can revert independently of the 404/`isUpToDate` work on that helper.
- Do not touch `utils.ts`.
- Fail-open-to-write when the GET has no `rules` (existing `hostReconciler.test.ts` case: `rewrites Role resourceNames when spec.secretRef changes`).

## Review-fix commit on this HEAD

`f6a2a157` — canonicalize Role **rule object keys** before the skip compare (`canonicalizeRoleValue`).

client-node `ObjectSerializer` rebuilds `V1PolicyRule` as `apiGroups, resourceNames, resources, verbs`. A raw `JSON.stringify` of authored vs live GET never matched, so the skip was dead and every reconcile still wrote. Arrays (`verbs`, `resourceNames`, `apiGroups`, `resources`) stay in **author order** — do not sort them.

The live-GET fixture (`asApiserverRole`) must rebuild `rules` in that key order. Assertions in `hostRoleNoopGate.test.ts` must stay **outside** the fail-quiet catch (otherwise a swallowed replace looks like a skip).

## Pins to verify

- **NOOP-ROLE-1:** create 409 + apiserver-shaped Role (uid/RV/managedFields, shuffled **label** keys, same rules) → replace ×0 and zero `Updated Role`. A GET without `rules` must fail-open (replace).
- **NOOP-ROLE-2:** existing equivalent Role, desired `secretRef` rotated → replace ×1 and one `Updated Role`.
- KEEP `test/hostReconciler.test.ts` (`rewrites Role resourceNames when spec.secretRef changes`).
- KEEP `hostDeploymentAnnotationMerge.test.ts` unchanged.

Please review product + tests + the key-canonicalize fix as one change, and report any remaining merge blockers.

zach88:
Took a deep pass through this one. It's a small, focused change and it holds up well: the no-op gate is correct and safe in the direction that matters. I actively tried to make `roleMatchesDesired` skip a write it shouldn't (rotated secretRef, extra/removed rule, extra/reordered verb, label-value tamper, empty/undefined rules) and every genuine drift still forces a `replaceNamespacedRole`. The only things I'd flag are test-coverage gaps, not code defects.

**🟡 Approve with nits**
- **Optional / non-blocking: 4**

## Optional (non-blocking)

The PR can merge without these. My suggestion is to fold the two test cases below in now since they're the same file and cover an invariant + a security boundary that currently rests on prose and comparator symmetry — but that's your call.

1. **[R1-M1] Convergence invariant has no test.** The body claims "if an older HCC wrote a different order, the gate fail-opens once then converges, doesn't oscillate," but no test exercises a reordered `verbs`/`resourceNames`. I verified the behavior by throwaway probe — a reordered verb (`['watch','get','list']` vs authored `['get','watch','list']`) triggers exactly one replace, after which HCC rewrites its fixed order and matches — so the invariant is real, it just only lives in prose. Add a reordered-verb case in `src/__tests__/hostRoleNoopGate.test.ts` asserting `replaceNamespacedRole` is called once.

2. **[R1-M2] Negative cases miss label-value drift and rule removal.** NOOP-ROLE-2/3/4 cover value rotation, rule addition, and verb addition, but not (a) a tampered label *value* — e.g. `clerum.io/host` changed to another host's name, which is exactly the "Host A must not read Host B" boundary the code comment guards — or (b) a removed rule / removed resourceName (the opposite direction from NOOP-ROLE-3's addition). I confirmed by probe both force a replace, so the code is right; the label-value case in particular deserves an explicit test rather than relying on the comparator's symmetry. Same file as #1.

3. **[R1-L1] Hand-built apiserver fixture.** `src/__tests__/asApiserverRole.ts` types `RECORDED_ROLE`/`asApiserverRole` by hand instead of deriving it from a real client-node deserialization. I judged it faithful and non-masking for this Role: the comparator sorts object keys on both sides, so the deliberately-shuffled label keys and attributeTypeMap rule-key order can't produce a false green in the missed-write direction, and the apiserver stores RBAC `labels`/`rules` verbatim. The only residual risk is the fixture silently drifting from real client-node output over time with no guard. Fine to merge; just noting it.

4. **[R1-L2] Circular assertion.** `src/__tests__/hostRoleNoopGate.test.ts:79` — `expect(Object.keys(existing!.rules![0])).toEqual([...])` asserts the shape `asApiserverPolicyRule` just produced (the test's own builder), not any real producer. Harmless; the load-bearing assertion is `replaceNamespacedRole not called`.

## What checked out clean

- **Every RBAC escalation still forces a replace** (reproduced, not assumed): resourceNames removed (grants read on all secrets), `verbs:['*']` (pod could rotate its own creds), extra cross-Host resourceName, `resources:['*']`, dropped `managed-by`/`host` label. Fail-open never fail-closes — missing rules/labels or any compare exception returns false → replace.
- **Rotated secretRef caught** — it lives in `rules[secrets].resourceNames[0]`, array order preserved, so a single-element change yields different canonical JSON → replace.
- **No convergence stall / oscillation** — HCC always writes a fixed field order, so any drift is one replace then matches.
- **HCC authors only name/namespace/labels/rules** on this Role (`hostReconciler.ts:906-940`), so ignoring server metadata/annotations hides nothing.
- **T3 has teeth** — ran the new test file against the merge-base: NOOP-ROLE-1 fails there (replace called, no gate) and passes at head, so the regression test genuinely covers the gate.
- **Suites/typecheck**: full service suite 1852 passed, `tsc --noEmit` clean. Minor: the body says "1830 tests" — real count at head is 1852 (the #486 sync added tests), worth updating.

alfredolopez80:
Medium + cheap Lows from https://github.com/evenfire-ai/evenfire/pull/487#issuecomment-5445573314 landed on `9517603d`.

- **[R1-M1]** NOOP-ROLE-5: live verbs `watch,get,list` vs authored `get,watch,list` → replace ×1.
- **[R1-M2]** NOOP-ROLE-6 tampered `clerum.io/host`; NOOP-ROLE-7 removed configmaps rule; NOOP-ROLE-8 removed one secret `resourceName` → each replace ×1.
- **[R1-L1]** FIXTURE-ROLE-1 pins fixture key order to live `V1PolicyRule.attributeTypeMap`.
- **[R1-L2]** NOOP-ROLE-1 now asserts authored keys ≠ live keys (not just the fixture's own output).
- Test-plan count updated: **1857**.

No product change.

alfredolopez80:
@claude please perform a **closing re-review and audit** of this PR on HEAD `9517603d322ce161300655f2bc9f5e22ac2e6895` (base `dev`).

This is **only** #487 / #460 G5 (Host Role no-op skip). Do **not** expand the review to #472, #486, `replaceWithConflictRetry`, `utils.ts`, NetworkPolicy apply, or `hcc_writes_total`. Do **not** merge from this comment.

GitHub CI on this HEAD is **34/34 SUCCESS**, including `host-context-controller`, CodeQL `Analyze (javascript-typescript)`, Real Postgres, E2E Vitest node-unit, and `Shell script syntax (bash -n)`.

Review the **full delta versus `dev`**, not only the latest commit. The latest commit (`9517603d`) is tests-only (no product change).

Your previous review on `f6a2a157` said ship-it. Since then three commits landed: `ee1b5f94` (NOOP-ROLE-3/4), `47af1037` (sync with merged #486), `9517603d` (zach88 Medium + cheap Lows). Re-audit the whole change as one unit.

## Required output format

Do **not** give only a global verdict. For **every** item below, reply with one line:

`ID — PASS | FAIL | STILL-OPEN — <one-sentence evidence from HEAD>`

If any item is FAIL or STILL-OPEN, say whether it is a merge blocker. End with a closing verdict: ship / do-not-ship, and list remaining blockers if any.

---

## Product constraints (must still hold)

- **P1** `ensureHostRole` already read the live Role, then always `replaceNamespacedRole` and logged `Updated Role` on every reconcile. The gate sits after the live GET and before the PUT.
- **P2** Skip the PUT only when authored `rbacLabels` and `rules` already match. Server metadata and **annotations are ignored** so last-applied noise does not force a write.
- **P3** Do **not** migrate this writer onto `replaceWithConflictRetry`. Do **not** touch `utils.ts`. `resourceVersion` is still stamped on the write path.
- **P4** Fail-open-to-write when the GET has no `rules` / no labels, or the compare throws (KEEP `test/hostReconciler.test.ts`: `rewrites Role resourceNames when spec.secretRef changes`).
- **P5** Security: a rotated `secretRef`, extra rule, extra verb, removed rule, removed `resourceName`, or tampered `clerum.io/host` must still replace. The gate must never suppress a needed resourceName rewrite (Host A must not read Host B).
- **P6** `canonicalizeRoleValue` (`f6a2a157`) sorts **object keys only**. Arrays (`verbs`, `resourceNames`, `apiGroups`, `resources`) stay in **author order** — do not sort them. client-node `ObjectSerializer` rebuilds `V1PolicyRule` as `apiGroups, resourceNames, resources, verbs`.
- **P7** Assertions in `hostRoleNoopGate.test.ts` stay **outside** the fail-quiet catch (a swallowed replace still records a mock call).
- **P8** KEEP `hostDeploymentAnnotationMerge.test.ts` unchanged in this delta.

---

## Prior Claude notes on `f6a2a157` — re-verify on this HEAD

- **C1** Fail-open is airtight (`!desired.rules`, `!existing.rules`, missing labels, or throw → write).
- **C2** Skip is true only when canonicalized `rules` JSON is byte-equal.
- **C3** Key-canonicalize is load-bearing; without it the skip is dead.
- **C4** `undefined` keys (e.g. `nonResourceURLs`) drop out symmetrically on both sides.
- **C5** No duplicate helper defs elsewhere in `src`; no scope leakage into #472/#486 helpers.
- **C6** Extra third-party label: gate does not engage and HCC still writes (same as pre-PR always-write). Confirm still true and still non-blocking.

---

## zach88 findings — validate each one closed on `9517603d`

Source: https://github.com/evenfire-ai/evenfire/pull/487#issuecomment-5445573314

- **[R1-M1]** Convergence invariant now has a test. **NOOP-ROLE-5:** live secrets verbs `watch,get,list` vs authored `get,watch,list` → `replaceNamespacedRole` ×1 and one `Updated Role`. Confirm it fail-opens once then would converge (HCC writes fixed author order; no oscillation). Confirm the test would go red if the comparator sorted verbs.
- **[R1-M2a]** **NOOP-ROLE-6:** tampered live `clerum.io/host` (other host name) → replace ×1. This is the Host A / Host B isolation boundary.
- **[R1-M2b]** **NOOP-ROLE-7:** removed live configmaps rule → replace ×1 (opposite of NOOP-ROLE-3 addition).
- **[R1-M2c]** **NOOP-ROLE-8:** removed one live secret `resourceName` → replace ×1 (privilege expansion if skipped).
- **[R1-L1]** **FIXTURE-ROLE-1:** `asApiserverPolicyRule` key order is pinned to live `V1PolicyRule.attributeTypeMap` (`apiGroups, nonResourceURLs, resourceNames, resources, verbs`). Confirm the fixture cannot silently drift from client-node without this test failing.
- **[R1-L2]** Circular-assertion fix: **NOOP-ROLE-1** now asserts authored keys `['apiGroups','resources','resourceNames','verbs']` **≠** live keys `['apiGroups','resourceNames','resources','verbs']`, not only the fixture's own output. Confirm this proves canonicalize must run.

---

## Pins already on the PR (re-verify each)

- **CREATE-ROLE-1:** successful create never reads or replaces.
- **NOOP-ROLE-1:** create 409 + apiserver-shaped Role (uid/RV/managedFields, shuffled **label** keys, same rules, last-applied annotation) → replace ×0 and zero `Updated Role`. GET without `rules` must still fail-open (KEEP case).
- **NOOP-ROLE-2:** existing equivalent Role, desired `secretRef` rotated → replace ×1 and one `Updated Role`.
- **NOOP-ROLE-3:** extra live rule → replace ×1 (`ee1b5f94`).
- **NOOP-ROLE-4:** extra live secrets `update` verb → replace ×1 (`ee1b5f94`).
- **KEEP-1** `test/hostReconciler.test.ts` (`rewrites Role resourceNames when spec.secretRef changes`) still present and still fail-opens.
- **KEEP-2** `hostDeploymentAnnotationMerge.test.ts` unchanged.
- **COUNT** Test-plan count is **1857**. Confirm the body matches HEAD.

---

## zach88 “checked out clean” — re-prove, do not assume

- **Z1** Every RBAC escalation still forces a replace: removed resourceNames (all-secrets get), `verbs:['*']`, extra cross-Host resourceName, `resources:['*']`, dropped `managed-by`/`host` label. Fail-open never fail-closes.
- **Z2** Rotated secretRef still lives in `rules[secrets].resourceNames[0]` and still replaces.
- **Z3** No convergence stall / oscillation: HCC always writes a fixed field order.
- **Z4** HCC authors only name/namespace/labels/rules on this Role (`ensureHostRole` body), so ignoring server metadata/annotations hides nothing.

Please review product + tests + canonicalize + each finding above as one change, and report any remaining merge blockers.

zach88:
Took another pass now that the test-hardening commit landed. This round is basically pure test coverage — `hostReconciler.ts` is byte-identical to the head I reviewed last round, so there's no behavior surface to regress. The four items I flagged in round 1 are all closed, and I verified the closures by mutation testing (patching the gate / dropping canonicalization) rather than just reading the asserts, so they're load-bearing, not green-by-accident. Nice work on NOOP-ROLE-6 — the host-label isolation case genuinely discriminates on the `clerum.io/host` value.

**🟡 Approve with nits**
- Optional / non-blocking: **2**

## Optional (non-blocking)

This PR can merge as-is. Both of these are follow-up polish on the tests, your call.

1. **[R2-L1] NOOP-ROLE-5 covers order-sensitivity but not the "converges" fixpoint** — `src/__tests__/hostRoleNoopGate.test.ts:168`. R1-M1 asked for "reordered verb → one replace → *converges*". The test nails the first half (live verb order ≠ author order forces exactly one replace) but never feeds the written body back through a second reconcile to assert the next pass is a no-op. What it actually proves is order-*sensitivity* — which is also the latent hot-loop risk if the apiserver ever reordered RBAC verbs. The "converges" clause is only covered implicitly by pairing with NOOP-ROLE-1 (HCC writes author order → next read skips). Either rename it to reflect order-sensitivity, or add a second reconcile pass over the just-written body asserting 0 replaces.

2. **[R2-L2] FIXTURE-ROLE-1 pins the declared map order, not a real serializer round-trip** — `src/__tests__/asApiserverRole.ts:57`, `src/__tests__/hostRoleNoopGate.test.ts:258`. `asApiserverPolicyRule` is still a hand-written key ordering; FIXTURE-ROLE-1 asserts its key list equals `V1PolicyRule.getAttributeTypeMap()`. Two things: (a) it never invokes client-node's `ObjectSerializer` to derive the fixture, so it pins the *declared* map order rather than the *emitted* order (small residual T1 gap); (b) it couples to a client-node internal that production doesn't consume — `canonicalizeRoleValue` sorts keys alphabetically, it doesn't depend on `attributeTypeMap` — so a minor `@kubernetes/client-node` bump that reorders that map would break CI without the gate being broken. If you want strict T1, build the fixture by round-tripping a `V1Role` through `ObjectSerializer.serialize/deserialize` instead of the manual if-ladder.

## What checked out clean

- Production (`hostReconciler.ts`) unchanged since last round — the RBAC gate that reproduced clean against 6 escalation probes in round 1 didn't move.
- All 5 new/reworked tests are load-bearing on the observable output (`replaceNamespacedRole` count + `Updated Role` log), not intermediate effects. Bypassing the gate (`if (true) return`) turns NOOP-ROLE-2..8 red; dropping key canonicalization turns only NOOP-ROLE-1 red — so the rewritten NOOP-ROLE-1 is genuinely non-circular now.
- Fixtures derived from the real builder (`desiredFromCreate` → `asApiserverRole`), not hand-authored (bar the R2-L2 residual).
- Version bump 0.1.259 → 0.1.260, no collision with dev.
- `npx vitest run src/__tests__/hostRoleNoopGate.test.ts` → 10 passed. `npx tsc --noEmit` → clean.

alfredolopez80:
@zach88 thanks — adversarial pass on both nits. Neither was a merge blocker (agreed). Landed the cheap valid one; left the other.

**[R2-L1] valid, landed in `bcbe6a41`.** The first half (reordered verbs → replace ×1) was already load-bearing. The “converges” clause was only implicit via NOOP-ROLE-1. NOOP-ROLE-5 now does a second `ensureHostRole` and feeds `replaceNamespacedRole`’s written body back through `asApiserverRole`. Pass 2 must be replace ×0, one `Updated Role` total, and live verbs back to authored `get,watch,list`. Still red if the comparator sorted verbs (both passes would skip).

**[R2-L2] not applying.** The residual is real as prose, not as a cheap T1 upgrade:

- Production `canonicalizeRoleValue` sorts object keys alphabetically. It does not read `attributeTypeMap` or `ObjectSerializer`.
- In `@kubernetes/client-node@1.4.0`, serialize walks `attributeTypeMap` in declaration order. Pinning the map **is** pinning the emitted key order; declared≠emitted is not a gap in this generator.
- `ObjectSerializer` is **not** on the public package export (`dist/index.d.ts`). A deep import would couple the fixture to a generated internal, which is worse than `V1PolicyRule.getAttributeTypeMap()` on the public class.
- A client-node bump that reorders the map would go red here **and** would change the live GET shape the fixture is meant to mimic. That false-red is the alarm that the hand-built if-ladder drifted, not a broken gate. NOOP-ROLE-1 already asserts authored keys ≠ live keys, so canonicalize cannot go dead behind a map reorder that happens to match author order (that case would fail the `not.toEqual` first).

No product change. Not merging from this comment — still waiting on a formal Approve, and public image-building merges are frozen until the /metrics ≥2-tick sample (~14:19Z).

zach88 (approved):

host-context-controller/src/hostReconciler.ts
```diff
@@ -4651,6 +4652,56 @@ export class HostReconciler {
...
+function canonicalizeLabelKeys(labels: Record<string, string>): Record<string, string> {
+  const canonical: Record<string, string> = {}
+  for (const key of Object.keys(labels).sort()) {
+    canonical[key] = labels[key]
+  }
+  return canonical
+}
+
+/** Sort object keys recursively. Arrays keep author order. */
+function canonicalizeRoleValue(value: unknown): unknown {
```
jozer-rami, line 4694: **Fifth copy of this function.** Identical body to `normalizeDeploymentValue` at line 4873 of this same file, `normalizeServiceValue` in `utils.ts:242`, and both `canonicalizeNetworkPolicyValue` (`networkPolicyReconciler.ts:123`) and `canonicalize` (`networkPolicyReconciler.ts:327`).

Five call sites now have to change in lockstep whenever the canonical form does (learning to preserve `Date`, handling a new apiserver default). That lockstep is the #307-class drift that makes a no-op gate silently stop skipping.

The PR body says "does not touch `utils.ts`" as a scope guard, but exporting one `canonicalizeValue` there and importing it is a *smaller* diff than this copy, and it puts `roleMatchesDesired` next to `serviceMatchesDesired`/`networkPolicyMatchesDesired` instead of leaving it as the one gate living in `hostReconciler.ts`.

Minor, while you're here: `if (entry !== undefined)` on the next line is dead - `JSON.stringify`, the only consumer, already drops undefined-valued keys. It's inherited verbatim from the siblings where it's equally dead.

host-context-controller/src/hostReconciler.ts
```diff
@@ -4651,6 +4652,56 @@ export class HostReconciler {
...
+    return (
+      JSON.stringify(canonicalizeRoleValue(desired.rules)) ===
+      JSON.stringify(canonicalizeRoleValue(existing.rules))
+    )
+  } catch {
+    return false
+  }
+}
+
+function canonicalizeLabelKeys(labels: Record<string, string>): Record<string, string> {
```
jozer-rami, line 4685: `canonicalizeLabelKeys` is redundant with `canonicalizeRoleValue`, which already sorts object keys recursively. Labels are a flat `Record<string, string>`, so `canonicalizeRoleValue(labels)` produces the identical result.

As written, `roleMatchesDesired` needs two different canonicalizers to compare two fields of the same object. Dropping this and comparing `{ labels: desired.metadata.labels, rules: desired.rules }` through a single canonicalize and a single `JSON.stringify` removes the function and one of the two code paths.

host-context-controller/src/hostReconciler.ts
```diff
@@ -4651,6 +4652,56 @@ export class HostReconciler {
...
+ * a live GET (#307). Rule arrays, verbs, and resourceNames stay in author order.
+ * Missing rules or labels, or any compare failure, returns false
+ * (fail-open-to-write).
+ */
+function roleMatchesDesired(desired: k8s.V1Role, existing: k8s.V1Role): boolean {
+  try {
+    if (!desired.rules || !existing.rules) return false
+    if (!desired.metadata?.labels || !existing.metadata?.labels) return false
+    if (
+      JSON.stringify(canonicalizeLabelKeys(desired.metadata.labels)) !==
```
jozer-rami, line 4671: **Exact-equality on the whole label map means the gate misses on any cluster with label-injecting admission.**

A live Role carrying one extra label (Kyverno/Gatekeeper `add-labels`, ArgoCD `app.kubernetes.io/instance`, a cost-allocation tag) mismatches here. HCC then PUTs a body whose labels are only the two `rbacLabels` keys, stripping the injected label; the webhook re-adds it; the next reconcile mismatches again. Those Hosts keep the full ~10-15 writes/min this PR exists to remove, and the label oscillates.

Not a regression - the strip-and-re-add loop predates the gate, since the replace was unconditional. But it's the difference between the fix landing everywhere and landing only on clusters with no mutating webhooks.

Comparing just `MANAGED_BY_LABEL` and `HOST_LABEL` - the same subset `isHccOwnedHostResource` already keys on - lets the gate fire and leaves the foreign label alone.

host-context-controller/src/__tests__/asApiserverRole.ts
```diff
@@ -0,0 +1,92 @@
...
+ * order so a missing key-canonicalize cannot hide behind a cloned builder object.
+ * FIXTURE-ROLE-1 pins this order against the live client-node class.
+ */
+export function asApiserverPolicyRule(rule: k8s.V1PolicyRule): k8s.V1PolicyRule {
+  const live: Record<string, unknown> = {}
+  if (rule.apiGroups !== undefined) live.apiGroups = rule.apiGroups
+  if (rule.nonResourceURLs !== undefined) live.nonResourceURLs = rule.nonResourceURLs
+  if (rule.resourceNames !== undefined) live.resourceNames = rule.resourceNames
+  if (rule.resources !== undefined) live.resources = rule.resources
+  if (rule.verbs !== undefined) live.verbs = rule.verbs
```
jozer-rami, line 63: These are reference copies, not clones, so the "live" Role shares nested array instances with the desired one: `existing.rules[i].verbs` **is** `body.rules[i].verbs`. Combined with `desiredFromCreate` handing back the reconciler's live `body` object (which `ensureHostRole` later mutates via `body.metadata!.resourceVersion = ...`), NOOP-ROLE-1 partly compares an object graph to itself.

Harmless today because `roleMatchesDesired` is pure, but a comparator that sorted, deduped, or normalized in place would compare equal to itself and the test would pass. `structuredClone` the desired Role before stamping it into apiserver shape.

host-context-controller/src/__tests__/asApiserverRole.ts
```diff
@@ -0,0 +1,92 @@
...
+    selfLink: '/apis/rbac.authorization.k8s.io/v1/namespaces/recorded-ns/roles/recorded-role',
+    labels: {
+      'clerum.io/host': 'recorded',
+      'clerum.io/managed-by': 'host-context-controller',
+    },
+  },
+  rules: [],
+}
+
+function shuffleLabelKeys(labels: Record<string, string>): Record<string, string> {
```
jozer-rami, line 44: Reversing only differs from sorted order because `rbacLabels` happens to emit its two keys in reverse-alphabetical order (`clerum.io/managed-by`, then `clerum.io/host`). Add a third label, or swap those two literals, and the reversed fixture can coincide with author order - at which point NOOP-ROLE-1 silently stops exercising label canonicalization while still passing green.

The test already guards this properly for rule keys (`expect(authoredKeys).not.toEqual(liveKeys)`). Worth the same explicit assertion for labels.

host-context-controller/src/__tests__/asApiserverRole.ts
```diff
@@ -0,0 +1,92 @@
...
+        ...desired.metadata?.annotations,
+      },
+      ownerReferences: desired.metadata?.ownerReferences,
+      selfLink: `/apis/rbac.authorization.k8s.io/v1/namespaces/${desired.metadata?.namespace}/roles/${desired.metadata?.name}`,
+    },
+    rules: (desired.rules ?? []).map(asApiserverPolicyRule),
+  }
+}
+
+export function updatedRoleLogs(log: ReturnType<typeof vi.spyOn>, needle: string): string[] {
```
jozer-rami, line 88: Third copy - `asApiserverService.ts:83` has `updatedServiceLogs` and `asApiserverNetworkPolicy.ts:123` has `updatedPolicyLogs`, identical bodies differing only in the hardcoded log substring. One shared `updatedLogs(log, ...needles)` in `test/__fixtures__` collapses all three.

host-context-controller/src/__tests__/hostRoleNoopGate.test.ts
```diff
@@ -0,0 +1,294 @@
...
+  createMockCustomApi,
+  createMockNetworkingApi,
+  createMockRbacApi,
+} from '../../test/__fixtures__/testMocks'
+import { HOST_LABEL } from '../constants'
+import { HostReconciler } from '../hostReconciler'
+import type { HostCRD } from '../types'
+import { asApiserverPolicyRule, asApiserverRole, updatedRoleLogs } from './asApiserverRole'
+
+function makeStubKc(): k8s.KubeConfig {
```
jozer-rami, line 22: Fifth verbatim copy of this Proxy stub - it's already in `hostServiceNoopGate.test.ts`, `hostNetworkPolicyNoopGate.test.ts`, `channelReaderServiceNoopGate.test.ts` and `hostReconciler.mounts.test.ts`. When `HostReconciler`'s constructor grows a kubeconfig-derived client the stub doesn't cover, that's five identical edits. Belongs next to `createMockRbacApi` in `test/__fixtures__/testMocks.ts`.

jozer-rami (commented):
Reviewed the gate end to end. **Sound and correctly scoped, two things worth fixing before merge.**

`ensureHostRole` was genuinely the only RBAC path doing an unconditional `replaceNamespacedRole` (`ensureHostRoleBinding` already returns on 409), so there's no sibling gap. The comparator covers exactly the fields the builder authors (`metadata.labels`, `rules`) with name/namespace as the read keys, so nothing authored escapes comparison, and the Host-A-can't-read-Host-B's-Secret invariant holds because a rotated `spec.secretRef` changes `resourceNames` and falls through to the replace. Fail-open-to-write is the right default.

---

### 1. The KEEP test no longer tests what it's named for

`host-context-controller/test/hostReconciler.test.ts:1028` (not in this diff, so no inline anchor):

```ts
rbacApi.readNamespacedRole.mockResolvedValue({ metadata: { resourceVersion: '7' } })
```

No `rules`, no `labels`. `roleMatchesDesired` therefore returns `false` on its **first line** (`if (!desired.rules || !existing.rules) return false`) before comparing anything.

So the test named *"rewrites Role resourceNames when spec.secretRef changes"* now passes only because the comparator fail-opens. If a future change made `roleMatchesDesired` wrongly return `true` for a rotated secretRef, this test stays green. The PR lists it as a deliberate KEEP ("read without `rules` still replaces"), which is a fine assertion to have, but it means the repo's one guard on the secretRef rewrite is now a fail-open check rather than a comparator check.

Suggest giving the mocked read the desired labels and rules carrying the *old* secretRef, so the comparator is the thing under test. Keep the current no-`rules` case as a second, separately named test.

### 2. Fifth copy of the same recursive key-sorter

`canonicalizeRoleValue` is line-for-line identical to four functions already in the tree, the nearest one 180 lines below it in the same file:

| Location | Name |
|---|---|
| `hostReconciler.ts:4873` | `normalizeDeploymentValue` |
| `utils.ts:242` | `normalizeServiceValue` |
| `networkPolicyReconciler.ts:123` | `canonicalizeNetworkPolicyValue` |
| `networkPolicyReconciler.ts:327` | `canonicalize` |

`utils.ts` already owns the two sibling gates this one is modeled on (`serviceMatchesDesired:189`, `networkPolicyMatchesDesired:267`), and `canonicalStringify`'s own docblock warns that "a divergence between the two canonical forms would be invisible until a rollout silently stopped happening" - which is exactly the failure mode a no-op gate has. Exporting one `canonicalizeValue` from `utils.ts` deletes four copies. Details inline.

---

### Explicitly not raising

The missing HCC-ownership check before the replace, the swallowed `console.error` on the read/replace path, the absent conflict retry, and the not-migrated-to-`replaceWithConflictRetry` inconsistency are all pre-existing on that path (the diff adds one line to the block) and are already deferred to G5 / #460. Agree with leaving them there rather than growing this PR.

Inline comments below are the two items above plus test-hygiene nits, none blocking.

PR #319, opened 2026-08-12
feat(mcp-host): OAuth per-user & shared-identity token injection (U4 + U6-host)

zach88:
Stacked on #317 (control-api base). Retargets to `dev` automatically once #317 merges.

> **Scope note (updated after review).** This branch absorbed the U5 desktop
> merges (reactive OAuth connect prompt + deep-link resume) via
> `feat/mcp-server-oauth-desktop`, so it now spans **U4 + the mcp-host half of
> U6 + U5** across mcp-host, control-api, host-context-controller, desktop-app,
> rpc-proxy and deploy. It is well over the soft PR budget (D1: 1 service / 25
> files); the natural cut is to land U5 (control-api / desktop / rpc-proxy) as
> its own PR once host-context-controller and mcp-host converge. Kept together
> here because the catalog rail admits OAuth servers by name only once #317's
> control-api changes land, so the isolation core and its consumers must ship
> coherently.

## What
Wires OAuth mcp-servers into mcp-host's connection layer. An OAuth mcp-server's
provider access token is now resolved **just-in-time, per connection, keyed by
the caller's identity**, so a single server can hold N per-user connections plus
a shared-context one. Covers U4 (per-user, the v1 core), the mcp-host half of U6
(shared identity / grantScope `context`), and U5 (desktop reactive consent:
`connect_required` suspension + deep-link resume). Also projects `spec.oauth`
from host-context-controller and adds the broker's nginx location.

## Why
The connectors the business needs (monday, clickup, google-workspace, office365,
bigquery, vercel) require OAuth 2.1 per user. control-api already brokers the
tokens (landed in #317); this is the mcp-host side that consumes them without
ever mounting a user token in the server pod, plus the desktop surface that lets
an end user grant consent on demand.

## How (key decisions)
- **Re-key the live-connection maps** (`clients`/`pendingAdmissions`) from
  `serverName` to `ClientKey = {serverName, principal}`, where principal is the
  `userId` (grantScope `user`) or a structural `SHARED` tuple (static +
  grantScope `context`). A new `byServer` index keeps reconcile and revocation
  reasoning per-`serverName`; the authoritative coordinator's
  desired/epoch/fingerprint is **not** re-keyed. Per-user partitions are
  admitted lazily via `ensureClient`.
- **Dispatch by `authKind`, not `grantScope`.** HCC derives a validated
  `authKind` enum (`static` | `oauth-user` | `oauth-context`) from the authority
[...]

PR #497, opened 2026-08-28
feat(connectors): tri-state connectors read-model and mcp-connectors endpoint (U1)

zach88:
## What
- New tri-state connector read-model (connected / disconnected / never-connected) in control-api, exposed through a `mcp-connectors` read endpoint.
- rpc-proxy route that serves that read-model to the desktop panel.
- Gateway allowlist entry so the rpc gateway forwards the new read path.

## Why
Phase 2 of the personal MCP-server OAuth work. The desktop panel needs to show, per agent, the full fleet of mcp-servers with their authorization state before the user runs anything. This PR provides the read surface the panel consumes. Unit U1 of the proactive-connectors spec.

## How
- The classifier composes grant existence with server OAuth resolution and does not drop servers that have no grant (unlike the rpc-proxy invocation lane, which filters them out), so "never-connected" and "disconnected" stay distinguishable.
- Grant keying stays server-side (keyed by the server's authoritative `spec.contextRef`), so **no client-supplied context identity is trusted**. The read-model does echo each agent's server-derived `contextRef`, exactly as the existing `/rpc/servers` already exposes `contextIds` — so this widens no disclosure surface (review R1-H5).
- **Tri-state semantics (review R1-SM1/SM2):** the state answers *"does this connector have a grant?"*, not *"can it run right now?"*. The classifier deliberately does **not** apply the invocation lane's viability gates (`spec.enabled === false`, the auth-type allowlist, a resolvable URL), so the classified set is a strict superset of `/rpc/servers` and the shared `mcp:servers:list` scope now spans two result-sets. This is the caller's own fleet, non-secret policy only — not a privilege escalation. Documented at `classifyConnector` and the rpc-proxy connectors route.
- The endpoint reuses the same auth gate as the existing user-token routes.

## Testing
- control-api: `services.access.mcpConnectors.test.ts`, `routes.rpcAccessUsersMcpConnectors.test.ts`, `routes.rpcAccessUsersMcpConnectors.auth.test.ts` (token-subject binding, R1-H3), realPostgres integration `services.access.mcpConnectors.realPostgres.integration.test.ts`, and gateway intent `network.gateway.intent.test.ts` (block-scoped method assertions, R1-H4).
- rpc-proxy: `controlApiRestService.connectors.test.ts` (typed 401/403 rejection + upstream abort signal), `rpc-connectors-route.test.ts` (401→401 / 403→403 mapping).
- Unit and integration suites green (rpc-proxy 485, control-api connectors lane 54 + 1 realPostgres skip).

## Risks / Follow-ups
- **Deploy dependency (review R1-H1):** the rpc gateway mounts its nginx ConfigMap with `subPath` and there is no reloader, so the new `location` does not reach the running pod on a plain `kubectl apply`. A companion change in `evenfire-infra` must `kubectl rollout restart deployment/control-api-rpc-gateway` (pattern of PR #422) and land **before or with** this deploy, or the endpoint ships returning 500. No fix for this lives in this repo.
- Rate-limiting / apiserver fan-out on this rail (review R1-SM3, non-blocking): bounded by the caller's authorized agents, not attacker-controlled surface — tracked as a follow-up issue, not fixed here. The per-request upstream timeout **is** fixed in this PR.
- Stacked PR. Base is `feat/mcp-server-oauth-host` (behind #319 then #317 then dev); retarget to dev as the parents merge.
- 16 files, under the D1 budget. Spans control-api, rpc-proxy, the deploy gateway and the shared REST contract types, but they form one read seam (endpoint plus its gateway allowlist plus shared types).
- No CRD changes.


deploy/base/control-plane/configmaps.yaml
```diff
@@ -142,6 +142,21 @@ data:
           proxy_set_header X-Forwarded-Proto $scheme;
         }
 
+        # spec 11 U1: proactive connectors panel — tri-state read model.
+        # rpc-proxy forwards the caller's rpc access token; control-api
+        # authorizes via requireRpcTokenUserMatch. Must be allowlisted here or
+        # rpc-proxy is 403'd at the edge (sibling of mcp-servers above).
+        location ~ ^/api/v1/rpc/access/users/[^/]+/mcp-connectors$ {
```
alfredolopez80, line 149: **H1 (High) — This `location` will not reach the running gateway. CONFIRMED against `keyper-labs/evenfire-infra`.**

Three facts stack:

1. `control-api-rpc-gateway.yaml:69-78` mounts `nginx.conf` with `subPath`. A ConfigMap mounted via `subPath` does not receive updates — the file inside the running pod never changes on disk.
2. There is no `configMapGenerator` anywhere in `deploy/` (`grep -rn 'configMapGenerator\|disableNameSuffixHash' deploy/` → empty), so `configmaps.yaml` is a flat resource. `kubectl apply` edits it in place, the Deployment pod template stays byte-identical, and no rollout fires.
3. No reload mechanism of any kind: no checksum annotation, no Stakater Reloader, no sidecar, no SIGHUP.

I verified the real pipeline rather than assuming. The ConfigMap **is** applied — the `Render assembled overlay + strip RBAC` step strips only RBAC kinds, never ConfigMaps. But the step that exists precisely for this problem, `Restart ConfigMap consumers that require pod recreation` (present in both `deploy-clerum-dev.yml` and `deploy-clerum-prod.yml`), restarts exactly four Deployments:

```
nginx-workflow-approval-gateway          (control-plane)
clerum-workflow-approval-request-reader  (channels)
host-context-controller-api-gateway      (control-plane)
control-api                              (control-plane)
```

`control-api-rpc-gateway` is not among them — **zero mentions in either workflow, and zero in the whole infra repo**. Every restart there is by explicit name; there is no namespace-wide or label-selector restart to catch it.

That step's own comment states this failure mode verbatim: *"mounts nginx.conf with subPath, so updated gateway routes are not visible to live pods until restart."*

**Consequence after deploy:** control-api and rpc-proxy roll (new image); the gateway does not. The stale nginx.conf has no matching regex, so the request falls through to `location / { return 403; }`. `fetchUserConnectorsFromControlApi` sees `!response.ok` and throws, and the client receives `HTTP 500 {"message":"Control API MCP connectors lookup failed (403)"}`. It does not self-heal — this persists until someone recreates the gateway pod manually.

**To be fair on ownership:** the mechanism is pre-existing. `mcp-servers`, the sandbox-ui locations and `mcp-oauth/authorize-url` were all added the same way. This PR does not introduce the gap — it is the first one where a *code path* depends on the ConfigMap change taking effect at the same time as the code deploy.

**Suggested fix.** PR #422 hit exactly this on the HCC gateway, and the fix was twofold — worth copying both halves:

1. In `evenfire-infra`, add to the restart step of **both** workflows, before `Verify rollout status across namespaces`:
   ```
   kubectl rollout restart deployment/control-api-rpc-gateway -n control-plane
   kubectl rollout status  deployment/control-api-rpc-gateway -n control-plane --timeout=300s
   ```
2. Extend `scripts/tests/test-pr422-hcc-gateway-rollout.sh` (or clone its pattern) to cover this gateway. Without the guard, that line can be dropped in a refactor with nothing failing.

Longer term, moving `configmaps.yaml` to a kustomize `configMapGenerator` would make this class of bug impossible rather than remembered — right now every `subPath` gateway is a line someone has to remember to add.

**This fix lives in another repo**, so the infra change needs to land before or alongside this deploy, or the endpoint ships broken.

rpc-proxy/src/services/controlApiRestService.ts
```diff
@@ -83,6 +83,96 @@ export async function fetchUserAllowedServersFromControlApi(
...
+  const response = await fetch(
+    `${controlApiBaseUrl()}/rpc/access/users/${encodeURIComponent(userId)}/mcp-connectors`,
+    {
+      method: 'GET',
+      headers: controlApiHeaders(rpcAccessToken),
+    }
+  )
+
+  if (!response.ok) {
+    throw new Error(`Control API MCP connectors lookup failed (${response.status})`)
```
alfredolopez80, line 189: **H2 (High) — Every non-ok response collapses to 500, which disables the desktop's token refresh. REPRODUCED.**

`next(error)` at `rpc.ts:241` reaches the handler in `app.ts:51-61`, which special-cases only `isUpstreamTimeoutError` → 504. Everything else becomes 500:

```
upstream=401 -> client status=500 body={"message":"Control API MCP connectors lookup failed (401)"}
upstream=403 -> client status=500 body={"message":"Control API MCP connectors lookup failed (403)"}
```

(reproduced against the real `createApp()`, stubbing only `global.fetch`)

Why this matters on the client: `AppService.shouldRefreshRpcToken` (`desktop-app/src/appService.ts:935-942`) returns `false` for 500 — it fires only on 401, or on 403 whose body contains `"missing scope"`. The RPC token has a **300 s TTL**, so the race is routine rather than exotic: the proxy validates at t₀ and passes, control-api validates at t₀+ε and it has already expired → 401 → surfaced as 500 → **the refresh never fires**, and the connectors panel stays broken until the user restarts the app. JWT key rotation and clock skew between the two services produce the same outcome.

There is direct evidence this already cost the team once: `isForbiddenSharedFilesListError` (`appService.ts:944-954`) exists *only* to rescue a 403 by string-matching `"(403)"` and `"forbidden"` inside a 500 message produced by this same defect on another route. It was patched in the client instead of at the source.

The correct pattern is already in this file, two functions below: `fetchHostConnectionFromControlApi:198-208` maps 403/404 → `null` and 401/403/409 → a typed `ControlApiHostAccessRejectedError`, which `rpc.ts:123-150` translates into real status codes. `rpc-proxy/src/routes/mcpOauth.ts:27-35` documents the policy explicitly.

In fairness, `fetchUserAllowedServersFromControlApi:54-56` has the same defect, so this is *consistent with `/rpc/servers`* — but the consistency is with the wrong side of a divergence this file already resolved better.

**Suggested fix:** a typed error for 401/403 (reuse `ControlApiHostAccessRejectedError` or add `ControlApiConnectorsRejectedError`), mapped in the route to real 401/403; everything else stays 502/500.

control-api/src/routes/rpc-access/users.ts
```diff
@@ -140,6 +143,35 @@ export function createRpcAccessUsersRouter(
...
 
+  // Proactive connectors panel read-model (spec 11 U1): the CLASSIFIED fleet
+  // per agent — `authorized` / `requires_setup` / `no_oauth`. Same gate as
+  // `/mcp-servers`; `req.params.userId` is authoritative (bound to the RPC
+  // token subject by `requireRpcTokenUserMatch()`) and flows only into the
+  // grant-presence key, never from a body. Agents derive from `getUserAgents`.
+  router.get(
+    '/rpc/access/users/:userId/mcp-connectors',
+    requireValidRpcAccessToken(),
+    requireRpcTokenUserMatch(),
```
alfredolopez80, line 154: **H3 (High) — This route documents a security invariant that no test exercises. REPRODUCED by mutation.**

The docblock above states that `req.params.userId` "is authoritative (bound to the RPC token subject by `requireRpcTokenUserMatch()`)" and flows "only into the grant-presence key, never from a body". I verified that claim is **true in the code** — `rpcAccessAuth.ts:63` rejects with 403 when `claims.sub !== userId`, and the handler never reads the body.

But nothing tests it. Two mutations, both surviving:

| Mutation | Result |
|---|---|
| Delete `requireRpcTokenUserMatch()` from this route | **GREEN** — 2/2 passed |
| `userId` from a query param: `(req.query.userId as string) ?? req.params.userId` | **GREEN** — 2/2 passed |

So the middleware preventing one user from reading another's connectors can be removed, or `userId` can be made client-controlled, and `routes.rpcAccessUsersMcpConnectors.test.ts` still passes. That suite has 2 tests and neither covers authorization.

**This is not a vulnerability today** — the middleware is present and correct. It is the absence of the net that would catch a future refactor removing it, on a route whose own comment advertises the guarantee.

**Suggested fix:** two tests — (a) a token for `alice` against `/users/bob/mcp-connectors` returns 403; (b) a `?userId=` parameter does not change the effective subject.

control-api/test/network.gateway.intent.test.ts
```diff
@@ -335,6 +335,14 @@ describe('network/gateway intent (manifest-level)', () => {
     const configmaps = read(`${BASE}/control-plane/configmaps.yaml`)
     const gatewayConf = docContaining(yamlDocs(configmaps), 'name: control-api-rpc-gateway')
     expect(gatewayConf).toContain('location ~ ^/api/v1/rpc/access/users/[^/]+/mcp-servers$')
+    // spec 11 U1: the proactive connectors read-model is gated by an rpc access
+    // token (requireRpcTokenUserMatch), not requireInternalService('rpc-proxy'),
+    // so the derived guard below does not cover it — assert its location here or
+    // rpc-proxy's GET /rpc/connectors 403s at the edge (the shipped regression).
+    expect(gatewayConf).toContain('location ~ ^/api/v1/rpc/access/users/[^/]+/mcp-connectors$')
+    expect(gatewayConf).toMatch(
+      /location ~ \^\/api\/v1\/rpc\/access\/users\/\[\^\/\]\+\/mcp-connectors\$ \{[\s\S]*?limit_except GET \{/
```
alfredolopez80: **H4 (High) — This assertion is vacuous: the `limit_except` block can be deleted and the test still passes. REPRODUCED by mutation.**

`[\s\S]*?` is not scoped to the matched block. It is non-greedy, but nothing stops it from running past the end of the `mcp-connectors` `location` and matching a `limit_except GET {` that belongs to a *different* `location` later in the file.

Mutation M11 — deleted the entire `limit_except GET { deny all; }` from the `mcp-connectors` block in `configmaps.yaml:150-152`, leaving the endpoint accepting POST/PUT/DELETE at the edge:

```
M11 remove limit_except from connectors block -> exit=0  total 41  passed 40  failed 0
```

What these two lines actually cover:

| Aspect | Covered? |
|---|---|
| The string exists in the YAML | **Yes, solidly** — `toContain`, anchored with `$` |
| Literal shape of the `location` directive | Partially — as text, not as nginx semantics |
| Permitted methods | **No — the assertion is vacuous** |
| **Order** of `location` blocks relative to the `location /` default-deny | **Not asserted at all** |
| That nginx actually applies it | No — manifest-level test; nginx is never started |

The ordering gap matters here because nginx resolves regex `location` blocks by first match in file order, so placement is part of the contract this test is meant to protect.

**Suggested fix:** scope the regex to the block (e.g. `[^}]*?`, or extract the block by parsing and assert against that), and add an assertion that this block precedes the `location /` default-deny.

alfredolopez80 (commented):
## Review: changes requested — 5 High · 13 Medium · 12 Low · 0 Blocker

Reviewed with 7 parallel agents against an isolated worktree of the PR head, scoped to the diff versus its declared base (`feat/mcp-server-oauth-host`). Findings were verified against the code — reproduced where possible — rather than read off the diff. Every claim below is marked **REPRODUCED** or **READ-ONLY**.

**The design is sound.** Classifying instead of filtering is the right call for a panel, authorization is equivalent to the sibling `/mcp-servers` route, and the three degradation paths are genuinely fail-closed: `authorized` is unreachable without a row in `oauth_grants`. What follows is what needs fixing before merge, not a redesign.

---

## High

### H1 — The gateway never picks up the new `location`; the endpoint returns 500 permanently
**CONFIRMED against `keyper-labs/evenfire-infra`.** See the inline comment on `configmaps.yaml`.

`deploy/base/control-plane/control-api-rpc-gateway.yaml:69-78` mounts `nginx.conf` with `subPath`, which does not propagate ConfigMap updates, and `configmaps.yaml` is a flat kustomize resource (no `configMapGenerator` anywhere in `deploy/`), so the pod template stays byte-identical and no rollout fires.

Verified in the real pipeline: the ConfigMap **is** applied (the render step strips only RBAC kinds), but the `Restart ConfigMap consumers that require pod recreation` step in **both** `deploy-clerum-dev.yml` and `deploy-clerum-prod.yml` restarts four Deployments, and `control-api-rpc-gateway` is not one of them — zero mentions of it anywhere in the infra repo.

Result after deploy: the stale nginx.conf falls through to `location / { return 403; }`, `fetchUserConnectorsFromControlApi` throws, and the client gets **HTTP 500** indefinitely until someone recreates the gateway pod by hand.

**This needs a companion PR in `evenfire-infra`, landing before or with this deploy.** PR #422 hit the identical problem on the HCC gateway; the fix there was the restart line plus a contract guard (`scripts/tests/test-pr422-hcc-gateway-rollout.sh`). Same mould applies here.

### H2 — A 401/403 from control-api collapses to 500, disabling the desktop token refresh
`rpc-proxy/src/services/controlApiRestService.ts:162-164` — **REPRODUCED**

`AppService.shouldRefreshRpcToken` returns `false` for 500; it only fires on 401, or on 403 whose body contains `"missing scope"`. The RPC token has a 300 s TTL, so the race is routine: the proxy validates at t₀ and passes, control-api validates at t₀+ε and it has expired → 401 → surfaced as 500 → **the refresh never fires** and the panel stays broken until the app restarts. Same for JWT key rotation or clock skew.

Worth noting: `isForbiddenSharedFilesListError` in `desktop-app` exists solely to rescue a 403 by string-matching `"(403)"` inside a 500 message from this same class of defect on another route. The correct pattern is already two functions below in the same file — `fetchHostConnectionFromControlApi` returns a typed `ControlApiHostAccessRejectedError` that the route maps to real status codes.

### H3 — The route documents a security invariant that no test exercises
`control-api/src/routes/rpc-access/users.ts:151` — **REPRODUCED by mutation**

Deleting `requireRpcTokenUserMatch()` from the route, or sourcing `userId` from a query param, both leave the suite **green**. Not a vulnerability today — the middleware is in place and works — but there is no net to catch a future refactor removing it. See inline.

### H4 — The nginx method assertion is vacuous
`control-api/test/network.gateway.intent.test.ts:343-344` — **REPRODUCED by mutation**

`[\s\S]*?` is not scoped to the block, so it can match a `limit_except GET {` belonging to a different `location`. Deleting the entire `limit_except GET { deny all; }` from the `mcp-connectors` block leaves the test green. See inline.

### H5 — The PR description states the opposite of what the code does
Independently reproduced by three agents.

> *"Grant keying stays server-side (keyed by `spec.contextRef`), so no context identity is echoed to the client."*

First half is true: `buildMcpServerGrantKey` uses the server's authoritative `contextRef`, never a body value. Second half is false: `AgentWithConnectors.contextRef` is emitted by control-api (`mcpInvocable.ts:82`, `:621`), **deliberately preserved** by the rpc-proxy sanitizer (`controlApiRestService.ts:133-140`), and returned verbatim at `rpc.ts:240`.

Mitigating: `/rpc/servers` already exposes `contextIds`, so this widens no disclosure surface. Either drop `contextRef` from the panel wire model if the client does not need it, or reword to *"no CLIENT-SUPPLIED context identity is trusted"*.

---

## Medium (13)

**Classification vs. viability** — `classifyConnector` applies none of the three viability gates `filterInvocable` applies (`spec.enabled === false`, auth-type allowlist, resolvable URL). A CR with `enabled:false` and a live grant is served as `authorized` while the invocation rail returns `[]`. Reproduced:

```
filterInvocable   → ['srv-oauth-nogrant']
resolveConnectors → ['srv-apikey', 'srv-disabled', 'srv-nourl', 'srv-oauth-nogrant']
```

Two consequences from one cause: users see a green connector that cannot run, and the returned set is a strict superset of `/rpc/servers` served under the **same** `mcp:servers:list` scope. Reusing the scope was reasonable while both rails showed the same thing. Worth an explicit decision: does the tri-state answer *"has a grant?"* or *"works?"* If the former, the read-model needs a separate availability dimension.

*(Caveat: that a disabled `McpServer` stays in `Context.spec.mcpServers` is inferred from the three filter call sites in `host-context-controller/src/k8sClient.ts`, not observed on a cluster.)*

**Cost** — no cache (deliberate and well justified) but also no rate limiting anywhere on this rail. Measured with real counters: 5 agents / 10 servers = 2 namespace LISTs + 5 GETs + 11 PG queries; 50 agents = **50 simultaneous GETs**. The grant checks run strictly sequentially (`maxInFlight: 1`) — 200 ms of 234 ms wall time in the large case. A bounded `Promise.all` fixes it. Dedup of the `union` does work as its comment claims: 5 agents on one context produce 10 queries, not 50.

**`fetch` without timeout** — `controlApiRestService.ts:154`, new in this PR (`git blame` `5d9b2290`). All four fetches in `mcpHostRestService.ts` and both in `mcpRpcService.ts` use `AbortController` + `config.upstreamTimeoutMs`; the three in this file use none. `config.upstreamTimeoutMs` already exists (default 8000 ms), so this is a three-line fix.

**Degradation log is inaccurate** — when `listResource('mcpservers')` fails the result is *zero connectors*, not `"connectors served without catalog classification"`. The sibling path (`loadAllowedNamesByContext`) has the accurate message. Both share the real gap: the 200 carries no degradation marker, so the panel cannot tell a catalog outage from an empty fleet — the exact ambiguity this PR exists to remove. Consider a `catalogDegraded` flag.

**Coverage** — 810 lines of tests produce **17 executable cases**. Mutation testing (18 mutations, 7 caught, 11 survived) shows five branches of `resolveConnectorsForAgents` are untested: both `catch` blocks that make the endpoint resilient can be deleted outright and the suite stays green, as can the oauth-ness gate, the missing-CR path, and the non-404 host branch. Connector ordering, rpc token forwarding, and `encodeURIComponent` on `userId` are likewise unasserted.

**realPostgres skip** — `describeRealPostgres = adminUrl ? describe : describe.skip` reports 3 skipped / 0 executed and green in an ordinary `npm test`. It is covered by the pre-existing `ci-public.yml` lane that sets `CONTROL_API_REAL_PG_REQUIRED=1` and fails on any `skipped`, and the new file joins that lane automatically by filename — so this is not a blocker. Flagging it only because the protection is a text grep over Vitest output.

**Silent sanitizer drop** — a connector with an unknown `status` disappears from the panel with no log. If control-api ever emits a fourth status, the panel quietly shows fewer connectors. (The sanitizer does correctly reject unexpected fields — mutation-verified.)

**Shared types** — the description mentions *"the shared REST contract types"*, but none exist: no `@clerum/*` package covers REST contracts, `rpc-proxy/package.json` declares no workspace dependency, and there are no cross-imports. `ConnectorEntry` and `UserConnector` are two hand-maintained definitions. No drift today (verified field by field); the only guard against future drift is a sanitizer that degrades silently.

**Comment accuracy** — 11 of 13 verifiable claims in the new docblocks are true. Two are not: *"Mirrors host-context-controller's `deriveAuthKind`"* diverges when `auth: {}` is present without `type` (schema-valid — the CRD does not mark `type` required), where HCC labels `static` and this omits it; and *"SOLELY from `oauthGrantExists`"* overstates, since `requires_setup` also arises from key-derivation failures that never touch the DB. Also `console.error` at `mcpInvocable.ts:396,428` (pre-existing, `5e6c990f`) survives in a file this PR edits heavily, against the repo's logging standard.

---

## Low (12, abbreviated)

`requireRpcTokenUserMatch` compares trimmed while the handler consumes the raw param — fails **closed** (parameterised `WHERE user_id = $1`, no privilege escalation), pre-existing, this PR adds the third consumer; `grantScope:'context'` with an empty `contextRef` pins to `requires_setup` forever with no log; a third `GrantScope` value would vanish silently in `sanitizeConnector`; `gfsSubject`/`namespace`/`active` are emitted but unconsumed while `displayName` is dropped by the sanitizer; unbounded fan-out concurrency against the apiserver; two child loggers for one flow, so grant-presence errors do not surface under `module: 'mcp-connectors'`.

---

- **Authorization**: same middleware pair as the sibling route; `claims.sub !== userId` → 403. No IDOR — `agentNames` comes from `getUserAgents`.
- **Secret hygiene**: `ConnectorEntry` carries only non-secret policy. No `auth`, `secretRef`, `clientIdRef`, tokens, or transport URL — unlike the RPC rail, which does return `url`.
- **Fail-closed**: verified on all three degradation paths. `authorized` is unreachable without a DB row.
- **The oauth gate** in `classifyConnector` matches `computeGrantPresence` exactly; both delegate to `resolveServerOAuth`. Already pinned by the PR's own 200-run `fast-check` property test.
- **The non-404 `throw`** is not a divergence from `resolveMcpServersForAgents` and produces no unhandled rejections (verified empirically).
- **The ad-hoc `DbClient`** matches the type and replicates the established idiom (8 other call sites).
- **Supply chain clean**: version bumps only, no third-party dependency added or changed. The PR also *fixes* a pre-existing lock/package desync.
- **The tri-state itself is genuinely covered**: inverting the classification fails 5 tests, after exhausting all 3 retries of control-api's `retry: 2`.

Suites executed: 61 total, 57 passed, 0 failed, 4 skipped (3 realPostgres + 1 pre-existing unrelated). All runs captured exit codes directly with `--reporter=json`, never through a pipe.

---

Happy to open the `evenfire-infra` companion PR for H1 if useful. Also worth confirming the retarget to `dev` once #319 and #317 land.

control-api/src/services/access/mcpInvocable.ts
```diff
@@ -410,3 +450,174 @@ export async function resolveMcpServersForAgents(
...
+/**
+ * Classify ONE server into the tri-state, given the authoritative grant-presence
+ * set (`grantPresent`, computed by `computeGrantPresence` — the SAME flavored
+ * key construction the rpc-proxy gate uses, D4). FAIL-CLOSED: a server that is
+ * not an OAuth connector (not `auth.type==='oauth'`, or no usable oauth id) is
+ * `no_oauth`, never `authorized`; an OAuth server absent from `grantPresent`
+ * (no grant OR a grant-read error, both swallowed fail-closed upstream) is
+ * `requires_setup`, never `authorized`.
+ */
+function classifyConnector(server: McpServerCR, grantPresent: ReadonlySet<string>): ConnectorEntry {
```
alfredolopez80, line 491: **SM1 (Medium, security) — Dropping the viability gates makes this endpoint return a strict superset of `/rpc/servers`. REPRODUCED.**

`classifyConnector` classifies purely on grant presence. It applies none of the three gates that `filterInvocable` applies on the invocation rail (`mcpInvocable.ts:208-234`):

- `spec.enabled === false`
- the `auth.type` allowlist
- `resolveServerUrl()` resolving to something usable

Reproduced side by side on the same fixture set (disposable test, since removed):

```
filterInvocable   → ['srv-oauth-nogrant']
resolveConnectors → [{"name":"srv-apikey",   "status":"no_oauth",       "authKind":"static"},
                     {"name":"srv-disabled", "status":"requires_setup", "provider":"google"},
                     {"name":"srv-nourl",    "status":"no_oauth"},
                     {"name":"srv-oauth-nogrant","status":"requires_setup","provider":"google"}]
```

Two distinct consequences from one cause:

**Correctness** — a CR with `enabled:false` and a live grant is served as `authorized` while the invocation rail returns `[]`. The panel shows a green, ready-looking connector that cannot run. Same for an OAuth server with `transport.type:'http'` and no `url` (also reproduced).

**Disclosure** — a desktop user whose RPC token already carries `mcp:servers:list` now receives the name and `provider` of servers that token could not read before: operator-disabled ones, `bearer`/`basic`/`apiKey` ones, and ones with no transport URL yet. No URL and no credentials are exposed — but the set of *names* is strictly larger. See my comment on `rpc.ts` for the scope side of this.

**Honest caveat on reachability:** that a disabled `McpServer` remains in `Context.spec.mcpServers` is *inferred* from the three filter call sites in `host-context-controller/src/k8sClient.ts:2676,4712,4725` — HCC filters disabled servers at materialisation rather than pruning the allowlist — not observed on a live cluster. If HCC does prune, the `enabled:false` half of this does not arise; the auth-type and missing-URL halves still do.

**There is a defensible reading where this is intentional**: the documented semantics at `:59` are "an OAuth server WITH a valid grant", and a disabled server does have a grant. If that is the intent, the read-model needs a separate availability dimension (`available` / `enabled`), because otherwise the panel is promising something it cannot deliver. Either way it is currently untested — no test covers a disabled server, an unresolvable URL, or a static-auth server.

Worth an explicit decision: **does the tri-state answer "has a grant?" or "works?"**

rpc-proxy/src/routes/rpc.ts
```diff
@@ -222,6 +223,27 @@ export function createRpcRouter(): Router {
...
 
+  // Proactive connectors panel read-model (spec 11 U1): per-agent classified
+  // fleet (`authorized`/`requires_setup`/`no_oauth`). Read of the same catalog
+  // rail as `/rpc/servers`, so it REUSES the `mcp:servers:list` scope (no new
+  // scope minted). `userId` is the signed session subject (`auth.sub`), never a
+  // body/query param.
+  router.get(
+    '/rpc/connectors',
+    requireRpcAuth,
+    requireScope('mcp:servers:list'),
```
alfredolopez80, line 243: **SM2 (Medium, security) — Reusing this scope silently widens what it discloses. REPRODUCED (verdict), READ-ONLY (scope inventory).**

Reusing an existing scope instead of minting `mcp:connectors:*` is the right instinct in general, and the comment above is accurate that no new scope is created. The problem is what the scope now covers.

`mcp:servers:list` is the same scope `/rpc/servers` uses at `rpc.ts:209`. Until this PR, that scope disclosed the **invocable** set — `filterInvocable` had already removed disabled servers, non-allowlisted auth types, and servers with no resolvable URL. This route serves the **classified** set, which as shown in my comment on `mcpInvocable.ts` is a strict superset of that, plus a per-server OAuth authorization state that the catalog never carried.

**Verdict: yes, this is a silent widening of the disclosure boundary.** Not a privilege escalation — the holder is a legitimately authenticated user reading their own agents' fleet, and the added state (`authorized` / `requires_setup`) is information *about that same user*, which is low sensitivity. But a token minted under the old contract now returns more than it did, with no version marker and nothing in the scope name signalling the change.

The underlying reason is structural, and it is the same one behind SM1: reusing the scope was correct while both rails showed the same set. It stops being correct once one rail filters and the other classifies. That divergence is the whole point of this PR, so the scope decision deserves to be revisited rather than inherited.

**Options, in increasing cost:**
1. Restore parity by applying the viability gates to the classified set (this also fixes SM1's correctness half) — then the scope reuse is straightforwardly correct again.
2. Keep the superset and mint a distinct scope for it.
3. Keep both as-is and document explicitly, in the route comment, that `mcp:servers:list` now spans two different result sets — so the next person to grant this scope knows what they are granting.

I would not block on this, but I would not leave it undecided either: option 3 costs one comment and removes the silent part, which is the part that actually bites.

rpc-proxy/src/services/controlApiRestService.ts
```diff
@@ -83,6 +83,96 @@ export async function fetchUserAllowedServersFromControlApi(
...
+        .filter((entry): entry is UserConnector => entry !== null)
+    : []
+  return { name, contextRef, connectors }
+}
+
+/**
+ * Fetch the proactive connectors read-model for a user (spec 11 U1). Projects
+ * the control-api payload down to the declared, NON-SECRET shape and drops any
+ * unexpected field — the inventory never transports `auth`/`secretRef`/tokens.
+ * Deliberately UNCACHED: the tri-state must reflect a just-completed
```
alfredolopez80, line 169: **SM3 (Medium, security) — Uncached by design is fine; uncached *and* unlimited *and* untimed is the problem. MEASURED.**

The justification in this docblock is correct and I would not change the decision: the tri-state must reflect a just-completed connect/disconnect, so a 30 s TTL like `/rpc/servers` uses would actively break the feature. (Small accuracy note: the catalog's cache does not live in `fetchUserAllowedServersFromControlApi`, which is also uncached — it lives in the `getCachedUserAllowedServers` wrapper in `mcpProxyService.ts:27-47`. The substantive claim is right; the citation points at the wrong function.)

What is missing is the compensation. Measured with real counters against the actual code path (3 tests, `total > 0`, not estimated):

| Scenario | k8s LIST | k8s GET | Postgres | PG concurrency |
|---|---|---|---|---|
| 5 agents / 10 servers / 5 contexts | 2 | 5 | 11 | **1** |
| 5 agents / 1 shared context | 2 | 5 | 11 | **1** |
| 50 agents / 40 servers | 2 | **50 simultaneous** | 41 | **1** |

Every request costs two full-namespace LISTs (`mcpservers`, `contexts`), one `getResource('hosts')` per agent fanned out with no concurrency bound, and one sequential `oauthGrantExists` per unique OAuth server. `getUserAgents` has no `LIMIT`.

Three things compound here, and it is the combination rather than any one of them:

1. **No cache** — deliberate, justified, not the issue by itself.
2. **No rate limiting** — `grep` across `rpc-proxy/src` and the `rpc-access` routes of control-api returns nothing. `rateLimiterService` in control-api covers approvals, GFS and external routes only, not this rail.
3. **No timeout** — the `fetch` just below (`:154`) has no `AbortController`, unlike all four fetches in `mcpHostRestService.ts` and both in `mcpRpcService.ts`, which use `config.upstreamTimeoutMs` (already defined, default 8000 ms).

A single authenticated client polling this endpoint generates two namespace-wide LISTs per request against the apiserver and holds a PG connection for the serialized grant checks, with nothing in the path to slow it down and no upper bound on how long a hung upstream can pin a socket.

**Credit where due — the dedup does work.** Five agents sharing one context produce 10 grant queries, not 50; the `union` comment at `:592-593` is accurate and I verified it.

**Suggested fixes, cheapest first:** bound the grant checks with a concurrency-capped `Promise.all` (that alone removes 200 ms of the 234 ms wall time in the 50-agent case), add `signal` from the existing `upstreamTimeoutMs`, and cap the host fan-out. Rate limiting is the bigger call and probably belongs to the rail rather than this PR — but it is worth naming now, since this is the first endpoint on it designed to be polled by a UI.

alfredolopez80 (commented):
### Addendum — the two security-flagged Mediums, inline

Following up on my earlier review with the security findings anchored to the code, as requested. Both come from the dedicated security pass and both are **reproduced**, not read off the diff.

Neither is a bypass. There is no IDOR, no secret leakage, and no fail-open path — I verified all three separately and they are clean. What these two describe is a **disclosure boundary that moved** without the scope moving with it, and a **cost amplification** with no limiter anywhere on the rail.

- **SM1** — `classifyConnector` drops the viability gates, so this endpoint returns a strict superset of `/rpc/servers` (inline on `mcpInvocable.ts`).
- **SM2** — that superset ships under the *same* `mcp:servers:list` scope (inline on `rpc.ts`).
- **SM3** — uncached by design, but nothing compensates: no rate limiting, no timeout (inline on `controlApiRestService.ts`).

SM1 and SM2 are one cause with two consequences; I split the anchors because the fix for each lives in a different file and they can be decided independently.

rpc-proxy/src/services/controlApiRestService.ts
zach88: Split this into its two halves:

**Timeout — fixed this round.** All four control-api fetches in `controlApiRestService.ts` now carry `AbortSignal.timeout(config.upstreamTimeoutMs)` (the same 8s budget the host rail already uses), so a hung upstream surfaces as a 504 via `isUpstreamTimeoutError` instead of pinning the socket. Regression test in `controlApiRestService.connectors.test.ts` asserts the fetch is bounded.

**Rate-limit / apiserver fan-out — deliberately out of this PR.** The host fan-out and `getUserAgents` are bounded by the number of agents a user is *authorized* for, not attacker-controlled input, so it's a scale concern (a user with hundreds of agents polling the panel), not a DoS surface. Tracking it as a follow-up issue on the rail rather than fixing it here — agreed it's worth naming now since this is the first UI-polled endpoint on it.

(Also correct on the cache citation — the TTL lives in `getCachedUserAllowedServers`, not the fetch; noted.)

zach88:
On H5 (the `contextRef` disclosure angle) — the security half is a non-issue, and I've corrected the body wording that overstated it:

- The `contextRef` echoed in the read-model is **server-authoritative** (derived from `host.spec.contextRef`), never a client-supplied value — so the actual invariant ("no client-supplied context identity is trusted") holds.
- It's not a new disclosure surface: `/rpc/servers` already returns `contextIds` for the same caller under the same scope. The panel wire model exposes the same class of identifier, nothing more.

So no code change on the disclosure side; the only real defect was the body claiming "no context identity is echoed," which was literally false. Body updated to say what the code does.

(The two genuinely-real security notes, SM1/SM2 — the classified set being a superset of `/rpc/servers` — were closed by an explicit owner decision: the tri-state answers "has a grant?" not "can it run?", documented at `classifyConnector` and the connectors route. Not an escalation: caller's own fleet, non-secret policy only.)

control-api/test/routes.rpcAccessUsersMcpConnectors.auth.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(svc.getUserAgents).not.toHaveBeenCalled()
+  })
+
+  it('200s when the token subject matches, keyed by the PATH user', async () => {
+    const res = await request(buildApp('alice')).get('/rpc/access/users/alice/mcp-connectors')
+    expect(res.status).toBe(200)
+    expect(svc.getUserAgents).toHaveBeenCalledWith('alice')
+  })
+
+  it('a ?userId= override cannot change the effective subject (still 403 cross-user)', async () => {
```
alfredolopez80, line 100: **H3 — still open. The two surviving mutations are the ones this test was written to catch. REPRODUCED.**

First, credit: this test is well built. It keeps `requireRpcTokenUserMatch` real (`importActual` + spread, replacing only the token middlewares), asserts a real HTTP 403 through supertest, and backs it with `expect(svc.getUserAgents).not.toHaveBeenCalled()`. Substituting the guard with a same-named no-op also goes red, which proves it tests **effect**, not textual presence. That is exactly right.

The gap is the **direction** of this case:

```
GET /rpc/access/users/bob/mcp-connectors?userId=alice     // token subject: alice
```

Token `alice`, **path** `bob`. `requireRpcTokenUserMatch()` compares the path param against the token subject → `bob !== alice` → **403 before the handler ever runs**. The handler never executes, so where it sources `userId` from is irrelevant — the case is green with and without the mutation.

Measured:

| Mutation | Result |
|---|---|
| `req.params.userId` → `(req.query.userId as string) ?? req.params.userId` | **GREEN**, 5/5/0/0 |
| `req.params.userId` → `req.body?.userId ?? req.params.userId` | **GREEN**, 5/5/0/0 |

The direction an attacker would use is the inverse: **path = your own subject** (to pass the guard) and **query = the victim**. A disposable probe asserting the safe behaviour, with the query mutation applied:

```
AssertionError: expected "vi.fn()" to be called with arguments: [ 'alice' ]
Received:
-   "alice",
+   "bob",
```

That is `alice`, authenticated and passing the guard cleanly, receiving `bob`'s connector read-model — with this suite green. The probe passes red→green (green on the clean tree), so it registers the behaviour rather than merely failing.

**To be clear: the current code is safe.** The guard is present and correct; on the unmutated tree the probe passes. What is missing is the net.

One consequence worth flagging: the header comment at `:16` states that *"sourcing `userId` from the query, must turn one of these red"*. Measured, that is not true today — the comment documents a guarantee the test does not provide, which is the kind of claim that stops the next reader from checking.

**Two assertions close it** (same `buildApp('alice')` harness, path = the token's own subject):

```ts
it('a ?userId= override does not change the effective subject', async () => {
  await request(buildApp('alice')).get('/rpc/access/users/alice/mcp-connectors?userId=bob')
  expect(svc.getUserAgents).toHaveBeenCalledWith('alice')
})

it('a body userId does not change the effective subject', async () => {
  await request(buildApp('alice'))
    .get('/rpc/access/users/alice/mcp-connectors')
    .send({ userId: 'bob' })
  expect(svc.getUserAgents).toHaveBeenCalledWith('alice')
})
```

Asserting on `resolveConnectorsForAgents` receiving the same `userId` would pin the grant-key path too.

control-api/test/network.gateway.intent.test.ts
```diff
@@ -335,15 +335,34 @@ describe('network/gateway intent (manifest-level)', () => {
...
+    // token (requireRpcTokenUserMatch), not requireInternalService('rpc-proxy'),
+    // so the derived guard below does not cover it — assert its location here or
+    // rpc-proxy's GET /rpc/connectors 403s at the edge (the shipped regression).
+    expect(gatewayConf).toContain('location ~ ^/api/v1/rpc/access/users/[^/]+/mcp-connectors$')
+    // Scope the method assertion to THIS block: `[^}]*?` cannot cross the block's
+    // closing brace, so deleting the connectors' own `limit_except GET {` makes
+    // this fail instead of silently matching a later location's limit_except
+    // (the vacuous-match defect the old `[\s\S]*?` had).
+    expect(gatewayConf).toMatch(
+      /location ~ \^\/api\/v1\/rpc\/access\/users\/\[\^\/\]\+\/mcp-connectors\$ \{[^}]*?limit_except GET \{/
```
alfredolopez80, line 348: **R2-1 (new, Medium) — the block scoping is fixed, but `deny all` itself is never asserted. REPRODUCED.**

The `[^}]*?` scoping works: deleting the whole `limit_except` block, widening it to `GET POST`, and moving the `location` after the `location /` default-deny all now fail. That closes H4 as reported, and the ordering assertion is a genuine addition.

What remains unasserted is the **body** of the block. Mutation:

```diff
  limit_except GET {
-   deny all;
+   allow all;
  }
```

Result: **GREEN**, 41 total / 40 passed / 0 failed.

The regex matches `limit_except GET {` and stops there, so a `limit_except` that permits everything reads identically to one that denies. The directive is present, correctly scoped, correctly ordered — and inert.

This is the same failure shape as the original H4, one level down: the assertion pins the shape of the guard rather than what the guard does. Extending the pattern to require `deny all` inside the braces (or asserting on the extracted block body) closes it, and would be worth applying to the sibling assertions that were generalised in `c6208370`, since they share the pattern.

rpc-proxy/src/services/controlApiRestService.ts
```diff
@@ -48,6 +69,7 @@ export async function fetchUserAllowedServersFromControlApi(
     {
       method: 'GET',
       headers: controlApiHeaders(rpcAccessToken),
+      signal: upstreamAbortSignal(),
```
alfredolopez80, line 72: **H2 — closed here, still live on the sibling route. REPRODUCED.**

The connectors half is genuinely fixed. Reproduced against the real `createApp()` with a valid RS256 token, stubbing only `global.fetch`:

| upstream | `/rpc/connectors` | body |
|---|---|---|
| 401 | **401** | `{"error":"Unauthorized"}` |
| 403 | **403** | `{"error":"Forbidden: user cannot read connectors"}` |
| 404 / 500 / 503 | 500 | generic |
| hang | **504** @ 302 ms | `{"error":"Gateway Timeout"}` |

`shouldRefreshRpcToken` is satisfied via the 401 path. The 403 carrying no `"missing scope"` is correct — that response means subject mismatch, not a scope problem, and refreshing the token would not help.

`fetchUserAllowedServersFromControlApi` on this same line, however, still collapses everything:

| upstream | `/rpc/servers` |
|---|---|
| 401 | **500** `Control API MCP server lookup failed (401)` |
| 403 | **500** |

Unlike `/rpc/connectors` — which has no consumer in the tree yet — `/rpc/servers` is in use today, so the 300 s token-expiry race that motivated H2 lands on the route where it actually costs something. The typed-error helper you added is directly reusable here.

Separately, and smaller: the `signal` you added on this fetch is not covered by any test. Removing it leaves **485 tests green** — only the connectors one is pinned by the new regression test. Same for the host-connection and host-wake fetches.

alfredolopez80 (commented):
# PR Review Summary — Round 2 (verification of the R1 fixes)

Re-verified at `c6208370` against the round-1 baseline `d285ff9e`. Because H3 and H4 existed *precisely because a green suite proved nothing*, I re-ran the exact mutations that survived in round 1 rather than accepting the commit messages. 11 mutations this round; both worktrees confirmed clean afterwards.

**3 resolved · 3 partial · 1 still open (other repo) · 2 new residuals.** No blockers, no regressions.

---

## Resolved (verified, not assumed)

**H4 — gateway method assertion.** Closed, and beyond what I asked for. All three mutations now fail:

| Mutation | R1 | R2 |
|---|---|---|
| Delete `limit_except` block entirely | GREEN | **RED** |
| `limit_except GET` → `GET POST` | — | **RED** |
| Move the block after the `location /` default-deny | — | **RED** |

The reorder case fails at `network.gateway.intent.test.ts:356` with `expected 6173 to be less than 6077` — that is the ordering assertion, which did not exist in round 1. Generalising the block-scoping to the sibling gateway assertions was the right call.

**SM3 — unbounded fetch.** Closed. All four fetches carry `signal` (audited at runtime with an instrumented `global.fetch`, not by grep), `config.upstreamTimeoutMs` is used with no magic number — proved by setting `RPC_PROXY_UPSTREAM_TIMEOUT_MS=300` and observing the cut at **302 ms** — and `isUpstreamTimeoutError` does recognise `TimeoutError`, so a hung upstream surfaces as **504**, reproduced, not 500.

**A correction on my side:** my SM3 comment said "the three in this file use none". There are **four**. The fourth is written `await (options.fetchImpl ?? fetch)(...)` and does not match a `fetch(` grep. Your "all four" was accurate and my count was not.

**H5 — PR description.** Closed. The new wording states what the code does, and both halves check out: the keying is server-authoritative, and `/rpc/servers` already exposes `contextIds` to the same caller under the same scope.

---

## Important Issues (4 pending)

**H3 — token-subject binding: still not pinned.** `[control-api/test/routes.rpcAccessUsersMcpConnectors.auth.test.ts]`

M12 (delete the guard) and M12b (replace it with a no-op) now go **RED** — real progress, and it proves the test checks effect rather than textual presence. But **M13 and M13b still survive green**, and those are the mutations the finding was about. Details and the reproduction inline.

**H2 — fixed on connectors, still live on `/rpc/servers`.** `[rpc-proxy/src/services/controlApiRestService.ts]`

Reproduced against the real `createApp()`:

| upstream | `/rpc/connectors` | `/rpc/servers` |
|---|---|---|
| 401 | **401** | **500** |
| 403 | **403** | **500** |
| hang | **504** @ 302 ms | — |

The connectors half is genuinely closed and `shouldRefreshRpcToken` is satisfied via the 401. The 403 without `"missing scope"` is correct — that is a subject mismatch, not a scope one. But the sibling route keeps the identical defect, and unlike `/rpc/connectors` it **has a consumer today**. Reasonable as a follow-up; flagging it so it is a decision rather than an oversight.

**R2-1 (new) — `deny all` is not asserted.** `[control-api/test/network.gateway.intent.test.ts]`

Changing `deny all;` to `allow all;` *inside* the `limit_except GET` survives green (41/40/0/1). The assertions verify the block exists and sits in the right place, but not that its body denies. Inline.

**H1 — unchanged, and it lives in `evenfire-infra`.** Not addressed this round, which is correct since the fix is not in this repo. Restating so it is not lost: neither `deploy-clerum-dev.yml` nor `deploy-clerum-prod.yml` restarts `control-api-rpc-gateway`, so the new `location` will not reach the running pod and the endpoint returns 500 until someone recreates it by hand. **The infra PR must land before or with this deploy.**

---

## Suggestions (2)

**R2-2 — the SM3 regression test asserts presence, not effect.** It checks `expect(init.signal).toBeInstanceOf(AbortSignal)`. Removing the signal from the *connectors* fetch goes red, so that one is pinned. Removing it from `fetchUserAllowedServersFromControlApi` leaves **485 tests green**. Three of the four signals are unguarded.

**SM1/SM2 — documented but not recorded.** The documentation is good and explicit: it names the superset and the scope's dual reach, which was exactly the "remove the silent part" option. Two gaps: the note is in `classifyConnector` and the rpc-proxy route but **not in the control-api route**, and **no test pins the decision** — a mutation inverting the behaviour leaves 34 tests green. The call itself ("has a grant?", not "can it run?") is yours to make and I am not re-litigating it; I am only noting that a future refactor could silently reverse it.

---

## Strengths

- The fixes are individually well-targeted, and every commit carries the finding ID — the review-to-fix trail is unambiguous.
- H4 was over-delivered: the ordering assertion was a gap I named but did not request.
- The new auth test is well built where it does reach: it keeps `requireRpcTokenUserMatch` real via `importActual` + spread, asserts a real HTTP 403 through supertest, and backs it with `expect(getUserAgents).not.toHaveBeenCalled()`. That it catches the no-op substitution is proof it tests behaviour, not shape.
- SM3 was fixed across all four call sites rather than just the one I flagged.
- The SM1/SM2 documentation resolves the substantive point I raised, not a narrower reading of it.

---

## Recommended Action

1. **Close H3** — two assertions, given inline. This is the only pending item I would hold the merge for.
2. **Sequence H1** with the `evenfire-infra` companion PR. Happy to open it.
3. **Follow-ups, not blockers:** R2-1, R2-2, `/rpc/servers` 401 mapping, and a test pinning the SM1/SM2 decision.

Verification method, for the record: every RED above exhausted all 3 attempts of control-api's `retry: 2`; every run captured its exit code directly (`--reporter=json`), never through a pipe; and no run reported zero tests.

rpc-proxy/src/services/controlApiRestService.ts
alfredolopez80: Timeout half: **verified closed**, and more thoroughly than I asked. I audited it at runtime rather than by grep — instrumented `global.fetch` and invoked all four exported functions:

```
SIGNAL-AUDIT [{"fetchUserAllowedServersFromControlApi":true},
              {"fetchUserConnectorsFromControlApi":true},
              {"fetchHostConnectionFromControlApi":true},
              {"requestHostWakeFromControlApi":true}]
```

`config.upstreamTimeoutMs` is genuinely used, no magic number: setting `RPC_PROXY_UPSTREAM_TIMEOUT_MS=300` cut at **302 ms**, not 8000. And `isUpstreamTimeoutError` does match `TimeoutError`, so a hung upstream surfaces as **504**, reproduced end to end.

Also: your count was right and mine was wrong. I said "the three fetches in this file"; there are four — the host-connection one is written `await (options.fetchImpl ?? fetch)(...)` and does not match a `fetch(` grep.

---

On the rate-limit half, I mostly agree, with one correction to the reasoning rather than the conclusion.

You are right that **N is not attacker-controlled** — the host fan-out and `getUserAgents` are bounded by what the caller is authorized for, so a user cannot inflate N to amplify. That disposes of the fan-out as an attack vector, and it is a fair reason to treat it as scale.

But the cost I measured was not N. It is the **two full-namespace LISTs** (`mcpservers` and `contexts`) issued on *every* request, which are constant and independent of N. A user with a single agent, polling in a loop, still produces two namespace-wide LISTs per request against the apiserver with nothing in the path to slow it down — `limit_req` is absent from the gateway ConfigMap, and `rateLimiterService` does not cover this rail. What *is* attacker-controlled is the request frequency, not N.

What keeps this bounded, and why I still read it as Medium rather than anything sharper: `CORE_POOL_CONNECTION_TIMEOUT_MS=2000` means requests fail fast instead of queueing, and `CORE_POOL_STATEMENT_TIMEOUT_MS=15000` caps each query, so it degrades loudly rather than hanging. The part worth keeping in the follow-up is that the PG pool (`CORE_POOL_MAX=10`) is **shared**, so the degradation is not confined to whoever is polling.

So: agreed it belongs on the rail rather than in this PR. Worth carrying into that issue that the constant per-request LIST cost is the interesting half, since a fan-out cap alone would not address it.

(And noted on the cache citation.)

PR #498, opened 2026-08-28
feat(connectors): end-user disconnect endpoint for mcp-server grants (U4)

zach88:
## What
- New end-user disconnect endpoint that removes an mcp-server OAuth grant, plus its rpc-proxy route and gateway allowlist location.
- Disconnect call-sites aligned to the converged `buildMcpServerGrantKey` helper.

## Why
Users need to revoke a connector they previously authorized, from the same panel. Unit U4 (backend) of the proactive-connectors spec. Aligning the grant key means disconnect deletes exactly the key the mint and read paths use, so it cannot miss or hit the wrong grant.

## How
- Disconnect branches by auth flavor the same way the existing grant deletion does.
- `buildMcpServerGrantKey` is now the single source used by mint, read and disconnect (D4), removing the duplicated keying that could drift.

## Testing
- control-api: `routes.internal.mcpOauthDisconnect.test.ts`, `services.access.mcpInvocable.grantGate.test.ts`, realPostgres integration `oauth.mcpDisconnect.realPostgres.integration.test.ts`.
- rpc-proxy: `routes/mcpOauthDisconnect.test.ts`.
- Unit and integration suites green.

## Risks / Follow-ups
- Stacked on `feat/mcp-connectors-panel-api` (#497).
- Adds a gateway ConfigMap location, so the nginx workflow-approval gateway needs a reload on deploy. No CRD.
- 11 files across control-api, rpc-proxy, deploy and shared types: one disconnect seam.


alfredolopez80 (commented):
## Review — U4 disconnect endpoint

Multi-agent review on a dedicated worktree, diffed against the **stacked base** `origin/feat/mcp-connectors-panel-api...HEAD` (merge-base `d285ff9e`), not `dev`. Dimensions: per-area code review, security, contract consistency, tests, architecture, comment veracity, and an adversarial pass. Findings are anchored inline below; this summary covers what could not be anchored to a changed line.

**The endpoint itself is well built**, and I want to be specific about what I verified rather than assumed. All 12 guards are fail-closed. `deleteOAuthGrant` is fully parameterised and matches the complete PK per variant, with an explicit `user_id IS NULL` on the `shared` branch so it cannot sweep the other flavor. `getUserContexts` is genuinely fail-closed — no `|| []`, no `?? []`; a DB error propagates to 500 *before* the delete. There is no IDOR: `userId` comes from `req.auth.sub` on every path, with a test that sends `{userId:'attacker'}` and asserts otherwise. And `buildMcpServerGrantKey` converges **field by field across all four call-sites** (mint, sweep, presence-gate, disconnect) **and matches the writer** in `oauth/callback.ts` — disconnect deletes exactly the key the mint writes and the presence-gate reads. Route prefixes are coherent across all four layers, which was the most likely failure mode for a change shaped like this and is not present.

**Counts: 4 Blocker · 5 High · 9 Medium · 8 Low.** Blockers are inline on `configmaps.yaml`, the `grantGate` test, and the token constants; the fourth is below.

### Blocker not anchorable to this diff — a refresh↔DELETE race resurrects the `user` grant

`control-api/src/oauth/tokenHelper.ts:61` and `:162`, `control-api/src/oauth/store.ts:118-136` — unchanged by this PR, but this PR creates the interleaving.

The refresh path persists via `upsertOAuthGrant` (`INSERT … ON CONFLICT DO UPDATE` for `user` grants), with no transaction and no existence recheck. A DELETE landing between the grant read and the refresh write is re-inserted with fresh tokens:

1. mcp-host asks for a token → `getAccessToken` reads the grant, sees it expired, fires the provider refresh (network round-trip, up to 15s).
2. User hits Disconnect → row deleted → 204 → the panel shows "disconnected".
3. The refresh returns OK → the upsert runs → **the row is back, with a new access token and a new refresh token.**

The presence-gate returns `exists:true` again, mcp-host's sweep evicts nothing, the broker keeps minting, and nothing logs the re-creation. The user believes they revoked; the system stays connected.

The reason I am confident rather than speculating is an asymmetry in your own code: for `shared` grants the refresh is a pure keyed `UPDATE` (commented *"Refresh path ONLY"*) which after a DELETE affects 0 rows and **does not resurrect**. One flavor is hardened against exactly this write-after-delete; the other is not. Coverage is zero — `grep -ciE 'race|concurrent|refresh|resurrect'` across the three new test files returns 0.

Suggested fix: make refresh persistence for `user`/`service` a keyed `UPDATE`, as it already is for `shared`, and treat 0 affected rows as `no_grant`.

### Other findings outside the changed lines

- **Disconnect does not revoke anything upstream.** No RFC 7009 call exists anywhere in control-api, rpc-proxy or mcp-host. The refresh token stays valid at the provider, issued access tokens work until expiry (indefinitely for providers that do not expire them), and the app remains listed as authorised in the user's account. Worse for `context`: one member's disconnect discards the *bootstrapper's* tokens while that identity stays authorised upstream. If this belongs to a later PR in the stack there is no marker saying so, and the endpoint comment says "revoke" unqualified.
- **`grantScope` is not CEL-guarded**, despite the comment saying *"immutable per server, CEL-guarded"* and the CRD saying *"Immutable once set"*. There is no `x-kubernetes-validations` rule for `grantScope` or `auth.type`, so the apiserver accepts the flip. Reproduced: after `user` → `context` the key changes shape, the DELETE matches **0 rows**, and the route returns **204 "success"** while the user's grant — and its refresh token — survives. `git blame` puts the premise at `94a241082` (2026-08-11), so this PR inherits it rather than introducing it.
- **No rate limiting anywhere in the chain** for a destructive operation: no limiter middleware in `rpc-proxy/src`, and no `limit_req`/`limit_conn` anywhere in `deploy/`. Each request costs one apiserver `getResource` plus two Postgres queries, uncached and unbounded.
- **~46 of the 127 new lines are literal copies of the mint route** — body validation, CR read + 404, `contextRef` cross-check, membership gate. The direct consequence is that two of the findings below are already duplicated across both surfaces, so each fix has to be applied twice or they diverge. Extracting `resolveMcpServerForUser(gateway, {...}) → {resolved} | {status, error}` would make the "D4" the comments claim actually true.
- **8 of 22 handler branches are untested**, including `server_missing_context` (0 grep hits in tests), `user` without `contextRef`, non-string `mcpServerName`/`contextId`, empty-string `userId`, a non-NotFound `getResource` error, the audit-log emission, and the gate ordering. rpc-proxy has no test for 500/429/503 propagation either — and an unlisted status is forwarded verbatim along with its content-type, HTML included.
- **The normative citations** (`spec 11 U4`, `mini-spec 05 §7`, `invariant §1.4`, `D4`, `F2`) point at documents that were **never committed** — verified absent from the tree, and `git log --diff-filter=D` over `--all` shows no deletion; `.gitignore:111-118` excludes those directories by design. In a public repo that leaves ~50 lines of comment whose authority rests on sources no external contributor can open. `F2` is the sharpest case: it is used in this repo with four unrelated meanings, so `(D4/F2)` is ambiguous even internally. `docs/features/control-ui-contract-test-hardening.md` is a committed mini-spec in a non-ignored directory — the precedent already exists.

### Test and gate status

All exit codes captured with `cmd > log 2>&1; echo "EXIT=$?"`, no pipes. `rpc-proxy` full suite: **EXIT=0**, 490 ran, 490 passed. `control-api`, the three PR files: **EXIT=0**, 24 ran, 21 passed, 3 skipped. `realPostgres` against an ephemeral Postgres 16: **EXIT=0**, 3 ran, 3 passed. `tsc --noEmit` and `npm run build`: **EXIT=0** for both packages. Every run had `total > 0`.

Two things worth flagging rather than glossing. First, the full `control-api` suite with **`--retry=0`** is **EXIT=1** with 2 failures (`routes.mcpSecrets.test.ts`, `workflows.shared.rateLimit.test.ts`); both are green in isolation, so it is cross-file contamination unrelated to this PR — but `retry: 2` in `vitest.config.ts:16` paints them green on every full run. Second, `prettier --check` fails (**EXIT=2**) on 156 baseline files, **none of them from this PR** — an environment issue, but I am not going to call a red gate green.

Mutation-verified: the three central invariants all go red under mutation (pinned `userId`, removed membership gate, 204→404), and the nginx allowlist guard fails correctly when the location is removed, when its verb is changed, and when the route is renamed — 53 tests ran, so it is not a green over zero.

### Description corrections

- *"the nginx **workflow-approval** gateway needs a reload"* → it is `control-api-rpc-gateway`. The workflow-approval gateway is the one that already restarts automatically.
- *"**11 files** … and **shared types**"* → 13 files, and no shared package is touched (`grep -rn "mcpServerName" packages/` → 0 results). The `{mcpServerName, userId, contextId?}` shape is hand-written in 4 places; the future desktop client will be the 5th.
- *"Unit and integration suites green"* → true with the default config; see the `--retry=0` note above.
- No consumer exists yet — no caller of `DELETE /api/v1/mcp-oauth/:name/grant` in `desktop-app`, `ui` or `mcp-host`. Consistent with "U4 (backend)", but worth stating that the endpoint ships inert.

### Decisions I cannot make from the code

1. Who may revoke a shared `context` grant? Today any `member`, with no owner/admin notion. Documented as deliberate — product call.
2. Does upstream revocation belong here, later in the stack, or explicitly out of scope?
3. Should an expelled member be able to revoke their own grant? Touches offboarding policy, not just this endpoint.
4. Is the boundary scanner wrong about dev fixtures? If so, the fix belongs in the scanner.
5. Does this PR block until the infra PR exists? The ConfigMap fix spans two repos and the restart list is frozen by a contract test.

control-api/src/routes/internal/oauth.ts
```diff
@@ -259,6 +261,131 @@ export function createInternalOAuthRouter(gateway: K8sGateway): Router {
...
+  //
+  // End-user "Disconnect <server>". rpc-proxy fronts this and forwards the
+  // session `userId` (auth.sub) over the mutually-authenticated seam
+  // (requireInternalService('rpc-proxy')) — NEVER a client-controlled param
+  // (invariant §1.4). The grant coordinate (oauthClientId + grantScope + the
+  // Context) is derived from the McpServer CR, never the body.
+  //
+  // Authorization BY FLAVOR (spec §2.3, mini-spec 05 §7):
+  //   - `user`    → the caller may only delete their OWN grant — the key is
+  //     built on the forwarded `userId`, so a cross-user delete is structurally
```
alfredolopez80: **High — this guarantee is contractual, not structural.**

`userId` arrives in the **request body** (see the destructuring ~20 lines below) and is accepted after a `typeof`/`length` check. `requireInternalService('rpc-proxy')` authenticates the *service*; it does not bind the token to a user. The property that actually holds is *"rpc-proxy, the only authorised caller, derives `userId` from `auth.sub`"* — which I verified is true — but that lives in **another process**.

The wording matters: "structurally impossible" switches off scrutiny at the one point where this endpoint trusts blindly. Anyone holding the service token can delete any user's grant by naming their `userId` (see my note on `config.ts:643` in the summary).

Suggested: *"a cross-user delete is impossible **for an end user**, because rpc-proxy — the only authorised caller — derives `userId` from `auth.sub`. control-api itself trusts the seam."*

control-api/src/routes/internal/oauth.ts
```diff
@@ -259,6 +261,131 @@ export function createInternalOAuthRouter(gateway: K8sGateway): Router {
...
+  //     impossible.
+  //   - `context` → any MEMBER of the server's Context may revoke the single
+  //     shared grant → blast-radius: the WHOLE Context is disconnected.
+  //
+  // The Context-membership gate runs for EVERY grantScope, using the SAME
+  // primitive as the authorize-URL mint and the callback bootstrap
+  // (`getUserContexts` → user_contexts), so membership authz lives in ONE place
```
alfredolopez80: **High — "ONE place (D4)" is false, in two independent ways.**

1. **The rule is written three times inline**, none factored out: here, in the authorize-URL mint (~`:194-200`), and in `oauth/callback.ts:364-376`. What is shared is the *data reader* (`getUserContexts`), not the *rule*.
2. **The callback bootstrap does not do what this says.** Its membership check sits **inside** `if (subject.grantScope === 'context')` — for `grantScope: 'user'` it never checks membership at all. So "runs for EVERY grantScope, same as the callback bootstrap" describes something that holds in only one of the two branches.

Worth contrasting with `buildMcpServerGrantKey`, which **is** a real single authority — I verified all four call-sites produce an identical key, and that it matches the writer in `oauth/callback.ts`. That is what D4 looks like when it is true.

The risk is concrete: the next person who needs to change the membership rule reads "ONE place" and edits one of three.

control-api/src/routes/internal/oauth.ts
```diff
@@ -259,6 +261,131 @@ export function createInternalOAuthRouter(gateway: K8sGateway): Router {
...
+        let server: McpServerAuthResource
+        try {
+          server = (await gateway.getResource(
+            'mcpservers',
+            mcpServerName,
+            config.mcpServersNamespace
+          )) as McpServerAuthResource
+        } catch (err) {
+          if (err instanceof K8sNotFoundError) {
+            return res.status(404).json({ error: 'server_not_found' })
```
alfredolopez80, line 323: **High — enumeration oracle: this 404 runs before the membership gate.**

The order is `getResource` → **404** → `auth.type` → **400** → `contextId` cross-check → **400** → *only then* membership → 403. Any authenticated user with `mcp:server:invoke` can therefore distinguish, for any name in the `mcp-servers` namespace:

- `404 server_not_found` → does not exist
- `400 not_oauth_server` → exists, not OAuth
- `403 context_membership_denied` → exists, is OAuth, not my Context

With a dictionary of names that maps the OAuth integration inventory of **every** Context in the cluster.

This contradicts the intent stated in the U5 comment in this same file: *"Without a universal gate, any user could start consent for (**and enumerate**) another Context's integration."* The gate exists — it just runs too late. It is also inconsistent with the privacy argument used ~30 lines above to justify the idempotent 204.

Pre-existing in the mint route and duplicated here; moving the membership gate directly after the CR resolution fixes both surfaces at once. **No current test fails if the order is inverted.**

control-api/src/routes/internal/oauth.ts
```diff
@@ -259,6 +261,131 @@ export function createInternalOAuthRouter(gateway: K8sGateway): Router {
...
+          }
+          throw err
+        }
+
+        const authType = server?.spec?.auth?.type
+        // `resolveServerOAuth` (not …Subject): revoke only needs the grant
+        // coordinate, not the client-secret refs — a server whose Secret rotated
+        // away must still be disconnectable. Gate on `auth.type==='oauth'` to
```
alfredolopez80: **High — this justification describes a scenario that cannot happen.**

Neither `resolveServerOAuth` nor `resolveServerOAuthSubject` reads a k8s Secret. `…Subject` only validates that `spec.oauth.provider` / `clientIdRef` / `clientSecretRef` **exist on the CR**. A Secret rotated or deleted from the cluster does not make it return null — so "a server whose Secret rotated away must still be disconnectable" is not the reason the two differ.

The **other** half of the comment is correct and sufficient: *"revoke only needs the grant coordinate, not the client-secret refs."* I verified that by exhaustive sweep — 81 field combinations, `violations=0`: `…Subject` is a strict subset, and `oauthClientId`/`grantScope`/`contextRef` never diverge. Using `resolveServerOAuth` here is the right call.

Risk of leaving it: someone "unifies" the two for D4 tidiness, believing the Secret case is load-bearing when nothing changes functionally. Suggested fix: drop the Secret clause, keep the coordinate reason.

*(Separately, the `auth.type==='oauth'` alignment claim in this same comment is **true** — I checked it is character-identical in the U1 classifier, the presence-gate and the mint.)*

control-api/src/routes/internal/oauth.ts
```diff
@@ -259,6 +261,131 @@ export function createInternalOAuthRouter(gateway: K8sGateway): Router {
...
+        }
+
+        // Context-identity servers: the AUTHORITATIVE Context is spec.contextRef.
+        // A body contextId, if present, must match it (cross-context guard).
+        if (resolved.grantScope === 'context') {
+          if (!resolved.contextRef) {
+            return res.status(400).json({ error: 'server_missing_context' })
+          }
+          if (typeof contextId === 'string' && contextId !== resolved.contextRef) {
+            return res.status(400).json({ error: 'context_mismatch' })
```
alfredolopez80, line 347: **High — this is an equality oracle for `spec.contextRef`.**

Because this cross-check runs *before* the membership gate below, a non-member can guess a `contextId` and read the answer off the status code. Reproduced against the route with a caller who was a member of neither:

```
contextId='ctx-GUESS'  → 400 {"error":"context_mismatch"}
contextId='ctx-SECRET' → 403 {"error":"context_membership_denied"}
```

So a wrong guess and a right guess are distinguishable, which confirms which Context any given server belongs to. Same fix as the 404 above: run the membership gate first.

No test covers this — the `context_mismatch` test and the non-member test never intersect.

control-api/src/routes/internal/oauth.ts
```diff
@@ -259,6 +261,131 @@ export function createInternalOAuthRouter(gateway: K8sGateway): Router {
...
+        }
+
+        // Universal Context-membership gate (every grantScope) — same rule +
+        // primitive as the authorize-URL mint (D4). `spec.contextRef` is
+        // CRD-required + singular; a server that somehow lacks it cannot be
+        // membership-verified → fail closed.
+        if (!resolved.contextRef) {
+          return res.status(403).json({ error: 'context_membership_denied' })
+        }
+        const { contextIds } = await getUserContexts(userId)
```
alfredolopez80, line 358: **High — this gate also blocks a user from revoking their _own_ grant, and there is no other path.**

For `grantScope: 'user'` the delete key is already bound to the caller's `userId`, so membership adds no authorisation here — only anti-enumeration. But it does add a failure mode: a user removed from the Context (or with `lifecycle_state != 'active'`, which `getUserContexts` filters on) gets **403** deleting a row whose `user_id` is themselves.

And there is no alternative route. `mcpserver` grants are neither listed nor deletable through the self-service surface — `listUserOAuthGrants` and `external/oauthGrants.ts:72` both filter `owner_kind = 'recipe'` — and no `deleteOAuthGrant` call-site runs on member removal.

Net effect: losing membership **permanently freezes** that user's encrypted refresh token and removes their only button to withdraw it. In a revocation endpoint, fail-closed fails toward credential **retention** — the inverse of the property it has in the mint this guard was copied from.

Options: allow self-delete unconditionally for `user` scope and keep the gate for `context`; or, to preserve anti-enumeration, return the idempotent `204` instead of `403`.

*(Verified clean, for the record: `getUserContexts` itself is properly fail-closed — no `|| []`, no `?? []`, a DB error propagates to 500 **before** the delete. That part is right.)*

control-api/src/routes/internal/oauth.ts
```diff
@@ -259,6 +261,131 @@ export function createInternalOAuthRouter(gateway: K8sGateway): Router {
...
+        }
+
+        // Build the flavored delete key from the SAME source of truth the
+        // grant-presence gate reads by (D4/F2). null ⇒ fail-closed.
+        const key = buildMcpServerGrantKey(resolved, {
+          mcpServersNamespace: config.mcpServersNamespace,
+          mcpServerName,
+          userId,
+        })
+        if (!key) {
```
alfredolopez80, line 370: **Low — unreachable branch, with the wrong error code.**

`buildMcpServerGrantKey` returns null in exactly two cases, both already intercepted above: `context` without `contextRef` (cut at the `server_missing_context` check ~25 lines up, and again at the `contextRef` gate), and `user` with an empty `userId` (cut by the body validation). Confirmed with a probe: a `context` server without `contextRef` answers from the earlier check, never from here.

Defence in depth is fine, but `server_missing_context` would be semantically **false** for the `user` case, which is the only one this could theoretically catch. The mint uses `invalid_request` in the same position and labels it *"Unreachable"* — worth matching, for consistency and so the next reader is not misled about which condition fired.

control-api/src/routes/internal/oauth.ts
```diff
@@ -259,6 +261,131 @@ export function createInternalOAuthRouter(gateway: K8sGateway): Router {
...
+          return res.status(400).json({ error: 'server_missing_context' })
+        }
+
+        await deleteOAuthGrant({ query: (text, values) => pool.query(text, values) }, key)
+
+        // Audit trail (sensitive: a revocation). Structured, pino-redacted; no
+        // tokens/secrets are read or logged here.
+        req.log?.info(
+          {
+            event: 'mcp_oauth_grant_revoked',
```
alfredolopez80, line 393: **Medium — the audit event for a destructive action records no actor, and fires even when nothing was deleted.**

Two separate issues:

1. **No `userId`, no `contextRef`.** For `grantScope: 'context'` — whose blast radius the comment above describes as *"the WHOLE Context is disconnected"* — this log cannot answer the one forensic question that matters: *who* disconnected the team. `correlationId` is not a substitute: `correlationIdMiddleware` accepts it from the client header when it is a valid UUID, so it is not a trustworthy identifier. `appendControlApiPermissionEventsInTransaction` already records `operatorSub` for access changes; an OAuth credential revocation is at least as sensitive.
2. **Emitted unconditionally.** `deleteOAuthGrant` returns `void` and never inspects `rowCount`, so this fires identically when 0 rows matched. Reproduced. An audit review would count revocations that never happened — and combined with the `grantScope` mutability issue in my summary, that is exactly the case where the row survives *and* the log says it was revoked.

Minimal fix: have `deleteOAuthGrant` return `rowCount`, and log `userId` plus `deleted: boolean`.

*(`oauthClientId` here is fine — it is `spec.oauth.id`, a logical identifier, not a credential. I checked the pino redact config covers the real secrets and that none are read on this path.)*

rpc-proxy/src/routes/mcpOauth.ts
```diff
@@ -116,5 +116,88 @@ export function createMcpOauthRouter(): Router {
...
+  //   - 401 → rpc-proxy↔control-api service-token misconfig → coerce to 502 so
+  //     the user does not read it as their own credential failure;
+  //   - 403 → legitimate `context_membership_denied` → propagate verbatim;
+  //   - 404 (`server_not_found`) / 400 (`not_oauth_server`, `context_mismatch`,
+  //     `server_missing_context`, `invalid_request`) → propagate verbatim;
+  //   - 204 → success (idempotent; empty body).
+  router.delete(
+    '/mcp-oauth/:mcpServerName/grant',
+    requireRpcAuth,
+    requireScope('mcp:server:invoke'),
```
alfredolopez80, line 145: **High — this scope is materially weaker than "the capability to invoke", for a destructive action.**

The comment above argues revoking is the inverse of connecting, so no new scope is needed. But the **invoke** rail applies the per-Context allowlist (`resolveInvocableMcpServersForContexts`, `services/access/mcpInvocable.ts:300`) **in addition to** membership. This route applies membership only.

The U5 comment in `internal/oauth.ts` states the difference explicitly: *"connect shares the `mcp:server:invoke` scope with invoke — which **additionally** requires the server to be in the caller's per-context allowlist."* For connect the omission is justified there by chicken-and-egg (no grant exists yet). **Disconnect has no chicken-and-egg** — the grant exists by definition — so the exclusion is neither justified nor documented.

Concrete case: user `U` is a member of `ctx-a`; server `github` lives in `ctx-a` but is not in any agent allowlist visible to `U`, so `U` cannot see or invoke it. `DELETE .../github/grant` → membership passes → **204, grant deleted**. `U` destroys a connection they could not use.

Related, and worth an explicit product decision rather than a code fix: for `grantScope: 'context'` the key carries no `userId`, and this scope is issued to **every** role including `member` (`utils/auth/rpcAuthToken.ts:26-45`). One request from any member revokes the shared connector for the whole Context, and recovery requires someone to redo the provider consent flow. That asymmetry — revoking needs nothing, restoring needs provider credentials — is not captured by "revoking is the inverse of connecting". It **is** documented and tested, so I am reading it as deliberate; flagging it as the thing most worth confirming.

rpc-proxy/src/routes/mcpOauth.ts
```diff
@@ -116,5 +116,88 @@ export function createMcpOauthRouter(): Router {
...
+            'content-type': 'application/json',
+          },
+          body: JSON.stringify({
+            mcpServerName,
+            userId, // from req.auth.sub — never the body
+            ...(contextId ? { contextId } : {}),
+          }),
+          signal: AbortSignal.timeout(config.upstreamTimeoutMs),
+        })
+      } catch {
```
alfredolopez80: **Low, but it is the root of three others — a bare `catch` with no binding and no log.**

An `AbortError` from the 8s timeout, an `ECONNREFUSED`, a TLS failure and a config `RangeError` all produce the same response and **zero** operator signal. For a timeout the message is simply false: control-api *was* reachable, it just did not answer in time. The app even has a handler that maps timeouts to 504 (`app.ts:51-54`) — this route never reaches it.

rpc-proxy does have a logging convention (`console.warn('[RPC_PROXY] …')` in `routes/rpc.ts`, `console.error` in `routes/desktopProxy.ts`), so this is not "no logger available". A **failed revocation** leaves no trace on either side, since control-api only audits successful ones.

Reproduced edge case that shows the cost: with an unparseable `RPC_PROXY_UPSTREAM_TIMEOUT_MS`, `AbortSignal.timeout(NaN)` throws **inside** this `try`, this `catch` swallows it, and every request to the route returns 502 with not one line of log.

```
[TIMEOUT_MS=abc → AbortSignal.timeout(NaN)] client sees → 502 {"error":"control_api_unreachable"}
```

Fix: bind the error, log it, and distinguish `AbortError` (→504) from a genuine connection failure.

*(Same pattern exists in the authorize-URL route above — `git blame` puts that one in the base branch, so this PR duplicates it rather than introducing it.)*

rpc-proxy/src/routes/mcpOauth.ts
```diff
@@ -116,5 +116,88 @@ export function createMcpOauthRouter(): Router {
...
+      // 401 = service-token misconfig — do NOT surface as the user's own auth
+      // failure. 403 IS a legitimate membership denial → propagate.
+      if (upstream.status === 401) {
+        res.status(502).json({ error: 'control_api_auth_failed' })
+        return
+      }
+
+      // Success: control-api answers 204 (no body). Forward it verbatim.
+      if (upstream.ok) {
+        res.status(upstream.status).end()
```
alfredolopez80: **Low — any 2xx is treated as success, not just 204.**

The contract documented ~50 lines above says *"204 → success (idempotent; empty body)"*, but this accepts anything `upstream.ok`. That matters because rpc-proxy does not talk to control-api directly — it goes through an nginx gateway (`RPC_PROXY_CONTROL_API_BASE_URL` points at `control-api-rpc-gateway`). A 200 produced by that intermediary becomes a **false "disconnected"** for the user with the grant still intact, and `.end()` silently discards the body that would have shown otherwise.

Reproduced:
```
[upstream 2xx that is NOT 204] client sees → status=200 content-type=null body=""
```

Fix: assert `upstream.status === 204` and coerce any other 2xx to 502.

For contrast, the mint route above pins `res.status(200)` and validates the response shape — which is why the *"mirrors the authorize-URL route exactly"* claim holds for the error paths but not for this one.

rpc-proxy/src/routes/mcpOauth.ts
```diff
@@ -116,5 +116,88 @@ export function createMcpOauthRouter(): Router {
...
+        return
+      }
+
+      // Forward the upstream JSON error verbatim.
+      const upstreamText = await upstream.text().catch(() => '')
```
alfredolopez80, line 224: **Low — two silent degradations on one line.**

`.catch(() => '')` turns a failed body read into an **empty body** that is still sent with `content-type: application/json`. Any client doing `JSON.parse(await res.text())` throws `SyntaxError: Unexpected end of JSON input`, turning a legitimate 403 into a desktop crash. Reproduced:
```
[error-body read fails] client sees → status=403 content-type=application/json body=""
```

And the upstream `content-type` is reflected verbatim below. With nginx in the path, an HTML error page reaches the desktop as `text/html`:
```
[nginx deny-all 403 HTML] client sees → 403 text/html "<html><head><title>403 Forbidden</title>…"
```
Note the worst part: by status alone that is **indistinguishable** from a legitimate `context_membership_denied` — which is exactly the failure the deploy issue on the ConfigMap produces.

I checked and ruled out reflected XSS here: nothing user-controlled reaches the response body (`mcpServerName` is validated and travels in the JSON body, never in the upstream URL, which is constant), and the client is Electron's main process, not a renderer. So this is robustness, not a vulnerability.

Fix: pin `application/json` and synthesise a structured error when the upstream body is missing or not JSON.

deploy/base/control-plane/configmaps.yaml
```diff
@@ -263,6 +263,20 @@ data:
           proxy_set_header X-Forwarded-Proto $scheme;
         }
 
+        # spec 11 U4: revoke an OAuth mcp-server grant (desktop "Disconnect").
+        # rpc-proxy is the sole caller; requireInternalService('rpc-proxy') gates
+        # it at control-api.
+        location = /api/v1/internal/mcp-oauth/grant {
```
alfredolopez80, line 269: **Blocker — this location will not be live after deploy. Confirmed against `evenfire-infra`.**

The block itself is correct — I verified it empirically against `nginx:1.30.1-alpine`, the exact image from the Deployment: `nginx -t` → `EXIT=0`; `limit_except DELETE` correctly denies GET/HEAD/POST/PUT/PATCH/OPTIONS (the implicit "GET enables HEAD" exception applies only to `limit_except GET`); query strings and doubled slashes still hit the exact match; and the block is byte-identical to its `sandbox-ui/oauth/grant` sibling after path normalisation.

The problem is that it never reaches a running pod:

1. `control-api-rpc-gateway.yaml:68-77` mounts `nginx.conf` with **`subPath`** — a ConfigMap volume mounted that way is **never** refreshed by the kubelet. Not eventually; never.
2. The ConfigMap is under `resources:` (not `configMapGenerator`), so no hash suffix and no rollout on apply. No `checksum/config` annotation, no reload sidecar (grep across `deploy/` → nothing).
3. In `keyper-labs/evenfire-infra`, both deploy workflows restart **exactly four** deployments in *"Restart ConfigMap consumers that require pod recreation"*: `nginx-workflow-approval-gateway`, `clerum-workflow-approval-request-reader`, `host-context-controller-api-gateway`, and `control-api`. **`control-api-rpc-gateway` is not one of them** — across that whole repo it appears only as a base URL in the `rpc-proxy` overlays.

There is a direct precedent. That workflow's own comment reads:

> `# PR #422 adds v2 locations to the HCC gateway ConfigMap. Its nginx.conf is mounted`
> `# via subPath, so restart and wait for the new pods before the final rollout gate`

**#422 is the same change as this one**, and it required editing both infra workflows, locked in by `scripts/tests/test-pr422-hcc-gateway-rollout.sh` (exact-count assertions per deployment, plus ordering). This is the third occurrence of the pattern and the first without a matching infra change.

**What the user sees:** the pod keeps serving the old config → this DELETE falls into `location / { return 403; }` → nginx returns 403 + HTML → rpc-proxy forwards it verbatim, because its contract says *"403 → legitimate `context_membership_denied` → propagate verbatim"* → **"you are not a member of this Context"**, when the truth is that the gateway was never restarted.

**Fix spans two repos:**
- `evenfire-infra`: add `rollout restart` + `rollout status` for `control-api-rpc-gateway` to both workflows, and extend the contract test so a future PR cannot silently drop them.
- Here: extend `assert_deploy_dev_restarts_subpath_gateway_after_apply` (`scripts/tests/test-gcp-scripts.sh:1195-1207`) to all **4** gateways that mount `nginx.conf` with `subPath`, not just one — otherwise this recurs on every PR touching one of those ConfigMaps.

Also: the PR description says the **workflow-approval** gateway needs the reload. That is the one gateway that already restarts automatically; the one you changed is `control-api-rpc-gateway`.

*(Credit where due: the CI allowlist guard `network.gateway.intent.test.ts` does cover this route — it derives routes from the code rather than a hardcoded list, which is why it is not in the diff yet still passes. I validated it is alive by triple mutation.)*

control-api/test/services.access.mcpInvocable.grantGate.test.ts
```diff
@@ -18,6 +18,17 @@ vi.mock('../src/oauth/store.js', async importActual => {
   }
 })
 
+// The resolver logs the fail-closed exclusion through the pino child logger
+// (`rootLogger.child({ module: 'mcp-connectors' })`), not console.*. Capture the
+// child's `.error` so the fail-closed-logging assertion survives the migration.
+const loggerMock = vi.hoisted(() => {
```
alfredolopez80, line 26: **Blocker — this block asserts nothing, and its comment says it does. Proven by a green mutation.**

The comment states: *"Capture the child's `.error` so the fail-closed-logging assertion survives the migration."* There is no such assertion — `loggerMock` appears only here and in the `vi.mock` below, **never in an `expect`**:

```
grep -n "loggerMock" …/grantGate.test.ts   → only lines 24 and 30
grep -rn "grant-presence check failed" control-api/test   → GREP_EXIT=1 (0 matches)
```

Mutation test — removed the `log.error` from the fail-closed path of `computeGrantPresence` (`services/access/mcpInvocable.ts:286-289`):

```
npx vitest run --retry=0 test/services.access.mcpInvocable.grantGate.test.ts
EXIT=0 | Tests 9 passed (9)      ← green with the invariant destroyed
```

The comment also names the wrong logger: it says `rootLogger.child({ module: 'mcp-connectors' })`, but the `log.error` under test is `child({ module: 'mcpInvocable' })`. `connectorsLog` is only used by `resolveConnectorsForAgents`, which this file does not exercise.

As it stands the mock silences the real logger for the whole file while reading as coverage. Either add the `expect(loggerMock.error).toHaveBeenCalledWith(...)` the comment promises, or drop the block — what should not stay is a comment claiming a guarantee that is not there.

*(This is the one place I would push back hardest, because the rest of the test work in this PR is genuinely strong: assertions on status, payload and exact SQL params, `deleteCalls()).toHaveLength(0)` on rejection paths, and no empty `toBeDefined()` anywhere. I mutation-tested the three central invariants — pinned `userId`, removed membership gate, 204→404 — and all three go red. This block is the outlier.)*

control-api/test/routes.internal.mcpOauthDisconnect.test.ts
```diff
@@ -0,0 +1,239 @@
...
+ * realPostgres integration test (T1/T4).
+ *
+ * T3: at the parent sha (5d9b22902) the route does not exist → Express answers
+ * 404, so both the 204 (delete) and 403 (authz) assertions fail there.
+ */
+
+// Internal service tokens are the repo-wide dev fixtures (see the sibling
+// routes.internal.* tests). Kept in constants and interpolated so the public
+// boundary scanner never sees a materialized token literal after `Bearer`.
+const RPC_PROXY_TOKEN = 'dev-rpc-proxy-token'
```
alfredolopez80, line 32: **Blocker — this change works around the secret scanner instead of the cause.**

The commit message for `314dff6a` states the intent plainly: *"Kept in constants and interpolated so the public boundary scanner never sees a materialized token literal after `Bearer`"*, and *"Token values unchanged"*.

The value still sits in this file, on this line, in a file that ships to the public repo. What changed is the **shape** the matcher looks for, not the fact it reports.

And it is not cosmetic: this fixture value is the same as the `internalServiceTokens` dev fallback in `control-api/src/config.ts:643` — a `||` onto a hardcoded dev map with **no `NODE_ENV` guard**. If that env var ever arrives empty, control-api accepts that credential, and it is the credential that authorises deleting grants with an **arbitrary** `userId`. The minimum-length check does not stop it. The scanner was right on substance.

Worth noting the asymmetry on that seam: the *client* side (`rpc-proxy/src/config.ts:176-180`) uses `requiredOrDevDefault` and **does** fail in production. The *server* side — the one that authorises — fails open. `git blame` puts that at `5d663a49` (2026-07-14), so it predates this PR; but this PR is the first to hang a destructive, free-`userId` operation off that seam.

Suggested: revert the interpolation and fix it properly — move fixture tokens to a helper outside the scanner's scope, or, if the scanner is wrong to treat a dev fixture as a secret, fix **the scanner**. Working around a guard leaves the next person to rediscover why it fired.

*(The other half of that commit — `postgres://user:pass@host` → `postgres://<user>:<pass>@<host>` — is a legitimate fix: the angle brackets declare it as a placeholder.)*

control-api/test/oauth.mcpDisconnect.realPostgres.integration.test.ts
```diff
@@ -0,0 +1,174 @@
...
+        ownerKind: 'mcpserver',
+        recipeNamespace: NS,
+        recipeName: 'gdrive',
+        userId: 'user-b',
+        oauthClientId: 'google-drive',
+      })
+    ).toBe(true)
+  })
+
+  it('context flavor: deletes the single shared grant for the whole Context', async () => {
```
alfredolopez80: **Medium — this case proves the delete happens, but not that it is narrow.**

The `user`-flavor test above does the right thing: it seeds two grants (`user-a`, `user-b`), deletes one, and asserts the other survives. This `context` case seeds a **single** shared grant and deletes it — there is no control row, so nothing here would catch an over-broad DELETE.

Concretely: if the `shared` branch of `deleteOAuthGrant` lost its `context_id = $4` or `recipe_name = $3` predicate, this test would still pass while the query wiped shared grants across Contexts.

I did verify the query itself is correct today — three parameterised statements, full PK per variant, explicit `user_id IS NULL` on the `shared` branch so it cannot sweep `user` rows. The gap is that the test would not notice if that changed.

Fix: seed a second shared grant (different `contextId`, or same Context with a different `recipeName`) and assert it survives — mirroring what the `user` case already does.

*(Also worth a look: the third case in this file is a pure unit assertion on `buildMcpServerGrantKey` and never touches the DB, yet it sits inside `describeRealPostgres` — so it is skipped in the ordinary matrix for no reason.)*

zach88:
Thanks for the deep review — went through every finding against the actual code and repro'd each one. Disposition below. The code fixes are committed locally (one commit per finding, each with a repro test that fails at the pre-fix head `314dff6a`); they'll land on the next push. Ran code-review + security + collateral over the fix diff — all clean.

## Addressed in this round

- **Refresh↔DELETE race resurrecting the `user`/`service` grant** — good catch, this was the real one. The refresh path now persists via a keyed `UPDATE … RETURNING id` for all flavors (new `refreshOAuthGrantTokens`), and `tokenHelper` maps 0 affected rows → `no_grant` instead of re-inserting. `upsertOAuthGrant` is now consent/insert-only. A DELETE landing mid-refresh no longer resurrects the row. Repro test: `oauth.tokenHelper.refreshRace.test.ts` (returns `ok`/resurrects at parent, `no_grant` after fix).
- **Audit event with no actor + fires on 0 rows** — `deleteOAuthGrant` now returns `rowCount`; the `mcp_oauth_grant_revoked` log carries `userId`, `contextRef` and `deleted: boolean`.
- **`grantGate` test asserts nothing** — added the real `expect(loggerMock.error)…` assertion and fixed the comment to the right module (`mcpInvocable`, not `mcp-connectors`). Mutation-verified: removing the fail-closed `log.error` now turns it red.
- **`context`-flavor integration test not narrow** — now seeds a sibling shared grant in `ctx-B` and asserts it survives the `ctx-A` delete, mirroring the `user` case.
- **rpc-proxy disconnect route** — bound+logged the bare `catch` and split timeout→504 from unreachable→502; success is now exactly-`204` (any other 2xx → 502); error forwarding is always `application/json` with a synthesized `{error}` when the upstream body isn't JSON (no more reflected nginx HTML that looks like a legit 403).
- **Unreachable null-key branch** → `invalid_request` (+ "Unreachable" comment), and the three comment-veracity nits reworded.

## Deploy blocker (B2) — real, but the fix lives in `evenfire-infra`

Confirmed the substance: `control-api-rpc-gateway` mounts `nginx.conf` via `subPath` (never refreshed without a pod restart), the ConfigMap is under `resources:` (no hash/rollout), and the deploy workflows restart only 4 deployments — `control-api-rpc-gateway` isn't one of them. So the new `location` block won't be live after deploy, and the DELETE would fall into `location / { return 403 }`. Since the fix is in the infra repo, tracked it there with the exact steps (add `rollout restart` for `control-api-rpc-gateway` to both deploy workflows + extend the contract test). **This needs to land in `evenfire-infra` before/with merge.**

Minor: the PR description says the *workflow-approval* gateway needs the reload — it's actually `control-api-rpc-gateway` (workflow-approval already restarts automatically).

## Follow-ups (real, but out of scope for U4 — filed as issues)

- **#503** — losing Context membership permanently freezes a user's own `user` grant (the membership gate 403s their own self-delete, and there's no alternative deletion path for `mcpserver` grants → orphaned encrypted refresh token). Needs an offboarding-policy call.
- **#504** — disconnect doesn't revoke upstream (RFC 7009); the refresh token stays valid at the provider. Either implement per-provider revocation or scope U4 explicitly as "delete local grant" and rename the audit event.

## False positives — flagging with evidence

1. **`grantScope`/`auth.type` not CEL-guarded** (the reproduced `user→context` flip → 0-row DELETE → 204): it *is* guarded. `charts/clerum-crds/crds/mcpserver.yaml:505-539` has `x-kubernetes-validations` immutability rules for `grantScope`, `id`, `provider` and `contextRef`, and closes the two-step bypass (`spec.oauth` can't be removed once set). The apiserver rejects the flip — the DELETE-0-rows path isn't reachable that way.
2. **`config.ts:643` dev-token fail-open in prod**: it's fail-*closed*. `config.ts:1220-1227` rejects any `dev-*` internal service token at production startup (throws → crash at boot). If `CONTROL_API_INTERNAL_SERVICE_TOKENS` is empty in prod the pod won't boot; it never accepts the dev token. So the arbitrary-`userId` delete premise doesn't hold in prod. (The test-fixture scanner nit stands as a process point, just not a security one.)
3. **"cross-user delete structurally impossible"** (oauth.ts:274): you're right the wording is too strong — control-api does trust the `userId` from the body; the property holds because the service token is infra-only and rpc-proxy derives `userId` from `auth.sub`. Reworded the comment to say "impossible for an end-user … control-api trusts the seam."

## By design (documented + tested, leaving as-is)

- Disconnect authorizing on Context membership only (weaker than the invoke allowlist), and any `member` being able to revoke a shared `context` connector — deliberate, documented at `oauth.ts:276-277` and tested. Flagging as the thing most worth a product confirm, but not changing it here.
- The enumeration oracle (404/400 before the membership gate): real but low — needs a team token + name-guessing, only leaks metadata, and it already exists identically in the authorize-url route (this PR doesn't introduce it). Can file separately if we want to move the gate earlier on both surfaces.

alfredolopez80:
## Round 2 — verification of the R1 fixes

Re-verified at `b0b2ffba` on a fresh worktree, against the head I originally reviewed (`314dff6a`). Standard applied throughout: a fix counts as real only if the new test **fails at the pre-fix head** and/or goes **red under mutation of the fix**. A real fix and a cosmetic one read identically, so nothing below is accepted on the strength of a commit message. All exit codes captured with `cmd > log 2>&1; echo "EXIT=$?"` — no pipes.

Thanks for the disposition — it made this round fast, and two of your rebuttals were correct.

## First: two of my R1 findings were wrong

I verified both against `314dff6a`, the exact head I reviewed. You were right on both, and I am withdrawing them.

**M-2 (`grantScope` not CEL-guarded) — false positive.** `charts/clerum-crds/crds/mcpserver.yaml:505` does carry the immutability rule for `grantScope`, plus `id`, `provider` and `contextRef`, plus the rule closing the two-step drop-and-readd bypass. It was already there at the head I reviewed.

Worth stating *how* I got it wrong, because the mechanism is the useful part: my analysis marked this **reproduced**. What it actually reproduced was the *consequence* — "if the flip happened, the derived key differs and the DELETE matches 0 rows" — which is trivially true. It never verified the *premise*, that the apiserver would accept the flip. Half a causal chain, carrying a verification mark that made it look rigorous. That is worse than an unmarked guess.

**M-4 (dev service-token fallback fails open) — false positive.** `control-api/src/config.ts:1220-1227` rejects any `dev-*` internal service token at production startup; the pod crashes at boot rather than accepting it. The seam is fail-**closed** in production. I looked at the `||` and never looked for a startup validator.

**Consequence for B-4:** its security premise *was* M-4 — "that token is the live credential authorising arbitrary-`userId` deletes in prod". That premise is false, so **B-4 drops from Blocker to a process nit**. What survives is only the point that a guard was worked around rather than its cause addressed. Not worth blocking a merge over, and the fixture tokens being unchanged is fine.

## Resolved — verified by mutation

**B-1 · refresh↔DELETE resurrection race.** `refreshOAuthGrantTokens` is a pure keyed `UPDATE` across all three flavors with `RETURNING id`; `tokenHelper.ts:173` maps 0 rows → `no_grant`. I confirmed **all** post-refresh persistence paths use it — the three remaining `upsertOAuthGrant` call-sites are in `callback.ts:263,274,400`, consent-after-fresh-code, which is legitimate: creating the row is the point of consent. The dangerous case was the automatic refresh with no human in the loop, and that path is closed. Making `upsertOAuthGrant` throw for `shared` is a good extra.

- **Pre-fix RED:** restoring the `314dff6a` versions of `tokenHelper.ts` + `store.ts` gives `EXIT=1`, `expected 'ok' to be 'no_grant'` — literally the resurrection bug.
- **Mutation RED ×2:** neutralising the 0-row mapping, and reverting the `user` branch to `INSERT … ON CONFLICT`. Both assertions in your new test are load-bearing.
- Per-flavor coverage confirmed for `user`, `service` and `shared` — I suspected `service` was a gap, measured it, and it is not.

**B-3 · the mock that asserted nothing.** This is the one that had to be mutation-verified, since in R1 it was green with the invariant destroyed. Removing the `log.error` from `computeGrantPresence` now gives `EXIT=1` (`1 failed | 8 passed`); in R1 the same mutation gave `EXIT=0`. The comment names `mcpInvocable` correctly now.

**M-5 · `context`-flavor delete isolation.** Verified against real Postgres 16, not declared blind: `EXIT=0`, 3 tests actually executed. Dropping `context_id = $4` from the `shared` branch turns it red — the `ctx-B` sibling disappears. Blast radius is detected.

**M-3 · audit with actor.** `deleted` is derived from a real `rowCount` (signature widened to `Promise<number>`), not hardcoded, and your test covers both states so it is not vacuous. I checked `utils/log/redact.ts` — `userId` is not in the redact list, so the field genuinely survives into the record. None of the other four call-sites break on the widened return.

**Exactly-204** and **`invalid_request` on the unreachable branch** — both confirmed; 200/202/205 → 502 with a warning, and the 204 pass-through is preserved.

## Partial — three specific gaps

**Comment veracity: 3 of 4.** The cross-user wording and the Secret-rotation clause are now accurate. Two gaps:

1. **"ONE place (D4)" survives 90 lines above.** You fixed the DELETE comment I anchored to (`:281-287`), but the identical claim is intact in the mint at `internal/oauth.ts:190-191`. Counted implementations of the rule (`getUserContexts` + `.includes` → deny): **three** — `:198-200` (mint), `:358-360` (disconnect), `callback.ts:372-375` (bootstrap). This one is partly on me: I anchored to one line and described the class in the body, so the fix followed the anchor.
2. **"Error propagation mirrors the authorize-URL route exactly" (`rpc-proxy/src/routes/mcpOauth.ts:135`) is now *more* false than in R1.** After your A/B/C fixes the two routes diverge in four new dimensions that did not exist before: 504-vs-502 on timeout, log-vs-silence, JSON-normalised-vs-verbatim-HTML, and exactly-204-vs-any-`ok`. Fixing code turns true comments false — worth a second pass over that block.

**rpc-proxy timeout/normalisation: the sibling route was left behind, and it is the one with a client.** Reproduced end-to-end with real fetch on Node v24.18.0 (matching `engines.node`): your detection is correct — `AbortSignal.timeout` yields `DOMException` with `name === 'TimeoutError'`, and `DOMException instanceof Error` holds, so the guard works. `[A1] 504 control_api_timeout` / `[A2] 502 control_api_unreachable`, both logged. Deciding JSON-ness by **parsing** rather than content-type is the robust choice, and I confirmed it ignores lying content-types in both directions.

But `mcpOauth.ts:80` (authorize-url) is still a bare `catch {`, and `:112-115` still forwards nginx HTML verbatim:

```
[authorize-url real timeout]  502 control_api_unreachable   warnings: []
[authorize-url nginx 403]     403 text/html  "<html>…403 Forbidden…nginx/1.25.3…"
```

That is the route that **does** have a desktop client wired (`rpcProxyClient.ts`), while disconnect has none yet — so the fix landed on the surface with no consumer and skipped the one with a consumer.

Also still open on that path: the `NaN` root cause is untouched (`config.ts:167` is still `Number(env || 8000)`, feeding **12** `AbortSignal.timeout` call-sites across rpc-proxy; a config error now surfaces as `control_api_unreachable` — better than silent, still misattributed), and `correlationId` is still forwarded verbatim to the client on a 500. Minor: a JSON **array** body passes through the `parsed && typeof parsed === 'object'` predicate unsynthesised.

## Still blocking

**B-2 — the gateway reload.** We agree entirely, and `deploy/` being unchanged is correct since the fix lives in `evenfire-infra`. Recording it as the one open blocker per your own line: *"needs to land in `evenfire-infra` before/with merge."* Nothing to argue, just not done yet.

## Where we still differ, narrowly

**H-3 (enumeration oracle).** Confirmed unchanged: `server_not_found`:323, `not_oauth_server`:337, `server_missing_context`:344, `context_mismatch`:347, all before the gate at :356/:360. Your three reasons are all true — it needs a team token, it leaks only metadata, and it pre-exists identically in authorize-url (I said that last one myself in R1). The narrow disagreement: you accept the substance, so filing it costs nothing, and moving the gate fixes both surfaces at once rather than leaving a known-and-accepted gap untracked.

**H-1 / H-2** — confirmed unchanged and documented, so reading them as deliberate. H-2 (any `member` revoking the shared connector for a whole Context) is the one I would most want an explicit product sign-off on, not a code change.

**#503 / #504** — both genuinely need a policy call rather than code. Filing them was the right move.

## Two small things this round introduced

- `store.ts:329` still says *"Distinct from the refresh path (`upsertOAuthGrant` above)"*. After B-1 the refresh path is `refreshOAuthGrantTokens`, so a previously-correct comment is now wrong. Pre-existing text (`a195b607`), newly false.
- `grantGate.test.ts:96-98`: `beforeEach` resets `grantExists` but not the logger spy. No false green today — the mutation confirms it — but a future test in that file that triggers a DB error would let the assertion pass on someone else's emission. A `mockClear()` closes it.

## Summary

| | |
|---|---|
| **Resolved (mutation-verified)** | B-1, B-3, M-3, M-5, exactly-204, `invalid_request` |
| **Partial** | comment veracity (3/4), rpc-proxy timeout + normalisation (disconnect only) |
| **Blocking** | B-2 — pending the `evenfire-infra` change |
| **Withdrawn (my error)** | M-2, M-4; B-4 downgraded to a process nit |
| **Deferred with issues** | #503, #504 |
| **Product decision** | H-1, H-2 |

The three findings that mattered are genuinely fixed, and fixed the hard way — I could put each one in the red, which is the only thing that distinguishes a real fix from one that reads like one. Once the infra change lands, from my side the remaining items are comment hygiene and the authorize-url parity, neither of which needs to gate this merge.

PR #499, opened 2026-08-28
feat(connectors): proactive connectors panel with connect and disconnect (U2/U3/U4)

zach88:
## What
- desktop-app Connectors panel that lists, per agent, each mcp-server with its tri-state status and connect/disconnect actions.
- Proactive connect flow plus an OAuth-return branch that refreshes the panel without an in-flight turn.
- End-user disconnect with a per-flavor confirmation.
- Panel redesigned as an agents-style table with context drill-in.

## Why
Units U2/U3/U4 (desktop) of the proactive-connectors spec. Before this, authorization was reactive and only surfaced mid-turn. This lets a user see and manage the whole fleet up front.

## How
- The panel reads from the same grant store as the LLM invocation lane, so it reflects the same authorization state the invocation lane enforces.
- Status is folded per grant, so sibling rows for the same grant show one coherent state.
- The OAuth return without a turn refreshes the panel instead of being a no-op.

## Testing
- `useConnectorsController.test.tsx`, `connectorRows.test.ts`, `connectorPresentation.test.ts`, `handleMcpOauthCompletion.test.ts`, `useNavigationController.test.ts`, plus page tests (`McpServersPage`, `ContextDetailsPage`, `TeamsPage`).
- desktop-app UI suite green.

## Risks / Follow-ups
- Stacked on `feat/mcp-connectors-disconnect` (#498).
- 27 files, single service (desktop-app), just over the 25-file D1 target. It is one coherent panel feature with its tests; splitting would ship a half-built panel.
- Host-side revocation still relies on idle eviction (push revocation deferred, covered by the spec-13 poll-sweep landed separately). No CRD.


alfredolopez80:
## Review: `CHANGES_REQUESTED` — 2 blockers, 4 high, 17 medium, 12 low

Reviewed at `595e7389`, isolated against the stacked base (`git diff origin/feat/mcp-connectors-disconnect...HEAD`), not against `dev`. Every finding below was verified against the code; findings marked **[REPRODUCED]** have a failing probe behind them.

**First, the part that matters:** this is well-built. The architecture is sound and I verified your central claim end-to-end — `resolveConnectorsForAgents` and `resolveInvocableMcpServersForContexts` converge on the *same* `computeGrantPresence` / `buildMcpServerGrantKey` / `oauthGrantExists`, so the panel really does read the same grant store the invocation lane enforces. The difference is policy (classify vs. filter), not source. The pure folding logic is well tested (10 of 16 mutations killed, all in the renderer), there's no privilege escalation, no secret leakage, no injection, no XSS, and the ~329 added comments are unusually truthful — I checked the strong claims (DELETE idempotency with 204, middleware-before-upstream ordering, `{confirmed:false}` on cancel, identity teardown) against the code, including the server side.

The problem is not shape. It's **error semantics**: the one destructive operation in this panel can fail — or vanish entirely — without either the user or the test suite noticing.

---

## Blockers

### B-1 · Every connect/disconnect failure is invisible to the user

`ui/src/pages/McpServersPage.tsx:199` and `:202` · `ui/src/pages/ContextDetailsPage.tsx:656` and `:670` — **[REPRODUCED]**

All four action call sites end in `.catch(() => undefined)`, and `useConnectorsController` exposes no action-error state: its only `error` (`:131`) is the `listConnectors` query error, never a write.

**Scenario.** User clicks *Disconnect* on an `oauth-context` connector, reads a native dialog saying *"removes access for every agent in this context"*, and confirms. `deleteMcpOauthGrant` returns 403 `context_membership_denied` or 502. The promise rejects; the page discards it. Measured in jsdom: spinner stops, **zero** `.status-banner` in the DOM, pill still reads `Authorized`, button re-enables. **The final screen state is identical to a cancelled dialog.** The grant is still live and the user believes they revoked it.

Three things make this blocking rather than a UX nit:

1. **The repo actively defends against this in the same capability.** The pre-existing call sites of `connectMcpServer` do log the failure — `components/agents/InFlightAssistantPlaceholder.tsx:121-125` and `components/agents/ChatThread.tsx:499-504` — and there are **two regression tests** written to prevent the degradation, whose comments literally say the assertion *"goes red"* if someone reverts to swallowing (`chatThread.connectFailure.test.tsx:141`, `inFlightAssistantPlaceholder.test.tsx:162`). The new panel is strictly weaker than code those tests exist to protect.
2. **The reporting mechanism already exists and is established in two places.** `FilesPage.tsx` takes `pushToast` as a prop (`:124`) and reports every revoke failure (`:376/379`, `:403/406`). In the very domain this PR touches, `hooks/domain/approvalDecision.ts:215,260` toasts on all failure branches.
3. **Zero error-path coverage.** Across all 9 test files in this PR: **0** occurrences of `mockRejected`, `rejects`, or `toThrow`.

**Root cause — and where the fix belongs.** `useConnectorsController()` takes **no arguments** (`:37`), unlike its sibling `useChannelNotificationPreferencesController({isAuthenticated, pushToast})` (`useAppController.ts:264-267`), and neither page imports `pushToast`. So `.catch(() => undefined)` isn't page-level carelessness — it's the only syntactic exit the hook's design leaves, since the alternative is an unhandled rejection.

The internal asymmetry confirms it: the hook already owns the action lifecycle (`pendingKey` at `:39` marks the start, the `finally` at `:92`/`:120` the end) but never records the *outcome*. And the render surface is already mounted — `McpServersPage.tsx:161` renders `<StatusBanner tone="error">`. The hook just never writes to it.

**Minimal fix:** add `const [actionError, setActionError] = useState<string|null>(null)` to the controller, populate it in a `catch` inside `authorize`/`disconnect`, return it; pages drop the four `.catch(() => undefined)` and render it in the banner they already have. Or `useConnectorsController({ pushToast })`, matching the siblings.

### B-2 · The destructive path can disappear entirely and the suite stays green

`src/appService.ts:4597, 4630, 4673` · `src/rpcProxyClient.ts:449` · `src/ipc.ts:1158, 1166` — **[REPRODUCED]** via mutation testing

This PR adds **253 lines of main process**, including a team-scoped destructive revocation with a native dialog, and **no test** under `desktop-app/src/__tests__/`. That directory already holds **55 suites**, and the exact template for this pattern — `appService.sandboxUiOauthConsent.test.ts` — exists and covers cancel/confirm/retry for the sandbox-ui sibling.

**6 of 16 mutations survived:**

| # | File | Mutation | Failures | Verdict |
|---|---|---|---|---|
| **M15** | `appService.ts:4648` | **`deleteMcpOauthGrant` never called — revoke is a total no-op** | **0** | **SURVIVED** |
| **M13** | `appService.ts:4646` | **destructive revoke ignores the user's CANCEL** | **0** | **SURVIVED** |
| **M14** | `appService.ts:4566` | **`connect` skips the shared-consent dialog entirely** | **0** | **SURVIVED** |
| **M11** | `useAppController.ts:650,758,835` | all three identity `connectorsData.reset()` removed | **0** | **SURVIVED** |
| **M12** | `useAppController.ts:385,507,521` | all three `connectorsData.refresh()` neutralised | **0** | **SURVIVED** |
| **M2 / M16** | `connectorRows.ts:48-51` | collision guards removed / `connectorRowKey` returns a constant | **0** | **SURVIVED** |

With M15 applied — the grant DELETE gone, disconnect doing nothing at all — the suite reports `Test Files 208 passed · Tests 2039 passed · EXIT=0`.

**M11 compounds it:** those three `reset()` calls are the only thing preventing one team's OAuth grants from leaking into the next, since `desktopQueryKeys.connectors` is identity-unscoped. Nothing tests them.

**Why this blocks together with B-1:** a broken disconnect is invisible to the tests *and* to the user. There is no detector left anywhere in the chain.

**Fix:** mirror `appService.sandboxUiOauthConsent.test.ts` for `disconnectMcpServer` — cancel → `{confirmed:false}` **and** `deleteMcpOauthGrant` not called; confirm → called with `(token, server, ctx)`; both `shared` copy variants; retry after 401. These must kill M13, M14 and M15.

---

## High

**H-1 · Two consecutive disconnects: the second revoked connector still shows "Authorized"** — `useConnectorsController.ts:57-67` + `:115-117` — **[REPRODUCED]**

`staleTime: 0` decides whether to *start* a fetch when the query is idle; it does **not** force a fresh one when a fetch is already in flight. `fetchQuery` dedupes, so the post-revoke refresh attaches to a fetch issued *before* the revoke and adopts its stale response. Measured against the installed `@tanstack/react-query@5.100.10`:

```
E2 DELETEs issued = 2
E2 total listConnectors calls = 2      (1 initial load + 1 refresh from row 1)
E2 revoked connector still listed? = true    pendingKey = null   error = null
```

Nothing has to fail — just two clicks, which is the panel's most natural flow. **Note:** `cancelRefetch: true` does **not** fix this; it was tested against the installed version and still resolves the stale value from a single invocation. Please don't close this with that one-liner.

**H-2 · After a failed initial load the panel is permanently broken** — `useConnectorsController.ts:47-51`, `McpServersPage.tsx:161-171` — **[REPRODUCED]**

The query is `enabled:false` and nothing triggers it on mount (no `useEffect`, no refresh button). `connectorsData.refresh()` only fires from post-auth bootstrap (`useAppController.ts:507,521`) and the OAuth deep-link return (`:385`). If `listConnectors` fails during bootstrap, mounting the panel emits no fetch, and the page renders the error banner **and** the `EmptyState` *"No connectors — No agents with connectors are available for your account yet."* simultaneously — two contradictory messages, no retry control. Only logout, team-switch or completing an OAuth escapes it. `refresh()` also swallows its own rejection, so `refreshAuthenticatedData`'s `Promise.all` resolves successfully while the panel is broken.

**H-3 · Layer inversion: `lib/` imports from `hooks/domain/`** — `ui/src/lib/connectorPresentation.ts:2`

It's the **only** one of 36 files in `ui/src/lib/` that imports from `hooks/` — the convention runs unanimously the other way (`useConnectorsController.ts:4` imports `connectorRowKey` *from* `lib/`, which is right). Works today via tree-shaking, but any test importing `connectorPresentation` now drags `@tanstack/react-query` and `react` into the graph. ~6 lines to fix: move `isSharedConnector`/`isActionableConnector` into `lib/connectorRows.ts` and re-export.

**H-4 · Double-authorization window** — `useConnectorsController.ts:84-93` — **[REPRODUCED]**

The `finally` releases the lock when the connect IPC resolves (i.e. when the browser opens), not when the deep-link returns. The grant only appears on return, so in that gap the row is un-busy and still `Requires setup`: a second click starts a **second OAuth flow**. Reproduced: `connectMcpServer` called twice on the identical row with no state change in between. On an `oauth-context` connector that's two team-wide consent prompts for one intended click.

---

## Medium (17)

**Correctness / data model**
- **`STATUS_RANK[unknown]` is `undefined`, and the defensive merge fails in exactly the direction it was written to prevent** — `connectorRows.ts:57-61,88`. Both `2 > undefined` and `undefined > 2` are `false`, so once an unknown status occupies the group slot **nothing can ever replace it**, including `authorized`. The comment at `:53-56` says the merge exists so "an existing grant is never hidden behind a stale row" — it does the opposite. **[REPRODUCED]**
- **`statusPresentation`'s `default:` labels any unknown status "No OAuth"** — `connectorPresentation.ts:22-23`. Combined with `isActionableConnector`, the row loses its button: a connector needing action renders as "managed by Secret, nothing to do here". Suggest an explicit `case 'no_oauth':` plus an exhaustive `default:` with `assertNever`. **[REPRODUCED]**
- **`oauth-user` grants produce N rows with N Disconnect buttons for one grant** — `connectorRows.ts:75`. The file's own docstring (`:7-11`) states `oauth-user` grants are keyed by `(server, userId)` and then dedupes by `(server, context)` anyway. Suggest keying on `contextRef` only when `isSharedConnector(connector)`. **[REPRODUCED]**
- **An agent with `contextRef: null` loses status and actions in the context tab** — `ContextDetailsPage.tsx:327`. The exact-equality filter drops context-less agents — precisely the case where context is irrelevant to the grant. The same grant *is* actionable from the top-level panel. Your own fixture has this shape (`McpServersPage.test.tsx:72-84`, `agent-orphan`). **[REPRODUCED]**
- **`pendingKey` is a single slot** — `useConnectorsController.ts:39`. Two *different* rows: the second action's `finally` clears the first row's busy state while its IPC is still in flight. A `Set<string>` fixes it. (Double-click on the *same* row is correctly protected — I checked, and discarded that as a non-issue.) **[REPRODUCED]**

**Security / consent**
- **The "affects the whole team" consent gate is decided by the renderer** — `ipc.ts:1152`, `appService.ts:4564`. In **connect** the dialog is conditional on a renderer-supplied flag (omit the 4th argument and it's suppressed entirely); in **disconnect** it's unconditional and `shared` only changes copy. Medium, not blocker: the dialog is a consent affordance, not the security boundary — per-flavour authz is server-side and independent (`rpc-proxy/src/routes/mcpOauth.ts:126-131`), so a non-member still gets 403. The user loses the warning, not control. Worth deriving `shared` main-side, or just always confirming on connect.
- **`isSharedConnector` is fail-open when flavour tags are missing** — `useConnectorsController.ts:18-22`. Both fields are optional in the type; absent → `false` → "personal" copy while revoking the whole Context. Not reachable today (`mcpInvocable.ts:496` always emits both, `:240` excludes fail-closed), but the default chosen for unknown input is the permissive one.
- **The revocation-latency copy is false for `oauth-context`** — `appService.ts:4682`. `oauth-user` has a 15-min idle backstop (`mcp-host/src/main.ts:825`); `oauth-context` has **none** — SHARED partitions are exempt from idle eviction (`main.ts:1282`) — and the sweep is fail-open by design (`main.ts:1249-1255`). If the sweep is down, "a few minutes" is unbounded for the shared flavour. Worth noting: your risk note *understates* the normal case (the hot sweep cuts in ~30 s, not by idle eviction) and *omits* the degraded one, which is where the actual risk lives.
- **`defaultId: 1` makes the destructive button the default: Enter revokes** — `appService.ts:4692`. It's copied from `requestSandboxUiOauthAuthorize` (`:4521`), but there it's an *additive* `type:'question'`. Suggest `defaultId: action === 'disconnect' ? 0 : 1`.
- **`assertSafeRouteSegment` missing on the only unguarded route segment in the file** — `rpcProxyClient.ts:453`. The helper is already imported (`:3`) and applied to 7 other segments (`:944`, `:971-973`, `:1024-1026`); `mcpServerName` is the only one without it. `encodeURIComponent` doesn't encode `..`, and fetch's URL normalisation rewrites the path. Bounded today (404), so Medium. The pre-existing sibling `requestMcpOauthAuthorizeUrl` (`:675`, blame `bdc099c3f`) has the same gap — worth fixing both together. **[REPRODUCED]**
- **IPC payload crosses as explicit `any`** — `ipc.ts:1144,1152,1168,1173`. `ipcMain.handle` declares `...args: any[]`, so `strict: true` can't see it. `sanitizeString` turns `{}` into `"[object Object]"` and passes the guard — the user would see `Disconnect "[object Object]" for the whole team?` and only then get a 400, spending the one human barrier on a request that could never work. Worse: `options.confirmShared` is property access on `any`, so a typo (`confirmshared`) compiles and silently disables the team dialog.

**Contract / presentation**
- **Agent display-name regression** — `McpServersPage.tsx:88`. `agentDisplayByName` went from **2 occurrences on the base to 0** at HEAD, while `ContextDetailsPage` still has 3. The removed comment read *"Visible agent name (spec.host) read directly from the producer-total map — no fallback (Decision #6)"*, and that decision has its own test (`src/__tests__/appService.agentDisplay.test.ts`). The Connectors panel is now the only surface showing the raw host identifier. Root cause is upstream: rpc-proxy drops `displayName` in `sanitizeAgentConnectors` (`controlApiRestService.ts:140`).
- **`ContextDetailsPage` ignores `error` and `loading`** — `:50-55` destructures only `agents`/`pendingKey`/`authorize`/`disconnect`, while `McpServersPage:144` takes all six. A failed read renders `—`, identical to "this server doesn't use OAuth", with no error surface and no retry.
- **A second convention for a problem the repo already solved** — `McpServersPage.tsx:77-131` uses 6 `stopPropagation` calls and 0 uses of `isEventFromNestedInteractive`, which lives in `lib/clickableRowProps.ts:62` — the same file this PR imports `clickableRowProps` from — and is already used by `FilesPage.tsx`. Correct today, but opt-in per control (every future button must remember) vs. opt-out on the row.
- **The drill-in shows a different name than the row you clicked** — `ContextDetailsPage.tsx:608` renders `server.name` raw while `McpServersPage.tsx:55` renders `formatMcpServerDisplayName(...)`. Click "monday", land on a table showing `mcp-monday-remote`. **[REPRODUCED]**
- **Capabilities removed in the rewrite aren't named in the PR description** — server URL (not recoverable: `RpcConnector` has no `url` field), the Source column, the "Unassigned context" label, and the diagnostic empty message that explained *why* it was empty. The objection isn't that they went — Status and Actions are worth more, and the table was tight — it's that the summary says "redesigned as an agents-style table" without listing them.
- **The error banner lost `role="alert"`** — `McpServersPage.tsx:161`. The base used `<div className="composer-error" role="alert">`; `StatusBanner` emits no ARIA, so screen readers no longer announce the error.

---

## Low (12)

`result?.confirmed` reads nullish as "cancelled" — unreachable today (the only two returns are `{confirmed:false/true}`; everything else throws) but a latent fail-silent, and this same PR uses the *opposite* convention in `approvalDecision.ts:35` · `usedByAgents[0] ?? ''` propagates an empty agent name as `hostRef` · `requestJson<RpcConnectorsResult>` is an unvalidated wire cast that breaks the file's own idiom (`wireObject`/`wireString` at `:39-74`) — this is the door M-1 and M-2 come through · `userId` is declared, produced and never consumed, and costs an extra `getMe` on the zero-agent branch · `loading` is `true` before the first fetch · proactive completions can auto-approve suspensions created up to 45 s later · `connectorsData.refresh` is captured in the closure despite the `:369` comment saying "read live" · `shared = true` default in `confirmSharedConnectorAction` · the destructive dialog shows before the session check · `url`/`mappingSource` still computed for columns no longer rendered · the blast-radius warning is only a `title` on a non-focusable span · 3 orphaned CSS classes.

---

## One structural note on the tests

`connectorRows.test.ts:133-170` computes its invariants by calling **the function under test**. I set `connectorRowKey` to return a literal constant — the most severe possible break — and the property test, which runs 100 random payloads and is described as *"the test is the spec"*, **stayed green**. Only two by-example tests catch it.

| Invariant | Real detection power |
|---|---|
| (1) total ordering | **Genuine** — reads `connector.name`/`contextRef` directly; killed M9 |
| (2) no duplicate keys | **Tautology** — rows come from a `Map` whose keys *are* `connectorRowKey(...)`, and `row.key` recomputes the same key |
| (3) no agent disappears | **Blind to over-grouping** — looks up by the same mutated key |
| (4) status = group max | **Blind**, same reason |

You clearly know the principle — the comment at `:92` explains that `STATUS_RANK` is re-declared locally rather than imported *"so a change to the module's ranking is caught, not mirrored"*. The same defence just needs to reach `connectorRowKey`, which is the symbol holding the whole grouping up.

Related mechanical cause for the error-path gaps: every double is `vi.fn(async () => value)`, which always resolves — so the error branches aren't merely untested, they're **unreachable by construction** from this suite.

---

## Gates run

Node 24, `npm ci` without `--ignore-scripts`, on an isolated tree:

| Gate | Result |
|---|---|
| Full suite | **207/207 files · 2036/2036 tests · 0 failed · 0 skipped · EXIT=0** |
| The 8 files cited as evidence | 47 tests, 8/8 present, 0 failures |
| `tsc -p tsconfig.json --noEmit` (the CI gate, `ci-public.yml:127`) | **EXIT=0** |
| `tsc -p ui/tsconfig.json --noEmit` (not a CI gate) | EXIT=2, 14 errors — **0 in files this PR touches** |
| `npm run build` · `verify:electron` | EXIT=0 |
| ESLint | not configured in this project |

Your *"desktop-app UI suite green"* claim is **literally confirmed** — with the caveat that it omits the 253 untested main-process lines, and that `TeamsPage.test.tsx` is cited as evidence while being a +1-line fixture change.

---

## Suggested order

**Blocking:** B-1 (action error channel + drop the four `.catch`), B-2 (main-process tests that kill M13/M14/M15).

**Follow-up, your call on what becomes an issue:** H-1 through H-4; then M-1/M-2 (exhaustive switch + unknown-status-safe rank), M-3/M-4 (flavour-conditional fold key), M-7 (fail-closed default), M-9/M-10 (`defaultId: 0`, `assertSafeRouteSegment`), M-12 (restore `agentDisplayByName`), and the property-test invariants.

**Cheap enough to fold into this PR:** the revocation-latency copy (M-8), the `queryKeys.ts:30-31` comment (it says "Invalidated" but the key is never invalidated — the mechanism is imperative refetch, exactly as your neighbouring comments explain), and naming the removed capabilities in the description.

**Worth a thought, not a blocker:** the navigation cluster (`NavigationContext.tsx`, `useNavigationController.ts` + test, `App.tsx`, tab hunks) doesn't depend on the read-model — none of it imports `RpcConnector` or `useConnectorsController`, and `ContextTab` already contained `'mcp-servers'` before this PR. Extracting it as a prior PR would leave this one at ~21 files, under the D1 target, without shipping a half-built panel. Your argument against splitting holds for the other three clusters — the main/IPC bridge in particular would land as unconsumed, untested code.

desktop-app/ui/src/pages/McpServersPage.tsx
```diff
@@ -1,168 +1,211 @@
...
-                      </span>
-                    </td>
-                  </tr>
+                {rows.map(row => (
+                  <ConnectorRowView
+                    key={row.key}
+                    row={row}
+                    busy={pendingKey === row.key}
+                    onAuthorize={input => {
+                      authorize(input).catch(() => undefined)
```
alfredolopez80: **Blocker B-1** — this swallows every `authorize` failure with no user-visible signal.

`useConnectorsController` exposes no action-error state: its only `error` (`useConnectorsController.ts:131`) is the `listConnectors` query error, never a write. So when this rejects, the spinner stops, the row is unchanged, and **nothing** appears — no banner, no toast, no log.

Two things make this blocking rather than a nit:

1. The pre-existing call sites of the *same* capability do log it — `components/agents/InFlightAssistantPlaceholder.tsx:121-125` and `components/agents/ChatThread.tsx:499-504` — and two regression tests exist specifically to keep them that way (`chatThread.connectFailure.test.tsx:141`, `inFlightAssistantPlaceholder.test.tsx:162`, whose comments say the assertion "goes red" if someone reverts to swallowing). This panel is strictly weaker than code the repo actively defends.
2. Across all 9 test files in this PR there are **0** occurrences of `mockRejected` / `rejects` / `toThrow`. Nothing exercises this path.

**The fix belongs in the hook, not here.** `useConnectorsController()` takes no arguments (`:37`), unlike its sibling `useChannelNotificationPreferencesController({isAuthenticated, pushToast})` (`useAppController.ts:264-267`), and neither page imports `pushToast` — so `.catch(() => undefined)` is the only syntactic exit this design leaves. Suggest returning an `actionError` from the controller (populated in a `catch` inside `authorize`/`disconnect`) and rendering it in the `<StatusBanner>` that already exists at `:161`.

desktop-app/ui/src/pages/McpServersPage.tsx
```diff
@@ -1,168 +1,211 @@
...
+                    }}
+                    onDisconnect={input => {
+                      disconnect(input).catch(() => undefined)
```
alfredolopez80: **Blocker B-1** (same defect, destructive path — see the note on the `authorize` line above).

This one is the costly case: the user has just confirmed a native dialog saying *"removes access for every agent in this context"*. If `deleteMcpOauthGrant` returns 403 `context_membership_denied` or 502, the rejection is discarded here and **the final screen state is identical to a cancelled dialog** — pill still `Authorized`, button re-enabled, no error anywhere. The grant is still live and the user believes it is gone.

desktop-app/ui/src/pages/ContextDetailsPage.tsx
```diff
@@ -541,70 +566,122 @@ export function ContextDetailsPage() {
...
+                        <td className="da-table__cell da-table__cell--right">
+                          {actionable && actionInput ? (
+                            connector?.status === 'authorized' ? (
+                              <Button
+                                color="danger"
+                                disabled={busy}
+                                loading={busy}
+                                onClick={event => {
+                                  event.stopPropagation()
+                                  disconnectConnector(actionInput).catch(() => undefined)
```
alfredolopez80: **Blocker B-1** (same defect as `McpServersPage.tsx:199,202`).

Worse on this surface: `ContextDetailsPage` destructures only `agents`, `pendingKey`, `authorize`, `disconnect` from the controller (`:50-55`) — no `error`, no `loading` — so unlike the top-level panel there is **no error surface at all** here, not even for the read. A failed `listConnectors` renders `—`, indistinguishable from "this server doesn't use OAuth".

alfredolopez80 (commented):
Inline anchors for the 2 blockers, 4 high and 6 security-medium findings from [my review comment](https://github.com/evenfire-ai/evenfire/pull/499#issuecomment-5456978136). The remaining 11 medium and 12 low findings stay in that comment rather than as inline noise.

Every anchor below was verified against the code at `595e7389`; the mutation and reproduction results come from an isolated tree, not the review worktree.

desktop-app/ui/src/pages/ContextDetailsPage.tsx
```diff
@@ -541,70 +566,122 @@ export function ContextDetailsPage() {
...
+                                Disconnect
+                              </Button>
+                            ) : (
+                              <Button
+                                color="primary"
+                                disabled={busy}
+                                loading={busy}
+                                onClick={event => {
+                                  event.stopPropagation()
+                                  authorizeConnector(actionInput).catch(() => undefined)
```
alfredolopez80: **Blocker B-1** (same defect, `authorize` branch).

Note the precedent one file over: `FilesPage.tsx` takes `pushToast` as a prop (`:124`) and reports every revoke failure (`:376/379`, `:403/406`). And in this very domain, `hooks/domain/approvalDecision.ts:215,260` toasts on all failure branches. The mechanism exists and is established in two places.

desktop-app/src/appService.ts
```diff
@@ -4576,6 +4586,118 @@ export class AppService {
...
+  /**
+   * Disconnect (revoke) an mcp-server's OAuth grant from the proactive panel
+   * (spec 11 U4). Host-bound to the agent whose row triggered it (RPC tokens
+   * require ≥1 hostRef; `mcp:server:invoke`, same as connect). Disconnect is
+   * destructive, so it ALWAYS asks for explicit confirmation via a native
+   * dialog (D-4); for an `oauth-context` connector the copy names the team-wide
+   * blast radius. `userId` is derived by rpc-proxy from `auth.sub`, never sent.
+   * Resolves `{ confirmed:false }` when the user cancels (nothing was revoked).
+   */
+  async disconnectMcpServer(
```
alfredolopez80, line 4630: **Blocker B-2** — this method, `confirmSharedConnectorAction` (`:4673`) and `getConnectors` (`:4597`) add **253 lines of main process** including a team-scoped destructive revocation with a native dialog, and this PR adds **no test** under `desktop-app/src/__tests__/`.

That directory already holds **55 suites**, and the exact template for this pattern exists: `appService.sandboxUiOauthConsent.test.ts` covers cancel / confirm / retry-after-401 for the sandbox-ui sibling.

I proved the consequence rather than asserting it. Three mutations here left the full suite green (`Test Files 208 passed · Tests 2039 passed · EXIT=0`):

| Mutation | Effect | Verdict |
|---|---|---|
| `deleteMcpOauthGrant` never called (`:4648`) | **revoke becomes a total no-op** | SURVIVED |
| revoke ignores the user's CANCEL (`:4646`) | confirmation dialog becomes decorative | SURVIVED |
| `connect` skips the shared-consent dialog (`:4566`) | team-wide consent silently bypassed | SURVIVED |

Combined with B-1, a broken disconnect is invisible to the tests **and** to the user — there is no detector left in the chain.

**Suggested fix:** mirror `appService.sandboxUiOauthConsent.test.ts` — cancel → `{confirmed:false}` *and* `deleteMcpOauthGrant` not called; confirm → called with `(token, server, ctx)`; both `shared` copy variants; retry after 401. Those must kill all three mutations above.

Minor, same method: the dialog at `:4641` runs *before* the session check, so an unauthenticated user confirms a team-wide revocation and only then gets a session error. `getConnectors:4598` already fails fast.

desktop-app/ui/src/hooks/domain/useConnectorsController.ts
```diff
@@ -0,0 +1,151 @@
...
+        const result = await window.clerum.rpc.disconnectMcpServer(
+          connector.name,
+          agentName,
+          contextId,
+          { shared }
+        )
+        // Only a CONFIRMED revoke changes the grant store; a cancelled dialog
+        // (`confirmed:false`) is a main-side no-op, so it must not refetch (U4).
+        if (result?.confirmed) {
+          await refresh()
```
alfredolopez80, line 129: **High H-1** — this refresh can silently merge with a fetch issued *before* the revoke.

`staleTime: 0` (`:62`) decides whether to *start* a fetch when the query is idle; it does **not** force a fresh one when a fetch is already in flight. `fetchQuery` dedupes, so this `await refresh()` attaches to the in-flight request and adopts its stale response. The correct refetch never comes into existence.

Reproduced against the installed `@tanstack/react-query@5.100.10`, with the panel's most natural flow — clearing two connectors in a row:

```
E2 DELETEs issued = 2
E2 total listConnectors calls = 2      (1 initial load + 1 refresh from row 1)
E2 revoked connector still listed? = true    pendingKey = null   error = null
```

Two confirmed DELETEs, one refetch. The second revoked connector stays listed as `Authorized` with an active Disconnect button. Nothing has to fail — just two clicks. Compounded by H-2 (no retry path) and B-1 (`error === null`, no signal).

**Please don't close this with `cancelRefetch: true`** — I tested it against the installed version and it still resolves the stale value from a single `queryFn` invocation.

desktop-app/ui/src/hooks/domain/useConnectorsController.ts
```diff
@@ -0,0 +1,151 @@
...
+  // Mirrors the sibling data-controllers (useMcpServersDataController /
+  // useContextsDataController): the query is app-coordinated, never
+  // self-enabling. `useAppController` owns the initial load (post-auth
+  // bootstrap) and the identity teardown (`reset` on logout / team-switch),
+  // so a nav to the panel only READS cache and a team-switch cannot leak the
+  // previous identity's OAuth authorization state (the key is identity-unscoped).
+  const query = useQuery({
+    queryKey: desktopQueryKeys.connectors,
+    queryFn: () => window.clerum.rpc.listConnectors(),
+    enabled: false,
```
alfredolopez80, line 59: **High H-2** — with `enabled: false` and no mount-time trigger, a failed initial load is permanent.

Nothing fires this query when the panel mounts: `McpServersPage` has no `useEffect` and no refresh control (grep for `retry|refresh|reload|useEffect` → 0 hits). The only callers of `refresh()` are post-auth bootstrap (`useAppController.ts:507,521`) and the OAuth deep-link return (`:385`).

So if `listConnectors` fails during bootstrap, opening the panel emits **no fetch**, and `McpServersPage` renders the error banner (`:161`) *and* the `EmptyState` "No connectors — No agents with connectors are available for your account yet." (`:167-170`) at the same time — two contradictory messages, no way to retry. The only escape is logout, team-switch, or completing an OAuth flow.

Related: `refresh()` swallows its own rejection at `:64-66`, so `refreshAuthenticatedData`'s `Promise.all` resolves *successfully* while the panel is broken — login reports nothing. (To be clear, the `catch {}` itself is defensible and its comment is true: I verified against `query-core@5.100.10` that the error does surface on the `enabled:false` observer. The issue is only that the caller can't tell.)

desktop-app/ui/src/hooks/domain/useConnectorsController.ts
```diff
@@ -0,0 +1,151 @@
...
+      setPendingKey(key)
+      try {
+        // The grant becomes present only after the OAuth deep-link returns; U3's
+        // completion handler refreshes the panel then. Nothing to refresh here
+        // (an unconfirmed shared dialog is a no-op main-side).
+        await window.clerum.rpc.connectMcpServer(connector.name, agentName, contextId, {
+          confirmShared: shared,
+        })
+      } finally {
+        setPendingKey(current => (current === key ? null : current))
```
alfredolopez80, line 104: **High H-4** — the busy lock is released when the connect IPC resolves, not when the deep-link returns.

`connectMcpServer` resolves once the browser has been opened; the grant only exists after `clerum://oauth-completed` comes back (as your own comment at `:85-87` says). But this `finally` clears `pendingKey` at the earlier moment, so in the gap the row is un-busy and still reads `Requires setup` — a second click starts a **second OAuth flow**.

Reproduced: `connectMcpServer` called twice on the identical row, with no state change and no refetch in between. On an `oauth-context` connector that's two team-wide consent prompts for one intended click.

Separately, `pendingKey` being a single `string | null` slot (`:39`) means two *different* rows interfere: the second action's `finally` clears the first row's busy state while its IPC is still in flight, re-enabling its button. A `Set<string>` addresses both. (Double-click on the *same* row is correctly protected — `Button` applies `disabled={loading || props.disabled}` — I checked and discarded that as a non-issue.)

desktop-app/ui/src/lib/connectorPresentation.ts
```diff
@@ -0,0 +1,44 @@
+import type { RpcConnector } from '../../../src/types'
+import { isSharedConnector } from '../hooks/domain/useConnectorsController'
```
alfredolopez80: **High H-3** — layer inversion: this is the only file in `ui/src/lib/` that imports from `hooks/`.

I measured it: of the 36 files in `ui/src/lib/`, a grep for `from '../hooks/|from '@hooks|from '../../hooks` returns exactly one hit — this line. The convention runs unanimously the other way, and this PR follows it correctly elsewhere: `useConnectorsController.ts:4` imports `connectorRowKey` **from** `lib/`.

It works today only because the bundler tree-shakes the hook. But any test importing `connectorPresentation` now drags `@tanstack/react-query` and `react` into the graph, and if anyone later adds a module-level effect to `useConnectorsController.ts`, `lib/` silently stops being pure.

`isSharedConnector` and `isActionableConnector` are pure predicates over `Pick<RpcConnector, …>` — they touch neither React nor `window`. Moving them into `lib/connectorRows.ts` (where `RpcConnector` and the pure folding logic already live) and re-exporting from the controller is about 6 lines.

desktop-app/src/appService.ts
```diff
@@ -4548,13 +4549,22 @@ export class AppService {
...
+    // panel asks main to show an explicit confirm dialog first. The reactive
+    // U5 path never sets this, so its behavior is byte-identical.
+    options?: { confirmShared?: boolean }
   ): Promise<void> {
     const server = String(mcpServerName || '').trim()
     const targetHostRef = String(hostRef || '').trim()
     if (!server) throw new Error('mcpServerName is required')
     if (!targetHostRef) throw new Error('hostRef is required')
     const ctx = contextId ? String(contextId).trim() || undefined : undefined
+    if (options?.confirmShared) {
```
alfredolopez80, line 4564: **Medium M-6** — the "affects the whole team" consent gate is decided by the renderer.

This dialog is **conditional** on a flag the renderer sends. `ipc.ts:1152` accepts it verbatim (`options.confirmShared === true`) without deriving it from anything authoritative, and `useConnectorsController.ts:75,88` is what computes it. Omit the 4th argument to `connectMcpServer` and there is **no dialog at all** — not degraded copy, no dialog.

Contrast with your own `disconnect` path at `:4641`, which is unconditional and where `shared` only selects the copy. That one is right.

**Why I filed this as Medium and not a blocker**, since the framing matters: the dialog is a *consent affordance*, not the security boundary. Per-flavour authz is server-side and independent (`rpc-proxy/src/routes/mcpOauth.ts:126-131` — control-api resolves server→context and enforces Context membership), so a non-member still gets a 403 even if the renderer lies. **There is no privilege escalation here** — the user loses the warning, not control. And the no-attacker variant (missing `authKind`) isn't reachable today: `control-api/src/services/access/mcpInvocable.ts:496` always emits `authKind` and `grantScope` together, and `:240` excludes the server fail-closed on an unknown `grantScope`.

It is fragile, though: rpc-proxy's sanitiser (`controlApiRestService.ts:119-124`) silently drops out-of-enum values **without rejecting the connector**, so the property rests entirely on the producer never changing. Deriving `shared` main-side (you already have `contextId` and the session), or simply always confirming on connect, removes the dependency — one extra dialog costs less than a silent team-wide grant.

desktop-app/src/appService.ts
```diff
@@ -4576,6 +4586,118 @@ export class AppService {
...
+    const message = shared ? `${verb} "${server}" for the whole team?` : `${verb} "${server}"?`
+    const detail = shared
+      ? action === 'connect'
+        ? `This connector is shared by the team. Authorizing it connects every agent in this context on your behalf.${latencyNote}`
+        : `This connector is shared by the team. Disconnecting removes access for every agent in this context.${latencyNote}`
+      : `This removes your authorization for this connector.${latencyNote}`
+    const opts = {
+      type: (action === 'disconnect' ? 'warning' : 'question') as 'warning' | 'question',
+      buttons: ['Cancel', verb],
+      defaultId: 1,
```
alfredolopez80: **Medium M-9** — `defaultId: 1` makes the destructive button the default: **Enter revokes**.

With `buttons: ['Cancel', verb]` and `type: 'warning'`, a reflexive Enter (or one queued from the chat input) confirms a revocation whose own `detail` says *"removes access for every agent in this context"*, on a method your docstring calls "destructive" (`:4625`).

This looks inherited rather than chosen: the same `defaultId: 1` appears in `requestSandboxUiOauthAuthorize` (`:4521`) — but there it's an *additive* `type: 'question'`, where Enter grants something the user just asked for. Carrying that default over to a revocation inverts which side is safe.

Suggest `defaultId: action === 'disconnect' ? 0 : 1`, keeping `cancelId: 0`.

desktop-app/src/appService.ts
```diff
@@ -4576,6 +4586,118 @@ export class AppService {
...
+  private async confirmSharedConnectorAction(
+    action: 'connect' | 'disconnect',
+    server: string,
+    shared = true
+  ): Promise<boolean> {
+    const { dialog, BrowserWindow } = await import('electron')
+    const win = BrowserWindow.getFocusedWindow() ?? undefined
+    const verb = action === 'connect' ? 'Connect' : 'Disconnect'
+    const latencyNote =
+      action === 'disconnect' ? ' It can take a few minutes to cut active sessions.' : ''
```
alfredolopez80: **Medium M-8** — this latency promise doesn't hold for `oauth-context`, and that's the flavour where it matters most.

| Flavour | Hot sweep | Idle backstop |
|---|---|---|
| `oauth-user` | ~30 s (`mcp-host/src/main.ts:1299`) | **15 min** (`main.ts:825`) |
| `oauth-context` | ~30 s (`mcp/manager.ts:627-631`) | **none** — SHARED partitions are exempt (`main.ts:1282`) |

And the sweep is fail-open by design (`main.ts:1249-1255`: *"a control-api/gateway blip must not tear down live sessions"*). So if the sweep is down — a sustained blip, or a misconfigured service token producing a permanent 403 (`grantExistenceClient.ts:104`) — every 30 s tick enters the `catch` and conserves all partitions, and for the shared flavour **there is nothing underneath**. Not "a few minutes": unbounded, until the pod restarts or a Context change replaces the manager. Same for a partition under continuous in-flight tool calls (`manager.ts:641-643`).

Worth noting against your risk note ("host-side revocation still relies on idle eviction"): it actually **understates the normal case** — the spec-13 hot sweep cuts in ~30 s, not by idle eviction, so revocation is much faster than you claim — and **omits the degraded one**, which is where the real exposure is. The clock is `CLERUM_MCP_STATUS_HEARTBEAT_INTERVAL`, default 30000 ms (`mcp-host/src/config.ts:674-678`), with no override in `charts/` or `deploy/`.

The copy fix is cheap and worth folding into this PR even if the `oauth-context` backstop is left to spec-13.

desktop-app/src/rpcProxyClient.ts
```diff
@@ -422,6 +423,45 @@ export class RpcProxyClient {
...
+   * optional `contextId` is only a hint for the `oauth-context` flavor — control
+   * -api resolves the authoritative Context server-side. Idempotent: a missing
+   * grant answers 204. Throws `ApiError` (401/403-missing-scope → refresh retry).
+   */
+  async deleteMcpOauthGrant(
+    rpcAccessToken: string,
+    mcpServerName: string,
+    contextId?: string
+  ): Promise<void> {
+    await requestJson<unknown>(
```
alfredolopez80, line 453: **Medium M-10** — `mcpServerName` is the only route segment in this file without `assertSafeRouteSegment`.

The helper is already imported at `:3` and applied to seven other segments — `hostRef`, `agent`, `chatId` at `:944`, `:971-973`, `:1024-1026`. Its own docstring says *"Encoding a segment is not sufficient"*, and that's exactly the gap: `encodeURIComponent` does not encode `.` or `..`, and fetch's WHATWG-URL normalisation rewrites the path before sending. Measured:

```
".."    -> /api/v1/grant                    (the mcp-oauth segment disappears)
"."     -> /api/v1/mcp-oauth/grant
sanitizeString("..") === ".."               // ipc.ts only does String(input||'').trim()
```

So `disconnectMcpServer('..', agent)` sends a DELETE with a valid `mcp:server:invoke` token to a different path, and the server-side `isValidK8sName` check (`rpc-proxy/src/routes/mcpOauth.ts:148`) never runs because the request never reaches that route.

Bounded today — rpc-proxy mounts only two DELETE routes and neither is reachable with a single `..` from this depth, so the result is a 404 — hence Medium rather than Critical.

**Detection duty:** the pre-existing sibling `requestMcpOauthAuthorizeUrl` (`:675`) has the same gap — `git blame` gives `bdc099c3f`, Diego Zacarias, 2026-08-11. Lower risk (a minting POST), but worth closing both in the same commit.

Fix: `assertSafeRouteSegment('mcpServerName', mcpServerName, { maxLength: 200, allowColon: false })`.

desktop-app/src/ipc.ts
```diff
@@ -1139,14 +1139,42 @@ export function registerIpcHandlers(service: AppService): void {
...
+    'rpc:connectMcpServer',
+    async (event, { mcpServerName, hostRef, contextId, options }) => {
+      assertTrustedSender(event)
+      const server = sanitizeString(mcpServerName)
+      const targetHostRef = sanitizeString(hostRef)
+      const ctx = sanitizeString(contextId) || undefined
+      // `confirmShared` is a UX-only flag (the server-side authz per flavor is
+      // authoritative). The panel sets it for `oauth-context` connectors so main
+      // shows a team-wide confirm dialog; the reactive U5 path omits it.
+      const confirmShared = options && typeof options === 'object' && options.confirmShared === true
```
alfredolopez80, line 1152: **Medium M-11** — `options` here is explicit `any`, so this narrowing is decorative and a typo silently disables the team dialog.

`ipcMain.handle` is declared `listener: (event, ...args: any[]) => …` (`electron/electron.d.ts:8790`), which makes the destructured `{ mcpServerName, hostRef, contextId, options }` at `:1144` **explicitly** `any` — `strict: true` cannot flag it, which is why `tsc` passes clean. `typeof options === 'object'` narrows nothing while `options` is `any`.

The concrete consequence: `options.confirmShared` is property access on `any`. Write `confirmshared` and it compiles, `confirmShared` is `false`, and the team-wide consent dialog at `appService.ts:4564` **never shows** (see M-6). Same for `options.shared` at `:1173`, which would silently flip the disconnect copy to the personal variant.

Related, on the same payload: `sanitizeString` is `String(input || '').trim()`, so `{}` becomes `"[object Object]"` and passes the `if (!server)` guard at `appService.ts:4638`. A renderer bug passing the whole connector object instead of `connector.name` shows the user `Disconnect "[object Object]" for the whole team?` and only then fails with a 400 — the one human barrier spent on a request that could never work.

Suggest typing `options` as `unknown` and validating with a helper in the style of `parseHostMessageRequest` (`ipc.ts:149`), which this file already has.

desktop-app/ui/src/hooks/domain/useConnectorsController.ts
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * A connector is "shared by the team" when its grant is keyed by Context, not by
+ * user (spec §1.3). Authorizing/disconnecting it affects EVERYONE in the
+ * Context, so both the confirm-dialog copy and the panel warning key off this.
+ */
+export function isSharedConnector(
+  connector: Pick<RpcConnector, 'authKind' | 'grantScope'>
+): boolean {
+  return connector.grantScope === 'context' || connector.authKind === 'oauth-context'
```
alfredolopez80: **Medium M-7** — this is fail-open: a connector missing its flavour tags is treated as personal.

Both `grantScope` and `authKind` are **optional** in the type (`desktop-app/src/types.ts:497-498`). If neither arrives, this returns `false` = "personal", and that answer drives the native dialog copy: the user reads *"This removes your authorization for this connector"* (`appService.ts:4689`) while the DELETE may be revoking the whole Context's grant. It also suppresses `scopeCaption` entirely (`connectorPresentation.ts:36-43`) and drops `contextId` from the call (`:79`, `:101`), so control-api loses its cross-check against `spec.contextRef`.

**Not reachable today** — I verified that `control-api/src/services/access/mcpInvocable.ts:496-499` always emits both fields together for any connector with `status !== 'no_oauth'`. So this is hardening against contract drift, not a live bug.

But the default chosen for unknown input is the *permissive* one, and rpc-proxy's sanitiser drops out-of-enum values without rejecting the connector (`controlApiRestService.ts:119-124`), so a new `authKind` would reach here silently. The conservative direction is to treat a missing flavour as shared (larger blast-radius copy) or refuse the action.

alfredolopez80:
# Follow-up review — `23099f18`

Re-validated every finding from [my earlier review](https://github.com/evenfire-ai/evenfire/pull/499#issuecomment-5456978136) against the new head. **Both blockers are closed**, and I verified the fixes rather than trusting the commit messages — the B-2 closure in particular is confirmed by re-running the mutations, not by the presence of a test file.

**The PR is no longer blocked from my side.** 3 high and 4 medium findings remain open, plus one scope note below.

---

## Resolved (8) — verified

| Finding | Evidence |
|---|---|
| **B-1** silent action failures | All four `.catch(() => undefined)` are gone. The controller now owns an `actionError` (`:48`), populates it in `catch` (`:102`, `:136`), clears it with `setActionError(null)` at the start of each action, and — importantly — **does not rethrow**, which is what makes the `void promise` call sites safe. It renders in **both** pages: `McpServersPage:171-172` and `ContextDetailsPage:561`. That last part matters: an `actionError` that is set but never displayed would have been the same defect wearing a different shape. |
| **B-2** untested destructive path | `appService.mcpConnectorDisconnect.test.ts`, 10 tests, green. I re-ran the three mutations that previously survived: **M13** (revoke ignores CANCEL) → `EXIT=1, 1 failed`; **M14** (connect skips the shared dialog) → `EXIT=1, 1 failed`; **M15** (`deleteMcpOauthGrant` never called) → `EXIT=1, 3 failed`. **All three now KILLED**, with the source restored byte-identical afterwards (`sha 024a5ea1`). This is the part I could not have confirmed by reading. |
| **H-3** layer inversion | Zero imports from `hooks/` anywhere in `ui/src/lib/`. `isSharedConnector` and `isActionableConnector` now live in `lib/connectorRows.ts:12,21`. |
| **M-8** false revocation-latency copy | Now *"It is not immediate — active sessions end on the next revocation sweep"*, with a comment naming the fail-open sweep. Accurate for both flavours. |
| **M-9** `defaultId: 1` | `defaultId: action === 'disconnect' ? 0 : 1` (`:4700`), `cancelId: 0` kept. |
| **M-12** display-name regression | `agentDisplayByName` restored — 7 occurrences in `McpServersPage`. |

**Gates:** full desktop suite `208/208 files · 2050/2050 tests · EXIT=0` (up from 207/2036) · `tsc -p tsconfig.json --noEmit` **EXIT=0**.

---

## Still open (7) — code verified unchanged

Each of these was re-checked against the new head, not assumed from the earlier run.

| Finding | State at `23099f18` |
|---|---|
| **H-1** `fetchQuery` dedup | `refresh` is **byte-identical** — same `staleTime: 0`, no forced fresh fetch. Two consecutive disconnects still leave the second revoked connector showing `Authorized`. Worth being explicit: **the B-1 fix does not cover this**, because in this scenario nothing throws, so `actionError` stays `null`. |
| **H-2** no retry path | `enabled: false` unchanged; still no `useEffect` and no refresh control in `McpServersPage`. A failed initial load is still terminal. |
| **H-4** double-authorization window | The `finally` still releases `pendingKey` when the connect IPC resolves rather than when the deep-link returns. |
| **M-6** renderer decides the team-consent gate | `if (options?.confirmShared)` still conditional at `:4564`. |
| **M-7** `isSharedConnector` fail-open | The function moved to `lib/connectorRows.ts` as part of the H-3 fix, but the **logic is unchanged**. GitHub marked that thread outdated because the code moved — please don't read that as a fix. |
| **M-10** missing `assertSafeRouteSegment` | Still absent from `deleteMcpOauthGrant`; still the only unguarded route segment in `rpcProxyClient.ts`. |
| **M-11** IPC payload as `any` | `ipc.ts:1152,1173` unchanged. |

The remaining 10 medium and 12 low findings from the original comment were not re-checked individually; assume they stand unless a commit touched them.

---

## One scope note

This PR grew from **27 files in one service** to **40 across four**: `control-api`, `rpc-proxy` and `external-rest-api` came in via the merge of `feat/mcp-connectors-panel-api`. Those 13 files — including `mcpInvocable.ts`, `routes/rpc.ts` and `controlApiRestService.ts`, which are the source of the very read-model this panel consumes — **are outside the scope of my review**. Nothing in this comment says anything about them either way.

Supply chain, on the other hand, is clean, and I checked because it now applies where it previously didn't: the two `package.json` changes are **version bumps only** (`0.1.506 → 0.1.509`, `0.1.69 → 0.1.71`), `+2/-2` lines in each lockfile, **no new dependencies**.

---

## Recommended action

1. **Nothing blocks merge from my side.** Both blockers are closed with tests that demonstrably fail when the behaviour regresses.
2. **H-1 deserves attention before this ships** — it needs no failure to trigger, just two clicks, and it is now the only remaining defect that silently misreports authorization state. Reminder from the original comment: `cancelRefetch: true` does **not** fix it; I tested that against the installed version.
3. H-2, H-4 and the four mediums are reasonable follow-up issues — your call on which become issues and which get folded in.
4. The `control-api`/`rpc-proxy` surface is worth a review by someone, since it is unreviewed and it is where the tri-state classification is produced.

One small note on the B-1 fix itself: `McpServersPage:170-172` renders `error ? … : actionError ? …`, so a stale query error takes precedence over a fresh action error. With H-2 still open, both can coexist — a failed read followed by a failed write would show the older message. `actionError ?? error` would favour the more recent event.

Nice work on the turnaround, particularly writing the main-process tests against the named mutations rather than around them.

desktop-app/ui/src/pages/McpServersPage.tsx
alfredolopez80: **Resolved at `23099f18`.** The `.catch(() => undefined)` is gone and the controller now owns an `actionError` (`useConnectorsController.ts:48`), populated in a `catch` at `:102`/`:136` and cleared with `setActionError(null)` at the start of each action. It renders at `McpServersPage:171-172`.

I checked the part that would have made this a hollow fix: the hook **catches without rethrowing**, so the `void promise` call site cannot produce an unhandled rejection, and the error actually reaches a rendered surface rather than just being stored. Both hold.

desktop-app/ui/src/pages/McpServersPage.tsx
alfredolopez80: **Resolved at `23099f18`.** Same fix as the `authorize` line above — the failure now lands in `actionError` and renders in the existing `StatusBanner`. A rejected revoke is no longer indistinguishable from a cancelled dialog.

desktop-app/ui/src/pages/ContextDetailsPage.tsx
alfredolopez80: **Resolved at `23099f18`.** `ContextDetailsPage` now destructures `actionError: connectorActionError` (`:61`) and renders it at `:561`, so this surface has an error channel it previously lacked entirely.

desktop-app/ui/src/pages/ContextDetailsPage.tsx
alfredolopez80: **Resolved at `23099f18`.** Same fix; the `authorize` branch reports through the same banner.

desktop-app/src/appService.ts
alfredolopez80: **Resolved at `23099f18` — and verified by mutation, not by the presence of the file.**

`appService.mcpConnectorDisconnect.test.ts` adds 10 tests. I re-ran the three mutations that previously survived:

| Mutation | Result |
|---|---|
| M13 — revoke ignores the user's CANCEL | `EXIT=1`, 1 failed → **KILLED** |
| M14 — `connect` skips the shared-consent dialog | `EXIT=1`, 1 failed → **KILLED** |
| M15 — `deleteMcpOauthGrant` never called | `EXIT=1`, 3 failed → **KILLED** |

Source restored byte-identical afterwards (`sha 024a5ea1`). Writing the tests against the named mutations rather than around them is exactly the right move — that is what makes this verifiable rather than assertable.

The minor point from this thread also landed: `defaultId: action === 'disconnect' ? 0 : 1` at `:4700`.

desktop-app/ui/src/lib/connectorPresentation.ts
alfredolopez80: **Resolved at `23099f18`.** A grep for `from '../hooks/|from '@hooks|from '../../hooks` across `ui/src/lib/` now returns zero hits. `isSharedConnector` and `isActionableConnector` live in `lib/connectorRows.ts:12,21`, which is where the pure folding logic and `RpcConnector` already were.

desktop-app/src/appService.ts
alfredolopez80: **Resolved at `23099f18`.** `defaultId: action === 'disconnect' ? 0 : 1` (`:4700`), with `cancelId: 0` kept. Enter no longer confirms the destructive branch, and the additive connect dialog keeps its original default.

desktop-app/src/appService.ts
alfredolopez80: **Resolved at `23099f18`.** The copy is now *"It is not immediate — active sessions end on the next revocation sweep"*, with a comment naming the fail-open sweep as the reason not to promise a bound. That is accurate for both flavours, including the `oauth-context` case that has no idle backstop underneath.

desktop-app/ui/src/hooks/domain/useConnectorsController.ts
alfredolopez80: **Still open at `23099f18`.** `refresh` is byte-identical to the reviewed version — same `fetchQuery` with `staleTime: 0`, no forced fresh fetch — so the dedup behaviour is unchanged.

Worth stating explicitly because the B-1 fix might look like it covers this: it does not. In this scenario **nothing throws** — both DELETEs succeed — so `actionError` stays `null` and the second revoked connector still renders `Authorized` with an active Disconnect button, silently. It needs no failure at all, just two clicks.

Standing reminder: `cancelRefetch: true` does not close this; I tested it against the installed `@tanstack/react-query@5.100.10` and it still resolves the stale value from a single `queryFn` invocation.

desktop-app/ui/src/hooks/domain/useConnectorsController.ts
alfredolopez80: **Still open at `23099f18`.** `enabled: false` is unchanged, and `McpServersPage` still has no `useEffect` and no refresh control (grep for `retry|refresh|reload|useEffect` → 0 hits). A `listConnectors` failure during bootstrap still leaves the panel showing the error banner and the "No connectors" empty state simultaneously, with no way to retry short of logout, team-switch or an OAuth round-trip.

desktop-app/ui/src/hooks/domain/useConnectorsController.ts
alfredolopez80: **Still open at `23099f18`.** The `finally` still clears `pendingKey` when the connect IPC resolves — i.e. when the browser opens — rather than when the deep-link returns and the grant actually exists. The second-click window is unchanged, and on an `oauth-context` connector it is still two team-wide consent prompts for one intended click.

desktop-app/src/appService.ts
alfredolopez80: **Still open at `23099f18`.** `if (options?.confirmShared)` is unchanged at `:4564`, so the connect dialog remains conditional on a renderer-supplied flag while the disconnect dialog stays unconditional.

Restating the calibration so this isn't over-read: this is a consent-integrity issue, not an authz hole. Per-flavour authorization is server-side and independent, so nobody gains a capability they didn't have — they lose the warning.

desktop-app/src/rpcProxyClient.ts
alfredolopez80: **Still open at `23099f18`.** `deleteMcpOauthGrant` still relies on `encodeURIComponent` alone; `assertSafeRouteSegment` remains applied to the other seven route segments in this file and not to `mcpServerName`. The pre-existing sibling at `:675` (`requestMcpOauthAuthorizeUrl`, blame `bdc099c3f`) is also still unguarded.

desktop-app/src/ipc.ts
alfredolopez80: **Still open at `23099f18`.** `ipc.ts:1152` and `:1173` are unchanged, so `options` is still explicit `any` and the `typeof options === 'object'` narrowing still narrows nothing. A typo in `confirmShared` would still compile and silently disable the team-consent dialog.

desktop-app/ui/src/hooks/domain/useConnectorsController.ts
alfredolopez80: **Still open at `23099f18` — and this thread is misleading, which is why I'm replying.**

GitHub marked it outdated because the code *moved*: `isSharedConnector` is now in `lib/connectorRows.ts:12` as part of the H-3 layer-inversion fix. But the body is unchanged:

```ts
return connector.grantScope === 'context' || connector.authKind === 'oauth-context'
```

Both fields are still optional in the type, so a connector arriving without flavour tags is still treated as personal. An outdated thread here reads like a fix and isn't one.

zach88:
Thanks for the thorough re-review. Two follow-ups from your comment:

## 1. Banner precedence nit — fixed (`58bb4113`)

`McpServersPage` rendered `error ? <read> : actionError ? <write>`, so a stale read error shadowed a fresh write error. Switched to `actionError ?? error`, so the more recent event wins — exactly your suggestion.

Class-check: the only other connectors page that mounts an error banner is `ContextDetailsPage:560`, and it renders `connectorActionError` alone (no read error is surfaced there), so there's no precedence collision to fix — one site changed.

Repro test at `desktop-app/ui/src/pages/__tests__/McpServersPage.test.tsx:263`: a confirmed disconnect whose refresh read throws (sets the query error), followed by a rejected disconnect (sets `actionError`) — asserts the banner shows the write message and no longer the stale read one. It fails against `ec0915cb` (times out on the write assertion because the read message is shown) and passes on the fix. `tsc -p tsconfig.json --noEmit` EXIT=0; connectors test files green (31/31).

## 2. H-1 (`fetchQuery` dedup) — tracked as #518

Agree it deserves attention but it's a distinct defect from the B-1 surface (nothing throws, so `actionError` stays null), so I've split it into its own issue rather than fold it in: **#518**. It captures the root cause (`refresh`'s `fetchQuery` dedup at `staleTime: 0`), your note that `cancelRefetch: true` doesn't fix it, and acceptance criteria that assert the observable row state after two back-to-back disconnects.

H-2, H-4 and the mediums stay as follow-ups per the ledger. The unreviewed `control-api`/`rpc-proxy` tri-state surface is noted separately — that's a review-ownership question, not a change to this PR.

PR #496, opened 2026-08-28
fix(hcc): land jozer Role-skip review nits after #487

alfredolopez80:
## Summary
- Follow-up to jozer-rami's review on #487 ([review 5051768660](https://github.com/evenfire-ai/evenfire/pull/487#pullrequestreview-5051768660)). Lands the review nits in product + tests. No GitHub issue; this PR is the correction.
- Share one `canonicalizeValue` (drop undefined, default key sort) for Role, Service, Deployment, and NetworkPolicy apply gates. Leave the egressSignature walker in `networkPolicyReconciler.ts` alone: it **keeps** undefined keys (#473 / #299). ASCII Kubernetes keys make `localeCompare` vs default `sort()` equivalent, so the old NP walker can delegate.
- Role skip now compares **authored label keys as a subset**. Extra live labels (admission / ArgoCD instance / cost tags) no longer force a PUT that would strip them. A missing or changed authored key still fail-opens. Hardcoding only `MANAGED_BY_LABEL` + `HOST_LABEL` would skip forever if `rbacLabels` grew a third key.
- Split the KEEP `hostReconciler.test.ts` fixture: secretRef rewrite now feeds a live Role with the **old** `resourceNames` + labels/rules so the comparator is under test; the no-`rules` fail-open stays a separately named case.
- Clone `asApiserverRole` / `asApiserverPolicyRule` (no shared nested arrays). Collapse `updated*Logs` and the five `makeStubKc` Proxies. NOOP-ROLE-1 asserts authored vs live **label** key order; NOOP-ROLE-9 pins an extra admission label as a skip.

Does **not** add an HCC ownership check before Role replace, unswallow `console.error`, add conflict retry, or migrate `ensureHostRole` onto `replaceWithConflictRetry` (G5 / #460). Those stay out of this PR.

**Image-building:** HCC version bumped `0.1.261` → `0.1.262`. Merge rebuilds the HCC image and resets `/metrics`. Do not merge while a merge freeze is in effect.

## Review checklist (jozer 5051768660)
- [x] KEEP secretRef rewrite actually exercises the comparator (live Role carries old `secretRef` + labels/rules)
- [x] KEEP fail-open (no `rules`) is a separately named test
- [x] Fifth recursive key-sorter extracted as `canonicalizeValue` (egress walker left distinct)
- [x] Drop `canonicalizeLabelKeys`; one canonicalize of `{ labels, rules }`
- [x] Extra live labels do not PUT
[...]

PR #495, opened 2026-08-28
feat(hcc): clerum-dev NetworkPolicy orphan-census wrapper

alfredolopez80:
## Summary

- Adds `scripts/ops/hcc-netpol-orphan-census-clerum-dev.sh` with a **static** public kubectl context placeholder (`gke_your-gcp-project_us-central1-a_example-dev`) so k8s-context-guard-v2 can inspect a literal `--context=` token. The live GKE project/cluster name is **not** in this repo (leak-guards).
- Reuses `hcc-netpol-orphan-census.sh` (no duplicated classifier).
- Generic script: `deployments` not `deploy`; fail-loud if mapper/host/rpc-proxy namespaces are missing.
- Typed membership: type ∈ {context-allow, rpc-proxy-egress, external-egress} and `managed-by` is null **or** `host-context-controller` (#484 excludes only present-and-foreign). Typed + missing managed-by stays eligible (controller repairable arm).
- Foreign typed reserved-name + owner label is **not** listed/orphaned and prints `VERDICT=AMBIGUOUS_PRESENT` (exit 5), never CLEAN — the controller aborts that pass as `ambiguous`.
- Does **not** build an HCC image. The post-#487 `/metrics` ≥2-tick sample is **not** a merge gate here.

Closes nothing. Feeds #478 (do not close until the series) and #484.

## Test plan

- [x] `bash -n scripts/ops/hcc-netpol-orphan-census.sh`
- [x] `bash -n scripts/ops/hcc-netpol-orphan-census-clerum-dev.sh`
- [x] `bash scripts/ci/check-no-infra-identifiers.sh` is clean
- [x] Wrapper file contains only the sanctioned public placeholder (no interpolated live context)
- [x] Missing namespace exits non-zero (never `VERDICT=CLEAN`)
- [x] Hermetic guards G1–G9 and verdicts V1–V8 (`SAMPLE_GAP_SEC=0`, no cluster)

PR #501, opened 2026-08-28
feat(hcc): give class D a defaults-only tick independent of the fleet pass

### LINKED ISSUE #488: feat(hcc): give the NetworkPolicy resync tick a defaults-only path — class D is fixed-size work delivered by a fleet-size pass
> ## ⚠️ Corrections — 2026-08-29, after #501 landed
>
> Three statements in the text below are wrong or were superseded. Read these
> first; the original wording is kept underneath so the reasoning history stays
> intact.
>
> **1. "Class D is the only class that fails toward OPEN" is FALSE.**
> Class A fails open too. The controller says so itself: `k8sClient.ts:2883`
> warns that with the full pass disabled, "dropped watch events will not
> self-heal **stale NetworkPolicy allows**", and the sweep's mass-delete guard
> states that a trip "leaves those orphans in place until an operator raises the
> env" (`networkPolicyReconciler.ts:88-99`). A surviving allow IS a residual
> permission.
> The accurate claim is narrower and still carries this issue: **class D is the
> only class whose drift removes ENFORCEMENT ITSELF** — no policy selects the
> pod, so the whole namespace goes unenforced — where class A leaks one
> *specific* allow. Blast radius, not direction.
> (Distinct and still correct: a *false positive* of the sweep — deleting a live
> allow — fails closed. That is #478's claim, about the sweep's action, not
> about drift.)
>
> **2. Acceptance criterion "It cannot run concurrently with a full convergence
> pass" was deliberately INVERTED by #501, and the non-goal about #481's
> coordination with it.**
> This is a decision, not an oversight: sharing the full-pass guard would have
> dropped the tick roughly 60% of the time — the duty cycle of the full pass —
> which is the exact pathology this issue exists to fix. The remedy would have
> become the symptom. #501 pins the inversion with a test
> (`k8sClient.test.ts:11552`).
> **But the mechanism originally given for that inversion — disjoint write sets
> — is the WRONG AXIS.** The full-pass timer runs with `ensureDefaults: true`
> (`k8sClient.ts:2878`), which executes the SAME `ensureDefaultPolicies()`
> (`networkPolicyReconciler.ts:1300`). The two lanes are therefore *not*
> disjoint over the defaults; they can write the same 15 objects concurrently.
> Disjointness (`networkPolicyReconciler.ts:234`) covers the **orphan sweep**, a
> lane the defaults-only tick never enters — and that is what #501's
> `defaults-only write set vs typed sweep` test actually pins.
> What makes the concurrency safe is **idempotence plus
> `replaceWithConflictRetry`'s 409 retry** (`utils.ts:81-88`). **Nothing
> currently pins that property for this scenario** — see the open item below.
>
> **3. Acceptance criterion "a deleted `deny-all-<ns>` is restored ... and a test
> proves it" is NOT met.** #501 declares it out of scope verbatim ("Out of this
> PR: ... minikube mutation of `deny-all-<ns>`"). The healing path is
> implemented and unit-pinned — the tests prove the tick *calls*
> `ensureDefaultPolicies()`; none proves a deleted policy *returns*.
> **Do not close this issue on that criterion until it is met or explicitly
> renegotiated**, notwithstanding the `Closes #488` in #501.
>
> ### Remaining open under this issue
> - The minikube mutation test (criterion 3). Minikube because deleting a
>   `deny-all` opens a whole namespace for the duration — unacceptable in the
>   shared dev cluster, fine in a throwaway local one.
> - A test pinning that a full pass and a defaults-only tick executing
>   `ensureDefaultPolicies()` concurrently is safe, resting on idempotence and
>   the 409 retry rather than on disjointness.
> - Activation: the knob ships at `0` and is absent from `deploy/base` and every
>   public overlay by design. Enabling it in `clerum-dev` at 300s is
>   `keyper-labs/evenfire-infra#68` / PR #69.

---

The periodic NetworkPolicy resync tick (#478/#481) exists for exactly one thing that no other path does: re-applying the L0/L1 isolation baseline. That work is **fixed-size**. It is currently delivered by a **fleet-size** operation, which is why choosing the interval is hard.

## The asymmetry

`ensureDefaults` does **not** gate the orphan sweep — that runs on every pass. It gates a single call, `networkPolicyReconciler.ts:1305`:

```ts
if (options.ensureDefaults !== false) {
  await this.ensureDefaultPolicies()
}
```

```ts
async ensureDefaultPolicies(): Promise<void> {
  await this.ensureDefaultDeny()             // L0: deny-all-<ns> in every runtime namespace
  await this.ensureInfrastructurePolicies()  // L1
  await this.ensureAllowContextMapperApi()
  await this.deleteLegacyStaticPolicy(...)   // x2 legacy permissive
}
```

And `start()` describes exactly this work (`k8sClient.ts:2665-2670`):

> L0/L1 isolation and removal of legacy permissive policies are mandatory bootstrap barriers. They are a **fixed-size safety operation, independent of fleet size**, and must succeed before HCC can watch or reconcile runtimes.

Coverage today:

| class | covered by | depends on fleet size? |
|---|---|---|
| A — orphans | the 8 event-driven call sites + startup | yes |
| **D — L0/L1 baseline** | **`start()` and the timer tick. Nothing else.** | **no** |

The event paths leave the flag `false` on purpose (`k8sClient.ts:897-899`), and the field defaults to `false` (`:900`).

## Why this matters for the interval

The tick calls `runInitialNetworkPolicyConvergence({ ensureDefaults: true })` — a full convergence pass over the whole fleet (364 managed policies on clerum-dev today). So the interval must be larger than a full pass duration, or ticks get skipped by the in-flight guard from #481.

Worst known pass: **17.1 min** (log) / **21 min** (infra), both **pre-#471**. Against the 1500 s the code comment names, that is a **1.19x** margin, i.e. the controller would be inside a full pass most of the time. The first dev enable is therefore going out at **3600 s** — a period chosen entirely by fleet pass duration, for work that has nothing to do with the fleet.

## Proposal

Give the tick a cheap path that re-applies the defaults **without** entering a full fleet convergence.

Sketch, not a design:

- a `runNetworkPolicyDefaultsOnly()` that calls `ensureDefaultPolicies()` under the same coordination the full pass uses (so it cannot race a pass in flight), and
- a separate, much shorter interval for it — minutes rather than the hour the fleet pass forces.

The full pass keeps its own interval for class A. The two stop being coupled.

## What this buys

1. **Class D heals in minutes instead of an hour.** That is the class that fails toward the OPEN side: in Kubernetes a pod selected by no NetworkPolicy is unrestricted, so a deleted `deny-all-<ns>` opens the namespace. Every other drift class in this family fails closed.
2. **The interval stops being a negotiation.** Right now the number is picked from unmeasured pass durations; a fixed-size operation needs no such argument.
3. **It removes the reason the enable was deferred at all.** The stability work in #471/#479/#480 made the pod stable, which removed the accidental class-D cleaner (the restart). A cheap tick restores it without making the fleet pass run continuously.

## Non-goals

- Not a replacement for the periodic full pass (#478) — class A still needs it.
- Not a change to `ensureDefaultPolicies()` itself, which is correct as written.
- Must not bypass the coordination that #481 added; a defaults-only tick that races an in-flight pass would reintroduce exactly what the skip guard removed.

## Acceptance

- A defaults-only tick re-applies L0/L1 without triggering a fleet pass — pinned by a test that fails if it starts one.
- It cannot run concurrently with a full convergence pass.
- A deleted `deny-all-<ns>` is restored within the short interval, and a test proves it (mutation required — minikube only, never dev).

Related: #478, #481, #472.

alfredolopez80:
## Summary
- Implements [#488](https://github.com/evenfire-ai/evenfire/issues/488): class D (L0/L1 baseline) is fixed-size work that was only reachable through a fleet-size `#478` pass. A separate opt-in timer re-applies `ensureDefaultPolicies()` without entering `fullReconcile` / `runInitialNetworkPolicyConvergence`.
- Knob `CONTEXT_MAPPER_NETPOL_DEFAULTS_RESYNC_SEC` via `requireCanonicalNonNegativeEnvInt`, **default 0**. Not shipped in `deploy/base` or public overlays. Non-canonical values fail loud at load.
- Defaults-only ticks do **not** share `initialConvergenceRuns.has('NetworkPolicy')` and must **not** increment `{reason="pass-in-flight"}`. Write sets are disjoint from the typed sweep (`policy-type ∈ {context-allow, rpc-proxy-egress, external-egress}`). Overlapping defaults-only ticks serialize with `{reason="defaults-only-in-flight"}`. `stop()` clears the timer **and** the in-flight lock.
- Each executed tick records `clerum_hcc_netpol_defaults_only_ticks_total` and `clerum_hcc_netpol_defaults_only_tick_duration_seconds` with closed `result=success|error`.
- Event-driven / startup full passes still consume+reset `netPolConvergenceEnsureDefaults` and stay `ensureDefaults: false`. Full-pass timer and `CONTEXT_MAPPER_NETPOL_RESYNC_SEC` are unchanged (do not lower 3600→1500).

**Base:** `origin/dev` already includes **[#496](https://github.com/evenfire-ai/evenfire/pull/496) MERGED** `4862d9c2`. This PR keeps those contracts (no sixth canonicalize walker; egress walker still distinct).

**Image-building:** HCC `0.1.263` → `0.1.266`. Merge rebuilds the HCC image and resets `/metrics` (post-#487 and post-#496). Do not merge while a merge freeze is in effect.

Closes #488

@claude please review

## Test plan
- [x] Default `0`: no defaults-only timer; advancing time does not re-apply L0/L1
- [x] Interval `> 0`: tick calls `ensureDefaultPolicies`, not `fullReconcile`
- [x] Full NetPol pass in flight: defaults-only **runs** and does **not** increment `{reason="pass-in-flight"}` (this pin failed on current `dev` before the product change)
- [x] Two overlapping defaults-only ticks: second serializes with `defaults-only-in-flight`
- [x] Event-driven full pass after a defaults-only tick still `ensureDefaults: false`
- [x] `stop()` clears the defaults-only timer and drops `netPolDefaultsOnlyRun`
- [x] Config fail-loud on non-canonical `CONTEXT_MAPPER_NETPOL_DEFAULTS_RESYNC_SEC`
- [x] Existing `#478` periodic-resync suite still green
- [x] Reconciler-level: `{name, policy-type}` written by `ensureDefaultPolicies` is disjoint from what the typed sweep lists/deletes
- [x] Tick success/error counters increment on the executed tick (not on start() bootstrap)
- [x] `npx tsc --noEmit -p tsconfig.build.json` in `host-context-controller`

Out of this PR: live minutes-scale interval; revisiting 3600 for class A; minikube mutation of `deny-all-<ns>`; changing the full-pass env; public overlays; closing #478.


alfredolopez80:
@claude please perform a **closing re-review and audit** of this PR on HEAD `386f5f15d8f74052d652c75512a004c2fa10eba9` (base `dev`).

This is **only** #501 / #488 (NetworkPolicy defaults-only / class D tick). Do **not** expand the review to #495, #490, #461, Role skip, `k8sClient.ts` apply-compare at the #490 site, or a sixth canonicalize walker. Do **not** merge from this comment.

Review the **full delta versus `dev`**, not only the latest commit. Commits: `ccaf6657` (feat, `Closes #488`) + `386f5f15` (T-A / T-B tests). HCC version **0.1.265**.

Local Grok review of this HEAD found **zero** PR-introduced defects. Re-prove that. Do not assume it.

## Required output format

Do **not** give only a global verdict. For **every** item below, reply with one line:

`ID — PASS | FAIL | STILL-OPEN — <one-sentence evidence from HEAD>`

If any item is FAIL or STILL-OPEN, say whether it is a merge blocker. End with a closing verdict: ship / do-not-ship, and list remaining blockers if any.

---

## Product constraints (must still hold)

- **P1** Knob `CONTEXT_MAPPER_NETPOL_DEFAULTS_RESYNC_SEC` via `requireCanonicalNonNegativeEnvInt`, **default 0**. Interval 0 → no defaults-only timer.
- **P2** Knob is **not** in `deploy/base` or any overlay. Enable stays an ops decision.
- **P3** `runNetworkPolicyDefaultsOnly()` → `ensureDefaultPolicies()` only. Never `fullReconcile`, never `runInitialNetworkPolicyConvergence`.
- **P4** Cheap tick must **not** consult `initialConvergenceRuns.has('NetworkPolicy')`. No shared skip with the fleet pass. No “full pass waits for defaults-only.”
- **P5** Overlap of two defaults-only ticks serializes with `{reason="defaults-only-in-flight"}`. Do not increment `pass-in-flight` for the cheap tick.
- **P6** A full pass in flight must still allow the cheap tick to run (disjoint write sets: L0/L1 vs typed sweep).
- **P7** Event-driven / startup passes stay `ensureDefaults: false`. Consume+reset of `netPolConvergenceEnsureDefaults` stays atomic.
- **P8** `stop()` clears `netPolDefaultsResyncTimer`. Do not lower the full-pass env 3600→1500. HARD REQ 2 stays on the full pass only.
- **P9** Do not add minikube mutation of `deny-all-<ns>`. Do not add a sixth `#496` canonicalize walker.

---

## Findings / hunts to re-verify on this HEAD

- **[D1]** Default 0: no defaults-only timer scheduled.
- **[D2]** Interval > 0: only `ensureDefaultPolicies`; enabled log; no fleet pass.
- **[D3]** Full pass in flight: cheap tick still runs; `pass-in-flight` unchanged.
- **[D4]** Overlap: serialize `defaults-only-in-flight`.
- **[T-A]** Throw releases the defaults-only lock; next tick re-applies L0/L1.
- **[T-B]** Held defaults-only does not starve the full-pass timer (dual-timer independence).
- **[D5]** Event-driven pass stays `ensureDefaults: false`.
- **[D6]** `stop()` clears `netPolDefaultsResyncTimer`.
- **[D7]** Knob absent from `deploy/` and overlays.
- **[D8]** Env fail-loud (`requireCanonicalNonNegativeEnvInt`); JSDoc documents disjoint write sets.
- **[D9]** Metric `reason` cardinality closed (no unbounded label).
- **[D10]** No shared `initialConvergenceRuns` skip; no “full pass waits for defaults-only.”
- **[D11]** No sixth canonicalize walker; no `k8sClient.ts` apply-compare change at the #490 site.
- **[D12]** Image bump is 0.1.265 only as required for this HCC change; `/metrics` reset after merge is expected (post-#487 / #496).

Hunt any remaining defect of any kind on the full delta vs `dev` (timers, locks, throw liveness, ensureDefaults contamination, deploy leakage, tests). Report new findings with severity. Closing verdict: ship / do-not-ship.

zach88 (approved):

zach88:
Took a deep pass through this one. Tight, well-scoped change — the single-flight lock, timer lifecycle, and metric labelling all hold up, and I independently re-proved the central "disjoint write sets" claim against the real code rather than taking it on faith. The only real gap is that this new safety invariant lives in JSDoc + prose but not in an executable test. Nothing here blocks a merge.

**🟡 Approve with nits**
- Optional / non-blocking: **4** (1 Medium, 3 Low)

## Optional (non-blocking)

The PR can merge without these — they're follow-up or author's-call.

1. **[R1-M1] The central safety invariant has no executable test** — `host-context-controller/src/k8sClient.test.ts:175-430` / `networkPolicyReconciler.ts`. Every new `#488` test mocks *both* `ensureDefaultPolicies` and `netPolFullReconcile`, so the write-set disjunction and concurrent-full-pass safety — the invariant this PR newly *depends* on (before it, `ensureDefaultPolicies` only ran serialized inside the single-flight pass) — are asserted nowhere. A future change that made `ensureDefaultPolicies` write a `context-allow`-typed policy, or made the sweep's safety inventory adopt a baseline policy-type, wouldn't turn any test red, even though it'd let a concurrent full pass and a defaults-only tick delete each other's writes. Suggest a reconciler-level test (in `networkPolicyReconciler.test.ts`, against a fake k8s API rather than a mock of the two methods) asserting the `{name, policy-type}` set written by `ensureDefaultPolicies` is disjoint from what the typed sweep lists/deletes — derived from the real functions, not hand-built.

2. **[R1-L1] Tests assert call counts, not observable state** — `k8sClient.test.ts` (`#488` block). The new tests assert `ensureDefaultPolicies` call counts (intermediate effect) rather than any K8s-observable result. Defensible at this seam — the tick's whole contract is "invoke `ensureDefaultPolicies`" — but combined with the mock it means the L0/L1 policies actually being present is never checked here. Reinforces R1-M1.

3. **[R1-L2] No metric on the defaults-only tick** — `k8sClient.ts:3094-3113`. A tick has no success/failure/duration counter; a failing tick is only visible via `console.error('[K8s] NetworkPolicy defaults-only tick failed', …)`. An enabled timer that fails persistently would be invisible to monitoring. Not promised by the body — observability gap only. (The full pass already has its metric family.)

4. **[R1-L3] `stop()` doesn't reset `netPolDefaultsOnlyRun`** — `k8sClient.ts:4646`. `stop()` clears the timer but leaves the in-flight lock set. On a `start()`-after-`stop()` of the *same* instance with a run still in flight, the lingering lock would make the first post-restart tick skip+join the stale run. Mirrors the existing pattern (`stop()` also doesn't clear `initialConvergenceRuns`), and restart-on-same-instance isn't an exercised path, so Low.

## What checked out clean

- **Disjoint write sets — the core claim — is TRUE.** `ensureDefaultPolicies` writes only `default-deny` / `infrastructure` / `allow-api` policy-types plus two name-targeted legacy deletes; the typed sweep only deletes `context-allow` / `rpc-proxy-egress` / `external-egress` orphans, and the safety inventory classifies a baseline policy as `unrelated` so it's never swept. A concurrent full pass is safe.
- **No "no-deny-all" window** — `applyNetworkPolicy` is create-or-replace with 409 retry, never delete-then-create; concurrent writes of the same policy are idempotent.
- **Single-flight lock** (`k8sClient.ts:3094-3113`) — no check-then-set race (no `await` between the guard and the assignment on Node's single thread), the try/catch means a throwing tick never rejects the returned promise, `finally` always releases, and the `=== run` identity guard stops a settling old run from nulling a newer lock.
- **Metric** — `defaults-only-in-flight` increments exactly on overlap; `pass-in-flight` is never touched from this path; `reason` cardinality stays closed.
- **Timer lifecycle** — registered only when interval > 0, cleared and nulled in `stop()`. Consume+reset of `netPolConvergenceEnsureDefaults` stays atomic, so event-driven/startup passes stay `ensureDefaults: false`.
- **Config validation** — `requireCanonicalNonNegativeEnvInt` rejects negatives, decimals, sci-notation, and suffixes; covered by tests.
- Knob is absent from `deploy/` and all overlays; no secrets in the two new infra logs.
- Suites: `vitest run src/k8sClient.test.ts src/config.netPolResync.test.ts` → **316 passed**; `tsc --noEmit` clean.

alfredolopez80:
@claude please perform a **closing re-review and audit** of this PR on HEAD `32a4e8d1e12d64570c8fe8536a30068fdfe5fd93` (base `dev`).

This is **only** #501 / #488 (NetworkPolicy defaults-only / class D tick). Do **not** expand to #495, #490, #461, Role skip, the #490 apply-compare site, or a sixth canonicalize walker. Do **not** merge from this comment.

Since your previous review on `386f5f15` (SHIP), this landed:
- `32a4e8d1` — zach88 nits: reconciler-level write-set disjoint test, tick success/error metrics, `stop()` clears `netPolDefaultsOnlyRun`
- zach88 review: https://github.com/evenfire-ai/evenfire/pull/501#issuecomment-5456833453

HCC version on this HEAD is **0.1.266**. Re-prove every item below. Do not assume the prior SHIP.

## Required output format

Do **not** give only a global verdict. For **every** item below, reply with one line:

`ID — PASS | FAIL | STILL-OPEN — <one-sentence evidence from HEAD>`

If any item is FAIL or STILL-OPEN, say whether it is a merge blocker. End with a closing verdict: ship / do-not-ship, and list remaining blockers if any.

---

## Product constraints

- **P1** Knob `CONTEXT_MAPPER_NETPOL_DEFAULTS_RESYNC_SEC` via `requireCanonicalNonNegativeEnvInt`, **default 0**. Interval 0 → no defaults-only timer.
- **P2** Knob is **not** in `deploy/base` or any overlay.
- **P3** `runNetworkPolicyDefaultsOnly()` → `ensureDefaultPolicies()` only. Never `fullReconcile` / `runInitialNetworkPolicyConvergence`.
- **P4** Cheap tick must **not** consult `initialConvergenceRuns.has('NetworkPolicy')`. No “full pass waits for defaults-only.”
- **P5** Overlap serializes with `{reason="defaults-only-in-flight"}`. Do not increment `pass-in-flight`.
- **P6** A full pass in flight must still allow the cheap tick (disjoint write sets).
- **P7** Event-driven / startup passes stay `ensureDefaults: false`. Consume+reset of `netPolConvergenceEnsureDefaults` stays atomic.
- **P8** `stop()` clears `netPolDefaultsResyncTimer` **and** `netPolDefaultsOnlyRun`. Do not lower 3600→1500. HARD REQ 2 stays on the full pass only.
- **P9** No minikube `deny-all-<ns>` mutation. No sixth `#496` canonicalize walker.

---

## Findings / hunts to re-verify

- **[D1–D6 / T-A / T-B]** Default 0; interval>0 only `ensureDefaultPolicies`; full pass in flight still runs cheap tick; overlap serialize; throw releases lock (T-A); dual-timer independence (T-B); event-driven `ensureDefaults: false`; `stop()` clears timer.
- **[R1-M1]** Reconciler-level test against a fake k8s API (not mocks of both methods): `{name, policy-type}` written by real `ensureDefaultPolicies` is disjoint from what the typed sweep lists/deletes; a planted `context-allow` orphan is deleted; baseline writes remain.
- **[R1-L1]** That test asserts K8s-observable store/deletes, not only call counts.
- **[R1-L2]** `clerum_hcc_netpol_defaults_only_ticks_total` + `clerum_hcc_netpol_defaults_only_tick_duration_seconds` with closed `result=success|error`. Bootstrap `start()` `ensureDefaultPolicies` must not increment the tick counters. A throwing tick increments `error`.
- **[R1-L3]** After `stop()`, `netPolDefaultsOnlyRun` is null; the next `runNetworkPolicyDefaultsOnly` starts a new apply and does not join a stale run.
- **[D7–D12]** Knob absent from `deploy/`; env fail-loud; metric `reason` cardinality still closed (`pass-in-flight`, `defaults-only-in-flight`); no shared skip; no sixth walker; image bump is 0.1.266.

Hunt any remaining defect of any kind on the full delta vs `dev`. Report new findings with severity. Closing verdict: ship / do-not-ship.

PR #517, opened 2026-08-29
feat(desktop-app): first-run tour for a signed-in user

andypavia:
Stacked on #458. Base is that branch, not `dev`, so the diff here is only the tour — retarget to `dev` once #458 merges.

Implements PR 1 + PR 2 (Units 1–4) of the authenticated onboarding spec: a modal tour shown once per install, right after a user first signs in.

## The deck

Six cards, the same six for everyone:

| | |
|---|---|
| Welcome to Evenfire | the product mark |
| Your agent | `your_agent.png` |
| One place for your files and your agents | `global_file_system.png` |
| Add your own connectors | `connectors.png` |
| Plugins that bring their own interface | `apps.png` |
| Say hi | `start.png` |

Fixed, not assembled from what the environment holds. Each card carries commissioned artwork, so a card only some users reach is artwork most people never see. Copy still adapts inside a card — the agent card names the user's agent, and the last card changes when no agent has been shared with them yet — but the deck itself does not.

## The modal

- Modal, not spotlight cut-outs. It anchors to no live DOM element, so a sidebar refactor cannot break it and it behaves the same while a surface is still loading.
- Focus is trapped and restored, the primary action takes focus on every step, position is exposed as text as well as dots, arrow keys walk the deck, Escape closes. A backdrop click deliberately does **not** — a stray click should not end something shown once per install.
- Last of the modals: it yields to plugin consent and deep-link confirmation, each of which is a user action or a security decision.

## Shown once, and only once

The seen flag lives in `localStorage`, read once per session rather than per render. React StrictMode double-mounts in development, and a naive write-on-paint burned the flag before anything painted — the tour was marked seen without ever being shown. There is a regression test that fails without the session-scoped read.

`VITE_EVENFIRE_TOUR_PREVIEW=true` (or `make local-app-tour`) replays the tour without reading or writing the flag, so previewing it leaves no trace and does not consume a real user's one showing.

## Notes for review

- **Artwork**: five PNGs, ~1.8 MB total, committed under `ui/public/tour/`. They ship with an opaque near-white background rather than transparent. Light theme matches the card and needs nothing; dark theme gets a plate behind the image so it reads as an illustration panel instead of a white hole. `--tour-art-plate` exists to be deleted once the assets ship transparent.
- **Two cards left the tour** when the deck was pinned: Approvals ("Nothing runs without your approval") and Worth knowing (notifications, command palette, several instances). Both were drawn-glyph cards with no artwork. Easy to bring back if either is worth commissioning for.
- **Copy**: this went through several rounds against the running app. The product speaks about connectors and plugins, never "context"; nothing claims the app is local; nothing asks the user to verify that the product works.

## Verification

- `tourDeck`, `useTourController`, `TourModal`, `App.deepLinks` — 62 tests, all passing.
- `npm run style-rules` clean; Prettier clean; renderer builds.
- `tsc --noEmit` produces no errors this branch did not already have on its base (24 pre-existing, unchanged).
- Driven by hand in the packaged dev app across both themes.


andypavia:
We need to have personal mcp-servers for the onboarding to be factual

PR #525, opened 2026-08-31
feat(codex): open sync on create and auto-fetch the catalog

alfredolopez80:
## Summary
- Create a ChatGPT subscription now closes the name dialog and opens the sync modal in the same gesture, then starts `device/start` (no pencil / Edit step).
- Control API auto-syncs the catalog when a grant becomes `connected` (device poll and browser callback), fail-soft so OAuth still returns 200/303; manual Sync catalog remains the retry.
- Device panel exposes a real ChatGPT `<a href>` (`target=_blank`, `noopener`) plus Copy for the user code.

## Test plan
- [x] `make test-codex-subscription-t0` (10 groups, no skips)
- [x] Hub units: create opens Sign in without pencil; link+copy; poll `ready` does not POST `catalog/sync`; device/start failure keeps the modal
- [x] API units: connected poll syncs before allowlist; sync failure still 200; callback sync failure still 303
- [ ] Playwright J3: Create stays on `/secrets/llm/subscriptions` with Sign in visible (no pencil)
- [ ] Playwright J4: `device/start` from Create, usable link + Copy, then Agents have one Credential select

### DIFF
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.521",
+  "version": "0.1.522",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.521",
+      "version": "0.1.522",
       "dependencies": {
         "@clerum/display-field": "file:../packages/display-field",
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/control-api/package-lock.json
+++ b/control-api/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.520",
+  "version": "0.1.521",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-control-api",
-      "version": "0.1.520",
+      "version": "0.1.521",
       "dependencies": {
         "@clerum/display-field": "file:../packages/display-field",
         "@clerum/image-policy": "file:../packages/image-policy",
--- a/control-api/package.json
+++ b/control-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.520",
+  "version": "0.1.521",
   "description": "Internal Clerum control plane REST API + UI for managing CRDs and secrets",
   "main": "dist/main.js",
   "engines": {
--- a/control-api/package.json
+++ b/control-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-control-api",
-  "version": "0.1.521",
+  "version": "0.1.522",
   "description": "Internal Clerum control plane REST API + UI for managing CRDs and secrets",
   "main": "dist/main.js",
   "engines": {
--- a/control-api/src/routes/admin/codexSubscription.ts
+++ b/control-api/src/routes/admin/codexSubscription.ts
@@ -29,6 +29,7 @@ import {
   type CodexOAuthDeps,
   CodexSubscriptionOAuthError,
   getCodexSubscriptionConnection,
+  isCodexOAuthErrorCode,
   pollCodexDevice,
   refreshCodexSubscriptionConnection,
   revokeCodexSubscription,
@@ -417,6 +418,11 @@ export function createAdminCodexSubscriptionRouter(
         )
         if (sync.connection) {
           result = { ...result, connection: sync.connection }
+        } else if (!sync.ok) {
+          result = {
+            ...result,
+            connection: { ...result.connection, catalogStatus: sync.catalogStatus },
+          }
         }
         if (await publishRuntimeAllowlistOrFail(res)) return
       }
@@ -461,26 +467,16 @@ export function createAdminCodexSubscriptionRouter(
         })
         return
       }
-      if (synced.reason === 'no_grant') {
-        res.status(404).json({ error: 'no_grant' })
+      if (synced.reason === 'no_grant' || synced.reason === 'disabled') {
+        res.status(404).json({ error: synced.reason })
         return
       }
       if (synced.reason === 'stale_revision') {
         res.status(409).json({ error: 'stale_revision' })
         return
       }
-      if (synced.reason === 'disabled') {
-        res.status(404).json({ error: 'disabled' })
-        return
-      }
-      if (synced.reason !== 'catalog_sync_failed' && synced.reason !== 'unavailable') {
-        sendOAuthError(
-          res,
-          new CodexSubscriptionOAuthError(
-            synced.reason as ConstructorParameters<typeof CodexSubscriptionOAuthError>[0],
-            synced.reason
-          )
-        )
+      if (isCodexOAuthErrorCode(synced.reason)) {
+        sendOAuthError(res, new CodexSubscriptionOAuthError(synced.reason, synced.reason))
         return
       }
       if (await publishRuntimeAllowlistOrFail(res)) return
--- a/control-api/src/routes/admin/codexSubscription.ts
+++ b/control-api/src/routes/admin/codexSubscription.ts
@@ -13,7 +13,6 @@ import {
   listOfferedCodexModelsForAssignment,
   pickCodexGrantModel,
   setCodexCatalogModelEnabled,
-  syncCodexSubscriptionCatalog,
 } from '../../services/codexSubscriptionCatalog.js'
 import {
   CODEX_UNASSIGNED_CONNECTION_KEY,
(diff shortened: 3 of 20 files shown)


alfredolopez80:
@claude closing re-review request (do not merge)

Head: `1a4b7d1726b966955b5397c0239ad4c908a3e180` on `feat/codex-subscription-connect-sync`.
Previous head reviewed: `3e7b12d16c22e0792a6f84a4503f3b0ac0f5e4c6`.
Base: `dev`.
CI on this head: all required checks passed (control-api, control-ui, desktop-app, Real Postgres, CodeQL, leak guards, repo hygiene).

## How this was reviewed

Read-only fan-out against disposable worktrees (base `307b65a3`, previous head `3e7b12d1`), then a confirmed-only fix commit. Lanes: Control API, Control UI, security, architecture, silent-failure, tests/E2E, frontend, code-review, refactoring, and regression (API 65/65 then 68/68 after the fix; UI 11/11 then 16/16). Local re-validation after the fix: `make test-codex-subscription-t0` passed (10 groups). Playwright J3/J4 was not re-run against a live Control UI (same non-goal as the original PR; those specs are not in this CI matrix).

Please re-review the **delta from `3e7b12d1` to `1a4b7d17`** plus whether the original create→sync plan invariants still hold.

## Confirmed findings that this commit fixes

1. **Closed catalog-sync `reason` (High).** `runCodexCatalogSync` no longer puts `err.message` on the result. Catch: `CodexSubscriptionOAuthError` → `err.code`; anything else → `catalog_sync_failed`. `isCodexOAuthErrorCode` is the closed set (`reauth_required` is not a member). `catalogSyncHandler` maps known OAuth codes through `sendOAuthError`; untyped / catalog failures are **503 `{ error: 'catalog_sync_failed' }`**, never a 400 that echoes internal text. Tests: helper “proxy down” now expects `catalog_sync_failed`; typed `provider_unavailable` stays 400; untyped `proxy down` on the admin route is 503 and must not appear in the body.

2. **Stale `ready` after reconnect + failed sync (Medium).** If poll auto-sync fails **without** `sync.connection`, the poll body still overlays `catalogStatus: sync.catalogStatus` so the Hub cannot toast “Connected — catalog synced” from a leftover ready row. New admin test covers leftover `ready` + `ok: false` without `connection`.

3. **Create→connect UI / tests (Medium).**
   - `closeEdit()` reloads the table (server may have persisted after ChatGPT approval; Cancel must not try to abort that persist).
   - `connectEpoch` increments on unmount.
   - Primary CTA is not labeled “Saving…” while device-connecting (`deviceConnecting`).
   - Device panel is `role="status"`; ChatGPT link is underlined only inside `.cu-codex-device-code`.
   - Hub unit now requires `role="dialog"` + Sync catalog on Create; create reject keeps the Add dialog; empty name does not POST; `rel` must include `noopener` **and** `noreferrer`; `href` matches the mocked `verificationUri`.
   - `sanitizeCodexDeviceStart` allowlists only `https://auth.openai.com/codex/device`.
   - J4 now requires `device/start` pathname `/connections/${grantKey}/device/start`, `href === deviceBody.verificationUri`, and `noreferrer`. `createGrant` rejects an empty `connectionKey`.
   - Callback / poll tests assert `runCodexCatalogSync(..., connectionKey, ...)`.

## Explicitly not changed (please do not treat as regressions)

- Browser callback stays **unconditional 303 `connected`** after persist (plan fail-soft). Catalog auto-sync failure still does not change the redirect.
- New catalog rows stay `enabled=true` before the allowlist ConfigMap (plan).
- Auto-sync stays in Control API (`devicePollHandler` + callback), not the UI. UI does not POST `catalog/sync` after a ready poll.
- No sync inside `persistGrantedTokens`. No combined create+OAuth POST. No `verificationUriComplete`.
- Cancel does **not** abort a ChatGPT-approved persist or ConfigMap publish; it aborts the client poll loop and reloads.
- Catalog write atomicity, transport `AbortSignal`, and `slow_down` / `* 40` deadline are pre-existing / non-goal.
- J6–J9 native `<select>` vs `LlmSecretSelect`, and callback return location `/llm-models/providers/codex-subscription` (Agents redirect), are pre-existing and outside this CI matrix.

## Files in `1a4b7d17`

- `control-api/src/services/codexSubscriptionOAuth.ts` — `isCodexOAuthErrorCode`; typed catch reason.
- `control-api/src/routes/admin/codexSubscription.ts` — OAuth-code-only mapping; poll overlay when sync fails without connection.
- `control-api/test/services.codexSubscriptionOAuth.test.ts`
- `control-api/test/routes.admin.codexSubscription.test.ts`
- `control-api/test/routes.auth.codexSubscriptionCallback.test.ts`
- `control-ui/components/CodexSubscriptionHub/index.tsx`
- `control-ui/app/globals.css`
- `control-ui/lib/codexSubscription.ts`
- `control-ui/components/__tests__/CodexSubscriptionHub.test.tsx`
- `control-ui/lib/__tests__/codexSubscription.sanitize.test.ts`
- `tests/e2e/playwright/control-ui/codex-subscription-connection.spec.ts`
- `tests/e2e/playwright/pages/codex-subscription.ts`
- `scripts/tests/test-codex-subscription-t0.sh` (includes sanitize suite)
- Hook version bumps: control-api 0.1.522, control-ui 0.1.477, playwright 0.1.37

Please confirm the High `reason` leak is gone, the poll overlay is correct, J4 is bound to the new grant, and the six plan invariants still hold. No merge requested from this comment.

zach88:
Took a deep pass through this one, tracing the catalog auto-sync end to end across control-api and control-ui plus a security sweep and a contract-consistency pass. This is in good shape: the fail-soft design holds, no token leaks, and the producer/consumer contracts line up byte for byte. The only things I found are three small non-blocking nits.

## Verdict

🟡 **Approve with nits**

Suites are green on the head (control-api 68 passed, control-ui 16 passed). This is my read from the review, not a formal GH approval.

## Optional (non-blocking)

1. **[R1-L1] Incomplete epoch guard in the `connected` branch** - `control-ui/components/CodexSubscriptionHub/index.tsx:222-231`. After a `connected` poll, the block does `await listCodexConnectionModels` and `await load()` and then fires `setEditModels` / `showToast` without re-checking `epoch !== connectEpoch.current`, unlike the rest of the loop. If someone closes the modal (closeEdit bumps the epoch) exactly during those awaits, a "Connected - catalog synced" toast fires and `setEditModels` runs against a closed modal. Purely cosmetic since the connection did succeed server-side and `setEditModels` is inert when `editing === null`, but worth a `if (epoch !== connectEpoch.current) return` before the toast for consistency.

2. **[R1-L2] Unexpected sync errors now surface as 503 instead of 500** - `control-api/src/routes/admin/codexSubscription.ts:483-490`. A raw non-OAuth error (e.g. DB down) during a manual "Sync catalog" is now caught by `runCodexCatalogSync` (fail-soft) and returned as `503 catalog_sync_failed`, where before `sendOAuthError` re-threw it as a 500. It doesn't break the UI (`handleSync` treats any non-2xx uniformly via `formatApiError`), so flagging it only as an intentional-looking behavior change in case the 500 was meaningful to any other consumer.

3. **[R1-L3] Two hardcoded `verificationUri` constants to keep in sync** - `control-api/src/services/codexSubscriptionOAuth.ts:44` and `control-ui/lib/codexSubscription.ts:6`. Same string is hardcoded on both sides and `sanitizeCodexDeviceStart` throws if they ever diverge, which would break device connect entirely. They match today; this is just a future-maintenance note (a shared const or a smoke test asserting equality would remove the footgun).

One test-quality note while I'm here: the `connection` fixtures in the control-api unit tests are hand-written rather than derived from `sanitizeSafeConnection`. Fine for a wrapper unit test that mocks the `syncCodexSubscriptionCatalog` boundary, just noting it so the pattern doesn't grow.

## What checked out clean

- **Fail-soft is real**: `runCodexCatalogSync` never re-throws, so the OAuth callback stays 303 and device poll stays 200 even when the catalog fetch fails (test-asserted).
- **No token leaks**: the `log.warn({ err })` goes through pino with redaction (not `console.*`), the catalog transport swallows fetch errors without dragging the access token (it rides in the POST body, not the URL), and every HTTP response returns a `SafeConnection` with no secrets. `assertNoLeak` covers each branch.
- **Contracts match end to end**: `catalogStatus` values, the manual-sync status codes (200/404/409/503 preserved vs base), `handleCodexBrowserCallback` returning a connection with `connectionKey`, `CodexOAuthDeps` carrying `db`/`encryptionKey` on both routers, `isCodexOAuthErrorCode` covering all 14 codes, and the `verificationUri` constant matching byte for byte between producer and consumer.
- **Auth guards intact**: `admin/index.ts` and `auth/index.ts` are untouched, and the `disabled` -> 404 check is preserved.
- **Body invariants covered**: "no pencil on create", "still 303 when sync fails", and the real `<a href>` with Copy all have tests. Playwright J3/J4 are still unchecked in the test plan, so those stay as manual verification on your side.

zach88 (approved):

alfredolopez80:
@zach88 thanks — adversarial pass on the three nits, then the cheap survivors only.

**[R1-L1] kept and fixed** in `aa251896`. The gap is real after `listCodexConnectionModels` / `load()`: Cancel bumps `connectEpoch` and a later toast could still fire. `setEditing` itself is same-turn as the last epoch check (no yield), so it cannot reopen a closed modal; the toast/`setEditModels` path could. Added epoch returns after both awaits, plus a hub unit that Cancels while models are held and asserts no “Connected — catalog synced”.

**[R1-L2] not changing.** Manual Sync landing on `503 catalog_sync_failed` for a raw/non-OAuth failure is the closed-reason contract from `1a4b7d17`, not a regression to undo. `runCodexCatalogSync` must not re-throw (poll stays 200, callback stays 303). Reintroducing a 500 would put internal errors back on the wire or bypass that helper. UI already treats any non-2xx the same. No issue — this is the intended mapping.

**[R1-L3] cheap lock only.** A shared package across control-api and control-ui is not worth it here. T0 now fails if `CODEX_OAUTH_DEVICE_VERIFICATION_URI` and `CODEX_DEVICE_VERIFICATION_URI` are not both exactly `https://auth.openai.com/codex/device`.

Hand-written `connection` fixtures in the API units: left as-is; not opening an issue for a wrapper-test pattern that already stops at the `syncCodexSubscriptionCatalog` mock.

Head: `aa251896`.

PR #528, opened 2026-08-31
feat: subscription setup flow, connect-side catalog sync, and secrets UI polish

miguelo-999:
## Summary

Reworks the ChatGPT subscription (grant) lifecycle UX in Control UI and aligns it with the connect-side catalog sync shipped in #525. Includes visual **and** behavioral changes.

### Subscription flow (behavioral)
- Single create dialog: name + full setup form (sign-in, enabled models, default model) in one surface — no more two-step modal pair with a duplicate name field.
- Confirming create makes the grant and starts the ChatGPT device sign-in immediately (device code card, verification tab opened synchronously in the click handler, copy actions for link and code).
- Catalog sync is no longer a manual step: it now rides the connect handshake (see #525 backend), with connect/copy failure surfaced via toasts; the redundant manual **Sync catalog** button was removed (lifecycle UI removal — matches dev behavior).
- Per-grant **model toggles are kept** (Enabled models grid + default model picker), backed by the existing catalog PATCH endpoints.
- Phase separation: successful grant creation can no longer be reported as \u201ccreation failed\u201d by a later table-refresh failure; refresh failures surface as partial-success toasts, and device sign-in always starts after a successful create. The sign-in card only claims a new tab opened when `window.open` actually returned one (popup-blocked falls back to the manual link).

### Navigation (behavioral + visual)
- Secrets LLM area flattened: sibling **LLM API Keys** / **LLM Subscriptions** tabs replace the nested scope + sub-tab stack.

### Visual
- Subscription dialogs restyled to match the agent **Models & creds** design language (`cu-llm-config` blocks, tags, hints).
- LLM sub-tab spacing doubled under the secrets header.

### Tests
- Unit: create \u2192 auto sign-in sequencing, partial-success regression (create succeeds / reload fails \u2192 sign-in still starts, accurate retryable state), device-code card + tab claim, model toggle + save patch, revoke via row menu.
- Fixtures now use the production `CODEX_DEVICE_VERIFICATION_URI` contract.


miguelo-999:
@claude review and provide greenlight if everything is solid this is mostly a cosmetic PR

josue-tech-94:
@miguelo-999  The PR is currently **blocked** with 2 must-fix findings and 2 should-fix items.

### Must fix

**R1-H1 — Existing-host lifecycle/stateless UI removal**

This PR removes the lifecycle/stateless editor, lifecycle status chip, `StatelessEnableRejected` banner, and the tests protecting those workflows from Host Overview.

There are two valid ways to resolve this:

1. **If lifecycle removal was not intentional:** restore the lifecycle view/edit/banner behavior and its tests.

2. **If lifecycle removal is intentional:** please get an explicit owner/product decision confirming that existing-agent lifecycle controls are intentionally being removed. Then:

   * update the PR description to clearly document the removal;
   * remove/correct stale comments saying existing-agent lifecycle support remains;
   * add replacement coverage proving existing `spec.lifecycle` data is preserved when saving Overview changes.

The important point is that this cannot remain an undocumented behavioral removal inside a PR described as visual subscription changes.

**R1-M1 — Subscription create/sign-in sequencing**

The new subscription flow can successfully create a grant and then fail during the list refresh before device sign-in starts.

That can leave the backend grant created while the UI reports the wrong failure state and never starts the ChatGPT device flow.

Please separate the phases so that:

* successful grant creation does not depend on the subsequent list reload succeeding;
* device connect/sign-in can still start after successful creation;
* a reload failure is reported as partial success / refresh failure rather than “creation failed”;
* the UI only claims the ChatGPT tab/sign-in started when that actually happened.

Please add a regression test covering:

`create succeeds -> post-create list reload fails`

and verify that sign-in still starts or that the user gets an accurate retryable partial-success state.

### Should fix

**R1-L1 — Device verification URI fixture**

`CodexSubscriptionHub` tests currently mock `https://chatgpt.com/device`, but production sanitization only accepts the canonical OpenAI Codex device URI.

Please use the real `CODEX_DEVICE_VERIFICATION_URI` / production-backed contract in these fixtures instead of an impossible response shape.

**R1-L2 — Typography tokens**

The newly added device setup CSS uses raw `font-size` / `font-weight` values. Please switch those to the existing Control UI typography tokens, unless there is an intentional design-system exception.

### PR description

The current description:

> Visual improvements to the ChatGPT subscriptions flows

is incomplete because this PR contains behavioral changes in addition to visual changes.

Please update the description to cover the subscription-flow changes, navigation changes, model management changes, and—if approved—the lifecycle UI removal.

If lifecycle removal is **not** approved, restore that functionality and leave it out of the updated PR description.

### Validation

The reviewed head passed:

* focused Control UI tests: **90 tests**
* full Control UI suite: **1801 passed / 1 skipped**
* `npm run build`

There were no must-fix security findings.

Once the findings above are addressed, we can run the next review round against the new head and specifically check the fix diff for regressions.

miguelo-999:
### R1-H1 is intentional! 
----

All four review items fixed, pushed (140218c0f), and the PR description updated on #528.

R1-M1 — create/sign-in sequencing: handleCreate is now phase-separated. Phase 1 is only the grant create — if it fails, you get "Could not create subscription" (accurate). Once it succeeds, sign-in always starts (handleConnect fires before anything else can fail), and a list-reload failure is reported as a partial-success toast ("Subscription created, but the list could not be refreshed."), never as a creation failure. Applied the same guard inside handleConnect and handleSaveEdit — a refresh failure after a successful connect/patch can no longer masquerade as "sign-in/update failed".

"Tab opened" claim: the card now only says "ChatGPT opened in a new tab" when window.open actually returned a window; if blocked it switches to "Open the ChatGPT verification page:" with the manual link.

Regression test added: create succeeds → post-create reload rejects → asserts startCodexDeviceConnect still fires, the partial-success toast appears, and neither "Could not create subscription" nor the raw error string is shown — then the flow completes through connect.

R1-L1: all test fixtures now use the production CODEX_DEVICE_VERIFICATION_URI contract instead of the impossible chatgpt.com/device shape.

R1-L2: .cu-device-setup styles now use --cu-font-size-sm/xs/xl and --cu-font-weight-semibold tokens (the 1.05rem oddity became xl 1rem — visually near-identical, fully tokenized).

PR description: rewritten to cover the behavioral subscription-flow changes (single dialog, auto sign-in, connect-side sync), navigation flattening, model-management retention, and the lifecycle removal (manual Sync catalog button — kept removed since it matches the new backend; model toggles were never dropped). 6/6 tests pass across 3 runs, tsc clean, tree clean.
@josue-tech-94

josue-tech-94 (approved):

Issue #526, opened 2026-08-31
feat(hcc): attribute clerum_hcc_writes_total and write_skips_total by clerum.io/policy-type


alfredolopez80:
`clerum_hcc_writes_total{kind="NetworkPolicy"}` answers "how many" but not "which". Today that cost an hour of log grepping to learn something the counter already had in hand.

Filed separately on purpose: `PR G` (the no-op gate PR) is fixed at `{kind}` only, and everything `policy-type`-shaped is sequenced last in the master plan because it is tooling capability, not an operational defect.

## The evidence

Measured on `clerum-dev`, 2026-08-31, read from `/metrics` via an ephemeral port-forward:

```
clerum_hcc_writes_total{kind="NetworkPolicy"}  →  783 in 33.3 min  =  23.5/min
```

That number is correct and useless on its own. `kind="NetworkPolicy"` lumps together four families the platform treats as separate concerns:

| family | `clerum.io/policy-type` | owner in the plan |
|---|---|---|
| `ctx-*` | `context-allow` | PR 8 (fan-out gate) |
| `ext-egress-*` | `external-egress` | #299 phases 2-4 |
| `rpc-egress-*` | `rpc-proxy-egress` | PR 8 |
| LlmHook policies | `llm-hooks-egress` | PR 10 |

To learn that those 23.5/min were **100% `ext-egress`**, the controller logs had to be captured with `--timestamps`, the window measured by hand, and the policy names parsed. The counter had the attribution available at the call site and discarded it.

This matters beyond convenience: the master plan's §2-QUINDECIES had to distinguish `ctx-*` from `ext-egress` writes to decide whether PR 7 and PR 8 still carry justification. That decision was made from logs because the metric could not make it.

## Design

**Add `policy_type` as a second label to BOTH counters, in the same change:**

- `clerum_hcc_writes_total` (`host-context-controller/src/metrics.ts:298-302`)
- `clerum_hcc_write_skips_total` (shipped by `PR G`)

Never one without the other. Comparing skips against writes is the entire reason both exist; a label on one and not the other makes `sum by (policy_type)` incomparable across them.

**Read it at the same call site as `kind`** — `host-context-controller/src/utils.ts:78`, where `next` is the merged object and `next.metadata?.labels?.['clerum.io/policy-type']` is already in hand.

**Validate against a closed allowlist — do not read the label raw.** This is required by an existing contract recorded in the body of #494 (closed as duplicate of #493):

[...]

Issue #473, opened 2026-08-27
HCC NetworkPolicy equality has three intentional predicates; do not unify without a dedicated contract


alfredolopez80:
`networkPolicyMatchesDesired` (full-object + apiserver default-fill, apply-path `isUpToDate`), `sameNetworkPolicySpec` (spec-only, no default-fill, safety-inventory filter plus Context revision canonicalize), and `egressSignature` (partial egress fingerprint plus `STATE_ANNOTATION` migration, #299 no-churn) answer different questions.

Unifying them would change uid/RV fence admission, #299 TTL no-op, or #214 author-order. Do not extract a shared canonicalize as a first step — that is the foot in the door.

Follow-up from PR #471 review (Fable KEEP_SPLIT). Not a merge blocker for #471.

PR #335, opened 2026-08-12
feat(egress): provider-CIDR egress rendering (issue #299 Phase 2)

alfredolopez80:
## Summary

Issue #299 **Phase 2 — Provider-CIDR Egress Rendering**. Renders Kubernetes
NetworkPolicy egress from provider-published CIDR ranges instead of the Phase-1
`/32` sliding window, closing the DNS-rotation gap for large provider IP pools.

The concrete failure this fixes (root-caused live in Phase 1): `api.github.com`
resolves to a single rotating IP out of `140.82.112.0/20`. A `/32` snapshot pins
one address; the next resolution rotates off it and egress is dropped. A
deterministic bidirectional TCP probe confirmed it (pinned IP OK 5/5,
non-pinned IP TIMEOUT 100%). The fix renders the **provider's published CIDR
block**, so every rotated address in the pool is covered.

## Design invariant — provider-agnostic by construction

Provider names (`github`, `google`, `aws`, `cloudfront`, `microsoft`) appear
**only** in:
- the data-only `@clerum/network-policy-core/providerRegistry` module,
- the control-api fetcher plugins,
- the seed ConfigMap, and
- fixtures / tests.

Core adapters, reconcilers, and validators are **provider-blind**. A CI grep-gate
(`scripts/ci/check-provider-generality.sh`) **derives** the forbidden name set
from the registry itself and fails if any provider name leaks into a
provider-blind file (17 gated files). Adding a new provider is a data change
(registry row + fetcher plugin + seed), never a code change to the egress engine.

## Components

- **`packages/network-policy-core`** — pure provider-blind functions
  (`validateProviderRanges`, `partitionIpsByProviderRanges`,
  `partitionCidrsByFamily`, `isAllowedExternalEgressCidr`,
  `parseProviderNetblocks`, `resolveProviderRanges`) + the `providerRegistry`
  data module (single source of truth for provider names/bounds and the
  grep-gate denylist).
- **`control-api`** — background cron fetch pipeline **off the hot path**:
  advisory-locked single writer (`pg_try_advisory_lock`), fail-static LKG,
  content-hash no-op-on-unchanged (zero ConfigMap churn), 900 KB CM budget,
  kill switch, bounded HTTP. Writes the `clerum-provider-netblocks` ConfigMap.
  Plus the McpServer `provider` egress validation arm (H4 dup `(dns,port)` guard).
- **`host-context-controller` + `workflow-recipes`** — provider egress branch in
[...]

PR #529, opened 2026-08-31
fix(hcc): skip no-op Deployment/ConfigMap/LlmHook-NP replaces at the remaining writers (#460)

### LINKED ISSUE #527: fix(hcc): skip no-op Deployment/ConfigMap/LlmHook-NP replaces at the remaining writers (#460)
Four HCC writers call `replaceWithConflictRetry` without the no-op predicate that three of their siblings already use. They rewrite the same objects every reconcile pass, and one of them does not go through the helper at all.

Pattern of #469 (Services) and #471 (NetworkPolicies), applied to what those two left behind. **Subsumes what the plan calls `PR 10`** (the LlmHook private writer).

## What is ungated today

| writer | call site | today |
|---|---|---|
| McpServer Deployment | `reconciler.ts:1519` | helper, **no** `isUpToDate` |
| nginx ConfigMap | `reconciler.ts:943-984` | **outside the helper** — create → 409 → read → unconditional replace, **no conflict retry** |
| SFS Deployment | `sharedFileSystemReconciler.ts:370` | helper, **no** `isUpToDate` |
| LlmHook Deployment | `llmHookReconciler.ts:773` | helper, **no** `isUpToDate` |
| LlmHook NetworkPolicy | `llmHookReconciler.ts:825-832` | helper, **no** `isUpToDate` — this is `PR 10` |

Already gated, for contrast: Host and channel-reader Deployments (`hostReconciler.ts:3419`, `:2586`), all four Service writers (#469), the 14 shared-helper NetworkPolicy callers (#471), the Host Role.

## Why it matters, measured

`clerum-dev`, steady-state window of 11.2 min on `sha-307b65a`, first 10 min of log discarded:

```
Updated Deployment       214    19.0/min    50 distinct objects    4.3 writes/object
Updated nginx ConfigMap   81     7.2/min    27 distinct objects    3.0 writes/object
```

Cross-checked against `clerum_hcc_writes_total` read from `/metrics`: `{kind="Deployment"}` at 31.3/min cumulative. The same object is being rewritten several times over a few minutes.

Second-order effect: an unconditional replace bumps `resourceVersion`, HCC watches that, and it recomputes context NetworkPolicies — the mechanic described in #460.

## The finding that makes this non-trivial

**`deploymentMatchesDesired` is not reusable as-is.** It exists (`hostReconciler.ts:5079-5084`) and works for Host *because the Host builder is exhaustive where the normalizer is not*. The new writers are not:

- McpServer and LlmHook probes omit `timeoutSeconds` and `failureThreshold`; the apiserver defaults them to `1`/`3` and the normalizer does not strip them.
- McpServer, LlmHook and SFS do not set `serviceAccountName`; the apiserver defaults it to `'default'`.

Wiring the predicate without addressing this produces a gate that is **permanently inert** — green in tests, useless in production, and indistinguishable from a gate that works.

So the PR adds three strips: `timeoutSeconds === 1`, `failureThreshold === 3`, `serviceAccountName === 'default'`.

## The contract that keeps one shared predicate safe

One predicate serving five writers is a coupling point. It is made safe by constraining the *form* of what may be added to it:

> **Safe-strip lemma.** A normalizer mutation of the form *"delete field X iff X === documented-default, applied identically to both sides"* **cannot hide any real change**, for any consumer present or future.
>
> *Proof.* For a transition A→B with A≠B: if B≠default the desired keeps B and equality fails; if B===default the desired loses it but the live object keeps A — which was ≠default, or there would be no transition — and equality fails anyway. All such a strip can do is turn a no-op into a skip, which is the feature.

Anything outside that form — unconditional deletes, value normalisation, per-writer branches — can hide changes, and a mutation sweep drives that red.

## Scope

**In**: move the predicate to `utils.ts` and export it; the three strips; wire it at McpServer, SFS and LlmHook Deployments; route the nginx ConfigMap through the helper with a new `configMapMatchesDesired`; `isUpToDate` on the LlmHook private NetworkPolicy writer; `clerum_hcc_write_skips_total{kind}`; a retroactive test suite for the Host gate, which today has none.

**Out, with reasons**: McpServer `status` (already short-circuits internally; a separate one-line commit fixes only its misleading log). GFS Deployments (gated upstream by `deploymentNeedsUpdate`'s template hash, and they own their rollout machinery). The `policy_type` metric label — that is #526.

## Acceptance

- [ ] The five ungated writers pass `isUpToDate`; a no-op reconcile issues **zero** replaces.
- [ ] A real change still writes: credential rotation rolls pods, an image bump replaces, a changed egress conf replaces, a changed NetworkPolicy rule replaces.
- [ ] **Every gate driven red**: remove the wiring, watch the no-op test fail, restore. A guard never seen failing is indistinguishable from a deleted one.
- [ ] Mutation sweep per builder (5): every leaf of the desired JSON tree, when mutated, makes the predicate return false. An undetectable path **fails printing the path** — never skipped.
- [ ] Writer inventory: all eight `replaceNamespacedDeployment` occurrences in `src/` are in one of two closed lists — gated, or exempt with a stated reason.
- [ ] **Fixture validated against a real apiserver in minikube** — `kubectl get deploy -o json` diffed against `asApiserverDeployment(builder(...))`, iterated to empty. A green `npm test` without this diff is not a verdict; it is the top risk of this PR.
- [ ] Post-deploy in `clerum-dev`: the `{kind="Deployment"}` write rate collapses, `write_skips_total` rises, and rotating an `envSecret` **still rolls a pod**.

## Implementation spec

Full specification — per-call-site instructions, the merge interaction, the nine-commit split, the test plan with each guard's red mutation, blast radius per writer, and the mandatory minikube step — lives in the repo at:

```
.local-notes/hcc-204-write-amplification-2026-08-26/claude-session-460/PR-G-IMPLEMENTACION.md
```

Read §0, §1, §2 and §8 before writing a line. The plan context is `PLAN-MAESTRO-FASES.md` §2-SEXDECIES (this PR's card) and §3-ORDEN (why it is first).

## Coordination

No code collision with any open PR. Two file-level intersections with #335 — `package.json` and `metrics.ts` — resolved by the #508 precedent: sibling PRs share the bump, the later merge rebases. Nothing to wait for.

alfredolopez80:
## Summary
- Implements [#527](https://github.com/evenfire-ai/evenfire/issues/527) / [#460](https://github.com/evenfire-ai/evenfire/issues/460): the remaining HCC writers that still replaced Deployments, the nginx ConfigMap, and LlmHook NetworkPolicies on every pass now go through `replaceWithConflictRetry` with a no-op predicate.
- Shared `deploymentMatchesDesired` lives in `utils.ts`. Safe-strip lemma only: delete field X iff X === documented default, applied identically on both sides. The strips that keep McpServer/LlmHook/SFS builders from leaving the gate inert are `timeoutSeconds===1`, `failureThreshold===3`, `serviceAccountName==='default'`, and empty `container.resources === {}` (kube always emits the key; mirrors the existing empty `securityContext` strip). No `periodSeconds===10`. No `imagePullPolicy` strip.
- New `clerum_hcc_write_skips_total{kind}` increments only when `isUpToDate` is true **and** `mutationAllowed()` (if set) is true, before return. `writesTotal` still increments only after a successful `replace()`. GFS and the stateless lifecycle writer stay ungated. LlmHook keeps its private NetworkPolicy helper (`mergeExisting: preserveObjectAnnotations`).
- Channel-reader retry-time ownership veto lives on `validateExisting` (throws) so an adoption race is not counted as a healthy skip.
- `writeStatusCondition` returns boolean and logs `Updated status conditions` only when a status patch actually happened.

**Base:** `origin/dev` `24d749d0` (includes #525). HCC image `0.1.271` → `0.1.284` via the src/ hook.

Closes #527

## Test plan
- [x] Host / channel-reader / McpServer / SFS / LlmHook: fixture ≠ desired (FIXTURE-1) and merged builder vs fixture is a no-op
- [x] `asApiserverDeployment` does not mutate the caller `desired` object
- [x] Leaf sweeps fail if any desired path is undetectable (desired + live-only spec)
- [x] Equality guards keep non-default timeout / failureThreshold / serviceAccountName; empty `resources: {}` matches omitted; `1000m` ≠ `1`
- [x] `NOOP-*` suites assert `replace` is not called; `IMAGE-*` / `ROTATE-*` / `CM-2` write once
- [x] SFS-3: live `restartedAt` skips; real image change keeps the marker
- [x] LlmHook NP: three callers skip; added egress writes once
- [x] Inventory: 8 `replaceNamespacedDeployment(` call-sites across 6 files — 5 gated / 3 exempt — plus 1 gated `replaceNamespacedConfigMap(`; identifier count must equal call count per file
- [x] SKIP-1 / SKIP-2 / SKIP-3 for `writeSkipsTotal` (skip increments; write and read-404 do not; stale `mutationAllowed` does not)
- [x] T0: `scripts/minikube/t0.sh` — 2012 HCC tests, `tsc -p tsconfig.build.json`, plus T0 contract/handoff scripts

Out of this PR: live apiserver §8 fixture diff (no Minikube); GFS Deployment gates; `policy_type` on skip metrics (#526); moving LlmHook NP onto shared `applyNetworkPolicy`; general Kubernetes quantity canonicalization (`1024Mi` vs `1Gi`). Residual lows from the independent review are tracked in [#530](https://github.com/evenfire-ai/evenfire/issues/530).

### DIFF
--- a/host-context-controller/README.md
+++ b/host-context-controller/README.md
@@ -380,7 +380,7 @@ a permissive policy — see the NetworkPolicy and GFS tables:
 | `CONTEXT_MAPPER_DESKTOP_RESOURCES_REQUEST_MEMORY` | `256Mi`                          | Desktop memory request.                                                                                                                                                                  |
 | `CONTEXT_MAPPER_DESKTOP_RESOURCES_REQUEST_CPU`    | `250m`                           | Desktop CPU request.                                                                                                                                                                     |
 | `CONTEXT_MAPPER_DESKTOP_RESOURCES_LIMIT_MEMORY`   | `4Gi`                            | Desktop memory limit.                                                                                                                                                                    |
-| `CONTEXT_MAPPER_DESKTOP_RESOURCES_LIMIT_CPU`      | `1000m`                          | Desktop CPU limit.                                                                                                                                                                       |
+| `CONTEXT_MAPPER_DESKTOP_RESOURCES_LIMIT_CPU`      | `1`                              | Desktop CPU limit. Author the apiserver-canonical form (`1`, not `1000m`) so the Deployment no-op gate can match.                                                                        |
 | `CONTEXT_MAPPER_DESKTOP_API_TOKEN`                | `` (empty)                       | Bearer token required by `GET /api/v1/desktop/:hostRef`. **When empty the check is skipped and the endpoint is unauthenticated** — a dev-mode convenience. Set it in any shared cluster. |
 
 ### channel-reader provisioning
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.280",
+  "version": "0.1.281",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.280",
+      "version": "0.1.281",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.279",
+  "version": "0.1.280",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.279",
+      "version": "0.1.280",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.273",
+  "version": "0.1.274",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.273",
+      "version": "0.1.274",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.278",
+  "version": "0.1.279",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.278",
+      "version": "0.1.279",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.271",
+  "version": "0.1.272",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.271",
+      "version": "0.1.272",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.283",
+  "version": "0.1.284",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.283",
+      "version": "0.1.284",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.281",
+  "version": "0.1.282",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.281",
+      "version": "0.1.282",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.277",
+  "version": "0.1.278",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.277",
+      "version": "0.1.278",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.282",
+  "version": "0.1.283",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.282",
+      "version": "0.1.283",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.272",
+  "version": "0.1.273",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.272",
+      "version": "0.1.273",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.276",
+  "version": "0.1.277",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.276",
+      "version": "0.1.277",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.274",
+  "version": "0.1.275",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.274",
+      "version": "0.1.275",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package-lock.json
+++ b/host-context-controller/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.275",
+  "version": "0.1.276",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-host-context-controller",
-      "version": "0.1.275",
+      "version": "0.1.276",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@clerum/codex-catalog-projection": "file:../packages/codex-catalog-projection",
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.279",
+  "version": "0.1.280",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.278",
+  "version": "0.1.279",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.277",
+  "version": "0.1.278",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.273",
+  "version": "0.1.274",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.280",
+  "version": "0.1.281",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.274",
+  "version": "0.1.275",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.275",
+  "version": "0.1.276",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.271",
+  "version": "0.1.272",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.282",
+  "version": "0.1.283",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.281",
+  "version": "0.1.282",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.276",
+  "version": "0.1.277",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.283",
+  "version": "0.1.284",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/package.json
+++ b/host-context-controller/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clerum-host-context-controller",
-  "version": "0.1.272",
+  "version": "0.1.273",
   "description": "Host Context Controller - provides McpServer CRDs to mcp-host via REST API",
   "main": "dist/main.js",
   "engines": {
--- a/host-context-controller/src/__tests__/asApiserverDeployment.ts
+++ b/host-context-controller/src/__tests__/asApiserverDeployment.ts
@@ -8,9 +8,15 @@ import type * as k8s from '@kubernetes/client-node'
  * a field, refresh this blob so the suite goes red until the comparator or
  * builder learns that default (PR G §8).
  *
- * Container `imagePullPolicy` is filled after overlay from kube
- * `DefaultImagePullPolicy` (Always for `:latest` / untagged; IfNotPresent
- * otherwise) because recorded containers are replaced by the desired list.
+ * Desired containers replace the recorded list (arrays overlay by clone).
+ * Top-level omitted container defaults are then copied from the recorded
+ * container — one list, not a hand-written twin of
+ * `normalizeContainerDefaults`. `imagePullPolicy` stays content-dependent
+ * (`DefaultImagePullPolicy`: Always for `:latest` / untagged; IfNotPresent
+ * otherwise). Nested defaults (ports, probes, env.fieldRef) only exist when
+ * those child objects are present and stay synthesized. Refresh the recorded
+ * container from a live GET so a new apiserver default turns the suite red
+ * (PR G §8 / #530).
  */
 const SERVER_TIME = new Date('2026-04-01T00:00:00.000Z')
 const SERVER_UID = '11111111-2222-3333-4444-555555555555'
@@ -70,6 +76,7 @@ export const RECORDED_APPSV1_DEPLOYMENT: k8s.V1Deployment = {
             name: 'recorded',
             image: 'recorded:1',
             imagePullPolicy: 'IfNotPresent',
+            resources: {},
             terminationMessagePath: '/dev/termination-log',
             terminationMessagePolicy: 'File',
           },
@@ -178,16 +185,33 @@ function applyPodSpecDefaults(podSpec: k8s.V1PodSpec | undefined): void {
   for (const volume of podSpec.volumes ?? []) applyVolumeDefaults(volume)
 }
 
+const RECORDED_CONTAINER_AUTHORING = new Set(['name', 'image'])
+const RECORDED_CONTAINER_CONTENT_DEPENDENT = new Set(['imagePullPolicy'])
+
+function recordedContainer(): k8s.V1Container {
+  const container = RECORDED_APPSV1_DEPLOYMENT.spec?.template?.spec?.containers?.[0]
+  if (!container) {
+    throw new Error('RECORDED_APPSV1_DEPLOYMENT is missing a recorded container')
+  }
+  return container
+}
+
 function applyContainerDefaults(container: k8s.V1Container): void {
-  if (container.resources === undefined) container.resources = {}
+  // Top-level omitted keys come from the recorded container so a refreshed
+  // blob (new apiserver default) fills without a parallel if-chain.
+  for (const [key, value] of Object.entries(recordedContainer()) as Array<
+    [keyof k8s.V1Container, unknown]
+  >) {
+    if (RECORDED_CONTAINER_AUTHORING.has(key) || RECORDED_CONTAINER_CONTENT_DEPENDENT.has(key)) {
+      continue
+    }
+    if (value === undefined) continue
+    if (container[key] === undefined) {
+      ;(container as unknown as Record<string, unknown>)[key] = structuredClone(value)
+    }
+  }
   if (container.imagePullPolicy === undefined) {
     container.imagePullPolicy = defaultImagePullPolicy(container.image)
-  }
-  if (container.terminationMessagePath === undefined) {
-    container.terminationMessagePath = '/dev/termination-log'
-  }
-  if (container.terminationMessagePolicy === undefined) {
-    container.terminationMessagePolicy = 'File'
   }
   for (const port of container.ports ?? []) {
     if (port.protocol === undefined) port.protocol = 'TCP'
--- a/host-context-controller/src/__tests__/asApiserverDeployment.ts
+++ b/host-context-controller/src/__tests__/asApiserverDeployment.ts
@@ -119,13 +119,17 @@ const REPLACE_OBJECT_KEYS = new Set([
 /** Defined overlay keys win; omitted keys keep the recorded apiserver default. */
 function overlayDefined(base: unknown, overlay: unknown): unknown {
   if (overlay === undefined) return base
-  if (overlay === null || typeof overlay !== 'object' || Array.isArray(overlay)) return overlay
-  if (base === null || typeof base !== 'object' || Array.isArray(base)) return overlay
+  if (overlay === null || typeof overlay !== 'object' || Array.isArray(overlay)) {
+    return structuredClone(overlay)
+  }
+  if (base === null || typeof base !== 'object' || Array.isArray(base)) {
+    return structuredClone(overlay)
+  }
   const out: Record<string, unknown> = { ...(base as Record<string, unknown>) }
   for (const [key, value] of Object.entries(overlay as Record<string, unknown>)) {
     if (value === undefined) continue
     if (REPLACE_OBJECT_KEYS.has(key) && typeof value === 'object' && !Array.isArray(value)) {
-      out[key] = value
+      out[key] = structuredClone(value)
       continue
     }
     out[key] = overlayDefined((base as Record<string, unknown>)[key], value)
@@ -175,6 +179,7 @@ function applyPodSpecDefaults(podSpec: k8s.V1PodSpec | undefined): void {
 }
 
 function applyContainerDefaults(container: k8s.V1Container): void {
+  if (container.resources === undefined) container.resources = {}
   if (container.imagePullPolicy === undefined) {
     container.imagePullPolicy = defaultImagePullPolicy(container.image)
   }
--- a/host-context-controller/src/__tests__/asApiserverDeployment.ts
+++ b/host-context-controller/src/__tests__/asApiserverDeployment.ts
@@ -4,8 +4,10 @@ import type * as k8s from '@kubernetes/client-node'
  * Apply documented apps/v1 defaulting onto a desired Deployment. Unlike
  * `asApiserverService`, this is not a recorded-blob overlay: it fills fields
  * the apiserver would default-fill, leaving every HCC-authored value intact.
- * Refresh this function when a live `kubectl get deploy -o json` reveals a
- * default the fixture does not model (PR G §8).
+ * Container `imagePullPolicy` follows kube `DefaultImagePullPolicy` (Always
+ * for `:latest` / untagged; IfNotPresent otherwise). Refresh this function
+ * when a live `kubectl get deploy -o json` reveals a default the fixture
+ * does not model (PR G §8).
  */
 const SERVER_TIME = new Date('2026-04-01T00:00:00.000Z')
 const SERVER_UID = '11111111-2222-3333-4444-555555555555'
@@ -96,6 +98,9 @@ function applyPodSpecDefaults(podSpec: k8s.V1PodSpec | undefined): void {
 }
 
 function applyContainerDefaults(container: k8s.V1Container): void {
+  if (container.imagePullPolicy === undefined) {
+    container.imagePullPolicy = defaultImagePullPolicy(container.image)
+  }
   if (container.terminationMessagePath === undefined) {
     container.terminationMessagePath = '/dev/termination-log'
   }
@@ -123,6 +128,21 @@ function applyProbeDefaults(probe: k8s.V1Probe | undefined): void {
   if (probe.successThreshold === undefined) probe.successThreshold = 1
   if (probe.failureThreshold === undefined) probe.failureThreshold = 3
   if (probe.httpGet && probe.httpGet.scheme === undefined) probe.httpGet.scheme = 'HTTP'
+}
+
+/**
+ * Mirror kube `DefaultImagePullPolicy` / `tags.Latest`: Always when the
+ * reference has no tag or the tag is `latest`; IfNotPresent for any other
+ * tag or a digest. This is not a full docker-reference parser.
+ */
+function defaultImagePullPolicy(image: string | undefined): 'Always' | 'IfNotPresent' {
+  if (!image) return 'Always'
+  if (image.includes('@')) return 'IfNotPresent'
+  const slash = image.lastIndexOf('/')
+  const name = slash === -1 ? image : image.slice(slash + 1)
+  const colon = name.lastIndexOf(':')
+  if (colon === -1) return 'Always'
+  return name.slice(colon + 1) === 'latest' ? 'Always' : 'IfNotPresent'
 }
 
 function applyVolumeDefaults(volume: k8s.V1Volume): void {
--- a/host-context-controller/src/__tests__/asApiserverDeployment.ts
+++ b/host-context-controller/src/__tests__/asApiserverDeployment.ts
@@ -1,0 +1,144 @@
+import type * as k8s from '@kubernetes/client-node'
+
+/**
+ * Apply documented apps/v1 defaulting onto a desired Deployment. Unlike
+ * `asApiserverService`, this is not a recorded-blob overlay: it fills fields
+ * the apiserver would default-fill, leaving every HCC-authored value intact.
+ * Refresh this function when a live `kubectl get deploy -o json` reveals a
+ * default the fixture does not model (PR G §8).
+ */
+const SERVER_TIME = new Date('2026-04-01T00:00:00.000Z')
+const SERVER_UID = '11111111-2222-3333-4444-555555555555'
+const SERVER_RESOURCE_VERSION = '1776125'
+
+export function asApiserverDeployment(desired: k8s.V1Deployment): k8s.V1Deployment {
+  const live = structuredClone(desired)
+  applyServerOwnedMetadata(live)
+  applyDeploymentSpecDefaults(live.spec)
+  return live
+}
+
+function applyServerOwnedMetadata(deployment: k8s.V1Deployment): void {
+  const name = deployment.metadata?.name ?? 'unnamed'
+  const namespace = deployment.metadata?.namespace ?? 'default'
+  deployment.metadata = {
+    ...deployment.metadata,
+    annotations: {
+      ...deployment.metadata?.annotations,
+      'deployment.kubernetes.io/revision':
+        deployment.metadata?.annotations?.['deployment.kubernetes.io/revision'] ?? '1',
+    },
+    creationTimestamp: SERVER_TIME,
+    generation: 1,
+    managedFields: [
+      {
+        apiVersion: 'apps/v1',
+        fieldsType: 'FieldsV1',
+        fieldsV1: { 'f:spec': { 'f:template': {} } },
+        manager: 'kube-apiserver',
+        operation: 'Update',
+        time: SERVER_TIME,
+      },
+    ],
+    resourceVersion: SERVER_RESOURCE_VERSION,
+    uid: SERVER_UID,
+    selfLink: `/apis/apps/v1/namespaces/${namespace}/deployments/${name}`,
+  }
+  deployment.status = {
+    observedGeneration: 1,
+    replicas: deployment.spec?.replicas ?? 1,
+    readyReplicas: 0,
+    updatedReplicas: 0,
+    unavailableReplicas: deployment.spec?.replicas ?? 1,
+    conditions: [],
+  }
+}
+
+function applyDeploymentSpecDefaults(spec: k8s.V1Deployment['spec']): void {
+  if (!spec) return
+  if (spec.replicas === undefined) spec.replicas = 1
+  if (spec.progressDeadlineSeconds === undefined) spec.progressDeadlineSeconds = 600
+  if (spec.revisionHistoryLimit === undefined) spec.revisionHistoryLimit = 10
+  if (spec.minReadySeconds === undefined) spec.minReadySeconds = 0
+  if (spec.paused === undefined) spec.paused = false
+  if (!spec.strategy) {
+    spec.strategy = {
+      type: 'RollingUpdate',
+      rollingUpdate: { maxSurge: '25%', maxUnavailable: '25%' },
+    }
+  }
+
+  const template = spec.template
+  if (!template) return
+  if (template.metadata && template.metadata.creationTimestamp === undefined) {
+    template.metadata.creationTimestamp = SERVER_TIME
+  }
+  applyPodSpecDefaults(template.spec)
+}
+
+function applyPodSpecDefaults(podSpec: k8s.V1PodSpec | undefined): void {
+  if (!podSpec) return
+  if (podSpec.restartPolicy === undefined) podSpec.restartPolicy = 'Always'
+  if (podSpec.dnsPolicy === undefined) podSpec.dnsPolicy = 'ClusterFirst'
+  if (podSpec.schedulerName === undefined) podSpec.schedulerName = 'default-scheduler'
+  if (podSpec.terminationGracePeriodSeconds === undefined) {
+    podSpec.terminationGracePeriodSeconds = 30
+  }
+  if (podSpec.enableServiceLinks === undefined) podSpec.enableServiceLinks = true
+  if (podSpec.preemptionPolicy === undefined) podSpec.preemptionPolicy = 'PreemptLowerPriority'
+  if (!podSpec.serviceAccountName) podSpec.serviceAccountName = 'default'
+  if (podSpec.serviceAccount === undefined) podSpec.serviceAccount = podSpec.serviceAccountName
+
+  for (const container of [...(podSpec.initContainers ?? []), ...(podSpec.containers ?? [])]) {
+    applyContainerDefaults(container)
+  }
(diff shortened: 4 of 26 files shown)


alfredolopez80:
@claude perform a **closing re-review and audit** of this PR at HEAD `173a813b38e32d9bb49cde78cf1f2a63a125ae3a` (base `dev` @ `307b65a38eb90e9484c6b429b987981d9477083b`).

This is **only** #529 / #527 / #460 (noop gates on the remaining HCC Deployment / nginx ConfigMap / LlmHook NetworkPolicy writers + `clerum_hcc_write_skips_total{kind}`). Do **not** expand to #335, #442, #526 `policy_type`, GFS Deployment gates, or Role direct PUTs. Do **not** merge from this comment.

There is **no prior Claude review** on this PR. Treat this as the first review **and** the closing audit: read the **full delta versus `dev`**, not only the latest commit.

CI on this HEAD is **green** (37/37, including `Build / test / typecheck (host-context-controller)` and `Real Postgres integration (Postgres 16)`). Re-check those jobs on **this** SHA before you treat CI as green.

HCC image hook is `0.1.279` on this HEAD.

## Product contract (must hold on this HEAD)

1. `deploymentMatchesDesired` is the single shared Deployment no-op predicate in `utils.ts`. Fail-open (`false`) when either `spec` is missing or compare throws.
2. Safe-strip lemma only: delete field X iff X === documented default, applied identically on both sides. The three new strips are `timeoutSeconds===1`, `failureThreshold===3`, `serviceAccountName==='default'`. **Do not** add `periodSeconds===10`.
3. One `canonicalizeValue` on the no-op path. Do **not** unify it with the `networkPolicyReconciler` walker that keeps `undefined` (#473 / #299).
4. `writeSkipsTotal.inc({ kind: next.kind ?? 'unknown' })` runs **only** when `isUpToDate` is true, before return. Label set is `{kind}` only — `policy_type` belongs to #526.
5. `writesTotal.inc` still runs **only after** a successful `replace()`.
6. GFS (`gfsK8sApi.ts`) and `statelessLifecycleExecutor.ts` stay **ungated**. Inventory: 8 `replaceNamespacedDeployment({` under `src/` excluding tests — 4 gated, 2 exempt.
7. LlmHook private NetworkPolicy helper must **not** move onto shared `applyNetworkPolicy` (it has `mergeExisting: preserveObjectAnnotations`; the shared helper does not). Do **not** add `validateExisting` there.
8. nginx ConfigMap goes through `replaceWithConflictRetry` + `configMapMatchesDesired` (whole object after server-owned metadata strip) + `preserveObjectAnnotations` + `mutationAllowed: isCurrent`.
9. `writeStatusCondition` returns boolean; log `Updated status conditions` only when a status patch happened.
10. `validateExisting` stays **before** `isUpToDate`. Do **not** put `ensureHostRole` on the helper.

## Findings to audit (local `/review-pr`)

Validate each against **this** HEAD. Say **landed / still open / accepted out of scope** with file:line.

- [x] Shared predicate + three safe-strips: Host/channel-reader still no-op vs fixture; McpServer/LlmHook/SFS gates are not inert.
- [x] Leaf sweeps: every desired leaf is detectable (Host, channel-reader, McpServer, SFS, LlmHook).
- [x] `NOOP-*` suites prove `replace` is not called; `IMAGE-*` / `ROTATE-*` / `CM-2` write once.
- [x] SFS-3: live `restartedAt` + identical desired skips; real image change keeps the marker.
- [x] LlmHook NP: three callers (image / service-target / host-egress) skip; added egress writes once.
- [x] SKIP-1 / SKIP-2: skip increments `writeSkipsTotal`; write and read-404 do not.
- [x] Inventory tripwire still matches 8 calls / gated+exempt lists.
- [x] Security: strips cannot hide non-default SA / probe / NP / nginx `data` / credentials-revision drift. Metric label stays `{kind}` only.
- [x] Channel-reader `isUpToDate` ownership veto (`freshOwnerHost !== host.name → true`) is **pre-existing** on `dev`. Do not expand this PR to move it onto `validateExisting` unless you find a **new** functional skip of real drift.

Do **not** treat passing unit tests as proof that `asApiserverDeployment` equals a live apiserver object. Call that residual risk, not a merge blocker, unless you can show a silent skip of authored drift.

zach88:
Took a deep pass through this one. The production code is clean and safe — the `hostReconciler`→`utils` move is byte-for-byte equivalent, the three new safe-strips hold the delete-iff-default lemma, and nothing fails open to skip. The one real problem is about the gate's *effectiveness* for stdio McpServers, plus the test methodology that hides it. Everything else checks out.

🔴 **Not ready to approve**

**1. [R1-H1] The new Deployment gate never skips for `stdio` McpServers (spurious write every reconcile), and no test catches it (T1 fixture coupling).** *Reproduced.*

The `stdio-bridge` container at `host-context-controller/src/reconciler.ts:1369` is the only container in the McpServer tree that **omits `imagePullPolicy`** — `copy-mcp-app` (1273), `egress-proxy` (1292), and the main container (1335) all set it. The apiserver defaults `imagePullPolicy` on the live object, and `normalizeDeploymentForComparison` in `utils.ts` does **not** strip it. So for every stdio McpServer, `deploymentMatchesDesired(desired, live)` returns `false` on every pass → `replaceWithConflictRetry` never skips → a `replace()` each reconcile. The new gate (`reconciler.ts:1504`, `isUpToDate: deploymentMatchesDesired`) delivers zero benefit for that server class: `clerum_hcc_write_skips_total` stays at 0 for them while `writesTotal` keeps climbing — exactly the churn this PR exists to kill.

To be precise on impact: this is a redundant apiserver *write*, not a pod rollout — the apiserver re-defaults `imagePullPolicy` to the same value so the persisted template is unchanged and the Deployment controller triggers no rollout. No correctness/security regression (these writers wrote every pass before this PR too). The harm is the optimization silently not applying to a whole class + extra apiserver/watch load + a mislabeled metric.

Why no test catches it: `src/__tests__/asApiserverDeployment.ts` is a hand-built fixture that synthesizes apiserver defaults from the **same list** `normalizeDeploymentForComparison` strips (its own docstring admits this), so any default neither side models is invisible — both sides match, the gate says no-op, green. On top of that `makeServer()` uses `streamableHttp`, so the stdio branch is never exercised. The suite is 135/135 green, which is the problem. Reproduced with an isolated test using the real `preserveDeploymentAnnotations`: control (live=fixture) → `true`; live with `imagePullPolicy` filled the way the apiserver fills it → `false` → spurious write.

Suggested fix (cheap): set an explicit `imagePullPolicy` on `stdio-bridge` like every sibling does (`server.spec.imagePullPolicy || config.mcpServerImagePullPolicy`) — one line, closes the divergence without touching the normalizer. And to keep the class covered going forward, derive `asApiserverDeployment` from a recorded real blob (`kubectl get deploy -o json`), the way `asApiserverService.ts` already does, and add a `stdio`-transport test case. You already flagged the live-apiserver fixture diff as out of this PR in the description — this is the concrete thing that gap was hiding.

## Optional (non-blocking)

The PR can merge without these; they're follow-up / author's call.

**2. [R1-M1] The inventory test gates at file granularity, not call-site.** `src/__tests__/deploymentWriterInventory.test.ts:46-62` — the "gated" check is `GATED.has(rel) && source.includes('deploymentMatchesDesired')`. `hostReconciler.ts` has two `replaceNamespacedDeployment` calls; the test only checks the string appears once in the file, not that **both** calls are gated. A second ungated call in an already-gated file would pass, caught only by the `toBe(8)` count tripwire (and only because the number changes). The regex `replaceNamespacedDeployment\s*\(\s*\{` also only counts inline-object calls; a `replace(params)` with a variable escapes the inventory. Consider asserting per call-site.

**3. [R1-M2] The sweep tests only cover leaves present in `desired`, blind to apiserver-added fields.** `src/__tests__/deploymentMatchesDesired.test.ts:110-119` and the sweeps in each `*DeploymentNoopGate` — `expectSweepDetectsEveryLeaf` walks only `desired`'s leaves, so by construction it can't cover a field the apiserver adds and `desired` lacks (the `imagePullPolicy` class from #1). Same root as #1; closes with the recorded fixture + a stdio case.

**4. [R1-L1] Hand-built `status.conditions` fixture in the new "does not log Updated status conditions" test.** `src/__tests__/reconciler.test.ts` — the object (incl. `lastTransitionTime: '2020-01-02…'`) is written by hand to match what the reconciler computes. Fine here since the status subresource has no apiserver defaulting, but it goes mute if the reconciler's reason/message/time format changes. Minor.

## What checked out clean

- **Security: safe.** `networkPolicyMatchesDesired` (new gate on the LlmHook NP) can't false-positive over different egress/ingress rules — rules compared verbatim, `canonicalizeValue` preserves array order, the only strip (`protocol==='TCP'`) and the `policyTypes` default-fill are symmetric. `configMapMatchesDesired` compares `.data`/`.binaryData` in full, so a tampered nginx ConfigMap (credential auth headers) gets rewritten. `write_skips_total` carries only `{kind}`, no sensitive data, and no new log prints bodies/headers/tokens. No matcher fails open to skip — all `catch → return false` (rewrite on doubt).
- **Core logic.** The three new safe-strips (`timeoutSeconds===1`, `failureThreshold===3`, `serviceAccountName==='default'`) hold the lemma. Gate coverage matches the description exactly (gated: host, channel-reader, McpServer dep, LlmHook dep + NP, SFS dep; intentionally exempt: GFS and `statelessLifecycleExecutor.ts`). `writeStatusCondition`→boolean propagates correctly across all 11 returns, and the `networkWrote || deploymentWrote` log guard is correct (that function only writes NetworkReady + DeploymentReady).
- **Description accuracy.** The "8 `replaceNamespacedDeployment({` — 4 gated, 2 exempt" line reads correctly (those are file counts; 8 call-sites split 5 gated / 3 exempt — both readings line up).
- **Typecheck:** no diff file produces a type error (the TS2307/TS7006 are the unlinked `@clerum/codex-catalog-projection` workspace, outside the diff).
- **Tests:** 9 new/modified files, 135 tests, all green.

alfredolopez80:
@claude perform a **closing re-review and audit** of this PR at HEAD `088305db6851c5938a74a2d16a2c61268e4b0af3` (base `dev` @ `24d749d08cfbdd22639f308755450910d55d9937`).

This is **only** #529 / #527 / #460 (noop gates on the remaining HCC Deployment / nginx ConfigMap / LlmHook NetworkPolicy writers + `clerum_hcc_write_skips_total{kind}`). Do **not** expand to #335, #442, #526 `policy_type`, GFS Deployment gates, or Role direct PUTs. Do **not** merge from this comment.

Read the **full delta versus `dev`**, not only the latest commit. After the first Claude pass on `173a813b`, zach88's review (issuecomment-5483500631) landed three follow-up commits: `cce48503` (stdio-bridge `imagePullPolicy`), `718b4bd3` (recorded Deployment overlay + live-only spec sweeps + writer-seeded status skip-log), `088305db` (per-call-site inventory tighten + ConfigMap inventory). There is also a merge of `dev` (`ff6101fb`, #525 Codex subscription only — not HCC product).

CI on this HEAD must be **green** before you treat it as green. Re-check `Build / test / typecheck (host-context-controller)` and `Real Postgres integration (Postgres 16)` on **this** SHA.

HCC image hook is `0.1.282` on this HEAD.

## Product contract (must hold on this HEAD)

1. `deploymentMatchesDesired` is the single shared Deployment no-op predicate in `utils.ts`. Fail-open (`false`) when either `spec` is missing or compare throws.
2. Safe-strip lemma only: delete field X iff X === documented default, applied identically on both sides. The three new strips are `timeoutSeconds===1`, `failureThreshold===3`, `serviceAccountName==='default'`. **Do not** add `periodSeconds===10`. **Do not** strip `imagePullPolicy` (Always vs IfNotPresent is not a single default).
3. One `canonicalizeValue` on the no-op path. Do **not** unify it with the `networkPolicyReconciler` walker that keeps `undefined` (#473 / #299).
4. `writeSkipsTotal.inc({ kind: next.kind ?? 'unknown' })` runs **only** when `isUpToDate` is true, before return. Label set is `{kind}` only — `policy_type` belongs to #526.
5. `writesTotal.inc` still runs **only after** a successful `replace()`.
6. GFS (`gfsK8sApi.ts`) and `statelessLifecycleExecutor.ts` stay **ungated**. Inventory: 8 `replaceNamespacedDeployment(` under `src/` excluding tests (5 gated / 3 exempt) plus 1 gated `replaceNamespacedConfigMap(` in `reconciler.ts`.
7. LlmHook private NetworkPolicy helper must **not** move onto shared `applyNetworkPolicy` (it has `mergeExisting: preserveObjectAnnotations`; the shared helper does not). Do **not** add `validateExisting` there.
8. nginx ConfigMap goes through `replaceWithConflictRetry` + `configMapMatchesDesired` (whole object after server-owned metadata strip) + `preserveObjectAnnotations` + `mutationAllowed: isCurrent`.
9. `writeStatusCondition` returns boolean; log `Updated status conditions` only when a status patch happened.
10. `validateExisting` stays **before** `isUpToDate`. Do **not** put `ensureHostRole` on the helper.

## zach88 findings — verdict required per ID

Say **CLOSED / PARTIAL / OPEN** with file:line. Do not LGTM globally.

- [x] **R1-H1** stdio-bridge omitted `imagePullPolicy` so the Deployment gate never skipped `stdio` McpServers. Tests were `streamableHttp`-only; fixture did not model IPP.
- [x] **R1-M1** inventory was file-granularity (`source.includes('deploymentMatchesDesired')`) and only counted `replaceNamespacedDeployment({`.
- [x] **R1-M2** mutation sweeps walked only `desired` leaves, blind to apiserver-added fields.
- [x] **R1-L1** hand-built `status.conditions` (reason/message/`lastTransitionTime`) in the skip-log test.

## Findings to re-audit (first Claude pass + local /review)

Validate each against **this** HEAD. Say **landed / still open / accepted out of scope** with file:line.

- [x] Shared predicate + three safe-strips: Host/channel-reader still no-op vs fixture; McpServer/LlmHook/SFS gates are not inert.
- [x] Leaf sweeps: every **desired** leaf and every **live-only spec** leaf is detectable (Host, channel-reader, McpServer HTTP + stdio, SFS, LlmHook).
- [x] `NOOP-*` suites prove `replace` is not called; `IMAGE-*` / `ROTATE-*` / `CM-2` write once.
- [x] SFS-3: live `restartedAt` + identical desired skips; real image change keeps the marker.
- [x] LlmHook NP: three callers (image / service-target / host-egress) skip; added egress writes once.
- [x] SKIP-1 / SKIP-2: skip increments `writeSkipsTotal`; write and read-404 do not.
- [x] Inventory tripwire: 8 Deployment calls, gated+exempt lists, `isUpToDate: deploymentMatchesDesired` or `return deploymentMatchesDesired(...)`, plus 1 ConfigMap call with `isUpToDate: configMapMatchesDesired`.
- [x] Security: strips cannot hide non-default SA / probe / NP / nginx `data` / credentials-revision drift. Metric label stays `{kind}` only.
- [x] Channel-reader `isUpToDate` ownership veto (`freshOwnerHost !== host.name → true`) is **pre-existing** on `dev`. Do not expand this PR to move it onto `validateExisting` unless you find a **new** functional skip of real drift.

Do **not** treat passing unit tests as proof that `asApiserverDeployment` equals a live apiserver object. Call that residual risk, not a merge blocker, unless you can show a silent skip of authored drift.

alfredolopez80:
## Independent review — 6 agents in parallel, isolated worktrees, plus a live-cluster measurement

HEAD `ff6101fb` · base `dev` · merge-base `24d749d0`

Six agents, three worktrees (one read-only, two disposable), no agent saw another's output. Partitioned by *failure type*, not by service: predicate/normalizer core · call sites & write path · apiserver-default inertia · test methodology (drive-red) · spec conformance & metrics · silent failure & security. A finding two agents reached from different directions is marked **[x2]**.

Full HCC suite on this HEAD: **97 files / 2004 tests, exit 0**, reproduced independently by three agents. `tsc --noEmit` exit 0.

---

## Verdict

**No Blocker. No Critical. The production code is correct and the safety argument holds.**

Every error path fails open to write. 30 tampered-live security scenarios all produce a write. All six guards were driven red and restored. Scope discipline is clean on all eight §12 prohibitions. The `hostReconciler`→`utils` move is faithful — a mechanical diff of the merge-base bodies yields exactly five differences, all declared.

**But the PR does not yet deliver its headline benefit.** The gate is permanently inert for **36% of live McpServer Deployments and 100% of SFS Deployments in `clerum-dev`** — measured on the cluster, not modelled. Direction is fail-open (a spurious write, never a silent skip), so this is an *effectiveness* gap, not a correctness or security one.

Recommendation: **two one-line changes before merge**, both inside the existing design.

---

## H1 — the gate is inert for McpServer and SFS  [x2 + live cluster]

`reconciler.ts:1171-1188` (`mcp-server`), `:1271-1282` (`copy-mcp-app`), `k8s/sharedFileSystemFactory.ts:335-341` (`init`) vs `utils.ts:394-412` · **High (effectiveness)**

`core/v1.Container.Resources` is a value struct, and Go's `omitempty` does not elide empty structs — so a live container **always** carries `resources`, rendered `{}` when unset. Three builders omit it; `normalizeContainerDefaults` does not strip it. `deploymentMatchesDesired` therefore returns `false` on every pass, forever.

Two agents reached this independently:

- **A3** enumerated five builders × ~35 apiserver defaults, then measured: `[stdio]`, `[streamableHttp]`, `[SFS]` → `gate vs real apiserver = false`; `[Host]`, `[channel-reader]`, `[LlmHook]` → `true`.
- **A5** came the other way — added `container.resources ??= {}` to the fixture and re-ran: `BASELINE EXIT=0, 45 passed` → `PROBE EXIT=1, 4 failed` (`NOOP-SFS-1`, `SFS-3`, `NOOP-MCPDEP-1`, `NOOP-MCPDEP-STDIO-1`). Same partition.

Both said plainly they could **not** verify the premise that the apiserver emits `resources: {}` (A5 hit the kubectl context guard and correctly did not route around it). Settled read-only against `clerum-dev` — 196 Deployments, 215 containers:

```
containers WITHOUT the resources key : 0 of 215
containers with resources == {}      : 42

McpServer Deployments : 86 total, 31 with >=1 container resources=={}  -> INERT (36%)
SFS (wfc-*)           :  4 total,  4 inert                            -> INERT (100%)
channel-reader        : 14 total,  0 inert                            -> effective
LlmHook               :  3 total,  0 inert                            -> effective
```

The live partition matches the agents' matrix row for row, **including the negative rows**. Affected objects span both McpServer classes — `mcp-server` (streamableHttp/sse) and `copy-mcp-app` (stdio). So `cce48503` closed the `imagePullPolicy` hole on stdio while the same object class stays inert through `resources`: that commit's stated goal is not achieved in production.

**Fix (one line, mirrors the existing pod-level `securityContext` strip at `utils.ts:384`, satisfies the §2 safe-strip lemma):**

```ts
if (Object.keys(container.resources ?? {}).length === 0) delete container.resources
```

If shipped as-is: `write_skips_total{kind="Deployment"}` will rise only from Host / channel-reader / LlmHook and the McpServer + SFS write rate will not collapse. That is also the cheapest falsification of this finding.

### Why this survived three prior reviews — the tie-breaker

**A2 independently reached the opposite conclusion.** It built its *own* apiserver defaulting model, deliberately not reusing the PR's fixture, and reported zero normalized diffs for `mcp-stdio`, `mcp-remote`, `llm-hook` and `sfs` — "all four Deployment gates genuinely reachable."

That negative is a **false negative, and it is the most informative result in this review**: A2's model fills `protocol`, `securityContext={}`, `replicas`, `revisionHistoryLimit`, `progressDeadlineSeconds`, `creationTimestamp`, probe fields, `httpGet.scheme`, `fieldRef.apiVersion`, volume modes and `imagePullPolicy` — but **not `resources`**. Two independently written models of the apiserver (the PR's fixture and A2's) share the same blind spot, and no amount of test-side ingenuity broke the tie. Only reading a live object did.

That is the strongest possible argument for the §8 live-apiserver fixture step, and it should be read as evidence for the procedure rather than as a mark against A2.

---

## H2 — the fixture mutates its own input, so the no-op tests are partly vacuous  [x2]

`__tests__/asApiserverDeployment.ts:110-133` (`overlayDefined`), introduced by `718b4bd3` · **Medium as a test defect — but it is the root cause of H1's invisibility**

`overlayDefined` returns overlay arrays **by reference**, so `asApiserverDeployment(desired).spec.template.spec.containers` **is** `desired.spec.template.spec.containers`. `applyContainerDefaults` then writes `imagePullPolicy`, `terminationMessagePath`, `terminationMessagePolicy`, port `protocol`, probe defaults and `fieldRef.apiVersion` **into the caller's `desired`**. Every `*NoopGate.test.ts` does `const existing = asApiserverDeployment(desired)` and then compares `desired` — which by then has silently acquired the very defaults under test.

- **A3** proved the consequence end-to-end: on the **genuinely buggy pre-fix builder** at `173a813b`, the test `predicate: merged builder vs fixture is a no-op` and **both** mutation sweeps stayed **green**. Only 2 of 13 tests caught the bug, and for collateral reasons.
- **A6** quantified the damage to the live-only sweep — the guard added specifically to catch "the apiserver added a field the normalizer strips":

```
HOST      : 11/23 live-only leaves exercised (12 never)
MCP-HTTP  : 14/22 ( 8 never)
MCP-STDIO : 14/26 (12 never)
SFS       : 12/23 (11 never)
```

Among the never-exercised leaves on MCP-HTTP are **two of the three strips this PR introduces** (`livenessProbe.timeoutSeconds`, `livenessProbe.failureThreshold`).

**Good news, verified:** A6 re-ran all five sweeps with the input protected (`asApiserverDeployment(structuredClone(desired))`) and got `undetectable === []` everywhere. **The normalizer is sound on all 43 leaves the PR never checks.** The guard is half inert; it is not hiding a live defect.

**Fix (one line):** `structuredClone` the overlay in `overlayDefined`. Then re-run the sweeps — they pass, and 43 more leaves become guarded.

**Do this one first.** While the fixture writes into `desired`, fixing `resources` leaves the next instance of the same class equally invisible. Four reviewers have now found four instances of it (zach88 `imagePullPolicy`; A3/A5 `resources`; A1 quantities; A2's own model reproducing the blind spot). That is structural, not four coincidences.

---

## H3 — a non-canonical quantity literal can never match the apiserver

`config.ts:719` (`CONTEXT_MAPPER_DESKTOP_RESOURCES_LIMIT_CPU='1000m'`) vs `utils.ts:351-396` · **High if exercised; not exercised in dev today**

The normalizer never canonicalizes Kubernetes quantities. `apimachinery` re-renders `1000m` as `1` on round-trip (the DecimalSI cache condition rejects a `000` suffix), so a Desktop Host's `resources.limits.cpu` differs forever. A1 reproduced the predicate half: `deploymentMatchesDesired('1000m','1') → false`; likewise `'1024Mi'` vs `'1Gi'`.

Live check: **no stored quantity in `clerum-dev` carries a `000m` suffix** (0 matches) and `cpu=1` appears 15 times — the canonicalization is real. **However**, Deployments by owner kind are 94 «no owner» (infra), 86 McpServer, 16 WorkflowRecipe: **no Host CRD Deployment exists in dev right now**, so the path is unexercised. It bites the first time a Desktop Host reconciles, and no test would show it.

`git blame` on the interacting default: `5d663a49` (2026-07-14) — pre-existing. This PR makes a latent non-canonical literal load-bearing for the first time. Fair as a follow-up rather than a merge condition.

---

## Medium

**M1 — `write_skips_total` counts an ownership refusal as a no-op skip  [x2]**
`utils.ts:74-77` + `metrics.ts:307-311` + `hostReconciler.ts:2604-2608`

`hostReconciler.ts:2604` uses `isUpToDate` as a **veto**, not an equality test: `if (freshOwnerHost && freshOwnerHost !== host.name) return true` — "never replace an object whose ownership changed". The veto is pre-existing on `dev` and correct. What this PR adds is a counter that cannot distinguish it from a healthy skip, while `metrics.ts:309` documents the counter as *"…because the merged object was already up to date"*, which is false on that path.

Scenario (A2, REPRODUCED — `skip counter before/after ownership refusal: 0 1`): a channel-reader Deployment is adopted by another Host, or two HostReconcilers race on the same name. Every reconcile hits the veto and increments `write_skips_total{kind="Deployment"}` while `writesTotal` stays flat — which reads exactly like a perfectly converged Deployment. **A dashboard or alert on `rate(write_skips_total)` renders an ownership-conflict storm as this PR's success signal.**

Fix, no behaviour change: move the veto to `validateExisting` (already runs first, `utils.ts:67`), or add a `reason` label (`up_to_date` vs `refused`). If `reason` is deferred alongside `policy_type` (#526), the help text should stop claiming "already up to date".

**M2 — inventory tripwire is blind to a new alias-form writer**
`__tests__/deploymentWriterInventory.test.ts:17` — A4 injected `const f = api.replaceNamespacedDeployment.bind(api); await f({...})` in a file on neither list: **tripwire silent, exit=0**. Four other injections (second ungated call in a gated file; unlisted file; ungated call inside a gated block; alias *refactor* of an existing call) were **all caught**, so this is a narrow residual hole, not a broken guard. One assertion closes it: require `identifierCount === callCount` per file (ground truth today: 9 identifiers vs 8 calls, the delta being a doc comment at `reconciler.ts:302`).

To be explicit about two things zach88's `R1-M1` raised: the regex is `/replaceNamespacedDeployment\s*\(/g`, so it **does** match `replace(params)` with a variable; and the gated check is **per call-site** — it resolves each call's enclosing `replaceWithConflictRetry` block, not "does the string appear in the file". `R1-M1` is resolved.

---

## Low

| # | Where | Finding |
|---|---|---|
| L1 | `reconciler.ts:963-972` + `utils.ts:62-65` | The 409→helper conversion turns a **loud 404 into silent success**. At merge-base a ConfigMap deleted between create-409 and read propagated and failed the reconcile; the helper now swallows it and `ensureConfigMap` resolves as if it succeeded — no ConfigMap, egress-proxy can't mount until the next full pass. Self-healing but silent, against the repo's fail-loud posture. A comment or a `console.warn` closes it. (A2, REPRODUCED) |
| L2 | `utils.ts:253-272`, `:336-350` | Foreign metadata **outside `annotations`** keeps a gate permanently false: both predicates compare the whole `metadata` block minus seven server-owned keys, while `mergeExisting` merges only annotations. A third party stamping a *label*, *finalizer* or extra *ownerReference* (a service mesh, a policy controller) makes `next !== existing` every pass — a permanent two-writer churn loop, the class #460 exists to kill. No such controller in the current clusters, hence Low. Worth a line in the docstring. (A2, REPRODUCED) |
| L3 | `utils.ts:74-77` | `writeSkipsTotal.inc` fires **before** the `mutationAllowed()` staleness check, so a superseded reconcile still counts as a skip — inflating the very number used to judge this fix, under exactly the watch churn it targets. (A6, REPRODUCED) |
| L4 | `metrics.ts:304-311` | `{kind}` alone cannot attribute a skip to a writer, and `kind="Deployment"` mixes four gated writers with GFS's ungated `applyDeployment` (contributes writes, structurally never a skip), permanently diluting any skip ratio. Deliberate here; it means #529 cannot be verified from this metric alone. (A6) |
| L5 | `mutateJsonLeaves.ts:3,10-14,61` | The sweep silently skips `null`, `undefined` and `{}` leaves, and its live-only half is `spec`-scoped, so live-only `metadata.*` is never swept. Unclaimed coverage, not falsely claimed. (A4, REPRODUCED) |
| L6 | `utils.ts:472-481` | `canonicalizeValue` walks any non-array object with `Object.keys()`, so a `Date` collapses to `{}` and two different instants compare equal. Latent today (Date fields are stripped or one-sided), but this is now the single canonicalizer for five gates. (A1, REPRODUCED) |
| L7 | `utils.ts:381-382` | `serviceAccount === serviceAccountName` is a **cross-field** conditional delete, outside the lemma's stated form. Provably non-hiding — but the new line 382 makes the **ordering** load-bearing and undocumented: swap them and the gate silently degrades to always-write. (A1) |
| L8 | `reconciler.ts:2066` | The `catch` treats every non-404 error alike, so a transport error raised *after* the apiserver committed returns `false` and suppresses the `Updated status conditions` log for a write that happened. Under-reports rather than over-reports — right direction; noted so the log isn't later read as a complete audit trail. (A2) |
| L9 | `reconciler.ts:1373-1374` | The stdio-bridge `imagePullPolicy` left operand is **provably dead** — `sanitizeCrdSpec` deletes `server.spec.imagePullPolicy` at `:990` before the builder runs (REPRODUCED: a CRD with `Never` yields `IfNotPresent`). Reads as though a CRD can influence platform pull policy; only call-ordering prevents it. Copied from the pre-existing pattern at `:1273`/`:1335` (`5d663a49`, 2026-07-14), not invented here. Also couples a *platform* sidecar to the *tenant* knob, whereas `egress-proxy` hardcodes `IfNotPresent`. Behaviour-neutral as deployed. (A3/A5/A6) |
| L10 | `utils.ts:96-119` | A foreign annotation on an HCC-owned object is merged into `next`, compares equal, and the gate **skips** — so it sits there with no write and no log. Not a regression (pre-PR code wrote it straight back; persisted state is identical), but the audit trail changes. Labels, which *are* security-load-bearing, are correctly **not** preserved. (A6, REPRODUCED) |
| L11 | `config.ts:282-284` | `getEnv` uses `??`, so `CONTEXT_MAPPER_MCPSERVER_IMAGE_PULL_POLICY=""` yields `imagePullPolicy: ""`, which the apiserver default-fills — reinstating the pre-fix inertness. Requires an operator to set the var empty. (A3, READ-ONLY) |
| L12 | PR body | *"8 `replaceNamespacedDeployment({` — 4 gated, 2 exempt"* mixes units: 8 is a **call-site** count, 4/2 are **file** counts (4+2=6≠8). Correct: *8 call-sites across 6 files — 4 gated files (5 call-sites), 2 exempt files (3 call-sites)*. Note the earlier closing review says "both readings line up"; only the file reading does. (A5) |
| L13 | PR body | Version claimed `0.1.279`, actual **`0.1.281`**; tests claimed 1991, actual **2004**. Both stale by two commits. The bump itself is a legitimate per-commit hook artifact — only the two `version` fields change, no dependency moved. (A5) |

---

## What checked out clean — verified by execution, not assumed

**Drive-red (A4), the primary methodological deliverable.** All six guards fail when their wiring is removed and return to green when restored, on a worktree verified clean between each:

| Guard | Wiring removed | Mutant |
|---|---|---|
| McpServer Deployment | `reconciler.ts:1506` | exit=1, 2 failed / 11 passed |
| nginx ConfigMap | `reconciler.ts:968` | exit=1, 1 failed / 4 passed |
| SFS Deployment | `sharedFileSystemReconciler.ts:370` | exit=1, 2 failed / 6 passed |
| LlmHook Deployment | `llmHookReconciler.ts:772` | exit=1, 1 failed / 6 passed |
| LlmHook NetworkPolicy | `llmHookReconciler.ts:833` | exit=1, 3 failed / 1 passed |
| skip metric | `utils.ts:75` | exit=1, 1 failed / 24 passed |

Symmetric check: forcing each predicate to `true` drives **48 / 5 / 16** tests red. `IMAGE-*`, `ROTATE-*`, `CM-2`, `WRITE-LLMNP-1` are load-bearing, not decorative. A5 ran 15 further mutations (M1–M12, R1–R3), all red — including removing each of the three new strips individually, which proves they are **necessary**, not decorative. A4 also introduced a real lemma violation (`delete spec.replicas`) and the sweep failed **naming the path**, as specified.

**Security (A6), proved not inspected.** 30 tampered-live scenarios against a passing control (so the assertions are falsifiable): `serviceAccountName: 'privileged-sa'` vs desired-omitted, the deprecated `serviceAccount` desync, `automountServiceAccountToken`, probe weakening and deletion, `strategy` flips both directions, `privileged: true`, `capabilities.add`, `runAsUser: 0`, `hostNetwork`/`hostPID`, image digest swap, injected sidecar, `secretKeyRef` repointed, secret volume renamed, `defaultMode` widened to `0777`. **Every one returns `false` → WRITE.** No strip hides a security-relevant change.

**nginx ConfigMap.** `.data` and `.binaryData` compared in full — no strip, no shallow compare. Proved by both A2 and A6, end-to-end through the real `ensureConfigMap` with a credential-bearing `proxy_set_header Authorization` conf: `proxy_pass` repointed at an exfil host → replace called once, written body byte-identical to the authored conf. Six further tamper vectors (injected header, trailing space, extra key, `binaryData`, stripped `ownerReferences`) all rewritten.

**Credential rotation still writes**, proved twice by A6 — predicate level and end-to-end through the real `ensureDeployment` — with a matching no-drift control proving the write is caused by the rotation, not by an inert gate.

**Helper ordering (A2, REPRODUCED).** `validateExisting` (`utils.ts:67`) provably precedes `isUpToDate` (`:74`); a throwing `validateExisting` never reaches the gate (`isUpToDate calls = 0`). The 409 retry **re-reads and re-evaluates the gate against the fresh object** — attempt 1 wrote on divergent state, attempt 2 skipped rather than clobbering a peer's write. Neither stale skip nor concurrent overwrite.

**`writeStatusCondition`** — all 12 exits audited line by line; exactly one `return true`, unreachable without a resolved patch; the `networkWrote || deploymentWrote` log gate is correct because the function has exactly two call sites and both are captured. 7 paths REPRODUCED.

**LlmHook private NetworkPolicy** kept its own `replaceWithConflictRetry` with `mergeExisting: preserveObjectAnnotations` (`llmHookReconciler.ts:830`); the shared helper (`utils.ts:431-459`) still has none — the difference is intact and load-bearing. `networkPolicyMatchesDesired` survived 13 adversarial cases with no false positive: rule reorder, `[]` vs `[{}]`, `policyTypes` defaulting both ways, `UDP`/`SCTP`, `ipBlock.cidr`, removed `except`, changed `podSelector`.

**Determinism.** No builder is nondeterministic — five desired objects byte-identical across two calls 5 ms apart. The only McpServer pod-template annotation is `clerum.io/credentials-revision`, a Secret digest.

**SFS `restartedAt` semantics** — a live marker skips and is preserved; a real image change writes once. Exactly the one-rollout contract claimed at `sharedFileSystemReconciler.ts:363-366`.

**Fail-open direction (A1, A6).** Every error path in the diff returns `false` (= write). No branch returns `true` from an exception or a missing spec, on either argument position, including a `structuredClone`-hostile input.

**The move is faithful (A1).** Mechanical diff of the merge-base bodies against `utils.ts` yields exactly five differences: `export` ×3; seven inline metadata deletes replaced by `stripServerOwnedMetadata`, whose deleted key set is **identical**; and the three declared strips. `preserveDeploymentAnnotationsExcept` survived intact at `hostReconciler.ts:5082-5111` with all call sites live — the trap the spec explicitly warned about.

**All 25 strips enumerated against the lemma (A1).** Every one conforms except the pre-existing cross-field `serviceAccount` delete (L7), provably non-hiding. **No fourth strip crept in** — no `periodSeconds === 10`, grep- and behaviour-verified. The `spec.strategy` compound delete is genuinely conditional and checks *every* field of the type (`V1DeploymentStrategy` has only `type`/`rollingUpdate`) before deleting.

**Scope discipline (A5), clean on all eight §12 prohibitions**, in both directions. `networkPolicyReconciler.ts` is absent from the diffstat — only its three test files change, `+1` line each (a mandatory `writeSkipsTotal` mock stub). GFS is not merely untouched: the inventory test **actively fails** if anyone ever wires the gate there. No dependency added, removed or bumped.

---

## Where the spec was wrong and the PR was right

`PR-G-IMPLEMENTACION.md` §4.1 asserted the McpServer builder authored `imagePullPolicy` explicitly. It did not, on the `stdio-bridge` sidecar. Had the implementer trusted the document, the stdio gate would have shipped inert. The implementation found it and fixed it — exactly what the brief asked for.

Beyond spec, in the PR's favour: the live-only spec-leaf sweep family, and an inventory test that resolves the enclosing `replaceWithConflictRetry` block *per call-site* rather than pairing by file as §7.2(c) only required.

---

## Acceptance criteria (#527)

**5 of 7 MET** (two exceeded) · **1 NOT MET** · **1 CANNOT VERIFY FROM CODE**

The one not met is criterion 6 — *"fixture validated against a real apiserver in minikube"* — which the PR body declares out of scope, honestly. The spec called it risk **R1** and *"el más probable del PR"*. **H1 is that risk, realised**, and the A2-vs-A3/A5 disagreement above shows no test-side substitute closes it.

---

## Recommendation

**Before merge — two one-line changes, both inside the existing design:**

1. **`structuredClone` the overlay in `asApiserverDeployment`** (H2), then re-run the five sweeps. Do this **first**: it makes the next item visible instead of invisible.
2. **Strip empty `resources` in `normalizeContainerDefaults`** (H1), mirroring the existing `securityContext` empty-object strip, and add a fixture case with `resources: {}` so the guard can fail.

**Follow-ups:** H3 (quantity canonicalization / the `1000m` default) · M1 (ownership veto vs skip metric — pairs naturally with #526) · M2 (inventory alias hole) · L1, L2, L3 · and the §8 live-apiserver procedure with CI enforcement, so a stale fixture is noticed rather than trusted.

**Post-deploy verification, if merged:** do not read `write_skips_total` alone. Read it **per kind**, and pair it with apiserver write rate or Deployment `metadata.generation` staying flat for a known-idle McpServer — that is the property the PR actually claims, and the direct falsification of H1.

alfredolopez80:
@claude perform a **closing re-review and audit** of this PR at HEAD `7df1dc1b55304170ba3fe7300992eef465694cce` (base `dev` @ `24d749d08cfbdd22639f308755450910d55d9937`).

This is **only** #529 / #527 / #460 (noop gates on the remaining HCC Deployment / nginx ConfigMap / LlmHook NetworkPolicy writers + `clerum_hcc_write_skips_total{kind}`). Do **not** expand to #335, #442, #526 `policy_type`, GFS Deployment gates, Role direct PUTs, or the leftovers filed as #530. Do **not** merge from this comment.

Read the **full delta versus `dev`**, not only the latest commit.

History since the last Claude APPROVE on `088305d` (issuecomment-5484024219):
- Independent 6-agent review at `ff6101fb` (issuecomment-5484247003) found H1/H2/H3 + M1/M2 + cheap lows. Claude's prior pass did **not** see those; they landed in `7df1dc1b`.
- Earlier follow-ups still in the delta: `cce48503` (stdio-bridge `imagePullPolicy`), `718b4bd3` (recorded fixture + live-only sweeps + writer-seeded status skip-log), `088305db` (per-call-site inventory + ConfigMap lock), `ff6101fb` (merge of `dev` / #525 Codex only — not HCC product).

CI on this HEAD is **green** (37/37). Re-check `Build / test / typecheck (host-context-controller)` and `Real Postgres integration (Postgres 16)` on **this** SHA before you treat CI as green.

HCC image hook is `0.1.283` on this HEAD.

## Product contract (must hold on this HEAD)

1. `deploymentMatchesDesired` is the single shared Deployment no-op predicate in `utils.ts`. Fail-open (`false`) when either `spec` is missing or compare throws.
2. Safe-strip lemma only: delete field X iff X === documented default, applied identically on both sides. The strips that keep builders from leaving the gate inert are `timeoutSeconds===1`, `failureThreshold===3`, `serviceAccountName==='default'`, and empty `container.resources === {}` (kube always emits the key; mirrors empty `securityContext`). **Do not** add `periodSeconds===10`. **Do not** strip `imagePullPolicy` (Always vs IfNotPresent is not a single default). **Do not** canonicalize Kubernetes quantities in the normalizer (`1000m` vs `1` must stay unequal).
3. One `canonicalizeValue` on the no-op path. Do **not** unify it with the `networkPolicyReconciler` walker that keeps `undefined` (#473 / #299).
4. `writeSkipsTotal.inc({ kind: next.kind ?? 'unknown' })` runs **only** when `isUpToDate` is true **and** `mutationAllowed()` (if set) is true, before return. Label set is `{kind}` only — `policy_type` belongs to #526.
5. `writesTotal.inc` still runs **only after** a successful `replace()`.
6. GFS (`gfsK8sApi.ts`) and `statelessLifecycleExecutor.ts` stay **ungated**. Inventory: 8 `replaceNamespacedDeployment(` under `src/` excluding tests (5 gated / 3 exempt) plus 1 gated `replaceNamespacedConfigMap(` in `reconciler.ts`. Identifier count must equal call count per file (alias/`bind` hole).
7. LlmHook private NetworkPolicy helper must **not** move onto shared `applyNetworkPolicy` (it has `mergeExisting: preserveObjectAnnotations`; the shared helper does not). Do **not** add `validateExisting` there.
8. nginx ConfigMap goes through `replaceWithConflictRetry` + `configMapMatchesDesired` (whole object after server-owned metadata strip) + `preserveObjectAnnotations` + `mutationAllowed: isCurrent`.
9. `writeStatusCondition` returns boolean; log `Updated status conditions` only when a status patch happened.
10. `validateExisting` stays **before** `isUpToDate`. Channel-reader retry-time ownership veto lives on `validateExisting` and **throws** (fail-loud). It must **not** return `true` from `isUpToDate` (that path inflated `write_skips_total`). Do **not** put `ensureHostRole` on the helper.

## Independent-review findings — verdict required per ID

Source: issuecomment-5484247003 at `ff6101fb`. Say **CLOSED / PARTIAL / OPEN / ACCEPTED OUT OF SCOPE** with file:line. Do not LGTM globally. If a finding is only partly landed, say what is still missing.

- [ ] **H2** `overlayDefined` returned overlay arrays by reference so `asApiserverDeployment(desired)` mutated caller `desired`. Must clone overlay leaves / replace-keys / arrays. Fixture-1 must prove the caller object is unchanged.
- [ ] **H1** Live kube always emits container `resources` (often `{}`). Builders omit it; the gate was inert for McpServer + SFS. Must strip empty `resources` in `normalizeContainerDefaults` (lemma-safe, mirrors empty `securityContext`). Fixture must fill omitted `resources: {}`. `NOOP-MCPDEP-*` / `NOOP-SFS-*` must still skip; `IMAGE-*` must still write.
- [ ] **H3** Desktop default `CONTEXT_MAPPER_DESKTOP_RESOURCES_LIMIT_CPU` authored as `1000m` (apiserver stores `1`). Cheap product fix is default `1` + README + a test that `1000m !== 1`. General quantity canonicalization (`1024Mi` vs `1Gi`) is **#530**, not this PR — do not add it.
- [ ] **M1** Channel-reader ownership veto must not increment `write_skips_total`. Veto on `validateExisting` (throw). Initial foreign-owner read may still skip before the helper. Retry-time ownership change must fail loud and leave `replace` uncalled on the second attempt.
- [ ] **M2** Inventory must fail if a file mentions `replaceNamespacedDeployment` without a matching call (`bind` / alias). `identCount === callCount` per file; files with zero calls still assert `identCount === 0`.
- [ ] **L1** Helper read-404 must `console.warn` (no longer silent success). Must not increment writes or skips.
- [ ] **L2** Docstrings on `deploymentMatchesDesired` / `configMapMatchesDesired` state that labels / finalizers / ownerReferences are compared in full.
- [ ] **L3** `writeSkipsTotal.inc` must not run when `mutationAllowed()` is false (SKIP-3).
- [ ] **L7** Comment documents that the deprecated `serviceAccount` alias is deleted **before** stripping `serviceAccountName === 'default'`.
- [ ] **L9** `copy-mcp-app` / `mcp-server` / `stdio-bridge` author `imagePullPolicy` from `config.mcpServerImagePullPolicy` only (dead `server.spec.imagePullPolicy ||` removed; sanitize already deletes the CRD field).
- [ ] **L12 / L13** PR body matches this HEAD: 8 Deployment call-sites across 6 files (5 gated / 3 exempt) + 1 ConfigMap; HCC `0.1.283`; 2010 HCC tests.

## zach88 findings — confirm still CLOSED

issuecomment-5483500631. Say **CLOSED / PARTIAL / OPEN** with file:line.

- [ ] **R1-H1** stdio-bridge authors `imagePullPolicy`; stdio NOOP/IMAGE still hold after H1/H2.
- [ ] **R1-M1** per-call-site inventory + `replaceNamespacedDeployment(` (not only `({`).
- [ ] **R1-M2** live-only spec-leaf sweeps still run and still detect after the fixture clone.
- [ ] **R1-L1** skip-log test is still writer-seeded (no hand-built `lastTransitionTime`).

## Prior Claude + local /review checklist

Validate each against **this** HEAD. Say **landed / still open / accepted out of scope** with file:line.

- [ ] Shared predicate + safe-strips (including empty `resources`): Host/channel-reader still no-op vs fixture; McpServer/LlmHook/SFS gates are not inert.
- [ ] `asApiserverDeployment` does not mutate caller `desired`; recorded blob still fills omitted `imagePullPolicy` and `periodSeconds=10` (fields the normalizer does **not** strip).
- [ ] Leaf sweeps: every **desired** leaf and every **live-only spec** leaf is detectable (Host, channel-reader, McpServer HTTP + stdio, SFS, LlmHook). After H2 the live-only sweep must actually exercise the leaves that used to be mutated into `desired`.
- [ ] `NOOP-*` prove `replace` is not called; `IMAGE-*` / `ROTATE-*` / `CM-2` write once.
- [ ] SFS-3: live `restartedAt` + identical desired skips; real image change keeps the marker.
- [ ] LlmHook NP: three callers skip; added egress writes once. Still **not** on shared `applyNetworkPolicy`.
- [ ] SKIP-1 / SKIP-2 / SKIP-3: skip increments `writeSkipsTotal`; write, read-404, and stale `mutationAllowed` do not.
- [ ] Inventory tripwire: 8 Deployment calls, gated+exempt lists, per-call-site gate, ident===call, plus 1 ConfigMap `isUpToDate: configMapMatchesDesired`.
- [ ] Security: strips cannot hide non-default SA / probe / NP / nginx `data` / credentials-revision drift. Empty-`resources` strip cannot hide a non-empty request/limit. Metric label stays `{kind}` only.

## Accepted out of scope (confirm still untouched — #530)

- General quantity canonicalization in the normalizer.
- `policy_type` / writer attribution on skip metrics (#526 / #530 L4).
- Sweep holes for `null` / `undefined` / `{}` / live-only `metadata.*` (#530 L5).
- `canonicalizeValue` collapsing `Date` to `{}` (#530 L6).
- Status-log under-report after a post-commit transport error (#530 L8).
- Foreign-annotation skip with no audit line (#530 L10).
- Empty `CONTEXT_MAPPER_MCPSERVER_IMAGE_PULL_POLICY=""` (#530 L11).
- Live apiserver §8 fixture diff (no Minikube).
- GFS Deployment gates; moving LlmHook NP onto shared `applyNetworkPolicy`.

## Additional findings

After the per-ID audit, say explicitly whether you found **anything new** on this HEAD that is not already CLOSED or filed in #530: a silent skip of authored drift, a lemma violation, an inert gate class, a metric lie, or a contract break.

Do **not** treat passing unit tests as proof that `asApiserverDeployment` equals a live apiserver object. Call that residual risk, not a merge blocker, unless you can show a silent skip of authored drift.

Verdict last: **APPROVE** only if every in-scope finding is CLOSED and you found no new merge-blocking defect. Otherwise **REQUEST CHANGES** and list the IDs. Do not merge.

zach88:
Took another pass through this for round 2, focused on whether the round-1 findings actually closed and whether the fixes broke anything on the way. Good news: they did close, and nothing regressed. The production logic is clean, security is solid, and the collateral audit came back empty. The only thing left is a test-methodology residue that's honestly more of a follow-up than a blocker.

**🟡 Approve with nits**

All four round-1 findings (H1 / M1 / M2 / L1) are closed and verified. This is my read from the review, not a formal GH approval.

None. No blockers, no regressions, no open High.

## Optional (non-blocking)

The PR can merge without this — it's a follow-up, your call.

1. **[R2-M1] Container-level default synthesis in the fixture is still hand-written (partial F2 of R1-H1).** `src/__tests__/asApiserverDeployment.ts:162-213` — the recorded blob (`RECORDED_APPSV1_DEPLOYMENT`) only supplies the top-level Deployment/PodSpec shape; its containers get replaced by `desired`'s and re-defaulted by hand in `applyContainerDefaults`, which is a hand-written twin of `normalizeContainerDefaults` (`utils.ts:410`). R1-H1's fix added `imagePullPolicy` to that list, so the instance is closed — but the class isn't: a future apiserver-added container default that neither list models (exactly the imagePullPolicy shape) would be invisible again, because the fixture won't add it and the test's `existing` won't have it. Honest mitigation: kube container defaulting is content-dependent (imagePullPolicy depends on the tag), so a recorded blob genuinely can't cover it — which by T1 is itself the finding. There's also a partial safety net: the two lists are twins, so if they drift the base no-op test goes red. Suggestion for whenever this gets picked up: anchor the fixture-add ↔ normalizer-strip pair to a real `kubectl get deploy -o json`, and add a test that diffs the synthesized container defaults against a recorded container so a new apiserver default can't fall out of both lists at once. Low priority — good candidate for a tracked issue rather than another round here.

## What checked out clean

- **Round-1 findings all closed and verified:**
  - **[R1-H1]** stdio-bridge now authors `imagePullPolicy: config.mcpServerImagePullPolicy` (`reconciler.ts:1371`), on par with its siblings, and the fix was generalized (F2) across every gated Deployment builder. A `stdio` transport case was added. Verified with a red-against-parent run: copying the new test over `173a813b` production yields 4 real assertion failures.
  - **[R1-M1]** the inventory test now binds each `replaceNamespacedDeployment(` call-site to its gate (plus an `identCount === sites.length` guard for the alias/bind form, and a new ConfigMap call-site lock) — a second ungated writer in an already-gated file now fails.
  - **[R1-M2]** live-only sweep coverage added across all 6 builders, so fields the apiserver *adds* (not just leaves present in `desired`) are now detectable.
  - **[R1-L1]** the status-conditions fixture is now derived from the real producer instead of hand-typed.
- **Security: clean (reproduced).** The no-op gates can't false-positive over security-relevant drift — NetworkPolicy egress/ingress compared verbatim (widened egress or dropped `Egress` policyType forces a write), ConfigMap `.data`/`.binaryData` compared in full (injected key or mutated `Authorization` forces a write), the `resources === {}` strip can't hide a real `resources.limits` disappearing, container `securityContext` isn't stripped (`privileged:true` forces a write). `imagePullPolicy` is now controller-config-only, dropping CRD-spec influence — a net attack-surface reduction. No fail-open-to-skip anywhere.
- **No regressions.** Dropping `server.spec.imagePullPolicy ||` from `reconciler.ts:1273/1334` is a no-op — `sanitizeCrdSpec` already deletes that field before containers are built, so the `||` branch was dead. The `resources === {}` strip only equates `{}`↔absent (identical in k8s), and `config.ts:719` `1000m`→`1` is just the apiserver's canonical form.
- **Suite + typecheck green:** 97 files / 2010 tests pass, `tsc --noEmit` exit 0.

zach88 (approved):

PR #279, opened 2026-08-06
feat: add Microsoft Teams organization integration

josue-tech-94:
## Summary

Adds a Microsoft Teams organization integration backed by Microsoft Entra ID. Administrators can connect a Microsoft tenant, import Teams and members, map access, and enable Microsoft sign-in for invited users across Profile UI and Desktop App.

## What changed

- Added persistent identity-provider connections, OAuth state, login codes, identity links, team mappings, setup sessions, and invitation access records.
- Added Microsoft OAuth authorization-code flow with PKCE, one-time hashed state, encrypted client secrets and refresh tokens, bounded login-code exchange, and strict return URL validation.
- Added resumable, concurrency-safe Teams and member imports with team mapping, access reconciliation, invitation delivery, and existing-member identity linking.
- Added a nine-step Microsoft Teams setup wizard and integration management pages under canonical Settings routes.
- Added Microsoft sign-in to Profile UI invitations, Profile UI login, and Desktop App login.
- Added validated Electron deep-link handling and restricted Microsoft authorization URLs to the official HTTPS login origin.
- Added a dedicated auth-proxy service for public OAuth callbacks, including service-token authentication, response limits, timeouts, non-root deployment settings, and default-deny NetworkPolicies.
- Added profile-aware minikube callback configuration so branch-scoped environments use their generated auth-proxy port.
- Added auth-proxy image publishing, public CI coverage, local build support, port-forwarding, and deployment manifests.
- Added tests covering OAuth validation, setup recovery, import serialization, invitation-backed identity linking, callback routing, team-name concurrency, desktop URL validation, and deployment contracts.

## Security

- OAuth state, PKCE verifiers, refresh tokens, client secrets, and login session codes are protected at rest or stored only as hashes where appropriate.
- OAuth state, login codes, and client flow bindings are single-use and fail closed.
- Web and desktop return URLs are allowlisted.
- Public callbacks traverse a dedicated least-privilege proxy and an authenticated internal route.
- Concurrent imports use renewable token-scoped leases and database uniqueness constraints.
- The auth-proxy runs without a service-account token, writable root filesystem, or Linux capabilities.

## Validation

- All affected package builds pass.
- Full Control API, Control UI, Desktop App, Auth Proxy, External REST API, and Profile UI tests pass.
- Public image-manifest, style-rule, shell-syntax, and base deployment rendering checks pass.
- Correctness and security reviews report no remaining actionable findings.

PR #531, opened 2026-08-31
feat(frontend): standardize table and list views

josue-tech-94:
## Summary

Standardizes the complete current Control UI and Profile UI table/list inventory on one
repository-local frontend component system while preserving app-owned data, routes, permissions,
and specialized interaction semantics.

The reconciled production inventory contains 71 Control UI and 7 Profile UI table/table-like
surfaces (78 total), plus CU-T56 tracked separately as non-table policy/form content. Every live
table/table-like surface now has a final disposition. The migration removes obsolete table systems
in the same change so `dev` does not retain competing contracts.

## User-visible behavior

- Primary record lists use a consistent title/description/toolbar, framed table, sticky header,
  body-only overflow, loading/empty/error states, sorting, row navigation, and far-right actions.
- Control UI toolbars order secondary CTAs, search, refresh, then the main CTA. Controls stay on one
  line when their combined minimum widths fit and wrap in DOM/focus order otherwise.
- Shared empty states default to `No data` and remain overridable with domain-specific copy.
- Durable record rows navigate to existing detail roots; interactive children do not accidentally
  activate the row. `View details` is first when a row also exposes a three-dot menu.
- Descriptions use bounded display with full-value hover/focus access where appropriate.
- Specialized permission, selection, hierarchy/file, and diagnostic views retain their domain
  behavior instead of being forced into ordinary record-table navigation.

## Final UI surface outcomes

| Area | Routes / access | Final landing result | Shared variant / behavior |
| --- | --- | --- | --- |
| External channels | `/external-channels` | Full-row hover/navigation and bounded far-right actions | Navigable record table |
| Agents | `/agents`, agent detail tabs | Stable record sorting, detail-root navigation, connector-count details including zero | Navigable records plus embedded access lists |
| Connectors | `/connectors` | Wider bounded descriptions, compact status columns, canonical actions, no obsolete metadata columns | Navigable record table |
| LLM models / discovery | `/llm-models`, `/llm-models/discovery` | Provider cells align logo/name; long catalogs retain body-only scroll and sticky headers | Standard and selection variants |
| Cost & usage | `/cost-and-usage/**` | Prices, budgets, and usage use shared states, sorting, and action placement | Standard and diagnostic variants |
| Secrets / LLM subscriptions | `/secrets/**` | Secrets and subscriptions use sibling routes, canonical states, and guarded row actions | Standard record tables |
| Plugins / workflow access / run details | `/plugins/**`, `/plugins/sdk` | Plugin records navigate normally; trigger members, trigger teams, and approval targets are sorted permission lists; run artifacts and workloads are shared diagnostic lists in producer/controller order | Navigable, permission, and diagnostic variants (CU-T77–CU-T81) |
| Marketplace / organization | `/marketplace/**` | Marketplace records use bounded descriptions; organization credentials use peer tabs and the common toolbar | Standard records and tabbed organization views |
| Users / teams / admins | `/users-and-teams/**` | Common table shell/actions, tab-level descriptions and CTAs, and conditional pending-invitation sections | Navigable and permission variants |
| Agent files / GFS | `/agent-files`; `/global-file-system` resource access dialog | Shared-filesystem records stay on `/agent-files`; global filesystem hierarchy and resource access stay on `/global-file-system`. Access grants/shares are principal-sorted permission records with guarded overflow revoke. | Shared-filesystem, hierarchy/file, and permission variants (CU-T57, CU-T61, CU-T82) |
| Traces / diagnostics | `/traces/**` | Diagnostic order, fields, and compact layouts remain explicit while using shared structure | Embedded/diagnostic variant |
| Profile UI members | Profile UI `/members`, member detail | Shared table/list structure with Profile-owned authorization and invitation behavior | Navigable and permission variants |
| Profile UI connected accounts / settings | Profile UI `/connected-accounts`, `/settings/**` | Repeated account/settings records use the shared system without copying Control UI components | Record-list and specialized settings variants |
| Shared table infrastructure | Both web apps | `@clerum/frontend-components` owns table/list primitives, viewport, sorting, truncation, states, and menus; all-disabled action menus disable and visually mute their trigger | 68 migrated viewport instances; zero direct app ownership |
| Removed Context / Resource surfaces | Historical `/contexts/**` and Resource views | Removed from active product topology; the current compatibility route remains where required by `dev` | No retired table family is revived |

## Shared architecture

- Adds private package `packages/frontend-components` / `@clerum/frontend-components` with
  `DataViewHeader`, `TableSearch`, `TableViewport`, `DataTable`, semantic row/cell/header/state
  primitives, `TruncatedText`, `RecordList`, `RowActionMenu`, and stable sorting utilities.
- `TableViewport` is the only application boundary for standardized table viewports. All 68 former
  direct application wrappers now use it, and style rules reject direct ownership in Control UI or
  Profile UI production markup and CSS.
- Uses app-owned CSS-variable bindings so Control UI and Profile UI retain distinct typography,
  themes, routes, data, permission, and domain contracts.
- Keeps narrow app adapters only where stable imports remain useful; they delegate to the shared
  package instead of maintaining parallel implementations.
- Installs the source-linked package explicitly in clean consumer CI and production image builds.

## Specialized variants

- **Standard record:** semantic rows, stable sorting, canonical states, and far-right actions.
- **Navigable record:** pointer/Enter/Space row navigation with interactive-child isolation.
- **Selection/permission:** checkbox, bulk-selection, editable permission, disabled, and revoke
  semantics without forced row navigation.
- **Hierarchy/file:** nested file/resource relationships and domain-specific disclosure.
- **Embedded/diagnostic:** chronology or producer/controller order, compact facts, and bounded
  diagnostic layouts where ordinary record behavior is inappropriate.

Ordinary record-detail expansion is removed where a durable detail route exists. File/tree and
specialized diagnostic disclosure remains because hierarchy is part of those interactions.

## Cleanup

- Removes zero-consumer `GrantsPanel` and Profile `EditableList` predecessors; `ProfileAdminPanel` was already absent at the merge base.
- Removes the retired Control `RowActions` mirror, app-local `tableSort`, obsolete LLM sub-tab
  wrapper, nested-tab styling, and superseded table/expansion CSS families.
- Retires the stale Context-specific MCP test fixture seam that produced CodeQL #1277 while
  preserving current agent-access, empty-state, binding, dialog, focus, and legacy-suppression tests.
- Replaces brittle production-path style-rule manifests with deterministic fixtures plus live normal
  and strict repository scans.

## API/backend work

Session Replay accepts only `oldest` / `latest` ordering at the authenticated Control API route. The
producer maps that allowlist to SQL direction, binds filters, anchors, and limits, includes order in
the query hash, and uses `last_occurred_at`, `host_ref`, and `session_id` as the deterministic cursor
continuation boundary. No other table adds a backend sorting API; fully loaded collections sort
locally and specialized producer-defined order remains explicit.

## Agent guidance

The provider-neutral frontend and app rules define the canonical viewport boundary, table/list
variants, sorting, navigation/action, state, description, tab, toolbar, and cleanup contracts.
Codex, Claude Code, Copilot, and Cursor files remain discovery adapters rather than semantic copies.

## Open PR coordination

No other PR was merged, cherry-picked, or otherwise integrated into this branch. Context removal is
already incorporated in current `dev`. Current known open overlaps are:

- **#252:** after synchronization, preserve its secret-row compare-and-swap deletion guard,
  disabled states, payload, and tests inside the canonical overflow-action contract.
- **#279:** after this standard is accepted, migrate its Microsoft import selection/review tables
  and integration list to the specialized selection/hierarchy and standard record contracts without
  losing select-all, editable-cell, listbox, or nested-review semantics.

### Final local validation

- `packages/frontend-components`: `npm test`; `npm run build` — passed.
- `control-ui`: `npm test` — 178 files / 1,933 tests passed; `npm run build` — passed with 60 routes.
- `profile-ui`: `npm test` — 75 Node tests plus 25 Vitest tests passed; `npm run build` — passed with
  12 routes.
- Focused `FrontendComponents`, `WorkflowAccessPanel`, `RecipeStatusContent`, `GfsGrantPanel`,
  `GfsBrowser`, and `McpServerTable` suites — passed.
- `node --test scripts/style-rules/rules.test.mjs` — 8/8 passed.
- `npm run style-rules`; `npm run style-rules -- --strict` — passed.
- Targeted Prettier over the completion range, release-version validation, commitlint, and
  `git diff --check` — passed.
- Canonical-reference resolution and guidance-graph acyclicity — passed.
- Independent production searches found zero raw app tables, direct app viewport-class ownership,
  stale production manifests, obsolete Context helpers, or `GrantsPanel` consumers.
- Independent code review: SAFE. Independent security review: SAFE. Completion-range collateral
  audit: CLEAN.

Authenticated route-level browser QA was unavailable for this completion pass. No browser result is
claimed; the route-by-route manual guide is recorded in Spec 009 for owner verification.

### GitHub CI / CodeQL

All 37 reported GitHub checks passed at exact head
`64a3bcc005275b2dcb6063b55e982309650c7f82`, including CI and CodeQL.

Code scanning reports zero open alerts for `refs/pull/531/merge`. Historical CodeQL #1277 remains
fixed; no alert was dismissed or suppressed.

## PR-size justification

This exceeds the usual file-count and one-service soft targets. The owner approved one complete
migration because shared APIs, every current consumer, interaction behavior, removal of parallel
systems, enforcement, tests, and guidance must agree in one landing state.

## Risks and reviewer focus

- Review keyboard/pointer isolation, action-menu focus and disabled states, description
  truncation/full-value access, sticky-header/body scrolling, and narrow-width behavior.
- Review permission/hierarchy/diagnostic exceptions for preserved ordering, data, confirmations,
  error handling, and authorization.
- Review Session Replay ordering/cursor continuation as the only backend sort contract.
- Authenticated route-level visual QA remains the principal manual-review gap.

## Checklist

- [x] All 78 current live table/table-like surfaces have a final disposition, with CU-T56 tracked separately as non-table policy/form content.
- [x] CU-T77–CU-T82 use their approved shared variants.
- [x] Every standardized table viewport uses `TableViewport`.
- [x] Raw tables, retired patterns, and direct viewport ownership are mechanically guarded.
- [x] Full shared-package, Control UI, Profile UI, style-rule, build, and static gates pass locally.
- [x] Independent code/security reviews are SAFE and collateral review is CLEAN.
- [x] This is not a new third-party MCP server or WorkflowRecipe.

## Follow-up

- Complete the route-by-route manual QA recorded in Spec 009 as desired.
- Formal merge-readiness review has completed through Round 2; R2-M1 was the sole remaining
  body-contract correction.
- Authors of #252 and #279 should synchronize after the canonical system is accepted and apply the
  adaptations above.
- Existing transitive application dependency advisories predate this branch and remain separate
  dependency-maintenance work; the new shared package reports no production vulnerabilities.


josue-tech-94:
@claude Please do a fresh full review of the current PR head against dev.

Please review the PR as it would land now, not the historical iterations. Treat the current PR description as the intended final contract and verify it against the implementation.

Pay particular attention to correctness, security, regressions, accessibility, test quality, and whether the shared table/list standard is applied consistently across Control UI and Profile UI.

The latest correction round specifically addressed:

user/team agent-access revocation correctly reconciling the underlying Context-derived authorization without removing unrelated/shared grants;
stale Event Explorer pagination responses after sort/filter/query changes;
readonly workflow trigger-member, trigger-team, and approval-target lists using the same shared ordered record-list contract;
retired-pattern enforcement for removed table/access infrastructure;
final table inventory/counts and route documentation;
meaningful regression proof for the Control UI diagnostic typography correction;
stronger ordering tests for workflow access, artifacts, and GFS grants;
valid ARIA list / listitem ownership in affected Control/Profile states.

Please independently verify those fixes rather than assuming they are correct.

Also check for collateral regressions caused by the fixes, especially:

authorization or access that remains after a visible revoke;
accidental removal of unrelated/shared access;
pagination cursor/order corruption;
sorting or tie-break drift;
row actions or navigation changing behavior;
lost diagnostic or permission data;
invalid accessibility semantics;
false-green tests;
stale or inaccurate PR-body claims.

Please distinguish PR-caused findings from pre-existing issues. For each actionable finding, include severity, exact file/location, reproduction/scenario, impact, and the smallest complete fix.

Avoid unrelated cleanup or stylistic suggestions unless they affect correctness, security, accessibility, or the stated PR contract.

If you find no merge-blocking issues, please say that explicitly.

zach88:
Took a deep pass through this one across all the touched surfaces (the shared package, the migrated tables/pages, the permission panels, the Session Replay backend change, security, and the Docker/CI packaging). Overall this is a solid, well-tested migration and the high-risk parts (backend ordering, secret handling, private-package packaging) all check out. The stuff worth stopping on is concentrated in two places: the profile-admin agent-access model, and determinism in the shared sort helper.

**🔴 Not ready to approve**
- Must resolve before approval: 2
- Optional / non-blocking: 11

1. **[R1-H1] Agents tab stopped rendering legacy scope-only grants (possible silent revoke)** — `control-ui/app/profile-admin/users/[userId]/page.tsx:1116` and `control-ui/app/profile-admin/teams/[teamId]/page.tsx:975`. The list used to render `grantedAgentNames`, the union of `assignedAgentNames` + `agentNamesForContextAccess(assignedContextIds, ...)`. The deleted comment spelled out why: "unioned with agents resolved from legacy scope-only mappings so nothing granted before this change disappears." Now it renders `assignedAgentNames` only. A member/team that can reach an agent solely through a legacy scope grant to an owned context (no entry in the agent mapping) drops out of the UI, and on the next save of any other agent `planAgentAccessUpdate` neither preserves nor derives that owned context, so the grant gets revoked without the admin ever seeing it. This looks deliberate (docstring on `planAgentAccessUpdate`, the r1-h1 fix, the compat suite), but it inverts a documented invariant and the suite only covers the write path (the `updateContexts` args), nothing asserts a legacy scope-only grant stays visible and revocable (T5). The real question: are legacy scope-only grants guaranteed already migrated into the agent mapping? If yes, this is fine, just drop the now-dead `agentNamesForContextAccess` export (`control-ui/lib/agentAccessCompatibility.ts:50`) and add a note. If not, please restore the union in the display (or add a test proving such a grant survives and can be revoked).

2. **[R1-H2] `stableSortRows` isn't deterministic for null/null pairs** — `packages/frontend-components/src/index.tsx:557`. `if (leftValue == null) return rightValue == null ? 0 : 1` returns `0` for null==null before reaching the `identity` tiebreak, so two rows with an empty field fall back to input order instead of a stable cross-render order. Repro (ran it, failed as expected): same two ids with `value: null`, input `[a,b]` yields `['a','b']` and input `[b,a]` yields `['b','a']`; with `value:'X'` both yield `['a','b']`. Visible impact is bounded (rows with an empty field visibly reshuffle on refetch, no data loss), but it's a real correctness bug in the piece ~60 consumers share. Fix: don't short-circuit to `0` on null==null, fall through to the `identity` + `index` tiebreak. Please close [R1-M2] with it (see below) since the current property test can't catch this.

## Optional (non-blocking)

The PR can merge without these; author's call / follow-up.

1. **[R1-M1] `RowActionsMenu` adapter drops `horizontalTrigger`** — `control-ui/components/RowActionsMenu/index.tsx:20`. The adapter destructures only `{ ariaLabel, actions }`, but `RowActionsMenuProps` (`control-ui/components/RowActionsMenu/types.ts:23`) still advertises `horizontalTrigger` and ~20 callers pass it (HostTable, SecretsTable, LlmPriceTable, RegistryApiKeysPanel, OwnedEntries, ProfileAdminHome, ...). It compiles (optional prop, silently ignored) so the horizontal-dots variant just disappears. Either honor it in the shared `RowActionMenu`, or delete the prop from the type and the ~20 call sites in this PR so TS flags the dead prop.

2. **[R1-M2] Property test mirrors the H2 bug, so it can't catch it** — `control-ui/components/__tests__/FrontendComponents.properties.test.ts:32`. The `rowOrder` oracle returns `0` for null==null just like the SUT, so "deterministic total order" passes with the bug present (verified: the suite stays green with H2 in). The "exact ties" sub-test only checks stability for one input order, not determinism across orders. When you fix H2, add a property that sorts the same multiset under two input permutations and asserts identical output, without mirroring the SUT in the oracle.

3. **[R1-M3] HostAccessTab: undeclared scope creep (grant/revoke now write Contexts)** — `control-ui/components/HostAccessTab/index.tsx:139`. The diff wires `applyAgentAccessCompatibilityUpdate` into all four grant/revoke paths, adding Context writes (`updateAdminUserContexts`/`updateAdminTeamContexts`) plus two new fetches (`getContexts`+`getHosts`) that didn't happen before (base doesn't call the helper). The wiring mirrors the pattern already in `users/[userId]/page.tsx` and is covered by payload-level tests, so it's correct, but it's an authorization-semantics change inside a PR whose body says "table standardization" (body/impl out of sync), and it widens the failure surface: a failing `getContexts()`/`getHosts()` now aborts a grant/revoke that previously only touched agents. Worth calling out in the body (or splitting), and confirming that abort-on-fetch-failure is intended.

4. **[R1-M4] Lost the "why is this disabled" tooltips + inert kebab** — `packages/frontend-components/src/index.tsx:367`. The `RowAction` type has no `title`/`tooltip`, so the migration to `RowActionMenu` dropped the `title=` strings that explained disabled reasons ("Only leaders can edit member permissions", `memberDeleteTooltip(...)`, "Current admin cannot be deleted"). And when every action in a row is disabled, `triggerDisabled = actions.every(...)` disables the whole three-dot button, so the user can't even open the menu to read why. The disable logic itself is correct, this is a UX/a11y regression. Fix: add `title?: string` (or `disabledReason`) to `RowAction`, render it on the menuitem, and repopulate the tooltips. Hits profile-ui members/settings and control-ui alike.

5. **[R1-M5] ControlAdminsPanel: "current admin cannot be deleted" guard changed shape with no test** — `control-ui/components/ControlAdminsPanel/index.tsx:572`. The old disabled button + delete button merged into one menu action with `disabled: currentAdmin || deletingAdminId === admin.id`. The guard is preserved in code (`handleDeleteAdmin` has no internal guard, the `disabled` is the only protection), but nothing asserts Delete stays disabled for the current admin. Since the guard's shape changed here, please make the invariant executable: open the current admin's menu, assert the menuitem is disabled and that `deleteControlAdmin` isn't called.

6. **[R1-L1] OwnedEntries dropped version collapsing** — `control-ui/components/PublisherView/OwnedEntries.tsx:310`. Base collapsed same-name versions into one expandable row ("+2 more"); now every version is a first-class row (test rewritten to match, so it's deliberate). It does contradict the body's "specialized views retain their domain behavior" line. Just confirm it's intended and align the body.

7. **[R1-L2] Pagination assertion is tautological** — `control-api/test/services.governedSessionReplayRepository.test.ts:246`. Hand-written fixtures with `query` mocked, so the merged list just replays the fixture's pre-baked order (T1/T4). The real boundary coverage lives in the integration test (gated behind `describeRealPostgres`), which is correct, so this doesn't block. Worth a comment that this test covers the emitted SQL, not pagination correctness.

8. **[R1-L3] Old cursors 400 during deploy rollover** — `control-api/src/services/tracing/governedSessionReplayService.ts:61`. The query hash went from `{filters}` to `{filters, order}`, so a cursor minted by the base head won't validate against the new one (UI shows an error on paging). Transient (ephemeral cursors, only the deploy window) and standard, flagging for completeness.

9. **[R1-L4] RegistryApiKeysPanel: spurious create-modal reopen** — `control-ui/components/RegistryApiKeysPanel.tsx:175`. `if (createSignal > 0 && view.kind === 'ready') setCreating(true)` reopens the modal if `view.kind` goes `error → ready` after a prior create. Low impact (only fires with an intermediate error). Prefer a consumed/edge signal over re-deriving from `view.kind`.

10. **[R1-L5] Ambiguous kebab aria-label for same-principal rows** — `control-ui/components/GfsGrantPanel.tsx:495` (also WorkflowAccessPanel, HostAccessTab). The kebab uses `ariaLabel="Actions for {label}"`, so two rows with the same principal (a grant + a share for the same user) share an aria-label; the revoke button used to distinguish the kind. Minor a11y regression and makes `getByRole` ambiguous as fixtures grow.

11. **[R1-L6] RegistryCatalog default order changed visibly** — `control-ui/components/RegistryCatalog.tsx:47`. From `sortKey=null` (registry delivery order) to `'name'` asc. The catalog isn't k8s-backed so it does arrive differently, low risk and consistent with the standardization, but it is a visible change. (McpServerTable does the same but the API already returns by name, so no visible change there.)

Nits: dead imports in profile-ui (`IconTrash` in members/page.tsx and members/invite/page.tsx, `IconPencil`+`IconTrash` in settings/TelegramVerificationPanel.tsx, `Button` in connected-accounts/page.tsx); `scripts/style-rules/rules.test.mjs` uses `@evenfire/frontend-components` but the real package is `@clerum/frontend-components`; `useTableSort` re-sorts every render if callers pass inline `accessors`/`identity`, and `sortBy` ignores `defaultDirection` when switching columns.

## What checked out clean

- **Session Replay backend**: deterministic, stable cursor in both directions; the `>`/`<` comparison matches the SQL direction; `order` is in the query hash; strict allowlist (400, no interpolation, so no SQL injection); limit clamped 1..200; default `latest` matches old behavior (no regression for an old client). Tracing suites 9/9, `tsc --noEmit` clean.
- **Security (full diff): no new vulns.** No XSS (React escapes, zero `dangerouslySetInnerHTML` in the package/TruncatedText); revoke/delete/rotate guards preserved (permission-disabled renders a real `<button disabled>`, keyboard nav filters `:not(:disabled)`); no secret/token leakage (SecretsTable/RegistryApiKeys render metadata/prefix only); supply chain clean (local `file:` package, lockfile pins dev-deps only with npmjs integrity).
- **Private-package packaging (the high-risk part): correct.** `COPY packages/frontend-components` before `npm ci` with materialization in both frontends, `.dockerignore` doesn't exclude it, CI installs the package's deps before the consumer, and `build-publish.yml` + `deploy/images.json` include it in source paths. `test-dockerfile-local-packages.sh` passes, profile-ui `next build` clean.
- **Cleanup claims verified**: `GrantsPanel`, `EditableList`, the retired `RowActions` import, and `lib/tableSort` all removed with zero residual consumers.
- **Suites reproduced green**: shared package + primitives (14), tables + big pages (283 across 22 files), permission/revoke panels (~111), profile-ui (75 node + 25 vitest), style-rules 8/8 + run-all with no false positives.

On PR size: 170 files / 4 code services is well past the usual soft budget, and I see the body justifies it as one complete migration. Fine here, just noting it's what makes each review round expensive.

josue-tech-94:
@claude Please do one final fresh review of the current PR head and current PR description against dev.

The implementation has already gone through two formal merge-readiness rounds. The latest formal review found no remaining source-code must-fix issue; its only blocker was stale PR-body/check-state wording, which has now been corrected.

Please independently verify the current landing state, with particular attention to:

legacy Context-only Agent grants remain visible and preserved for both users and teams until explicitly revoked;
explicit revoke still removes only the correct compatibility-owned Context access and preserves shared/unowned access;
stableSortRows remains deterministic for nullish/nullish ties using the stable identity tie-break;
the corrected property tests actually prove that behavior rather than mirroring the implementation;
horizontal row-action trigger variants still work;
disabled row actions retain accessible explanations without weakening authorization guards;
current-admin self-delete remains impossible;
Registry API-key create signals are consumed once and do not reopen after an error → ready transition;
workflow workloads preserve producer/controller order;
no collateral regression was introduced by those corrections.

Also verify that the PR description is now factually accurate for the exact current head, especially the CI/CodeQL section and review/follow-up state.

Please distinguish PR-caused issues from pre-existing or optional hardening. Do not repeat previously closed findings unless they actually reproduce on the current head.

If you find an actionable issue, include severity, exact file/location or body section, concrete scenario/reproduction, impact, and the smallest complete correction.

If you find no merge-blocking issue, please say that explicitly.

zach88 (approved):

zach88:
Went through round 2 with a focus on what the fix commits actually closed and, more importantly, what they might have broken. Good news: this round is clean. All seven findings from round 1 that got a fix are verified closed by reproduction (the regression tests fail against the previous head `84e2c51` and pass at `64a3bcc`), and the collateral pass over the whole fix range came back with zero regressions. What's left open is all non-blocking, plus one Medium that's hit its re-triage threshold and needs a call from you.

**🟡 Approve with nits**
- Optional / non-blocking: **8**

Nothing. No open blockers, and the collateral audit found no regressions introduced by the fix commits (contract changes are all additive/optional, full control-ui suite stays green at 1933 tests).

## Optional (non-blocking)

The PR can merge without these; a couple need a design call from you rather than code.

1. **[R1-M3] HostAccessTab grant/revoke now writes Contexts (undeclared scope) - re-triage.** `control-ui/components/HostAccessTab/index.tsx:139`. Grant/revoke now also write Contexts and fire two new fetches (`getContexts` + `getHosts`) that didn't happen before, so a failed fetch now aborts an operation that previously only touched agents. This is the same compatibility mechanism the R1-H1 fix doubled down on across the PR, so it's more entrenched now. It's been open two rounds, so per our process it either escalates or gets closed with a written justification, not left floating. Your call: if writing Contexts on grant/revoke is the intended compatibility migration and abort-on-fetch-failure is acceptable, let's close it in writing and sync the PR body to say this PR changes authorization semantics (materializing context-only grants), not just "standardize tables". If abort-on-fetch-failure isn't acceptable, this becomes a blocker.

2. **[R1-L1] OwnedEntries dropped version collapsing.** `control-ui/components/PublisherView/OwnedEntries.tsx:310`. Each version is now a first-class row instead of the old "+2 more" collapse. Reproduced. It's deliberate but it contradicts the body line about specialized views retaining domain behavior. Just confirm intent and align the body.

3. **[R1-L6] RegistryCatalog default sort changed visibly.** `control-ui/components/RegistryCatalog.tsx:47`. Default went from registry delivery order (`sortKey=null`) to `name` asc. Since the catalog isn't k8s-backed, it arrives in a different order, so this is a real visible change. Low risk and consistent with the standardization, just confirm.

4. **[R1-L2] Pagination assertion is tautological.** `control-api/test/services.governedSessionReplayRepository.test.ts:246`. Hand-built fixtures with a mocked `query`, so the merged list just replays the fixture's pre-baked order (T1/T4). The real boundary coverage lives in the Real Postgres integration test, which passed in CI at head, so this doesn't block. Worth a comment noting this test covers the emitted SQL, not pagination correctness.

5. **[R1-L3] Stale cursors 400 during deploy rollover.** `control-api/src/services/tracing/governedSessionReplayService.ts:61`. The query hash now includes `order`, so a cursor minted by the base head won't validate against the new one and paging shows an error. Transient (ephemeral cursors, only the deploy window) and standard, flagging for completeness.

6. **[R1-L5] Ambiguous kebab aria-label between rows of the same principal.** `control-ui/components/GfsGrantPanel.tsx:495` (+ WorkflowAccessPanel, HostAccessTab). Two rows for the same principal (a grant + a share of the same user) now share `Actions for {label}`. Minor a11y regression and ambiguous for `getByRole` in tests.

7. **[R2-L1] First property test still has a circular oracle.** `control-ui/components/__tests__/FrontendComponents.properties.test.ts:32`. The `rowOrder` oracle re-implements the SUT's logic, so its pairwise assertion can't catch a bug shared by both. Not a real gap here because the concrete null/null case is covered by the new independent test (lines 67-100, verified failing against `84e2c51`). Future cleanup: assert invariants derived from the contract, not from the SUT.

8. **[R2-L2] profile-ui `disabledReason` wiring has no test of its own.** `profile-ui/app/members/[userId]/page.tsx:277` (+ `members/page.tsx`). The three menu items wire `disabledReason` onto the shared component, but there's no profile-ui test asserting the reason renders and the action stays inert. The shared behavior is covered in control-ui, so risk is low, it's just prop wiring over an already-tested component. A light test (open a no-permission member's menu, assert `aria-disabled` + reason + no-op) would close it.

## What checked out clean

Verified, not assumed:

- **The seven fixes, by reproduction.** R1-H1 (legacy scope-only agent grants restored in both display and write path, four T5 tests asserting visible-and-revocable), R1-H2 (stableSortRows falls through to identity tiebreak on null/null), R1-M1 (`horizontalTrigger` forwarded and honored), R1-M2 (new independent property test), R1-M4 (`disabledReason` additive, trigger focusable with a reason), R1-M5 (test asserts the current admin's Delete is `aria-disabled` and `deleteControlAdmin` isn't called), R1-L4 (create signal consumed by value, second legit create still reopens). Every new regression test fails against `84e2c51`.
- **No regressions.** Contract changes are additive/optional (`RowAction.disabledReason?`, `RowActionMenu.triggerVariant?='vertical'`, `RowActionMenuItem.disabledReason?`); no call site stops compiling or changes default. Full control-ui suite green (178 files / 1933 tests), profile-ui 25 tests green, shared package 16/16.
- **Security: SAFE for the fix range.** The agent-grants fix doesn't widen authorization (the union is the effective access the subject already had; revoke re-derives and drops the revoked agent's context). `disabledReason` renders as React-escaped text with static reasons or an already-visible display name, no injection or leak. Lockfiles only re-resolve `@clerum/frontend-components` as a local `file:` dep, no new third-party packages.
- **CI green at `64a3bcc`:** 37/37 checks including CodeQL (zero open alerts), Real Postgres integration, and the leak guards.

One domain note to keep conscious (not a bug, it's the R1-H1 fix design): revoking or adding any agent now promotes context-only grants to direct agent grants, which is exactly what prevents the silent revoke. It's codified and covered by the four new tests.

PR #532, opened 2026-09-01
feat(desktop): polish new conversation landing

miguelo-999:
## Summary

- Refresh the new-chat landing title and center the agent selector.
- Tune composer, recent-session, spacing, and visual hierarchy for the desktop home experience.
- Align the Files navigation icon with Control UI and update focused coverage.

## Testing

- Desktop suite (Node 24, `npm test` from `desktop-app`): 240 files / 2304 tests passed, including the new `test/browser/newChatComposerGutter.browser.test.ts` responsive coverage.
- R3-M1 is covered by an automated Chromium computed-geometry regression test (not a manual visual check): real-browser layout measurement at 1100px / 1220px / 1440px viewports asserts the new-chat composer keeps its `--space-3` horizontal inset (and the 1024px responsive cap) in both the full-screen chat route and the docked chat drawer (460px). Verified red on the pre-fix CSS, green after.
- `npm run verify:electron` passed in the author's Node 24 environment (Electron runtime resolved); the review environment could not independently rerun it because its Electron executable is missing.

## Scope note

The control-api real-Postgres test fix that briefly rode on this branch has been split into its own PR (#649, `fix/realpg-pool-error-guard`), so this PR is Desktop-only again.

## Checklist

- [ ] This is not a new third-party MCP server / WorkflowRecipe (those go to the registry).


miguelo-999:
@claude bro pls check this

miguelo-999:
@claude bro please review this PR, check the previous reviews and provide your own independent review if there's feedback sort it by P0, P1, etc and if everything is good provide explicit green light. This one is mostly cosmetic so keep that in mind.

miguelo-999:
Addressed the review feedback in 927b37a0:

- **Dead `IconContexts` import** — removed from `SidebarNav/index.tsx`.
- **`SESSION_PREVIEW_LIMIT` 9 → 5** — intentional; aligns the in-thread collapse threshold with `SIDEBAR_SESSION_PREVIEW_LIMIT = 5` so both previews match.
- **Broadened `.nav-link` selector** — intentional; Files row hover now matches the chat row styling. Eyeballed in the running app.

Validation: Node 24, `npm run verify:electron` OK, full desktop unit suite green (239 files / 2294 tests).

miguelo-999:
@claude fixed, please check and provide feedback or green light

josue-tech-94:
Round 3 found one issue that should be fixed before merge:

### `R3-M1` — New-chat composer loses its responsive gutter

**Files involved:**

* `desktop-app/ui/src/components/agents/AgentWorkspace.tsx`
* `desktop-app/ui/src/styles.css`
* responsive/browser coverage for the new-chat landing

`AgentWorkspace.tsx` now wraps the inline composer in `.agent-workspace-new-chat`. Because of that extra wrapper, the existing `<=1200px` CSS rule in `styles.css` no longer matches the composer as a direct child, so the new-chat composer can become full-width and lose its horizontal gutter on laptop/narrower layouts.

**What needs to be preserved:** at `<=1200px`, the new-chat composer should retain the same token-based horizontal inset it had before this change.

**Suggested fix — not prescriptive:** extend the existing breakpoint selector so it also covers something like:

`.agent-workspace-body-slot > .agent-workspace-new-chat > .composer-inline`

using the same existing token calculation, while keeping the current direct-child selector for the other composer path.

That selector is only a suggested implementation. If there is a cleaner solution that preserves the same responsive behavior and follows the existing desktop UI rules, use that instead.

**Test/coverage:** please add viewport/browser coverage at `<=1200px` that proves the new-chat composer keeps its horizontal inset. Also verify the result in both the full-screen and drawer chat layouts. This part is important because the existing unit tests do not exercise the responsive behavior.

No changes are currently requested for the other reviewed files. The Files icon/test alignment, landing title/display-name behavior, selector interaction, renderer build, and style-rule checks all looked good.

After the fix, please also update the PR `Testing` section with the validation actually run, including the responsive visual check. `verify:electron` can remain noted separately if the local Electron runtime issue still prevents that check.

miguelo-999:
@josue-tech-94 

The fix: added .agent-workspace-body-slot > .agent-workspace-new-chat > .composer-inline to that media rule's selector list (same token calc, direct-child selectors kept), with a comment explaining why the arm exists.

Coverage: new desktop-app/test/browser/newChatComposerGutter.browser.test.ts (real Chromium + real styles.css, following the existing responsiveUtilityGutter pattern) — 5 tests: sub-cap slot at 1100px (inset exactly --space-3 = 14px), the 1024px cap case, the exact 1220px boundary, the 460px docked drawer (same shared ChatPage DOM), and a wide-viewport guard that the 920px cap still owns above the breakpoint.

Verified red/green: stashing the CSS fix makes 4 tests fail with exactly the reported symptom (expected +0 to be close to 14); with it, 5/5 pass. Full desktop suite: 240 files / 2299 tests green, verify:electron OK (Node 24).

josue-tech-94:
The Desktop responsive issue from the previous round is fixed and verified. The new browser regression test goes red on the previous head and green on the current head, so no further Desktop landing changes are requested for `R3-M1`.

There are two remaining items:

### `R4-M1` — Must fix: Real-Postgres test hides unexpected pool errors

**File:**
`control-api/test/services.pluginWorkloadSdkFinalization.realPostgres.integration.test.ts`

The new pool error listener currently does:

`dbPool.on('error', () => {})`

for the duration of the integration suite. That suppresses every idle-client pool error, even though the intended exception appears to be only the expected PostgreSQL `57P01` shutdown error during teardown.

This creates a false-green risk: an unexpected connection failure during the test phase could be swallowed while later assertions continue and the lane still reports green.

**What needs to change:** only the expected teardown error should be ignored. Unexpected pool errors, especially while tests are running, need to remain observable and fail the lane.

**Suggested implementation — not prescriptive:** track whether teardown is in progress and suppress only `57P01` during that window; otherwise surface/rethrow/record the error through whatever mechanism best fits the existing test harness.

That exact mechanism is only a suggestion. Please use the cleanest pattern consistent with the existing control-api/Postgres tests.

**Required coverage:** prove both sides:

* expected teardown `57P01` remains clean;
* a non-`57P01` pool error during the test phase is surfaced and causes the lane to fail.

### `R4-L1` — Should fix: explain or split the control-api scope

The effective PR now spans both `desktop-app/**` and `control-api/**`.

Since the PR is primarily Desktop landing polish, please either:

* split the control-api package/test changes into a separate PR, which is preferred if they are independent; or
* add a concise explanation to the PR description for why the control-api change needs to ship together with the Desktop work.

This is primarily a PR-scope/reviewability issue, not a product defect.

### PR description updates

Please also update the Testing section:

* Desktop suite count: **240 files / 2304 tests**.
* Describe the responsive check as an **automated Chromium computed-geometry regression test**, rather than a manual visual check.
* After `R4-M1` is fixed, update the control-api/Postgres validation wording so it does not rely on “no unhandled errors” while the current listener suppresses all pool errors.
* Keep the `verify:electron` result as the author's successful Node 24 run; this review environment could not independently rerun it because its Electron executable is missing.

Everything else reviewed cleanly: the gutter fix preserves the token inset, existing direct-child behavior, and wide-layout cap; the UI-focused tests/build/style checks pass; and no new security or production API/persistence boundary issue was found.

miguelo-999:
All three items done:

R4-M1 — fixed in PR #649. Replaced the blanket dbPool.on('error', () => {}) with createPostgresPoolErrorGuard (control-api/test/poolErrorGuard.ts): it suppresses only 57P01 while teardown is in progress, records everything else, and the suite calls assertNoUnexpectedErrors() at the start of afterAll (test-phase errors fail the lane) and again after teardown. Both sides proven:
- test/poolErrorGuard.test.ts (6 tests): expected teardown 57P01 stays clean; non-57P01 in test phase, 57P01 outside the window, and non-57P01 during teardown are all recorded and rethrown; first error kept; real Pool wiring exercised via pool.emit('error', …).
- End-to-end red proof: a temporarily emitted synthetic ECONNRESET on the suite's pool during the test phase failed the file in afterAll with all 12 tests passing.
- Full realPostgres lane re-run with the guard active: 33 files / 155 tests green.

R4-L1 — split (preferred option). The control-api fix now lives in its own PR, #649 (fix/realpg-pool-error-guard, base dev), carrying the original 57P01 absorb plus the guard. PR #532's control-api delta was reverted — its file list is now Desktop-only, and a scope note in the body points to #649.

PR #532 Testing section updated: 240 files / 2304 tests (rerun on the final head), the responsive check described as an automated Chromium computed-geometry regression test, the control-api bullet removed (superseded by #649, whose wording describes the guard coverage rather than "no unhandled errors"), and verify:electron kept as the author's Node 24 run with the review-environment caveat.

@josue-tech-94

josue-tech-94 (approved):

PR #543, opened 2026-09-01
fix(profile-ui): default internal API to in-namespace Service DNS

jozer-rami:
## Why

Shared-trial tenant `jose-test` on smoke-8: profile-ui 500s on password set and `/api/v1/me` with

```
Failed to proxy http://external-rest-api.profiles.svc.cluster.local:8091/...
ENOTFOUND external-rest-api.profiles.svc.cluster.local
```

MCC already sets `EXTERNAL_REST_API_INTERNAL_URL` to `profiles-<slug>`. Next **bakes** `rewrites()` at image build, so the dedicated-cluster FQDN is frozen in. SSR invitation GET can still work (runtime env); browser POSTs through `/external-rest-api` do not.

## What

Default the rewrite + `serverApi.ts` to `http://external-rest-api:8091` (in-namespace Service name). That DNS works in dedicated `profiles` **and** shared `profiles-<slug>`. Runtime env still wins when Next evaluates it.

Landing this is not enough for `jose-test` until a new profile-ui image is built (dev `build-publish`) and pinned/rolled by MCC.

## Evidence

- `profile-ui` `npm test`: 58 node:test + 13 vitest passed, including new `internalApiUrl.test.ts`.

**Do not merge until Jose says go.**

### DIFF
--- a/profile-ui/app/constants/serverApi.ts
+++ b/profile-ui/app/constants/serverApi.ts
@@ -1,3 +1,7 @@
 export const EXTERNAL_REST_API_INTERNAL_URL =
   process.env.EXTERNAL_REST_API_INTERNAL_URL ||
-  'http://external-rest-api.profiles.svc.cluster.local:8091'
+  // In-namespace Service name: works in dedicated `profiles` and shared
+  // `profiles-<slug>`. Do not default to `.profiles.svc.cluster.local` —
+  // Next bakes this rewrite at image build, so that FQDN 500s on shared
+  // tenants (ENOTFOUND). Runtime env still wins when Next evaluates it.
+  'http://external-rest-api:8091'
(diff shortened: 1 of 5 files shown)


jozer-rami:
Closing: this is the wrong layer.

Evenfire OSS is a self-host product. Its default
`http://external-rest-api.profiles.svc.cluster.local:8091` is the documented
self-host topology (namespace `profiles`). MCC must not push shared-tenant
DNS into that default — anyone self-hosting should keep the OSS contract.

MCC already injects `EXTERNAL_REST_API_INTERNAL_URL` for `profiles-<slug>`,
but Next bakes rewrites into `.next/routes-manifest.json` at **image build**.
Live jose-test confirmed the baked destination is still
`http://external-rest-api.profiles.svc.cluster.local:8091/:path*` even when
the pod env is `profiles-jose-test`. Runtime env is ignored for `/external-rest-api/*`.

The MCC-side fix (same pattern as control-ui FIX B) is to remap that OSS
hostname inside the profile-ui pod via `hostAliases` to the tenant
`external-rest-api` Service ClusterIP. That stays in MCC.

Not merging.

PR #544, opened 2026-09-01
feat(wrc): record the resolved egress set per host so the WRC lane has a history (#299)

alfredolopez80:
The WRC egress lane leaves no record of what it resolved. HCC does, and the difference is why one lane can be reasoned about over weeks and the other only in the present.

> **This PR was reworked after an adversarial review of its first cut.** The review found that the original line did not deliver the reconstruction it argued for, and that its central justification was false. Both are fixed here, and the corrections are listed at the bottom rather than quietly dropped.

## The asymmetry

`host-context-controller` has emitted this per DNS binding since the initial commit (`5d663a49`, 2026-07-14):

```
[NetPol] Resolved api.github.com → [140.82.112.5/32, 140.82.112.6/32, …]
```

With retained logs, that makes its lane reconstructable: which addresses a host actually used over weeks, whether its universe is closed, whether it drifts.

`workflow-recipes` has had **its own resolution path since that same initial commit** (`fqdnResolver.ts`) and simply never got the equivalent record; #299 later added the sliding-window accumulator on top of it. So there is no difference in arrival to explain the gap — `wl-egress-*` and `ui-egress-*` are observable only in the present, through whatever the live annotation happens to hold, and nothing documents that as intentional.

## Why it matters now

Measured on `clerum-dev`, after the overlap raise landed: **every window whose DNS round exceeds its retention is in the WRC lane.** The half of the fleet that needs history is exactly the half that has none.

The gap is concrete. For the HCC lane, 28 days of logs answer the question directly:

| host | distinct IPs in 28 days | spread |
| --- | --- | --- |
| `api.github.com` | **6**, always the same 6 | all inside `140.82.112.0/22` |
| `mcp.stripe.com` | **8**, unchanged across every sampled window | 2 `/24` blocks |

That is a closed, stable universe — and knowing it changes the remedy, because a host whose addresses do not actually move needs a prefix, not a longer window.

The same question cannot be asked of `oauth2.googleapis.com`, which currently resolves to 13 addresses across 13 distinct `/24` blocks, because there is no series to ask it of. Whether that universe is closed or growing decides whether a curated catalog fixes it or merely postpones the problem.

## What this adds

One record per **declared FQDN**, emitted through the service's structured logger:

```json
{"level":"info","component":"wrc","recipeName":"acme","msg":"resolved external egress set",
 "policy":"ui-egress-acme","fqdn":"api.github.com",
 "cidrs":["140.82.112.4/32","140.82.113.4/32"],"ports":[443]}
```

Four decisions, each of which a test pins:

**Per FQDN, not per policy.** `egress.external[]` is a list, so one policy routinely covers several hosts. A flat union of their addresses cannot say which host used which — the only question the history is for. Keyed by `source.fqdn`, this matches HCC's shape.

**After the write, over the enforced set.** Emitted once `createOrReplace` / `applyNetworkPolicy` has landed, and given the post-M3 (`isBlockedExternalIPv4`) filtered set that was actually rendered. A record that disagrees with the policy would attest to an allowance that never existed — the worst failure mode for something whose purpose is post-hoc reconstruction.

**On change only.** `changed` is `acc.changed`: true iff the `(fqdn,ip,port,protocol)` set differs from the persisted state, so a timestamp-only refresh is silent. It is **one of the three terms** the H4 write gate keys on, so every record corresponds to a write that landed — the converse does not hold, since a renewal (audit M1) writes without one.

**An empty set is recorded, not suppressed.** A declared host that resolved to nothing emits `cidrs: []`. A set collapsing to empty is the most consequential event in the series and must not read as "nothing happened".

**Through `observability/logger.ts`**, the wrapper this service already has and which the repository logging standard names. That also gives the record a `LOG_LEVEL` switch and emits a JSON object rather than a string a consumer would have to regex.

## Verification

| | |
| --- | --- |
| `workflow-recipes` suite (`npm test`, the CI command) | **2533 passed**, 29 skipped, 0 failed, `EXIT=0` |
| `reconciler.test.ts` | **310 passed** (302 before, **+8**) |
| `npm run build` (`tsc -b`) and `npx tsc --noEmit` | both `EXIT=0` |

**Mutation matrix — 11 mutations, run against the full 310-test file. All 11 killed, zero survivors.** (The review of the first cut found 4 survivors out of 6.) The source was restored byte-identically after every mutation.

| mutation | killed by |
| --- | --- |
| suppress the emission entirely | 6 tests |
| remove the change gate | *does NOT record a renewal write* |
| drop the address-wise sort | *records the ui lane … exact enforced cidrs* |
| revert to a plain lexicographic sort | same |
| drop the dedup | *keeps a port-only change visible* |
| rename the `msg` field | 6 tests |
| flatten, ignoring per-FQDN grouping | *attributes each address to its own host* |
| skip a host that resolved to nothing | *still records a declared host that resolved to no addresses* |
| silence the workload lane | *records the workload lane too* |
| drop the `ports` field | 5 tests |
| record the **pre**-filter set | *never records an address the M3 filter kept OUT* |

The blocked-address test is the one worth reading: it rehydrates an RFC1918 entry through the state annotation — the only route by which one can reach the accumulator, since fresh DNS fails a whole host closed — and asserts it never reaches the record.

## Corrections to this PR's first cut

| claim | status |
| --- | --- |
| "workflow-recipes has no logging wrapper" | **False.** `src/observability/logger.ts` exists, imported by 13 files, 5 in `src/reconciler/`. Now used. |
| "WRC gained its own resolution path later, in #299 Phase 1" | **False.** `git ls-tree 5d663a49` shows `fqdnResolver.ts` in the same initial commit. Corrected above — it makes the case stronger, not weaker. |
| "the equivalent line" (HCC parity) | **Was not equivalent** — per-policy vs per-binding. Now per-FQDN. |
| "a logged line and a written policy stay in step" | **Overstated.** One-directional; the JSDoc now says so and a test pins it. |
| "109 existing `console.*` calls" | 108 before this PR. Moot — the record no longer uses `console.*`. |
| `176 of 1510` sample in permanent source comments | Removed from source; unreproducible point-in-time measurements belong here. |

`k8sClient.ts:1284` carries a deliberate same-issue invariant — *"Counts and duration only — never the resolved FQDNs/IPs or any Secret."* It is not overridden: that line summarises a **fleet-wide sweep**, where naming every host of every recipe would dump the fleet onto one line with no attribution. Its comment now points at where the per-host record lives instead, so the two do not read as contradictory.

## Scope

No behaviour change to any policy: nothing is written, read, or shaped differently. The only runtime effect is a log record on set change. `warnEgressAccumulator` is warnings-only again, so its JSDoc is accurate again. The `package.json` / `package-lock.json` entries are the repository's CI-enforced version bump.

Related: #299, and the exposure work in #335 / `evenfire-infra#81`.

### DIFF
--- a/workflow-recipes/package-lock.json
+++ b/workflow-recipes/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "workload-recipes-controller",
-  "version": "0.1.320",
+  "version": "0.1.321",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "workload-recipes-controller",
-      "version": "0.1.320",
+      "version": "0.1.321",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1009.0",
--- a/workflow-recipes/package-lock.json
+++ b/workflow-recipes/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "workload-recipes-controller",
-  "version": "0.1.321",
+  "version": "0.1.322",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "workload-recipes-controller",
-      "version": "0.1.321",
+      "version": "0.1.322",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1009.0",
--- a/workflow-recipes/package-lock.json
+++ b/workflow-recipes/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "workload-recipes-controller",
-  "version": "0.1.319",
+  "version": "0.1.320",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "workload-recipes-controller",
-      "version": "0.1.319",
+      "version": "0.1.320",
       "license": "SEE LICENSE IN LICENSE",
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1009.0",
--- a/workflow-recipes/package.json
+++ b/workflow-recipes/package.json
@@ -1,6 +1,6 @@
 {
   "name": "workload-recipes-controller",
-  "version": "0.1.319",
+  "version": "0.1.320",
   "description": "Workload Recipes Controller (WRC) - WorkloadRecipe lifecycle management and MCP server interface",
   "main": "dist/main.js",
   "engines": {
--- a/workflow-recipes/package.json
+++ b/workflow-recipes/package.json
@@ -1,6 +1,6 @@
 {
   "name": "workload-recipes-controller",
-  "version": "0.1.320",
+  "version": "0.1.321",
   "description": "Workload Recipes Controller (WRC) - WorkloadRecipe lifecycle management and MCP server interface",
   "main": "dist/main.js",
   "engines": {
--- a/workflow-recipes/package.json
+++ b/workflow-recipes/package.json
@@ -1,6 +1,6 @@
 {
   "name": "workload-recipes-controller",
-  "version": "0.1.321",
+  "version": "0.1.322",
   "description": "Workload Recipes Controller (WRC) - WorkloadRecipe lifecycle management and MCP server interface",
   "main": "dist/main.js",
   "engines": {
--- a/workflow-recipes/src/k8sClient.ts
+++ b/workflow-recipes/src/k8sClient.ts
@@ -1282,6 +1282,12 @@ export class WorkflowRecipeWatcher implements WorkflowRecipeProvider {
     })
 
     // Counts and duration only — never the resolved FQDNs/IPs or any Secret.
+    // This line summarises a FLEET-WIDE sweep: naming every host of every recipe
+    // here would dump the whole fleet's egress onto one line, every pass, with no
+    // way to attribute an address back to a policy. The per-host record lives
+    // where the set is decided and written instead — `logResolvedEgressSet` in
+    // workflowRecipeReconciler.ts, one entry per FQDN, only when the set changed
+    // and only once the policy has landed (#299). Secrets stay out of both.
     console.log(
       `[WR-K8s] External egress refresh: re-enqueued ${enqueued}/${selected.length} recipe(s) in ${Date.now() - started}ms`
     )
--- a/workflow-recipes/src/reconciler/reconciler.test.ts
+++ b/workflow-recipes/src/reconciler/reconciler.test.ts
@@ -3856,88 +3856,436 @@ describe('WorkflowRecipeReconciler', () => {
     expect(npCall.body.spec.egress).toHaveLength(2) // internal + external
   })
 
-  // issue #299 — the WRC lane must leave a resolution history in the log.
+  // issue #299 — the WRC lane must leave a per-host resolution history.
   //
-  // HCC has emitted `[NetPol] Resolved <fqdn> → [ips]` since its initial commit,
-  // so its lane can be reconstructed from Cloud Logging: which addresses a host
-  // used over weeks, whether its universe is closed, whether it drifts. WRC had
-  // no equivalent, so `wl-egress-*` and `ui-egress-*` were observable only in
-  // the present through the live annotation — and that is exactly the lane where
-  // the currently-exposed hosts sit. These tests pin both the emission and its
-  // change-gating, since the value of the line is the history it accumulates.
-  it('logs the resolved egress set for the ui lane on first render', async () => {
-    const logSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-    const kc = new k8s.KubeConfig()
-    const rec = new WorkflowRecipeReconciler(kc, undefined, {
-      fqdnLookup: async () => ({
-        kind: 'ok',
-        ipv4: ['93.184.216.10', '93.184.216.11'],
-        ipv6: [],
-        ttlSeconds: 300,
-      }),
-    })
-    const recipe = makeRecipe({
+  // HCC emits `[NetPol] Resolved <fqdn> → [cidrs]` per DNS binding, so its lane
+  // replays from retained logs: which addresses a host used, whether its
+  // universe is closed, whether it drifts. WRC has had its own resolution path
+  // since the same initial commit and simply never got the equivalent record,
+  // so `wl-egress-*`/`ui-egress-*` were observable only in the present through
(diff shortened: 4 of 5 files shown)


alfredolopez80:
@claude — closing re-review and full audit requested.

This PR was reworked after an adversarial review of its first cut found six blocking problems, including that its own stated justification was false. I want an independent audit of the rework, not a confirmation of it. Please attack it.

## What changed and why

The first cut (`a360f52b`) added one `console.log` inside `warnEgressAccumulator`, emitting a flat union of resolved IPs per policy, before the write. The review found:

1. **The justification was false.** It claimed `workflow-recipes` has no logging wrapper. `src/observability/logger.ts` exists, has since the initial commit, is imported by 13 files (5 in `src/reconciler/`), and the repository logging standard names this service among those that must not bypass it.
2. **No per-host attribution.** HCC emits per DNS binding (`networkPolicyReconciler.ts:2229`); the first cut emitted per policy and discarded `EgressEntry.fqdn`. `egress.external[]` is a list, so multi-host is the normal case — the flat union could not answer the only question the history exists for.
3. **The record could contradict the policy.** It fired before the M3 `isBlockedExternalIPv4` filter and before the write, so a rehydrated RFC1918 address from the state annotation would appear in the record and not in the policy.
4. **`N→0` was silent.** `entries.length > 0` suppressed the set-collapses-to-empty event.
5. **A false history claim.** "WRC gained its own resolution path later, in #299 Phase 1" — `git ls-tree 5d663a49 workflow-recipes/src/reconciler/` shows `fqdnResolver.ts` in the same initial commit as HCC's line.
6. **An unacknowledged same-issue invariant.** `k8sClient.ts:1284` reads *"Counts and duration only — never the resolved FQDNs/IPs or any Secret"*, `git blame` → `1ac021146`, same author, same issue.

The rework (`a89d6a31`) emits one structured record per declared FQDN, through `createLogger('wrc', recipe).info(...)`, **after** the write lands, over the post-M3 filtered set, with `cidrs: []` for a host that resolved to nothing.

## Where I most want you to push back

- **Is "after the write" actually correct?** I claim the divergence between record and policy is now zero because both lanes end in an `await` and the early returns (null policy, no-op gate) precede it. Find a path where the record still misrepresents what is enforced — a partial write, a `replace` that failed after a successful `create`, a workload loop iteration that throws mid-way leaving earlier records standing.
- **Is `k8sClient.ts:1284` genuinely reconciled, or did I rationalise around it?** I argued that line summarises a fleet-wide sweep where per-host naming would be an unattributable dump, while the new record is per-policy at the point of decision. Is that a real distinction or a convenient one? The invariant also says "or any Secret" — confirm nothing secret can reach `cidrs`/`ports`/`policy`/`recipeName`.
- **Volume.** Per-FQDN multiplies records versus per-policy. `LOG_LEVEL` gates it, but the default is `info`, so it is on. Bound the worst case against `externalEgressRefreshIntervalSeconds` / `REFRESH_FLOOR_SECONDS` / `maxEntries` and tell me if `info` is the wrong default.
- **The tests.** I ran an 11-mutation matrix against the full 310-test file and all 11 died, zero survivors (the prior review found 4 survivors of 6). Design mutations I did not think of. In particular: is the renewal test (`does NOT record a renewal write`) actually exercising `renewalDue`, or does it pass for another reason? Is the blocked-address test really rehydrating through `parseState`, or is the entry dropped earlier for an unrelated reason?
- **`compareCidrByAddress`.** New helper, no direct unit test — only exercised through the reconciler. Malformed input, IPv6, a `/32`-less string: does it misorder or throw?
- **Did I overreach?** The diff grew from +115/−3 to +539/−3. Argue that some of it should not be in this PR.

## Verification claims to check rather than trust

- `npm test` in `workflow-recipes`: 2533 passed, 29 skipped, 0 failed, EXIT=0 — **on the merged tree**, after syncing 20 commits from `dev`.
- `npm run build` (`tsc -b`) and `npx tsc --noEmit`: both EXIT=0.
- 37/37 CI checks green on `a89d6a31`.
- Every mutation restored byte-identically; `git status --porcelain` empty.

Note that `npm run lint` fails repo-wide (ESLint 9, no `eslint.config.js`) — pre-existing, not in CI, not touched here. Flag it if you disagree with leaving it.

Please report blocking / non-blocking / refuted, with `file:line` and a concrete failure scenario for each finding, and say explicitly which of my claims above you verified versus took on trust.

zach88:
Took a deep pass through this one across three angles: the reconciler logic, the security posture of logging resolved egress data, and the quality of the new test suite. Short version: this is clean, and the tests genuinely earn their keep.

**🟢 Ready to approve**
- Optional / non-blocking: **0**

## What checked out clean

**Feature logic** — verified end to end:
- Logger wiring is right: `createLogger('wrc', recipeName)` + `log.info(msg, fields)` lands `policy/fqdn/cidrs/ports` at top level, and none of those keys hit the key-based redaction.
- `logResolvedEgressSet` reads `.source.fqdn/.cidr/.port` off `ResolvedExternalEgressInput` and the accumulator actually emits those, so the per-host filter matches — records don't silently come out empty.
- The set logged is the post-M3 (`isBlockedExternalIPv4`) filtered array — the same one handed to the policy builder — so the record can't attest to an allowance the policy never made.
- `changed` is exactly `acc.changed`, so every emission implies a landed write, and the emit sits after `createOrReplace`/`applyNetworkPolicy` succeeds (no-op/delete/error paths return before it). Both lanes pass their own declared/resolved/changed with no argument swap.

**Security** — no new exposure. The emitter only ever receives `policy/fqdn/cidrs/ports` (no tokens, no raw state annotation); those values are already cluster-readable policy data, not Secrets. FQDNs go in as structured JSON values, so a `\n`/`{` in an adversarial hostname can't forge a log line. The `k8sClient.ts:1282` "counts and duration only" invariant isn't contradicted — that's a fleet-wide sweep, a deliberately different construct.

**Tests** — ran the suite: **310 passed** in `reconciler.test.ts`. Ran the new tests against the merge-base too: 6/8 fail correctly without the feature; the 2 negatives (`does NOT record a RENEWAL write`, `does NOT re-record when unchanged`) pass in the parent as any absence-of-new-behavior test must, but their positive controls are load-bearing and discriminate on HEAD. The hand-built `rehydrated` fixture in the M3 test is byte-compatible with what `serializeState`/`parseState` actually produce, and there's no fresh-DNS path that could reach that state, so the annotation fixture is the honest way to model it.

Nice touch on the PR body owning the corrections from the first cut rather than quietly dropping them — I checked the load-bearing technical claims against the head and they hold.

Merge whenever CI is green.

zach88 (approved):

PR #545, opened 2026-09-01
fix: close the subscription modal after finishing setup

miguelo-999:
Fix the modal remaining open after successful subscription


miguelo-999:
@claude pls provide feedback or green light I mean this is something tiny we should be good to go, but provide your review pls

josue-tech-94 (approved):

PR #547, opened 2026-09-01
feat(desktop-app): shrink chat-view session tabs Chrome-style

zach88:
## What
The full-screen chat tab strip pinned every session tab at a fixed 220px and let the row overflow into horizontal scroll as sessions piled up. The tabs now behave like Chrome: they share the available width and shrink together from a 220px ceiling down to a 96px floor, and only past that floor does the existing horizontal scroll take over.

## Why
With many open sessions the fixed-width strip forced horizontal scrolling almost immediately, making it hard to see and reach tabs. Shrinking-to-fit keeps far more tabs visible and navigable before scroll is ever needed, matching the behavior users already expect from browsers.

## How
- `.chat-view-tab` switches from `flex: 0 0 220px` / fixed `width` to `flex: 1 1 var(--chat-tab-max)` bounded by `min-width: var(--chat-tab-min)` (96px floor) and `max-width: var(--chat-tab-max)` (220px ceiling); the list drops `width: max-content` for `width: 100%` so the tabs shrink within the viewport instead of expanding the row.
- The active tab shrinks slower (`flex-shrink: 0.4`) so it stays legible when the strip is crowded.
- Sizing lives in two layout tokens, `--chat-tab-max` / `--chat-tab-min`, next to the existing single-use layout tokens in `tokens.css`.
- The strip's bottom shelf moves off the scrolling list onto the non-scrolling `.chat-view-tabs` container (`z-index: -1`, showing through the gaps between tabs) so it no longer slides away when the overflowing tabs scroll. Each inactive tab now draws its own matching bottom border and the active tab omits it to notch through into the surface below.
- The chat drawer's session switcher is a dropdown (`ChatSwitcher`), a different surface that does not compete for width, and is left untouched.

## Testing
- `ChatTabs` unit tests (3) and `ChatViewWorkspace` tests (5): all pass. The style-contract assertions were updated to the new flex/min/max model and now also pin the active-tab `flex-shrink`, the relocated shelf, and the inactive-tab bottom border.
- Native chat-workspace geometry validation: `cd desktop-app && npm run test:native:chat-workspace-geometry` passes. The native Electron fixture was updated for the elastic tabs and now covers, through measured Chromium geometry/computed style, the 220px ceiling, the 96px floor, tabs shrinking to fit before the strip overflows, horizontal scroll starting only after even the active tab reaches the floor, the active tab's slower shrink, and the non-scrolling shelf/notch treatment staying correct with `scrollLeft` applied. Verified it fails against the pre-PR CSS (merge base `47b191c`) and passes on head, so it genuinely guards the new contract.
- Repo style-rules checker: passes (exit 0, no findings).

## Risks / Follow-ups
- Purely presentational (CSS + tokens + style-contract tests); no logic, auth, or data paths touched. Live visual confirmation in the running Electron app is recommended as a final check of the shelf/notch across scroll positions.

### DIFF
--- a/desktop-app/test/native/chat-workspace-geometry.fixture.cjs
+++ b/desktop-app/test/native/chat-workspace-geometry.fixture.cjs
@@ -17,12 +17,16 @@ async function geometry(window) {
     const workspace = document.querySelector('.chat-view-workspace').getBoundingClientRect()
     const tabs = document.querySelector('.chat-view-tabs').getBoundingClientRect()
     const active = document.querySelector('.chat-view-tab.is-active').getBoundingClientRect()
+    const inactive = document.querySelector('.chat-view-tab:not(.is-active)').getBoundingClientRect()
     const scroller = document.querySelector('.chat-view-tabs__scroller')
     const surface = document.querySelector('.chat-view-surface').getBoundingClientRect()
     const surfaceStyle = getComputedStyle(document.querySelector('.chat-view-surface'))
     const activeTab = document.querySelector('.chat-view-tab.is-active')
     const activeStyle = getComputedStyle(activeTab)
-    const listSeamStyle = getComputedStyle(document.querySelector('.chat-view-tabs__list'), '::after')
+    // The bottom shelf now lives on the non-scrolling .chat-view-tabs container
+    // (::after) so it stays under the whole visible strip when the tabs
+    // overflow and scroll — it used to live on the scrolling list.
+    const seamStyle = getComputedStyle(document.querySelector('.chat-view-tabs'), '::after')
     const inactiveStyle = getComputedStyle(document.querySelector('.chat-view-tab:not(.is-active)'))
     const inactiveBorderReferenceStyle = getComputedStyle(
       document.querySelector('[data-inactive-tab-border-reference]')
@@ -35,6 +39,7 @@ async function geometry(window) {
     )
     return {
       activeBottom: active.bottom,
+      activeWidth: active.width,
       activeBackground: activeStyle.backgroundColor,
       activeBackgroundImage: activeStyle.backgroundImage,
       activeBorderBottom: activeStyle.borderBottomColor,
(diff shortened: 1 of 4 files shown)


josue-tech-94:
@zach88 one blocker to address before this is ready to merge: **R1-B1**.

The CSS changes look good, but the existing native Electron geometry fixture still asserts the old tab behavior and is now failing.

Please update:

`desktop-app/test/native/chat-workspace-geometry.fixture.cjs`

The stale assertion is around:

`desktop-app/test/native/chat-workspace-geometry.fixture.cjs:141`

The fixture currently expects the old model where inactive tabs have no bottom border and the shared shelf/seam belongs to the scrolling tab list. This PR changed that contract in:

* `desktop-app/ui/src/styles.css`
* `desktop-app/ui/src/styles/tokens.css`

The native fixture should be updated to validate the new behavior:

* tabs have a **220px max width**
* tabs shrink down to a **96px minimum width**
* tabs shrink to fit before horizontal overflow starts
* horizontal scrolling only starts after tabs reach the 96px floor
* the active tab shrinks slower than inactive tabs
* the shelf is owned by the non-scrolling `.chat-view-tabs::after`
* inactive tabs draw the matching bottom border
* the active tab keeps the notch/open bottom
* the shelf/notch remains correct when the tab strip is horizontally scrolled

Please keep these assertions based on observable Chromium/Electron geometry/computed behavior where possible, rather than only matching CSS strings.

To reproduce the current blocker:

`cd desktop-app && npm run test:native:chat-workspace-geometry`

That command currently fails on this PR head, while it passes at the merge base.

After updating the fixture, please verify:

`cd desktop-app && npm run test:native:chat-workspace-geometry`

and keep the existing `ChatTabs` / `ChatViewWorkspace` tests and style-rules checks green.

Once it passes, please also update the PR Testing section with:

* Native chat-workspace geometry validation: `npm run test:native:chat-workspace-geometry` passes, covering the 220px ceiling, 96px floor, active-tab slower shrink, overflow-after-floor behavior, and non-scrolling shelf/notch treatment.

That should close **R1-B1**.

zach88:
Thanks for the catch — **R1-B1** is fixed in `e52fa29b4`.

The native geometry fixture (`desktop-app/test/native/chat-workspace-geometry.fixture.cjs`) was the only stale piece; the CSS is unchanged. Two assertion clusters were asserting the old contract, so both got repointed:

- **Inactive-tab bottom border** — was asserting `border-bottom-style: none` / `0px`; now asserts `solid` / `1px` with the color matching `--chat-view-selected-border` (the shared seam), in both dark and light.
- **Shelf ownership** — moved the pseudo-element read from `.chat-view-tabs__list::after` to the non-scrolling `.chat-view-tabs::after` (and added a `z-index: -1` check).

New measured-geometry assertions cover the rest of your list, all via real `getBoundingClientRect` / computed style rather than raw CSS strings:

- 220px ceiling and 96px floor (measured tab widths at wide/narrow, plus the `min/max-width` tokens);
- tabs shrink to fit before overflow (intermediate width where widths sit strictly between 96 and 220 and `scrollWidth <= clientWidth`);
- horizontal scroll starts only after the floor (at the overflow width both active and inactive tabs sit at ~96px);
- active tab shrinks slower (`flex-shrink 0.4`, and measured active width > inactive width when both shrink);
- shelf/notch stay correct with the strip scrolled (force `scrollLeft > 0`, then re-check the shelf geometry and the painted notch/seam).

Verification:

- `cd desktop-app && npm run test:native:chat-workspace-geometry` → passes on head.
- Ran the updated fixture against the pre-PR CSS (merge base `47b191c`) and it **fails** (`'none' !== 'solid'`), so it actually guards the new contract instead of golden-ing the old one.
- `ChatTabs` / `ChatViewWorkspace` vitest tests and the style-rules check stay green.

The PR Testing section is updated with the native-geometry line. Let me know if you'd like anything else on it.

josue-tech-94 (approved):

PR #548, opened 2026-09-02
feat(desktop-ui): collapse global search and tune header/panel spacing

zach88:
## What

Header and layout polish for the desktop renderer (`desktop-app/ui`).

### Collapsible global search
- The header global-search now collapses to a 40px magnifier icon and expands to the full field on focus **or** while its results dropdown is open.
- `is-open` state on `AppHeader` means "there is a query / open results", not merely "was focused". Focus-time expansion comes from `.global-search:focus-within` (independent of `is-open`); `is-open` only keeps the field expanded while a query/results dropdown is present. Focusing an empty field, or clearing the query, collapses back to the 40px idle box even for a keyboard user who tabs away (no outside `mousedown` fires).
- Collapsed glyph is centered in the square; expanded glyph pins left and drops back to the in-field size.
- `prefers-reduced-motion: reduce` disables the width transition.
- Responsive variants: in the full-width header row the field never collapses (icon stays left-pinned); narrow breakpoint retunes the expanded width via custom props.

### Surrounding spacing
- `left-nav` padding and logo `margin-bottom` tightened.
- `content-panel--agent-chat`: top padding restructured to clear the absolute top-bar, with matching top-bar `top`/`right` offsets and a mobile fallback.
- `content-panel` with an open app: top padding `--space-4` → `--space-2`.
- `content-panel--chat-drawer-open`: utilities-width gutter `36px` → `16px`.
- Removed a stray `border-radius` on a selected chat-view element.

## Scope
- 1 service (`desktop-app/ui`), 4 files: 2 implementation files (`AppHeader/index.tsx`, `styles.css`) plus 2 test files (`responsiveUtilityGutter.browser.test.ts`, `AppHeader/__tests__/searchCollapse.test.tsx`) — within the PR budget.
- CSS + JSX state logic in `AppHeader`. No CRD, no API, no security surface.

## Verification
- Pre-commit passed: prettier, commitlint, `style-rules` (`da-*`) clean.
- `AppHeader` vitest suite green (13 tests incl. `searchCollapse.test.tsx`); the new regression tests fail against the pre-fix head, confirming they cover the collapse bug rather than shield it.
- The two spacing values (`--space-2` sandbox padding, `16px` chat-drawer gutter) and the search state change were run through `code-reviewer` + `security-reviewer`: no blockers, SAFE.

### DIFF
--- a/desktop-app/test/browser/responsiveUtilityGutter.browser.test.ts
+++ b/desktop-app/test/browser/responsiveUtilityGutter.browser.test.ts
@@ -22,12 +22,16 @@ function launchOptions() {
   return { headless: true }
 }
 
-async function headerGeometry(width: number, drawerOpen = false): Promise<HeaderGeometry> {
+async function headerGeometry(
+  width: number,
+  drawerOpen = false,
+  searchOpen = false
+): Promise<HeaderGeometry> {
   if (!browser) throw new Error('Browser must launch before measuring responsive utility geometry')
   const page = await browser.newPage({ viewport: { width, height: 800 } })
   await page.setContent(`
     <div class="content-panel${drawerOpen ? ' content-panel--app-notification-drawer-open' : ''}">
-      <div class="top-bar"><div class="header-left"><div class="global-search"></div></div></div>
+      <div class="top-bar"><div class="header-left"><div class="global-search${searchOpen ? ' is-open' : ''}"></div></div></div>
       <section class="page">
         <header class="apps-page-header">Apps</header>
         <header class="sandbox-ui-mounted-header"><button>Back</button></header>
@@ -62,8 +66,14 @@ describe('responsive utility gutter', () => {
   it('uses the compact closed-header reservation at tablet widths and retains drawer-open space', async () => {
     const closed = await headerGeometry(1100)
     const open = await headerGeometry(1100, true)
+    const searchExpanded = await headerGeometry(1100, false, true)
 
-    expect(closed.searchWidth).toBe(352)
+    // Idle, the global-search collapses to the 40px magnifier; it grows to its
+    // tablet expanded width (clamp(160px, 32vw, …) = 352px at 1100) only on
+    // focus/open. The utility gutter still reserves for the *expanded* search
+    // (352 + space-2 + 36 = 398) so expansion never collides with page content.
+    expect(closed.searchWidth).toBe(40)
+    expect(searchExpanded.searchWidth).toBe(352)
     expect(closed.mountedPaddingRight).toBe(398)
     expect(closed.mountedPaddingRight).toBeLessThan(open.pagePaddingRight)
     expect(open.mountedPaddingRight).toBe(0)
--- a/desktop-app/ui/src/components/AppHeader/__tests__/searchCollapse.test.tsx
+++ b/desktop-app/ui/src/components/AppHeader/__tests__/searchCollapse.test.tsx
@@ -1,0 +1,110 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { cleanup, render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { AppHeader } from '../index'
+
+// The collapsed idle contract: `.global-search` may only carry `is-open` while
+// there is query/result state. Focusing an empty field or typing-then-clearing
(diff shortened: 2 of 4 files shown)


josue-tech-94:
**Findings**

1. **Severity:** medium  
   **File and lines:** `desktop-app/ui/src/components/AppHeader/index.tsx:626`, `:635-639`, `:646`  
   **Problem:** `searchOpen` can remain true when there is no query, no dropdown, and no focus. `onFocus` always sets it true, and `onChange` also sets it true even when the user clears the field. The dropdown only renders when `searchOpen && hasSearch`, so after focusing an empty search or clearing the query and tabbing away, `.global-search.is-open` remains applied even though the search should be idle/collapsed.  
   **Why it matters / concrete failure scenario:** Keyboard user tabs into search, types nothing or clears the query, then tabs away. No outside `mousedown` fires and there is no blur cleanup, so the desktop/tablet search remains expanded indefinitely. That breaks the PR’s central contract that the 40px collapsed state represents the closed idle state.  
   **Recommended fix:** Treat `searchOpen` as “results/open query state”, not “has ever focused”. For example, only set it true from `onChange` when the trimmed value is non-empty, and/or close it on blur when focus leaves the search and there is no query. Focus alone can still expand via `:focus-within`.  
   **Whether a test should be added:** Yes. Add an AppHeader interaction test for focus-empty-tab-away and type-clear-tab-away verifying `.global-search` does not retain `is-open` after focus leaves with an empty query.

2. **Severity:** low  
   **File and lines:** PR description, `What` / `Scope` sections  
   **Problem:** The PR description says the change is “across two files” and “2 files”, but the PR changes 3 files, including `desktop-app/test/browser/responsiveUtilityGutter.browser.test.ts`.  
   **Why it matters / concrete failure scenario:** Reviewers use the PR body as the review contract. The stated scope understates the test change and is inconsistent with GitHub’s changed-file list.  
   **Recommended fix:** Update the PR body to say 3 files, or distinguish “2 implementation files plus 1 test file.”  
   **Whether a test should be added:** No.

## PR description / updates

Needs update. Copy-ready replacement for the scope line:

```text
- 1 service (`desktop-app/ui`), 3 files: 2 implementation files plus 1 browser
  layout test file — within the PR budget.
```

## Validation

I read the PR body, metadata, complete diff, surrounding `AppHeader`, `useClickOutside`, responsive CSS, sandbox/header/drawer layout code, and related AppHeader/browser tests. I attempted local validation, but the temporary worktree had no installed dependencies: `npm run verify:electron` failed because Electron was not installed, and there is no `test:browser` script. I did not modify code or publish anything to GitHub. Temporary review worktree was removed.

## Verdict

REQUEST CHANGES

The main reason is the PR’s core state contract is not actually true: `searchOpen` can stay true after empty focus or query clearing, so the search remains visually expanded while idle. The updated browser test checks synthetic CSS widths, but it would not catch that real AppHeader state drift. The PR body also needs a small scope correction.

zach88:
Thanks for the catch — both addressed in f567620eb.

**1. `searchOpen` stuck open (medium).** Fixed. Redefined `searchOpen` to mean "there is a query / open results" instead of "was ever focused", so the collapse tracks the idle state:
- `onChange` opens only when the trimmed value is non-empty (clearing the field now collapses it).
- `onFocus` opens only when a query already exists (focusing an empty field no longer forces `is-open`).
- the `searchFocusRequestId` effect no longer forces it true; it derives open-ness from the live input value.

Focus-time expansion is unchanged — it comes from `.global-search:focus-within` in `styles.css`, independent of `is-open`, so an empty focused field still widens visually but collapses the moment focus leaves. No `onBlur` handler needed.

Added `AppHeader/__tests__/searchCollapse.test.tsx` covering both scenarios you named (focus-empty-then-tab-away, type-then-clear) and asserting the observable `.global-search` class. Verified the tests go red against the pre-fix head, so they reproduce the bug rather than pin it.

While wiring the effect I found (and fixed in the same commit) a narrow follow-on: with the naive version, Escape-then-reopen-shortcut wouldn't reopen the dropdown when a query was still present, because `focus()` is a no-op on an already-focused input and `onFocus` wouldn't re-fire. Reading the live value in the effect covers that.

**2. Scope says 2 files (low).** Updated the PR body — it now reads 4 files (2 implementation + 2 test).

josue-tech-94 (approved):

Issue #579, opened 2026-09-03
bug(wrc): the coordinator-to-gfs NetworkPolicy guard compares JSON.stringify against the server view, so it can never match (latent)


alfredolopez80:
> **Latent, not currently firing.** The defect is verified in code and reproduced in isolation, but it writes nothing today because no recipe in `clerum-dev` declares `gfs.publishTargets`, so zero such policies exist. It activates the moment one does.

## Summary

`ensureCoordinatorGfsNetworkPolicyIfEnabled` guards its replace with an ad-hoc comparison that can never be true, even when the policy has not changed. Once a recipe declares `gfs.publishTargets`, that policy gets one `PUT` per reconcile pass forever — roughly 214 per hour per recipe at the current cadence.

## The defect

`workflow-recipes/src/reconciler/workflowRecipeReconciler.ts:4982`

```ts
if (labelsMatch && JSON.stringify(existing.spec) === JSON.stringify(policy.spec)) {
  return existing
}
```

`existing.spec` is the apiserver's view; `policy.spec` is what the builder constructed. `JSON.stringify` preserves insertion order, and the two orders differ:

| source | key order |
|---|---|
| `buildCoordinatorGfsNetworkPolicy` (`resourceBuilder.ts:530`) | `podSelector`, `policyTypes`, `egress` |
| apiserver (checked against live NetworkPolicies in `clerum-dev`) | `egress`, `podSelector`, `policyTypes` |

Reproduced in isolation — identical content, both key orders:

```
contenido idéntico: NO
  builder : {"podSelector":{"matchLabels":{"clerum.io/recipe":"r"}},"pol
  servidor: {"egress":[{"to":[{"podSelector":{"matchLabels":{"app":"cont
```

The comparison is therefore always false and the guard never closes. Server-side defaulting would compound it, but key order alone is enough.

## Why it is not firing

`:4963` returns early when `recipe.spec.gfs?.publishTargets` is empty. No recipe in `clerum-dev` sets it:

- policies matching `*coordinator-to-gfs*` in the cluster: **0**
- `[WR-Reconciler] Updated NetworkPolicy` lines for that name in one hour: **0**

So there is no PUT storm to point at — only a guard that will not do its job when the feature is used.

## Relationship to #575

Same class, different mechanism, and worth keeping apart:

- **#575**: `applyNetworkPolicy` passes no `idempotency`, so there is no gate at all. ~12 200 `PUT`/h today.
[...]

Issue #542, opened 2026-09-01
prod(clerum): HCC unconditional replaces saturated the apiserver (GKE REPAIR_CLUSTER 2026-09-01)


jozer-rami:
## What happened?

On 2026-09-01 UTC, GKE prod `clerum` (`gke_eventfire-491421_us-central1-a_clerum`, project `eventfire-491421`, zone `us-central1-a`) went through an API-server saturation event. GKE ran `REPAIR_CLUSTER` twice. Calico could not program new pods. HCC crashlooped. User-facing `/health` recovered after the second repair, but **HCC is still writing ~150 objects/min on an unchanged fleet**.

This is an investigation issue. Do **not** treat the cluster as currently dead: at ~16:13 UTC the API server `/readyz` passed, almost all app Deployments were `1/1`, and:

- `https://api.prod.evenfire.ai/health` → 200 `{"status":"ok"}`
- `https://rpc.prod.evenfire.ai/health` → 200
- `https://profile.prod.evenfire.ai` → 200
- `https://app.prod.evenfire.ai` → 302 Cloudflare Access (expected)

No prod mutations were made during the debug (kubectl-guard on, read-only). Last `deploy-clerum-prod.yml` success was **2026-08-13**. HCC image on prod: `us-central1-docker.pkg.dev/eventfire-491421/clerum/host-context-controller:sha-21d9a7d`.

Please investigate the write storm and the watch-ERROR crash as the two load-bearing defects. The rest of this issue is evidence.

## Version / commit

- Prod HCC: `host-context-controller:sha-21d9a7d` (`sha256:c48bb8bdd41c96ac6160787e760fa182896b528f0e762a3d3cc26f8af543ed0e`)
- Last prod promotion: `deploy-clerum-prod.yml` on evenfire-infra, 2026-08-13
- Source pointers below are from current `dev` / this worktree and match the behaviour observed on prod. Confirm against `sha-21d9a7d` before treating a line number as gospel.

## Environment

- Cluster: GKE `clerum` (prod), `eventfire-491421` / `us-central1-a`
- Context: `gke_eventfire-491421_us-central1-a_clerum`
- 1 node: `gke-clerum-default-pool-ae96e515-8kt8` (`e2-standard-8`, created 2026-08-29T19:56:21Z, k8s `v1.34.9-gke.1610001`)
- Pod pressure: **101/110** pods on the node
[...]

PR #580, opened 2026-09-03
fix(wrc): converge NetworkPolicies from live state (#575)

### LINKED ISSUE #575: bug(wrc): applyNetworkPolicy skips the spec-hash gate — ~12.2k wasted PUT/hour (largest of four ungated NetworkPolicy writers)
> **Corrected 2026-09-03** after a design pass re-verified every claim against the code. Four statements in the original text were wrong: the count of gated callers, "the only writer", the size of the effect, and why `wl-egress` behaves differently. The corrected figures are below; the mechanism itself held.

## Summary

`applyNetworkPolicy` does not opt into the spec-hash idempotency gate that the same file implements and uses for five other object kinds. Every WRC NetworkPolicy that funnels through it is rewritten on every reconcile pass without comparing anything: roughly **12 200 wasted `PUT` requests per hour** in `clerum-dev`, of objects whose content is provably unchanged.

It is not the only writer with this defect — it is the largest of four (see *Other ungated writers*). This is the same mechanism as #542, where HCC writing ~150 objects/min over an unchanged fleet saturated the production apiserver and GKE ran `REPAIR_CLUSTER` twice.

## The defect

`workflow-recipes/src/reconciler/workflowRecipeReconciler.ts:4879`

```ts
private async applyNetworkPolicy(policy: k8s.V1NetworkPolicy, namespace: string): Promise<void> {
  const policyName = policy.metadata!.name!
  await this.createOrReplace(
    () => this.networkingApi.createNamespacedNetworkPolicy({ namespace, body: policy }),
    async () => {
      const existing = await this.networkingApi.readNamespacedNetworkPolicy({ name: policyName, namespace })
      this.assertInternalDependencyPolicyOwnership(policy, existing, namespace)
      policy.metadata!.resourceVersion = existing.metadata?.resourceVersion
      return this.networkingApi.replaceNamespacedNetworkPolicy({ name: policyName, namespace, body: policy })
    },
    `NetworkPolicy "${policyName}" in ${namespace}`
    // ← no 4th argument
  )
}
```

`createOrReplace` (`:3731`) takes a fourth parameter, `idempotency`, which stamps the manifest with a SHA-256 of its content and compares it against the annotation on the live object before replacing:

```ts
if (idempotency && (await this.applyIsNoop(idempotency))) {
  console.log(`[WR-Reconciler] ${label} unchanged (spec-hash match); skipping update`)
  return
}
```

**Five** call sites pass `idempotency` — Deployment (`:5479`), headless Service (`:5526`), CronJob (`:5623`), Job (`:5662`), DaemonSet (`:5700`). `applyNetworkPolicy` does not. The gate is not missing; it is not opted into.

Six call sites funnel through this one method — `:4283` (ui-ingress), `:4433`/`:4448` (wr-intdep egress/ingress), `:4660`/`:4687` (wl-egress, wl-ingress), `:6221` (oauth-broker egress) — which is why every family behaves identically.

## Evidence

Measured on `clerum-dev`, read-only, 2026-09-03.

**1 · The gate works, and NetworkPolicy is the only kind it never fires for.** Spec-hash skips in one hour:

| kind | skips |
|---|---:|
| `Deployment` | 1 592 |
| `Service` | 487 |
| **`NetworkPolicy`** | **0** |

**2 · The content is provably unchanged.** One affected policy, dumped, left 50 s (≈3 `PUT`s), dumped again:

```
t0 resourceVersion: 1785183402281791006
t1 resourceVersion: 1785183402281791006   ← unchanged
spec identical: True · labels identical: True · annotations identical: True
```

The apiserver absorbs the identical `PUT`: no etcd write, no `resourceVersion` bump, **no watch fan-out**. The cost is apiserver CPU — auth, admission, validation, decode/encode — and priority-and-fairness budget.

A design pass since confirmed this holds by construction, not just for the sampled policy: the builders order targets, sources and ports deterministically (`internalDependencyNetworkPolicies.ts:117,181`; oauth `[...optedIn].sort()`), with no `Date`, no `Math.random`, and no iteration over a container with variable order. The hash converges — unlike #413, where two manifests never do.

**3 · Volume.** Controller logs, one hour:

| family | live policies | `PUT`/h |
|---|---:|---:|
| `wr-intdep-egress-*` | 37 | 5 123 |
| `wr-intdep-ingress-*` | 26 | 3 688 |
| `ui-ingress-*` | 12 | 1 593 |
| `wl-ingress-*` | 16 | 1 313 |
| `wf-*` | 3 | 506 |
| `wl-egress-*` | 52 | 227 |
| **total** | | **12 450** |

Not uniform: of the 63 `wr-intdep-*` policies, **50 receive ~214 `PUT`/h each** — one every 17 s, matching the reconcile cadence — and 13 receive none.

**`wl-egress` is not on a different path.** It reaches the same `applyNetworkPolicy` at `:4660`, but only when `egressWriteNeeded(existing, policy) || wlEgressRenewalDue || wlEgressStateChanged` (`:4640-4660`) already decided a write is due. Its 4,4 writes/policy/h are intentional TTL renewals, not the gate at work.

## Expected effect: ~12 200 `PUT`/h, not 24 400

The original text claimed the failed `POST` disappears too. It does not. `createOrReplace` runs `createFn()` **before** the gate; the gate lives inside the `catch` on 409. Per apply:

- today: `POST` (409) + `GET` + `PUT`
- with the gate: `POST` (409) + `GET`

Only the `PUT` goes away — roughly **36 700 → 24 400 requests/hour** on this path, a 33 % reduction, not the 24 400 avoided that the first version claimed. Removing the `POST` requires read-first in `createOrReplace`, which changes the contract for all six gated classes and the default test mocks; that is a follow-up, not this issue.

## Other ungated writers (inventory, so the remaining work is visible)

NetworkPolicy writers that do **not** pass `idempotency`:

| site | note |
|---|---|
| `:4879` `applyNetworkPolicy` | this issue — 6 call sites, largest volume |
| `:3040` webhook-gateway (3 policies) | same defect; fix is routing through `applyNetworkPolicy` |
| `:4212` ui-egress | ungated, but pre-filtered by #299, so it only writes when a write is due |
| `:4968` coordinator-to-gfs | has an ad-hoc gate comparing `JSON.stringify(existing.spec)` against the desired spec — the server-defaulted view, so it may never converge and have a PUT storm of its own. Unmeasured |

Beyond NetworkPolicy: of 17 `createOrReplace` calls, **12** pass no `idempotency` (Service ×2, ConfigMap ×2, Secret ×2, gateway Deployment/Service, NetworkPolicy ×4).

The run lane (`workflow/workflowReconciler.ts:4504`) has its **own** `applyNetworkPolicy`, also ungated, under the same service account. It shares no helper with the recipe reconciler — not `createOrReplace`, not `applyIsNoop`, not `safeDelete` — so it is separate work.

## Fix direction

Pass `idempotency` in `applyNetworkPolicy`, in the shape the five gated classes already use. Two constraints:

1. **`assertInternalDependencyPolicyOwnership` currently runs inside the `replace` path.** A gate that returns first would skip it. It belongs inside `readExisting`, i.e. on the read the gate itself evaluates, with the result cached only after the assertion passes so the replace path reuses the object the gate looked at.
2. **`applyIsNoop` must stop swallowing everything.** Its `catch { return false }` would turn an ownership veto raised in `readExisting` into a write. A read *failure* (HTTP status, retryable transport) should still fall through to the replace; anything else must propagate.

The gate stays loud: `unchanged (spec-hash match); skipping update` is what keeps the ~16 s reconcile cadence visible once the writes stop. A silent gate would make the symptom of #460 invisible.

## First run

The first pass writes once per policy to stamp the annotation — a real write this time. Order of 100–300 WRC policies, spread over the reconcile cadence, so 1–2 minutes. HCC does not watch NetworkPolicies, so there is no fan-out. No staged rollout needed.

## Version

- Cluster `clerum-dev` (`gke_eventfire-491421_us-central1-a_clerum-dev`), measured 2026-09-03 07:40–08:20Z
- `control-plane/workflow-recipes`, ReplicaSet `workflow-recipes-54db484dbd`
- Code re-verified against `origin/dev` at `e5afb0de`; no cluster mutations were made

alfredolopez80:
## Current synchronization status — 2026-09-09

Head `32087f6ef3dffb4722ca3614e085da008d2f30f1` incorporates `origin/dev`
at `f8460e9d6c85eaf8f3666274b2376d1fae6795fd`. This is a merge into the
PR branch only. The PR remains open and must not be merged as part of this task.

The six merge conflicts were resolved while preserving the recipe-wide
pre-DNS permission contraction from dev, live-state no-op/repair from #580,
recipe ownership on the same validated snapshot, and bounded retry behavior.
Regression coverage now includes failures before and after DNS and explicit
ownership/no-op/replacement checks for the intersecting policy families.

- Local WRC: 2,773 passed / 29 expected skips; build PASS.
- Local HCC: 2,072 passed; network-policy core: 53 passed.
- All 75 PR hermetic contracts and the public-boundary guard PASS.
- GitHub CI on this synchronized head: **37/37 checks PASS**, including CodeQL.
- **Fresh T0/T1/T2 runtime certification is required and not yet complete:**
  the local container runtime is unavailable. The earlier head's evidence and
  reviews below are historical and do not certify this merged candidate.
- The code delta against updated dev adds no production source change outside
  WRC and no production manifest/RBAC/schema change. A live-cluster comparison
  still requires a verified target context; no production deployment or
  resource mutation was performed, and zero production impact is not asserted.

---

Closes #575.

## Problem and intended outcome

WRC's create-first NetworkPolicy path attempted a POST and unconditional PUT
for existing policies on repeated reconciliation. The issue reported roughly
12,450 redundant PUTs per hour. A hash-only skip could hide live enforcement
drift, stale owner UIDs or an ownership veto.

This implementation uses live Kubernetes state as the convergence authority:
skip genuinely converged objects without suppressing repair when the parent
reconciles. It does not introduce a new autonomous drift watcher or timer.

## Implementation

- Read-first NetworkPolicy-specific convergence; create only after confirmed
  absence. Normalize Kubernetes defaults and compare live spec plus WRC-owned
  metadata. The legacy hash is provenance, not no-op authority.
- Validate ownership and replace using the same snapshot/resourceVersion.
  Repair missing/stale same-recipe webhook owners; reject foreign ownership
  and unexpected owner references on historically ownerless families.
- Preserve external labels, annotations and finalizers during drift repair,
  while updating/removing WRC-owned fields according to the desired state.
  Admission-added metadata alone does not cause churn.
- Keep terminating-object and create/replace 404/409 races retryable. Both
  workflow and non-workflow paths schedule the existing bounded requeue.
  Transient workload-egress pre-read failures, including 503/ECONNRESET, use
  the same classification; permanent failures such as 403 do not become an
  endless retry. Watcher integration proves recovery after status-only events.
  A recovered transient read may intentionally issue one version-guarded
  replace; ordinary healthy steady-state reconciliation remains read-only.
- Use WRC's structured, redacting logger and preserve ownership-veto identity.

All six in-scope families are covered: UI ingress, workload ingress/egress,
internal dependency, OAuth broker egress and webhook gateway. The real cluster
aggregate observes nine concrete policies across those families.

## Regression and test-quality coverage

- Unit tests cover live drift with intact hashes, normalization, metadata and
  owner repair/vetoes, lifecycle ordering, read/create/replace races, retry
  semantics and exact zero create/replace calls for converged policies.
- Four live cluster suites exercise positive service routes, attributable
  negative controls, repair and recovery, and a terminating-policy race that
  heals without a second parent-spec event.
- Negative probes distinguish connection observations from executor failures.
  Narrow complementary permissions and the same-source allow/deny/allow
  sequence isolate the policy under test. BusyBox uses native verbose timeout
  diagnostics; silent failures and post-connect read timeouts fail the harness.
- No-write observation begins before the trigger, spans the complete window,
  checks policy/controller identity and all controller replicas, requires an
  exact reconcile receipt, and rejects writes even if a no-op also occurs.
  Stable resourceVersion alone is not claimed to prove zero API writes.
- Public CI runs 75 hermetic contracts, including real aggregate execution,
  missing/failed children and mutations that skip execution or swallow errors.
  This is not represented as live CNI execution.
- Fixtures use collision-safe creation, metadata-only ownership ledgers,
  UID-preconditioned cleanup and verified dynamic port-forwards. Parent
  replacement or deletion failure cannot authorize foreign-child deletion.
  Webhook signing Secrets receive their required recipe-owner label before
  create; the real producer regression rejects removal of that label.

## Scope and compatibility

- WRC version against the current base: `0.1.345` → `0.1.347`.
- No dependency graph, CRD, API, database schema/migration, RBAC or production
  manifest change. Other Kubernetes object kinds keep their generic behavior.
- No production source change in HCC, Control API, External REST API, RPC Proxy,
  MCP Host, GFS, Control UI or Desktop. No browser surface changed.
- No first-deploy annotation-stamping wave and no unrelated object migration.
- The transient pre-read and metadata-loss defects were reproduced as
  pre-existing and explicitly included in the corrective scope; they are not
  misattributed as regressions introduced by the original PR.

## Historical local evidence — before dev synchronization

Exact candidate: `8d4384b82885c3feb3c1bac8fd36a2546cae8365`.
Base: `c28fa3b54c691292f97c1a9128738399d501c1ec`.

- WRC: 2,651 passed / 29 expected skips; build/typecheck PASS.
- 75 hermetic contracts, Bash syntax, ShellCheck and public-boundary checks PASS.
- Full isolated development certification: T0 PASS, T1 PASS, T2 PASS.
- Real PostgreSQL: 35/35 selected physical files, 163/163 tests, zero pending.
- Deployed NP-08 authorization, required readiness and user-facing Health PASS.
- NetworkPolicy aggregate: all four suites PASS, including OAuth 41/41 and
  webhook 24/24. Final fixture-resource and matching port-forward record
  inventories were empty.
- Independent Astra code and local-evidence review: no confirmed pending
  defects in scope. This does not replace public CI or maintainer approval.
- Playwright E2E: NOT_RUN; no browser surface is changed by this PR.

An earlier cross-service RPC unit run observed one unexplained 400/404
assertion. Focused and full reruns, including the subsequent full gate, passed
without changes. The entire RPC source/config/lockfile tree is identical to the
base; the incident is recorded as uncertainty, not attributed to this PR.

These are public-safe summaries: no local paths, profile identifiers,
endpoints, generated ports, credentials or raw operational logs are included.

## Historical public review — before dev synchronization

- GitHub CI on this exact head: **37/37 checks PASS**.
- [Fresh Claude re-review](https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5550848403)
  completed successfully against the exact base/head: no blocker, critical,
  high or medium findings; three non-gating informational/optional notes.
  Verdict: merge with green CI.
- GitHub still requires a formal maintainer approval (`REVIEW_REQUIRED`).
  Claude's comment is not a formal GitHub approval. No merge has been performed.


alfredolopez80:
@claude Full re-review requested. This PR has already been through a four-agent internal review with issue #581 in context; every finding is listed below with its disposition, so you can check each one against the code rather than rediscovering it. Please verify point by point — including the claims I make about what was fixed.

CI is green on `223162a7`.

---

# Part 1 — What this PR actually changes

## The defect

`workflow-recipes/src/reconciler/workflowRecipeReconciler.ts` implements a spec-hash idempotency gate via `createOrReplace(createFn, replaceFn, label, idempotency?)`. The fourth parameter is **positional and optional**. Five callers pass it — Deployment, headless Service, CronJob, Job, DaemonSet. Twelve do not.

`applyNetworkPolicy` was the largest of the twelve. Every NetworkPolicy family funnelling through it — `ui-ingress-*`, `wl-ingress-*`, `wr-intdep-*`, `wf-*-oauth-broker-egress`, and the three webhook-gateway policies — is rebuilt deterministically from the recipe on every reconcile pass, so the unconditional replace was one `PUT` per policy per ~16 s.

Measured on `clerum-dev`, read-only:

| kind | spec-hash skips in one hour |
|---|---:|
| `Deployment` | 1 592 |
| `Service` | 487 |
| **`NetworkPolicy`** | **0** |

12 450 `PUT`/h from this path. Of the 63 `wr-intdep-*` policies, 50 took ~214/h each — one every 17 s, matching the reconcile cadence — and 13 took none.

One affected policy, dumped, left 50 s (≈3 `PUT`s), dumped again:

```
t0 resourceVersion: 1785183402281791006
t1 resourceVersion: 1785183402281791006   ← unchanged
spec identical: True · labels identical: True · annotations identical: True
```

The apiserver absorbed every `PUT`: no etcd write, no `resourceVersion` bump, no watch fan-out. The cost was apiserver CPU and priority-and-fairness budget — the same mechanism as #542.

## The three code changes

**1. `applyNetworkPolicy` (`:4879`) passes `idempotency`.**

**2. The webhook gateway's three policies (`:3040`)** had the same defect inline and now route through the same method. The log label changes from `WebhookGateway NetworkPolicy "x"` to `NetworkPolicy "x" in <ns>`.

**3. The ownership assertion moved, and this is the part worth your attention.**

`assertInternalDependencyPolicyOwnership` ran inside the replace path. A gate that returns first would skip it — meaning a foreign object carrying a matching hash would be silently accepted. It now runs inside `readExisting`, the read the gate itself evaluates, and the result is cached **only after** the assertion passes:

```ts
const readExistingAndAssertOwnership = async (): Promise<k8s.V1NetworkPolicy> => {
  const fetched = await this.networkingApi.readNamespacedNetworkPolicy({ name: policyName, namespace })
  this.assertInternalDependencyPolicyOwnership(policy, fetched, namespace)
  assertedExisting = fetched          // AFTER the assertion, never before
  return fetched
}
```

The invariant is not "one read". It is **no `PUT` without an assertion on the exact version written**.

## Why `applyIsNoop` was deliberately left alone

An earlier draft narrowed its `catch` so an ownership veto would propagate instead of falling through to the replace. That broke an existing test — `falls open and replaces when the existing object cannot be read (never silently skips)` — which injects a bare `Error('transient read timeout')` and asserts the write still happens. That expectation predates this change and is a real requirement, so the draft was wrong, not the test.

No rule keyed on the *shape* of an error can separate "I could not read" from "I read it and I veto it": both arrive as exceptions. The catch stays total; the veto surfaces from `replaceFn`, which re-reads and re-asserts. **T5c** pins that for NetworkPolicy so the idea cannot return quietly.

---

# Part 2 — The four-agent review

Four agents in parallel over an isolated worktree at `11db8118`, each given issue #581 and told to classify every finding as "fix in #580" or "covered by #581", and to verify #581's own claims against the code.

## 2A — Findings fixed in this PR

### F1 · High · **T6 was vacuous** — REPRODUCED

T6's second-pass assertion was purely negative:

```ts
expect(mockNetworkingApi.replaceNamespacedNetworkPolicy).not.toHaveBeenCalledWith(
  expect.objectContaining({ name })
)
```

"This policy was not written" is true in two different worlds: the gate ran and chose not to write, **or the code never ran at all**.

**Reproduction**: patched the gateway policy loop to execute only on the *first* reconcile pass — deleting outright the behaviour T6 protects. Result: `318/318`, **T6 green**.

**Control**: the other mutation (stamping a timestamp into the gateway policies, making the builder non-deterministic) *does* kill T6. So its determinism claim was sound; only the liveness half was missing.

**Blast radius**: #581 specifies T6's shape as a mandatory `expectConvergesOnSecondPass` helper across seven PRs and six object families. Extracting it as written would have propagated the blind spot.

**Fix** — both halves now required:

```ts
// Negative half: the gate closed, no write went out.
expect(replaceNamespacedNetworkPolicy).not.toHaveBeenCalledWith(expect.objectContaining({ name }))
// Positive half: prove pass 2 actually REACHED this object.
expect(readNamespacedNetworkPolicy).toHaveBeenCalledWith(expect.objectContaining({ name }))
expect(logged).toContain('unchanged (spec-hash match); skipping update')
```

The read is the witness the behaviour cannot fake: to decide the hash matches, the gate must have fetched the object.

**Please verify**: that the positive half genuinely fails under the non-execution mutation, and that the log assertion is not satisfiable from another code path.

### F2 · Low · **T4b added** — the untested other side

T4 pinned the ownership *veto*. Nothing pinned that an internal-dependency policy WRC **legitimately owns** passes the gate. An assertion that vetoed every intdep policy would have left T4 green while stopping the whole family from converging and setting an `OwnershipConflict` condition every pass.

T4b asserts the apply resolves, no replace is issued, and **one** read occurs — contrast T4's two, which is the signature of the veto path.

### F3 · Low · **T4 pins the error class, not just the message**

`reconcileInternalDependencyPolicies:4449` classifies with `error instanceof NetworkPolicyOwnershipConflictError` to set the CRD condition `OwnershipConflict`. A bare `Error` with the same text would set `PolicyApplyFailed` while passing a message-only assertion.

The class is module-local. Exporting it for a test's benefit would be production code shaped by a test, so the test pins `error.name` instead.

### F4 · Low · **Hash assertions were weak**

Three `expect(...[SPEC_HASH]).toBeTruthy()` — a stamp present but *wrong* would pass, and would skip the write for the wrong reason. Now `toBe(computeSpecHash(plainPolicy()))`.

### F5 · Low · **`applyIsNoop` discarded the error it swallowed**

`catch { return false }` — no variable, no log. Falling through to the replace is correct, but the diagnosis was lost: revoked RBAC or a persistently failing apiserver looked exactly like a healthy cache miss, with the gate silently saving nothing.

`git blame` → `5d663a49`, 2026-07-14. Preexistent; this PR multiplies its radius to NetworkPolicies. Now logs at `warn` with the object label (the `label` is threaded in from `createOrReplace`, which already has it). The file uses `console.*` throughout (109 calls) and its `createLogger` requires a recipe name not in scope in this generic helper, so a second convention here would have been worse than consistency.

### F6 · Low · **`existing` renamed to `assertedExisting`**

Its real meaning is "the object that passed the assertion", and `undefined` means two different things (read failed, or assertion vetoed). The name now carries that, and half the explanatory comment stops being necessary.

### F7 · Low · **T6 did not pin *which* three policies**

`expect(sealed.size).toBe(3)` stays green if a refactor emits zero gateway policies and three from another family. Now asserts the sorted names.

### F8 · Low · **T5/T5b/T5c were declared out of order**, and T5b's non-redundancy was undocumented

Reordered. T5b (`{ code: 'ECONNRESET' }`) and T5c (codeless `Error`) look redundant given the total catch, but every draft that regressed this did so by keying on `code`; the codeless shape is the one that broke the Deployment test. Now stated in the test.

### F9 · Medium · **The deploy guide implied NetworkPolicy drift is visible**

It said drift "is still visible: the recipe `status` reports the workload `degraded`". True for a Deployment. For a NetworkPolicy there is no replica health to degrade, and nothing detects it — verified: `pruneStaleInternalDependencyPolicies:4476-4511` compares *names* under a `labelSelector`, never content; `deploy/scripts/verify-networkpolicies.sh` is deploy-time and checks existence only, for the overlay's policies rather than WRC's; there is no audit CronJob; HCC does not watch WRC's policies.

### F10 · Medium · **The guide's list of gated objects read as exhaustive**

It is not. Now stated explicitly:
- `wl-egress-*` routes through the same method but carries a `clerum.io/egress-fqdn-resolved-at` timestamp regenerated every pass, so **its hash never matches and the gate never closes for it**. What keeps it quiet is #299's `egressWriteNeeded`.
- `ui-egress-*` and `coordinator-to-gfs` do not use the spec-hash at all (#299 and #579).
- The webhook gateway's **own** ConfigMap, Deployment and Service remain ungated.

### F11 · Medium · **The PR body's arithmetic was impossible**

It claimed "36 700 → 24 400 requests/hour, a 33 % reduction" with 12 200 `PUT`/h against a measured 12 450. In a create-first design every `PUT` is preceded by a `POST`, so `POST ≈ PUT` and those figures cannot coexist. Corrected in the body to **24 900 → 12 450, a 50 % reduction**, with each number's provenance stated: 12 450 is measured, the `POST` count is inferred by construction.

## 2B — Deliberately deferred to #581, with reasons

### D1 · Medium · REPRODUCED · The ownership veto loses its identity if the **second** read fails

Veto raised by read #1 → `applyIsNoop` swallows it by design → `replaceFn` re-reads to re-assert → that read returns 503 → the **503** propagates, not `NetworkPolicyOwnershipConflictError`. The write never happens (the foreign object is safe), but `reconcileInternalDependencyPolicies:4449` classifies by `instanceof`, so the CRD condition reads `PolicyApplyFailed` instead of `OwnershipConflict`. The operator loses the signal that another controller owns the object.

This is the structural residue of leaving `applyIsNoop` alone. It disappears under #581's contract, where `assertExisting` runs once, outside the gate's try/catch — not as a separate fix.

### D2 · High · REPRODUCED · Drift no longer auto-reverts

A policy mutated to `podSelector: {}` + `ingress: [{}]` with the annotation left intact survived ten reconcile passes with **zero** writes; the pre-#580 code repaired it on the first. This is the accepted, documented cost of any idempotency gate, and #581 bounds it with `resyncAfterMs`.

### D3 · Medium · Freezing the annotation defeats the gate at zero cost

`computeSpecHash` excludes its own annotation, so the stamp already on a live policy *is* the hash WRC will recompute while the recipe is unchanged. Read, edit the `spec`, leave the annotation alone: skipped forever. Permission required is `update`/`patch` on `networkpolicies` in the namespace — held by HCC and WRC, **not** by tenant workload service accounts. Integrity and auditability, not a tenant escape. **#581 does not currently mention this**; reported on the issue.

### D4 · Medium · `wl-egress` cluster-local: this gate can cancel a stronger one

For `wl-egress-*` with **only cluster-local** bindings, `wlStateAnnotations` stays `undefined`, the manifest is deterministic and the hash is stable — so the gate can close. But `egressWriteNeeded` (`:4758`) compares *enforced content* and may already have decided a write is due. A weaker gate downstream then vetoes it.

Scope is narrow: with external egress the `resolved-at` timestamp changes the hash every pass and the gate never closes. Two of the four agents placed this in #581 (PR-C touches this exact method and should route the family to `always-write`); one placed it here. I left it documented rather than adding a special case that PR-C would immediately undo. **This is the judgement call in this PR most worth a second opinion.**

### D5 · Medium · Convergence is proven only for the gateway

T1/T2/T3 prove the gate *mechanism* with a synthetic policy. T6 proves *builder determinism*, but only for the gateway family. The other four families that reach `applyNetworkPolicy` are deterministic by inspection, not by test. The savings figure therefore rests on inspection for four of six families.

### D6 · Low · `ownerReferences` excluded from the hash

Removing an ownerRef does not flip the hash, so the policy is never rewritten and survives deletion of the recipe. Not confirmed whether the builders set ownerRefs at all.

### D7 · Low · Untested branches

Create failing with a non-409; the gate read failing **and** the re-read failing; `replace` itself failing. The contract in #581 reorders these paths.

## 2C — Corrections to #581 itself

Posted on the issue (#581, comment `5525521745`), since six PRs are specified against that inventory.

1. **"The twelve callers without a gate are *exactly* the ones that do something else in `replaceFn`" is false for three of twelve** — `:2982`, `:2999`, `:3040` do nothing else, and they are precisely the three this PR fixes. The issue's own table marks all three "—" in the *Assertions in the write path* column. This sentence is the issue's causal explanation, so the "exactly" has to go.
2. **`assertClusterEnforcesExternalEgress` is not in `ui-egress`'s write path** — it is called at `:4174`, before the `createOrReplace` at `:4207`. It never receives `existing`, so it does not migrate to the `assertExisting` slot.
3. **`applyNetworkPolicy (6 call sites)` is the pre-#580 count** — 7 after this PR; totals become 16 sites / 6 gated, and every line number shifts.

Plus two design additions: `resyncAfterMs` of one hour is not sufficient for a NetworkPolicy and the default should not be uniform; and recomputing `computeSpecHash(existing)` from the live object closes failure mode 4 without SSA — which is what HCC already does for the same kind (`host-context-controller/src/utils.ts:471` → `networkPolicyMatchesDesired`).

---

# Part 3 — Tests

Nine, each with the mutation that kills it. All ten declared mutations were run during review and each killed its target.

| test | pins | dies to |
|---|---|---|
| T1 | no replace on a matching hash; the skip line is emitted; the stamp is the **exact** hash | dropping the 4th argument; silencing the log |
| T2 | replaces on a differing hash, one read, live `resourceVersion` | forcing `specHashUnchanged` true; dropping the read cache (→ 2 reads) |
| T3 | one write to seal a policy with no hash, then skips | skipping when the annotation is absent |
| T4 | a matching hash cannot shortcut the ownership assertion; the rejection is the conflict **class** | assertion only in `replaceFn`; cache assigned before the assertion |
| T4b | an intdep policy WRC **already owns** passes the gate in one read | an assertion that vetoes every intdep policy |
| T5 / T5b / T5c | HTTP, transport and **codeless** read failures all fall open to the replace | reintroducing the narrowed catch |
| T6 | the gateway's three policies converge, **and pass 2 actually ran** | reverting the `:3040` hunk; making the builder non-deterministic; **removing execution entirely** |

The most informative result: with the *faithful* T4 mutation — the assertion still guards the write, it merely stops running on the gate's read — **T4 is the only test in the repo that catches it**. The preexisting `fails closed instead of replacing an existing non-WRC internal-dependency policy` stays green.

The skip line `unchanged (spec-hash match); skipping update` is load-bearing: it is what keeps the ~16 s cadence visible once the writes stop. A silent gate would make the symptom of #460 invisible.

---

# Part 4 — Gates

| gate | result | exit |
|---|---|---|
| `workflow-recipes` `npm test` | **2592 passed** / 29 skipped (+9 over `dev`'s 2583) | 0 |
| `npx tsc --noEmit` (includes tests) | clean | 0 |
| `npm run build` (`tsconfig.build.json`, excludes tests) | clean | 0 |
| `packages/network-policy-core` | **42 / 42** | 0 |
| CI on `223162a7` | green | — |

All exit codes read from a captured log, never from a wrapper's notification. `total > 0` verified on every suite. `npm run build` uses `tsconfig.build.json`, which excludes tests; CI's separate `tsc --noEmit` does not — both were run, because a previous PR went red twice on exactly that gap.

`npm run lint` is **not runnable** in this package: ESLint v9 flat-config has not been migrated. Preexistent and unrelated to this PR, but stating it rather than implying lint passed.

---

# Part 5 — What I would like scrutinised

1. **D4** — the `wl-egress` cluster-local case. Two agents said defer, one said fix here. Is documenting it and leaving the regression to PR-C the right call, or should this PR route that family to `always-write` now?
2. **F1's fix** — is the read a sufficient liveness witness, or can it be satisfied without the gate running?
3. **F5** — `applyIsNoop` is shared by all six gated classes. Is a `warn` on every pre-read failure the right level, given it now fires for Deployment, Service, CronJob, Job and DaemonSet too?
4. **The decision not to touch `applyIsNoop`'s catch.** D1 is its known cost. Is accepting a misclassified CRD condition until #581 lands acceptable, or does it warrant a narrower fix now?
5. **The four measured numbers** — 12 450 `PUT`/h, 63 policies, 50 at ~214/h, the unchanged `resourceVersion` — came from read-only audit-log queries against `clerum-dev`, not from anything reproducible in CI. They are the weakest evidence in this PR.

zach88:
Took a deep pass through this one across correctness, tests and security. The core change is solid: the "no PUT without an ownership assertion on the exact version being written" invariant holds on every path I traced, the WRC↔HCC ownership boundary is intact (actually reinforced), and the nine new tests are genuinely good. Nothing open blocks the merge; what's left is doc/log polish.

**🟡 Approve with nits**
- Optional / non-blocking: 4

One thing worth flagging so it's on the record, not missed: the spec-hash gate turns off drift auto-revert for NetworkPolicies (before this PR `applyNetworkPolicy` rewrote every ~16s and reverted any hand-edit; now a matching `spec-hash` annotation skips the write even if the live `spec` was altered). That's a real defense-in-depth reduction, but it only lands for someone who already holds RBAC to edit NetworkPolicies, the body documents it as "permanent and silent", and the content-based enforcement for NP is deferred to #581. Reviewed and accepted as-is pending #581, so it doesn't gate here.

## Optional (non-blocking)

1. **[R1-M1] Doc overstates what `verify-networkpolicies.sh` covers.** `docs/deploy/workflow-recipes-guide.md:340` says the script "only checks that the overlay's policies exist". It actually also does content drift detection (`equivalent_policy_specs?` → `drift.txt` → `exit 1`), so a reader who opens the script will think the doc is wrong. The paragraph's conclusion (nothing sweeps drift on WRC-managed NPs) is still correct, just for a different reason: recipe-derived policies (`ui-ingress-*`, `wl-ingress-*`, `wr-intdep-*`, gateway) don't live in any overlay, so the script never evaluates them. Suggest rewording to something like "the script only covers overlay-rendered policies; WRC recipe-derived policies aren't in any overlay, so it never inspects them (neither existence nor drift)". Cheap fix and it keeps operators from reading that NP drift is visible when it isn't.

2. **[R1-L1] An ownership veto is logged as a read failure.** `workflowRecipeReconciler.ts:3781` + `:4928`. For a foreign-owned `wr-intdep-*` policy, the gate read throws `NetworkPolicyOwnershipConflictError`, `applyIsNoop` swallows it and logs `"idempotency pre-read failed, falling through to update"` (misleading, it was a veto not a read failure), then `replaceFn` re-reads and re-asserts and throws loud. End state is correct (loud conflict, no write), so this is just diagnostic noise plus one extra read on an already-alarming path. Optional: special-case `instanceof NetworkPolicyOwnershipConflictError` in the warn, and skip the warn when it's a veto.

3. **[R1-L2] PR body table calls it "T7" but the code (and your own "T4b added" note) call it T4b.** There's no T7 in the diff; anyone mapping the table to the code won't find it. Rename the row to T4b.

4. **[R1-L3] Stale line refs in the body.** `applyNetworkPolicy (:4879)` is at `:4884`/`:4906` in the head. Minor prose drift (`:3040` for the gateway hunk still matches).

## What checked out clean

- **Central invariant.** `assertedExisting` is assigned only after `assertInternalDependencyPolicyOwnership` returns without throwing, and `replaceFn` reuses that exact object's `resourceVersion`, so the PUT carries the version the gate evaluated. A foreign-owned policy with a matching hash can't be skipped: the assertion runs inside `readExisting`, before `specHashUnchanged` is computed. Veto path re-reads and re-asserts (T4); own intdep policy converges (T4b).
- **Stamp/hash order.** `stampSpecHash` runs before `createFn`; the `resourceVersion` mutation in `replaceFn` happens after and is in `VOLATILE_METADATA_FIELDS`, so it never influences the hash. `spec-hash` and `pre-deploy` annotations correctly excluded.
- **Family coverage.** `ui-ingress`, `wr-intdep-*` egress/ingress, `wl-ingress`, oauth-broker-egress and the 3 gateway policies all funnel through `applyNetworkPolicy`. `wl-egress` routes through it but its manifest carries `clerum.io/…resolved-at` (part of the hash), so the gate never closes for it and `egressWriteNeeded` is the real gate. Body claim confirmed accurate. Gateway ConfigMap/Deployment/Service and `ui-egress`/`coordinator-to-gfs` stay ungated, doc documents all of them correctly.
- **Gateway migration** changes only the log label (nothing consumes it); create body still stamped, error handling preserved by `createOrReplace`.
- **No secret leak** in the new `console.warn(..., error)`: the k8s client error doesn't store the request `Authorization` header, and NetworkPolicy specs carry no secrets.
- **Tests.** Fixtures derived from the real producer (`computeSpecHash`/`stampSpecHash`), no hand-written hashes; T4 asserts the error class not just the message; T6 carries a real liveness witness; T5/T5b/T5c aren't redundant. All 9 new tests fail against the merge-base (verified, so no vacuous ones), and the full suite is 319/319.

alfredolopez80:
@zach88 All four nits from your round are in `80218fc4`. Branch is level with `origin/dev`, 37/37 checks green.

## 1 — `verify-networkpolicies.sh`: the doc's reason was wrong

You were right, and the failure mode was worse than a plain inaccuracy. I wrote that the script "only checks that the overlay's policies exist"; it compares specs too (`equivalent_policy_specs?` → `drift.txt` → exit 1). The paragraph's *conclusion* was correct and its stated *reason* was false — which is the worst of the three outcomes, because a reader who opens the script concludes the doc is unreliable.

Corrected to what is actually true: the script does detect content drift, but only over policies an overlay renders. Recipe-derived policies are authored by the reconciler and live in no overlay, so it never evaluates them at all.

## 2 — the ownership veto announced as a read failure

Taken. An ownership veto is not a read failure, and `applyIsNoop` was saying it was.

For a foreign-owned `wr-intdep-*` policy the gate's read **succeeds** — the caller's assertion on the result is what throws. So the old warn put a wrong diagnosis in the log immediately before `replaceFn` re-read, re-asserted, and raised the conflict with its own accurate message. Two lines about the same event, the first one wrong.

The warn is now skipped for `NetworkPolicyOwnershipConflictError` specifically:

```ts
if (!(error instanceof NetworkPolicyOwnershipConflictError)) {
  console.warn(`[WR-Reconciler] ${label}: idempotency pre-read failed, falling through to update:`, error)
}
return false
```

Nothing is swallowed: the veto still reaches the caller loud, which is what T4 has always pinned.

**T4 now also asserts that line is absent**, scoped to the string rather than to `console.warn` at large — other warnings on this path are legitimate, and a blanket assertion would break on them for unrelated reasons. Mutation-verified: reinstating the unconditional warn kills T4 and **only** T4.

**Not taken from the same finding — the extra read on the veto path.** That is the known cost of leaving `applyIsNoop`'s catch total, documented in the PR body, and it disappears under #581's contract rather than by a local change here. Fixing it locally would mean narrowing the catch, which is the thing this PR deliberately does not do.

## 3 and 4 — body drift

Both real. The test table said `T7` where the code says `T4b` (renamed during implementation, the table was not), and `applyNetworkPolicy` is at `:4906`, not `:4879`. Corrected in the body.

## Gates

Read from captured logs, never a wrapper's exit:

| | |
|---|---|
| `npm test` | **2592 passed** / 29 skipped |
| `npx tsc --noEmit` | exit 0 |
| `npm run build` | exit 0 |
| GitHub checks | 37 / 37 |

https://claude.ai/code/session_016eFKFk67BF9Gdqy5PtsSUV

zach88:
Round 2 pass. This one's short: the delta since round 1 is a single product commit (`80218fc41`) that closes the round-1 findings, and it lands clean. The rest of the range between heads is just a `dev` merge that doesn't touch `workflow-recipes`, so no semantic conflict in the reconciler. Ran a correctness + security + collateral pass over it — all three agree.

**🟡 Approve with nits**

## Round-1 findings — all resolved
- **[R1-L1] veto no longer announced as a read failure** — the `instanceof NetworkPolicyOwnershipConflictError` guard in `applyIsNoop` (`workflow-recipes/src/reconciler/workflowRecipeReconciler.ts:3782`) is precise. Verified the end-state is preserved: on a veto `applyIsNoop` still returns `false`, `assertedExisting` stays uncached, `replaceFn` re-reads/re-asserts and the conflict surfaces loud (`console.error` + the `OwnershipConflict` CRD condition). A genuine read failure (403/500/ECONNRESET) isn't that class, so it still hits the warn — T5/T5b/T5c pin that. T4's new assertion is scoped to the one string, which is the right call.
- **[R1-M1] doc claim on `verify-networkpolicies.sh`** — rewrite is accurate. Confirmed against the script: it does compare specs (content drift → `drift.txt` → exit 1), but only over overlay-rendered policies; recipe-derived ones (`ui-ingress-*`, `wl-ingress-*`, `wr-intdep-*`, gateway trio) live in no overlay, so it never evaluates them.
- **[R1-L2] / [R1-L3] body drift** — table says `T4b`, `applyNetworkPolicy` ref is `:4906`. Fixed.

## Optional (non-blocking)
PR can merge without this — author's call / follow-up.

1. **[R2-L1] `warnSpy` restored outside a `try/finally`** — `workflow-recipes/src/reconciler/reconciler.test.ts:925` (spy) / `:959` (`mockRestore`). The root `describe`'s `beforeEach` only does `vi.clearAllMocks()`, which clears `mock.calls` but doesn't restore the spy's implementation. If any assertion between `:937-946` threw, the no-op `console.warn` stub would leak into later tests. Benign today (only bites once T4 is already red, so it can't mask a green), and T1 (`:863-865`) already uses `finally` — so it's just an inconsistency. Fix: wrap from `:925` in `try/finally`, or register `vi.restoreAllMocks()` in an `afterEach`.

## What checked out clean
- **Central invariant holds** — no PUT without an ownership assertion on the exact version written; the round-2 change only touches what gets logged in the exception branch, not the gate logic.
- **T3 verified executably** — copied the new T4 assertion onto the round-1 head (no guard) and it fails there (`AssertionError: … not to contain 'idempotency pre-read failed'`). Genuine regression test, not a golden.
- **Security observability preserved** — a `wr-intdep-*` hijack attempt still produces a `console.error` + the `OwnershipConflict` CRD condition; only the diagnostically-false warn line goes away. No secret leak in the conditional warn or the test.
- **Suite** `reconciler.test.ts` → **319/319** green.

Heads up on context (not a round-2 finding): the accepted High from round 1 — the gate turning off drift auto-revert for NetworkPolicies — is still a conscious tradeoff tracked for #581; the round-2 delta doesn't touch or worsen it.

alfredolopez80:
@claude Please perform a fresh, full re-review of the current PR head
`8ddbe65ea350313cdc02baa81bd4f5fa8ae72408` against base
`c28fa3b54c691292f97c1a9128738399d501c1ec`.

Do not treat the previous review conclusions or the current PR description as
authority. Re-derive the behavior from the current diff and verify every claim
below. Please report line-specific findings with severity and explain whether
each is introduced by this PR. If no material issue remains, say so explicitly
and state whether you recommend merge.

## 1. Original issue and intended outcome

Verify that the PR still solves #575: WRC must stop issuing redundant
NetworkPolicy updates for objects that are genuinely converged, without
suppressing repair of live enforcement drift or weakening ownership/lifecycle
boundaries.

The final design is intentionally different from the first spec-hash-only
revision. The legacy hash may remain as provenance, but live normalized state
must be the no-op authority.

Please verify:

- read-first behavior and create only on a confirmed 404;
- canonical live spec comparison, including Kubernetes defaults;
- comparison of WRC-authored metadata without churn from admission-added
  metadata/finalizers;
- use of the same live snapshot/resourceVersion for validation and replace;
- no first-deploy annotation-stamping write wave;
- stable no-op behavior after convergence;
- actual drift still forces repair.

## 2. Revalidate the prior public review history

Please revisit the prior review threads rather than assuming their old
dispositions still apply:

- @alfredolopez80 full review request:
  https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5525758278
- @zach88 round 1:
  https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5528809132
- @alfredolopez80 round-1 disposition:
  https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5530089270
- @zach88 round 2:
  https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5535344956

In particular, verify the final status of every item raised there:

- T6/liveness-witness and exact gateway-family coverage;
- the positive owned-internal-dependency path (T4b);
- ownership-conflict error class/condition semantics;
- exact rather than truthy hash assertions;
- deploy-guide accuracy for overlay versus recipe-authored policies;
- the misleading ownership-veto/read-failure diagnostic;
- stale test/body references;
- `warnSpy` restoration/isolation from the round-2 review.

The older reviews accepted several risks as follow-up work. They should not be
assumed accepted now; verify whether the current head actually closes them:

- a matching/frozen hash hiding live NetworkPolicy drift;
- ownership-veto identity being lost to a second read;
- cluster-local workload-egress content repair being cancelled by a weaker
  downstream gate;
- stale/missing ownerReferences and garbage-collection lineage;
- insufficient coverage of NetworkPolicy families and create/read/replace
  failure branches.

Also identify any prior conclusion that is now obsolete because the
implementation was redesigned.

## 3. Revalidate the Codex audit and corrective implementation

Codex independently audited the prior head and attributed five PR-introduced
findings:

1. **High:** the spec-hash seal could hide divergent live enforcement, including
   a workload-egress path where a stronger content decision was vetoed.
2. **Medium:** webhook-gateway policies could retain a stale or missing current
   WorkflowRecipe owner UID and later be removed by garbage collection.
3. **Medium:** a raw Kubernetes error was emitted outside WRC's structured,
   redacting logger boundary.
4. **Low:** an ownership veto could be swallowed and replaced by a second-read
   infrastructure error, losing its conflict classification.
5. **Low:** a warning spy could leak when an assertion failed before restore.

Please verify that all five are genuinely fixed in production code and pinned
by meaningful tests, not merely hidden by changed expectations.

During implementation, further edge cases were found and corrected. Please
audit these too:

- workload-egress early-return bypass of metadata, owner, and lifecycle drift;
- write churn from admission-added labels/annotations/finalizers;
- create-409 and replace-404/409 races;
- terminating-object ordering relative to ownership checks;
- workflow preparation versus ACTIVE non-workflow handling of transient
  NetworkPolicy infrastructure failures;
- transport errors that expose only a message/code shape;
- unexpected ownerReferences on historically ownerless policy families;
- structured safe logging and documentation matching the actual state machine.

Please pay special attention to:

- foreign versus repairable ownership;
- same-recipe stale/missing gateway owners;
- desired gateway owner identity/UID prerequisites;
- terminating resources;
- optimistic concurrency and TOCTOU behavior;
- whether any catch/fallback can turn a trust-boundary rejection into a write;
- whether retryable errors preserve phase/status/requeue semantics;
- whether sanitized logs can expose request/response bodies, headers,
  credentials, or private endpoints;
- whether normalization can cause either a permanent false no-op or a write
  storm.

## 4. Regression and blast-radius review

Please inspect the complete base-to-head diff, not only the newest module, and
look for new correctness, security, operability, performance, and test-quality
problems.

Confirm or refute the claimed blast-radius boundary:

- WRC NetworkPolicy convergence and its tests/docs/version are the intended
  product scope;
- no CRD, public API, database schema/migration, RBAC, dependency graph, or
  production manifest contract changed;
- no production source in HCC, Control API, External REST API, RPC Proxy,
  MCP Host, GFS, Control UI, or Desktop changed;
- other Kubernetes object kinds retain their existing generic idempotency
  behavior;
- the new read-first path does not introduce an unacceptable API-load,
  availability, or rollout regression relative to the problem it solves.

Please also check that the package-lock change is version-only and that the
documentation does not promise coverage the implementation does not provide.

## 5. Sanitized validation evidence

The following is a public-safe summary only; no local paths, environment
identifiers, endpoints, credentials, or raw logs are being published:

- Local exact-head developer gate: **T0 PASS, T1 PASS, T2 PASS**.
- T1 Real PostgreSQL: **35/35 selected files, 163/163 tests, zero pending**.
- Deployed NP-08 Host-to-HCC authorization journey: **PASS**.
- All required deployments in the isolated development profile were Ready.
- Focused convergence/reconciler tests: **341/341**.
- Full workflow-recipes suite: **2,614 passed / 29 skipped**.
- Shared network-policy-core: **42/42**.
- Cross-service suites for External REST, RPC Proxy, MCP Host, HCC, Workflow
  SDK, Approval Reader, Control UI, and Desktop passed.
- Desktop ran with Node 24 and a verified Electron runtime.
- GitHub CI on the exact head: **37/37 reported checks passed**, including WRC,
  Real PostgreSQL, E2E node-unit, leak guards, manifest/hygiene checks, all
  service/package jobs, and CodeQL.

No browser product surface changed, so Playwright was not presented as executed.
No manual destructive live injection of drift-with-preserved-hash or an owner-GC
race was performed; those scenarios rely on focused deterministic tests. Please
say whether that leaves a material verification gap and, if so, specify the
smallest public-safe test needed before merge.

## Requested conclusion

Please return:

1. findings first, ordered by severity, with current-head line references;
2. a point-by-point disposition of the prior @zach88 and @alfredolopez80 items;
3. a point-by-point disposition of the five Codex findings and the additional
   edge cases;
4. a blast-radius assessment;
5. an explicit final verdict: **merge**, **merge after fixes**, or **do not
   merge**.

alfredolopez80:
@claude Please perform a fresh, exhaustive review of PR #580 at exact head
`8968b58499fd1e3614bf69284db4b384187a4285` against current `dev`. Please derive
the full base-to-head diff yourself; do not rely only on the earlier review of
`8ddbe65e`.

This pass must explicitly revalidate:

1. Every prior review thread and concern raised by @zach88 and @alfredolopez80,
   including whether each resolution remains correct after the two coverage and
   retry commits.
2. The original Codex findings and fixes: live state rather than the legacy hash
   as convergence authority; safe webhook owner-UID repair and foreign-owner
   rejection; structured/sanitized logging; ownership vetoes not hidden by a
   second read; and isolated test spies.
3. The independent follow-up findings and fixes:
   - retryable NetworkPolicy lifecycle/API races now return a deterministic
     `requeueAfterMs` in both workflow and non-workflow paths;
   - watcher integration and a live terminating-policy journey prove recovery
     without a second parent-spec event;
   - no-churn assertions require an exact structured witness bound to policy,
     namespace, and family before checking stable `resourceVersion`;
   - OAuth/webhook cleanup propagates failures and preserves the original exit
     status.
4. All six in-scope WRC NetworkPolicy families and nine concrete policies:
   UI ingress, workload ingress/egress, internal-dependency ingress/egress,
   OAuth-broker egress, and the three webhook-gateway policies.
5. New-bug and blast-radius risk: other WRC objects, retries/status semantics,
   Kubernetes ownership/GC, write churn, and every production service outside
   WRC. Confirm that no CRD, RBAC, schema, migration, production manifest,
   dependency graph, or browser behavior was unintentionally changed.
6. Test quality: look for vacuous assertions, hidden skips, mock leakage,
   nondeterminism, unsafe E2E shortcuts, flaky polling, incomplete cleanup, and
   public CI claiming more than its static contract executes.

Public-safe exact-head evidence available for cross-checking:

- T0 PASS; full WRC suite 2,627 passed / 29 expected skips; build/typecheck PASS.
- Independent focused verification: 462/462.
- T1 PASS: 35/35 Real PostgreSQL files, 163/163 tests, zero pending.
- T2, deployed NP-08 authorization, required deployment readiness, and
  user-facing health: PASS.
- Live NetworkPolicy aggregate: service routes 18/18, internal dependency
  28/28, OAuth 41/41, webhook 26/26.
- Independent local audit at this SHA: 0 blocker / critical / high / medium /
  low findings.
- Playwright was not run because this PR changes no browser surface.

Please report every finding with severity, file/line, concrete failure mode,
and recommended fix. End with explicit answers to: prior issues fixed, no new
bugs found, blast radius acceptable, tests sufficient for the stated scope,
and merge yes/no. GitHub CI is still required to complete on this exact SHA.

alfredolopez80:
@claude The previous review job failed before analysis. Please perform a fresh,
exhaustive review of PR #580 at exact head
`b5771338952888c020e09ad7d8e0f769dc981bcd` against current `dev`. Please derive
the full base-to-head diff yourself; do not rely only on the earlier review of
`8ddbe65e`.

This pass must explicitly revalidate:

1. Every prior review thread and concern raised by @zach88 and @alfredolopez80,
   including whether each resolution remains correct after the two coverage and
   retry commits.
2. The original Codex findings and fixes: live state rather than the legacy hash
   as convergence authority; safe webhook owner-UID repair and foreign-owner
   rejection; structured/sanitized logging; ownership vetoes not hidden by a
   second read; and isolated test spies.
3. The independent follow-up findings and fixes:
   - retryable NetworkPolicy lifecycle/API races now return a deterministic
     `requeueAfterMs` in both workflow and non-workflow paths;
   - watcher integration and a live terminating-policy journey prove recovery
     without a second parent-spec event;
   - no-churn assertions require an exact structured witness bound to policy,
     namespace, and family before checking stable `resourceVersion`;
   - OAuth/webhook cleanup propagates failures and preserves the original exit
     status.
4. All six in-scope WRC NetworkPolicy families and nine concrete policies:
   UI ingress, workload ingress/egress, internal-dependency ingress/egress,
   OAuth-broker egress, and the three webhook-gateway policies.
5. New-bug and blast-radius risk: other WRC objects, retries/status semantics,
   Kubernetes ownership/GC, write churn, and every production service outside
   WRC. Confirm that no CRD, RBAC, schema, migration, production manifest,
   dependency graph, or browser behavior was unintentionally changed.
6. Test quality: look for vacuous assertions, hidden skips, mock leakage,
   nondeterminism, unsafe E2E shortcuts, flaky polling, incomplete cleanup, and
   public CI claiming more than its static contract executes.
7. The final head-only delta after the prior request: verify that using the
   reserved test finalizer `e2e.invalid/hold-deletion` preserves the terminating
   race, that helper mode `100755` is correct, and that both public guards are
   now satisfied without weakening them.

Public-safe exact-head evidence available for cross-checking:

- T0 PASS; full WRC suite 2,627 passed / 29 expected skips; build/typecheck PASS.
- Independent focused verification: 462/462.
- T1 PASS: 35/35 Real PostgreSQL files, 163/163 tests, zero pending.
- T2, deployed NP-08 authorization, required deployment readiness, and
  user-facing health: PASS.
- Live NetworkPolicy aggregate: service routes 18/18, internal dependency
  28/28, OAuth 41/41, webhook 26/26.
- Independent local audit at this SHA: 0 blocker / critical / high / medium /
  low findings.
- Public identifier and repository-hygiene guards: PASS locally on this SHA.
- Playwright was not run because this PR changes no browser surface.

Please report every finding with severity, file/line, concrete failure mode,
and recommended fix. End with explicit answers to: prior issues fixed, no new
bugs found, blast radius acceptable, tests sufficient for the stated scope,
and merge yes/no. GitHub CI is still required to complete on this exact SHA.

alfredolopez80:
@claude Please perform a fresh, exhaustive review of PR #580 at exact head
`8d4384b82885c3feb3c1bac8fd36a2546cae8365` against base `dev` at
`c28fa3b54c691292f97c1a9128738399d501c1ec`. Declare any subsequent change to
the PR head or base before reviewing, and identify the exact pair you audit.
Re-derive the complete base-to-head diff. Earlier approvals, comments and test counts are
context, not proof that this revised implementation is correct. This is a
review-only request: do not modify files, commits, settings or merge the PR.

The issue remains #575: eliminate redundant WRC NetworkPolicy writes for
genuinely converged objects without suppressing live drift repair or weakening
ownership, lifecycle and service availability.

## Prior review history

Please give an explicit disposition for every prior item, including obsolete
conclusions after the redesign:

- [Alfredo's original request](https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5525758278).
- [Zach round 1](https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5528809132).
- [Alfredo's round-1 resolutions](https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5530089270).
- [Zach round 2](https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5535344956).
- [Codex audit and redesign request](https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5541509769).
- [Previous coverage/retry re-review request](https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5544864190).

This includes T6/liveness, exact gateway-family coverage, the owned internal
dependency path, ownership-conflict semantics, exact assertions, overlay versus
recipe-policy documentation, ownership-veto diagnostics, stale references and
spy restoration. The earlier accepted hash-only drift tradeoff must not be
carried forward: live normalized state is now authoritative on reconciliation.

## Codex and independent Astra corrections to verify

1. Original Codex findings: intact hashes cannot hide live enforcement drift;
   webhook policies repair missing/stale same-recipe owner UIDs but reject
   foreign ownership; diagnostics stay in WRC's structured/redacting logger;
   an ownership veto cannot be replaced by a second-read error; failed test
   assertions cannot leak spies. Validate all six policy families, including
   the workload-egress prefilter and final apply using the same authority.
2. Retry and lifecycle: confirmed absence is required for create; ownership
   validation and replace use one live snapshot/resourceVersion; terminating
   objects and 404/409 races retain the bounded retry. Verify workflow and
   non-workflow status/requeue paths plus watcher recovery without another
   parent-spec event. Raw transient errors from the workload-egress pre-read,
   including HTTP 503 and ECONNRESET, must not terminalize the workflow; a
   permanent 403 must not become an endless retry.
3. Metadata ownership: drift repair must preserve external labels,
   annotations and finalizers while updating WRC-owned fields. Verify removal
   of obsolete owned egress metadata and the explicit owner-reference vetoes.
   The pre-read retry and external-metadata defects were reproduced as
   pre-existing; they were explicitly included in this corrective scope, not
   misattributed as new PR regressions.
4. Negative test attribution: remote executor errors, missing tools, silent
   failures and invalid responses must fail the harness, not count as denied
   traffic. Inspect nonce-bound probe results, narrow complementary fixture
   permissions and same-source allow/deny/allow controls, including OAuth and
   undeclared sibling egress. The BusyBox regression uses native verbose nc
   diagnostics and rejects post-connect read timeouts as denial evidence.
5. No-write evidence: the baseline starts before the parent trigger and
   observation covers the full window. Validate policy UID/resourceVersion,
   controller pod/container identity, logs from all replicas, exact recipe
   completion and rejection of writes even when a no-op also appears. Exact
   zero create/replace calls are asserted separately in unit tests; stable
   resourceVersion alone is not presented as proof of zero API writes.
6. Executable CI contract: run the real aggregate against instrumented children
   and verify context, order, completion receipts, each child failure, missing
   children and denied lease. Mutations replacing execution with `:` or
   swallowing errors with `|| true` must fail. Hermetic CI is not live CNI E2E.
7. Fixture safety: collision-safe creation, unique identities, UID-preconditioned
   deletion and owned dynamic port-forwards must remain fail-closed. Independent
   review found a replacement-parent cleanup race: verify parent ownership
   before and after child discovery, staged enrollment, no new discovery for
   an absent parent, and no child deletion after a parent veto/delete failure.
   Check corruption, absent-parent retries and cleanup error propagation.
   Also verify the webhook signing Secret carries its required recipe-owner
   label in the initial create body, not only the test run label. The real
   producer contract must reject removing that ownership and must keep Secret
   data out of the metadata ledger; WRC's fail-closed Secret boundary is intact.

## Public-safe exact-head validation

- T0 PASS; WRC 2,651 tests passed / 29 expected skips; build/typecheck PASS.
- All 75 hermetic contracts, shell checks and public-boundary checks PASS.
- T1 PASS: 35/35 selected physical files, 163/163 tests, zero pending.
- T2, deployed NP-08 authorization, required readiness and Health PASS.
- All four live NetworkPolicy suites PASS; OAuth 41/41 and webhook 24/24.
- Final fixture-resource and matching port-forward record inventories empty.
- Independent Astra code and local-evidence seal on this exact head: YES,
  with no confirmed pending in-scope defects. Public CI remains a separate gate.
- Playwright E2E NOT_RUN because no browser surface changed.

One earlier RPC unit run observed an unexplained 400/404 assertion for malformed
pagination. The focal file passed 47/47 and the complete suite passed 500/500,
including the subsequent full gate, without changes. The complete RPC tree is
identical to base. Preserve that uncertainty; do not invent a cause or assume
that a passing rerun erases the earlier result.

## New bugs, blast radius and evidence quality

Review the full diff for regressions, concurrency/TOCTOU, garbage collection,
retry/status behavior, rollout load, security, and tests that could pass
vacuously. Trace correctly configured UI/API, sibling, internal-dependency,
OAuth and webhook routes before and after repair, including negative controls
and recovery.

Confirm or refute the scope boundary: no CRD, RBAC, database schema/migration,
production manifest, dependency graph or browser behavior changed; no unrelated
service production source was changed; generic handling of other Kubernetes
object kinds is unchanged. Source-file scope alone is not proof of zero runtime
impact. WRC's patch version is now 0.1.346.

Report findings first with severity, current file/line, concrete failure mode,
base/head attribution and recommended fix. Then provide a disposition matrix
for the prior Zach/Alfredo, Codex and Astra items. End with explicit conclusions:
prior issues corrected; any new bugs; acceptable service blast radius; adequate
tests for the stated scope; and merge / merge after fixes / do not merge.

Treat T0, T1, T2, live NetworkPolicy E2E, public CI and Playwright as distinct
evidence lanes. Public summaries intentionally omit local paths, profiles,
endpoints, generated ports, credentials and raw operational logs. Do not
request or publish those sensitive details. Missing execution evidence is
unknown, never a passing result.

alfredolopez80:
@claude Please perform a fresh, complete re-review and independent audit of PR #580 at head `650f74e03125d68a98f22288215a920b1a2d9761`, against `dev` at `f8460e9d6c85eaf8f3666274b2376d1fae6795fd`. Previous favorable reviews must not be treated as approval of this head.

Review the full PR, not only the latest commit. Reconcile every finding from prior reviews by zach88 and alfredolopez80, the Codex fixes, and the independent Fable review summarized below. Classify each as fixed, still reproducible, pre-existing/out of scope, or unsupported, with code/test evidence.

Specifically verify:

1. Issue #575’s intended live convergence: equivalent policies avoid unnecessary mutations, real drift is repaired, and transient errors, creation races, termination and ownership conflicts preserve safe retry behavior.
2. Pre-DNS contraction preserves external metadata/finalizers, rejects foreign lifecycle owners, and retains subtraction ordering and UID/resourceVersion fencing, including read-back races.
3. The E2E observer detects contraction writes and read-retry events independently of no-op witnesses and stable UID/resourceVersion. Check historical/unrelated-event controls and the documented family coverage.
4. Fable H1/H2: composition tests positively witness final apply; T3 positively witnesses its second read. Verify that disabling those phases makes the tests fail.
5. Fable H3: the shared metadata prefilter also reaches the dedicated UI egress writer. Verify metadata/finalizer preservation and late foreign-owner/termination vetoes on its fresh read, without changing its dedicated cadence. Check the eight new UI tests and the live metadata-repair/traffic regression.
6. Independently reassess Fable H4–H6: observer coverage limits, conservative comparison behavior, intentional read-recovery writes and negative assertions. Do not accept prior dispositions without checking the code.
7. Audit blast radius and regressions for correctly configured existing services: UI/backend, workload dependencies, OAuth and webhooks; DNS/renewal semantics, tenant boundaries, permissions, metadata ownership and lifecycle behavior. Distinguish demonstrated compatibility from untested assumptions.

Reported validation for this exact head: CI 37/37 successful; WRC 2,807 passed with 29 expected skips; T0/T1/T2 passed; real PostgreSQL 164/164 tests across 35 files with no skips; NP08 authorization and all four NetworkPolicy service E2E suites passed. Playwright browser journeys were NOT_RUN. Treat these as supporting evidence, not a substitute for your independent audit.

`coordinator-to-GFS` remains outside this PR’s converted convergence path and is tracked by #579 and #581/PR-C. Check that the scope boundary is accurate; report any newly introduced cross-path effect.

Please return a merge-readiness recommendation, remaining blockers by severity, missing validation, and any required follow-up. Keep operational evidence private: do not request or publish credentials, cluster/profile identifiers, private endpoints, local paths or raw logs. Do not merge or enable auto-merge; the purpose is to support the required external approval.

zach88:
Took a full pass over this one across correctness, tests, security and collateral. Heads up that this head is a **complete redesign** of what the earlier rounds reviewed: the spec-hash gate is gone, replaced by a read-first live-convergence state machine (`networkPolicyConvergence.ts` + rewritten `applyNetworkPolicy`). The good news is that redesign closes the round-1 High **by construction** — the "drift auto-revert turned off, deferred to #581" risk no longer exists, live drift is compared on the full spec and repaired. The core logic and the ownership boundary hold up under reproduction; what's left is test-quality polish and a couple of small design calls.

🟡 **Approve with nits**
- **Optional / non-blocking: 9** (3 medium, 6 low)

Nothing. No blocker/critical/high, no security issue, and the collateral pass came back clean — zero regressions. The hard invariant ("no PUT without an assertion on the exact resourceVersion being written") holds on every path, the WRC↔HCC ownership boundary is deny-by-default and intact, and the ~12k-PUT/hour steady-state churn does not come back (all reproduced).

## Optional (non-blocking)

The PR is mergeable without these. The first three are the ones I'd actually spend the half hour on before merge, since they're the debt the next round would otherwise trip over.

1. **[R3-M1] No property test for the central idempotency invariant.** `workflow-recipes/src/reconciler/networkPolicyConvergence.test.ts`. The whole PR rests on "a converged policy produces zero writes", but the roundtrip `decide(family, desired, buildNetworkPolicyReplacement(desired, live)) === unchanged` isn't pinned as a test — it only lives in point cases and the body prose. I reproduced that roundtrip in a throwaway and it does return `unchanged`, so the invariant genuinely holds and nothing's hidden; it just isn't encoded. `fast-check` is already in the repo (`planContextReplacement.property.test.ts`), and a fuzz here would also have caught R3-L1. Suggest an idempotency/composability property over the two writers that share `buildNetworkPolicyReplacement`.

2. **[R3-M2] Hand-authored fixtures in the contract tests.** `networkPolicyConvergence.test.ts:12`/`:32` and `reconciler.test.ts:1668`/`:1683`/`:1696`. `desired` stands in for the real manifest builders and `live` is a hand-mutated `structuredClone(desired)`, so these certify your idea of the shapes rather than what the builders emit / what the apiserver defaults. It's mitigated (the ingress no-op cases use the real builders, and I reproduced the workload-egress no-op against the real producer), but `uiEgressMetadata.test.ts` already shows the right pattern — it derives `live` from `buildUiEgressNetworkPolicy`. Worth anchoring at least the core convergence cases to the real builder / an `ObjectSerializer` round-trip.

3. **[R3-M3] A fixture self-test goes green with a missing dependency.** `scripts/tests/test-wrc-fixtures.sh:88` and `:93` — `[[ "$(rg -c '^delete ' ... || true)" == '' ]]`. If `rg` (ripgrep) isn't a bash-resolvable binary, the substitution yields empty via `|| true`, the check passes unconditionally, and the script still prints PASS / exits 0. Reproduced on a box without a visible `rg`: `line 88: rg: command not found` yet `WRC_FIXTURES_CONTRACT_PASS`. The `must_reject` on the line above still guarantees the "refuse deletion" property, so only the "didn't even attempt the DELETE" sub-check is silently disabled — but that's the file that claims to be hermetic and POSIX-only. GitHub's ubuntu runners ship ripgrep so CI isn't a false green today, but any dev box without it is. Fix: `! grep -q '^delete ' "$TEST_ROOT/calls"` (note `grep -c` prints `0`, not empty, so a naive `s/rg/grep/` breaks the happy path).

4. **[R3-L1] Defaulted `policyTypes` isn't sorted.** `networkPolicyConvergence.ts:166-170`. The defaulted branch returns `['Ingress','Egress']` unsorted while the explicit branch sorts, so a side that omits `policyTypes` and one that carries it explicitly never compare equal → `replace/live-drift` every reconcile. Unreachable today (every builder emits an explicit single-element `policyTypes`), and the bug direction is always toward false drift, never toward hiding real drift, so no correctness/security exposure — but it's a one-line foot-gun a future builder would silently trip. Sort the defaulted value too. Reproduced with a throwaway.

5. **[R3-L2] UI-egress wasn't migrated to the convergence path.** `workflowRecipeReconciler.ts:4613` (`reconcileUiEgressPolicy` still calls the legacy `applyOwnedRecipeNetworkPolicy`). Not a regression — that path keeps all four guarantees — but it's the one recipe-scoped family that doesn't get the unified convergence decision (owner-repair, conflict reasons, terminating classification), so it'll read as inconsistent next time someone looks. Either migrate it or drop a line saying why it stays out.

6. **[R3-L4] The stuck-terminating requeue isn't actually bounded.** `workflowRecipeReconciler.ts:5672`/`2163`/`3003`. A NetworkPolicy stuck `Terminating` degrades + requeues at a flat 5s forever with no backoff/cap. It self-heals once deletion completes and mirrors the existing DNS-SERVFAIL cadence, so it's not a regression — just flagging that the "bounded retry" the description mentions isn't bounded here. Conscious call.

7. **[R3-L3] Dead branch.** `workflowRecipeReconciler.ts:5697` — the `retry` action is unreachable because the `deletionTimestamp` guard above throws first. Drop it or mark it belt-and-suspenders.

8. **[R3-L5] Unused export.** `networkPolicyConvergence.ts:184` `normalizeNetworkPolicyForComparison` has zero consumers repo-wide and no test. Remove it, or use it in place of the inlined calls and cover it.

9. **[R3-L6] Test asserts a veto by message, not class.** `uiEgressMetadata.test.ts:166-168` matches an error-message substring where `reconciler.test.ts:1760` asserts the error class. Prefer class + reason for a refactor-resilient contract.

Two design choices I want on the record but that I'm not filing as findings: `applyNetworkPolicy` forces one replace on an `unchanged` decision when the first pre-read failed transiently and recovered (`:5698`, `reason:'read-recovery'`) — defensible, a flaky read shouldn't be trusted into a silent skip; and the label-owned families go fully deny-by-default (any foreign `ownerReference` → conflict, WRC stops repairing drift and surfaces it loudly rather than blindly overwriting a foreign-owned object). Both are intentional and safe.

## What checked out clean

- **Round-1 High closed by construction** (reproduced ×3): emptied `podSelector` / widened rules → `replace/live-drift`. Full spec compared; normalization only neutralizes k8s defaults (policyTypes, `protocol:'TCP'`, empty arrays, `_from`↔`from` symmetrically).
- **Version fencing**: every `replace` writes `buildNetworkPolicyReplacement(desired, existing)` with the resourceVersion from the same read snapshot, guarded by a `!resourceVersion` throw. No new TOCTOU — a concurrent race is 409→retry, not a silent overwrite.
- **Ownership boundary**: foreign controller owner → `conflict`; `repairable-owner` is reachable only for `webhook-gateway` with all four identity labels present/equal AND the live controller owner matching apiVersion/kind/name (a stale-UID refresh, not cross-controller adoption). Spoofing labels only makes WRC overwrite the rogue policy with the restrictive one — no privilege gain.
- **Zero writes in steady state** (the 12k-PUT/hour fix) reproduced against the real producer.
- **`normalizeIngressFrom`** correct for `@kubernetes/client-node` 1.4.0 (ingress uses `_from`, egress uses plain `to`; no `_to` to coalesce).
- **Logging** is structured + sanitized (`safeInfrastructureErrorFields`), no raw `console.*`, no secret-bearing payloads; no RBAC/manifest/CRD change.
- **E2E harness** is genuinely evidence-based: the no-write observation requires a per-policy/family no-op witness + an exact reconcile receipt bound to recipe/namespace/uid/generation + rejection of any write log + stable UID/rv across all controller replicas — not resourceVersion inference. Negative controls are attributable (prove reachability with a temporary complementary permission before denying). Fixtures are collision-safe with UID-preconditioned cleanup.
- **Docs** check out against the code; the round-1 `verify-networkpolicies.sh` inaccuracy stays fixed.
- **Suites**: `networkPolicyConvergence.test.ts` + `uiEgressMetadata.test.ts` → 20 passed; `reconciler.test.ts` (442) + `k8sClient.test.ts` → 559 passed (579 total), once the workspace package is built. `tsc -b` clean on the touched files.

zach88 (approved):

alfredolopez80:
@claude Please perform an independent verification re-review of the changes addressing zach88’s R3 review: https://github.com/evenfire-ai/evenfire/pull/580#issuecomment-5602300215

Verify current head `f37ce9b311f7155d75961c9174c27dbd39c7139a` against base `f8460e9d6c85eaf8f3666274b2376d1fae6795fd`. Inspect the latest correction commit and its interaction with the full PR. Do not treat earlier favorable reviews as verification of these changes.

Please independently validate each disposition:

- **R3-M1/M2:** producer-backed property tests use real builders and ObjectSerializer, distinguish simulated API defaults from serialization, and prove replacement idempotency/composition, unchanged decisions, drift detection, metadata preservation and ownership vetoes. Check for vacuity and over-constrained generators.
- **R3-M3:** the no-DELETE assertion accepts only grep’s no-match exit status. Verify that actual DELETE calls, unreadable logs and unavailable dependencies fail the assertion, and that the self-controls genuinely test those paths.
- **R3-L1:** inferred and explicit policyTypes are normalized consistently, with a regression that detects the former mismatch.
- **R3-L2:** the dedicated UI egress path remains intentionally outside the converted read-first contract; verify the documented boundary and retained protections.
- **R3-L3:** the defensive retry branch is now explained; ensure the earlier termination guard still preserves termination-before-ownership ordering.
- **R3-L4:** verify the complete reconciler→watcher path: exponential backoff from a 5-second base, capped at 60 seconds, with no total attempt limit. Challenge this disposition if the actual caller bypasses that behavior.
- **R3-L5:** confirm the removed WRC export has no consumers and the separate HCC implementation is unaffected.
- **R3-L6:** verify the existing error-name plus message assertions from `50da8c4d1`.

Reported evidence for this head: WRC 2,810 passed with 29 expected skips; 153 shell contracts passed; CI 37/37 successful; T0/T1/T2, real PostgreSQL 164/164 across 35 files with no skips, NP08 and four service NetworkPolicy E2E suites passed. Playwright browser journeys were NOT_RUN. Independently assess coverage; these results are not a substitute for your review.

Check for new regressions, security/ownership changes and blast radius. Return remaining findings with severity and evidence, disposition of all nine R3 points, missing validation and a merge-readiness recommendation. No new follow-up issue was created: broader migration is already tracked in #581, and the remaining notes did not establish a new costly defect.

Keep operational evidence private: no credentials, cluster/profile identifiers, private endpoints, local paths or raw logs. Do not merge or enable auto-merge.

Issue #584, opened 2026-09-03
bug(wrc): the run lane fires ~1.4k unconditional NOT_FOUND DELETE/hour on every reconcile pass (not finalization-only)


alfredolopez80:
## Summary

The workflow **run lane** issues **~1 436 unconditional `DELETE` requests per hour** against NetworkPolicies on `clerum-dev`, all returning `NOT_FOUND`. That is 9.5 % of the WRC service account's total delete volume.

Same defect as #582, different file. Split out because `workflow/workflowReconciler.ts` is owned by **#581 PR-G** and the two changes must not collide.

## Measured

Read-only, `clerum-dev` only, 1-hour audit-log window, from `system:serviceaccount:control-plane:workflow-recipes`:

| policy family | count/h |
|---|---:|
| `*-mcp-servers-egress-internet` + `*-mcp-host-to-codex-proxy` | **1 436** |

All status 5 (`NOT_FOUND`). Total across all lanes in the same window: 15 115.

## The cause

Two call sites in `workflow-recipes/src/workflow/workflowReconciler.ts` fire a `DELETE` on **every** `applyWorkflowNetworkPolicies` pass, regardless of whether the object exists:

- `pruneLegacyMcpServersInternetEgressPolicy` — `:2803`
- the codex-proxy policy delete — `:2979`

Both go through `safeDelete:4458`, which swallows the 404 and, for anything else, logs without failing.

## Correction to earlier notes

An earlier version of #582 and of the write-amplification master plan asserted that the run lane's deletes were **finalization-only**, living in `reconcileDelete:3202`. That is true of *those* deletes but not of these two: `:2803` and `:2979` sit on the steady-state reconcile path. The 1 436/h is the proof.

## The fix

Same shape as #582 — know what exists rather than asking the apiserver to reject the request.

`pruneLegacyMcpServersInternetEgressPolicy` is, by its name, a **migration**: it removes a policy a previous version authored. A migration that runs forever is a migration that never completed. Two options, and the choice needs the file's owner:

1. **Read first** — `safeDelete` becomes read-then-delete, skipping when absent. Trades 1 436 `DELETE`/h for 1 436 `GET`/h. Cheapest change; does not remove the request.
2. **Desired-set prune**, as #582 does for the recipe lane — one `LIST` per pass, delete only what exists and is no longer wanted. Removes the requests entirely, and would let the legacy prune retire once no such policy remains anywhere.

[...]

Issue #583, opened 2026-09-03
bug(wrc): a workload that gains transport or becomes the UI workload leaves its wl-egress/wl-ingress policy permanently unreapable


alfredolopez80:
## Summary

A workload's `wl-egress-*` / `wl-ingress-*` NetworkPolicy becomes **permanently unreapable** once that workload gains `transport` or becomes the recipe's UI workload. Nothing writes it again and nothing deletes it.

Latent, not a live incident: it contributes **zero** to the delete storm in #582, and there are currently **no** such stale policies for a live recipe on `clerum-dev`. Filed because #582 introduces a desired-set prune that would otherwise silently start deleting these, and the decision to *not* do that needs a record.

## The mechanism

`reconcileWorkloadEgressPolicies` (`workflowRecipeReconciler.ts:4533`) skips two classes of workload in both of its loops:

```ts
for (const w of workloads) {
  // Skip MCP + UI workloads — covered by other code paths.
  if (w.transport) continue
  if (uiWorkloadId && w.id === uiWorkloadId) continue
```

`:4544`, `:4545` in the egress loop; `:4674`, `:4675` in the ingress loop.

The comment is accurate about *why*: those workloads get **different** policies — `ui-egress-*`, `ui-ingress-*`, and the MCP route. But it does not mean their `wl-*` names are written elsewhere. **Nothing writes them.** Verified repo-wide: `workloadEgressPolicyName` / `workloadIngressPolicyName` appear at exactly four sites —

- `workflowRecipeReconciler.ts:4598` — inside the lane, **after** the skip
- `:5052`, `:5064` — the two reap helpers (removed by #582)
- `resourceBuilder.ts:1731`, `:1805` — the two builders, themselves called only from the lane, also after the skip

So a `wl-egress-<recipe>-<workload>` for a `transport` or UI workload can only exist if that workload **once was** a normal workload with `egressBindings`, and then changed. After that, no code path writes it and no code path reaps it: the `continue` at `:4544` happens *before* the reap branch.

## Why it is not the storm

The producer is a **spec edit**, not the reconcile loop. It is not a produce-and-delete cycle, so it emits no repeated requests. Measured on `clerum-dev`: **zero** `wl-egress` stale for a live recipe. The 15 115 `DELETE`/h of #582 come entirely from the *other* branch — `bindings.length === 0` on a normal workload, firing every pass.

## Why #582 does not fix it

[...]

Issue #581, opened 2026-09-03
refactor(wrc): make the idempotency contract mandatory — four failure modes from one optional parameter


alfredolopez80:
## Summary

WRC applies every managed object through `createOrReplace`, whose fourth parameter — the one that decides whether to compare before writing — is **positional and optional**. Five of seventeen call sites pass it. The default for any new caller is "write unconditionally", and nothing asks the author to decide.

That default has produced four distinct failure modes, three of which are live and measured. This issue proposes replacing the helper with a contract where the same mistakes are not expressible.

Umbrella for the migration; #579 is closed by PR-C below.

## The four failure modes

**1 · No gate at all.** The fourth argument is not passed, so every pass writes. #575 measured 12 450 `PUT`/h from `applyNetworkPolicy` alone. Twelve of seventeen callers are in this state.

**2 · A gate written by hand that can never close.** Three sub-causes with different remedies:

| | example | why it never matches |
|---|---|---|
| **2a** compares against the server's view | **#579** | the builder inserts `{podSelector, policyTypes, egress}`, the apiserver returns `{egress, podSelector, policyTypes}`. `JSON.stringify` preserves insertion order, so identical content produces different strings |
| **2b** two desired manifests for the same object in one pass | **#413** | stamped with one, compared against the other |
| **2c** the desired manifest is not deterministic | `wl-egress`, run-lane NP | carries a `resolved-at` timestamp, so the hash changes on its own |

**3 · The gate skips an assertion that lived in the write path.** The ownership check — "is this policy mine or another controller's?" — sat inside `replaceFn`. A gate that returns first never runs it. Latent today in the ConfigMap/Secret callers (#571) and `coordinator-to-gfs`: anyone adding `idempotency` to them "like Deployment does" reproduces it.

**4 · The seal has no provenance.** The hash certifies *"what I want has not changed since the last time **I** wrote"*, not *"the live object equals what I want"*. If another writer changes `spec` without touching the annotation, the gate compares my hash against my own annotation, they match, and it stays silent forever over an object that no longer resembles the desired state. HCC only touches annotations excluded from the hash today, so it does not happen — but nothing prevents it.

The same property means **manual drift is no longer reverted**. For a Deployment that is visible (replica health degrades). For a NetworkPolicy nothing shows it: the recipe stays `active` and the enforcement object is wrong.

## Why it happened

[...]

Issue #578, opened 2026-09-03
ops(hcc): a persistent egress reconcile fault raises no metric and still counts as ready at the convergence gate


alfredolopez80:
## Summary

A persistent non-DNS fault in external-egress reconciliation is now invisible to metrics. The only egress metric is a gauge that counts retry ladders stuck at their cap, and the fail-static path never reaches the ladder, so the gauge stays at zero while the fault repeats on every resync.

The signal still exists — a `False` condition on the CRD and a `console.error` per pass — but nothing an alert can be built on.

## Mechanism

Before #567, a non-DNS fault threw. `k8sClient` caught it, called `scheduleExternalEgressRetry`, and a fault that never cleared eventually pinned the ladder at its cap, which raises `clerum_hcc_external_egress_retries_at_cap` (`metrics.ts:338`).

After #567, a fault with a live policy to serve returns normally. `retryCompletion?.complete()` (`k8sClient.ts:4107`) then clears the ladder, so:

- no retry is scheduled,
- `clearRetryAtCap` runs,
- the gauge never rises,
- and the startup convergence gate (`externalEgressConvergenceCoordinator.ts:243-251`) counts the server as `ready=true`.

That last one is the sharper edge: a server whose egress reconcile faults on every pass is indistinguishable, to the convergence gate, from one that reconciled cleanly.

## Why this was not part of #567

Adding a counter is a few lines; making it useful is not. A metric with no dashboard and no alert is a second place to look after someone already noticed the problem some other way. The right shape — counter by reason, or a gauge for "servers currently reconciling with a fault", and where it is surfaced — is a decision about the alerting story, not a line of code.

## Suggested direction

A counter incremented on the fail-static path, labelled by whether the fault was blocking, plus whatever the convergence gate should report for a server in that state. The gate treating a faulting server as `ready` is arguably the defect worth fixing first: it is a correctness question about what "ready" means, not an observability gap.

Found by a closing audit of #567 (read, not reproduced — the mechanism is by code inspection of the retry and gate paths).

Related: #567, #513.

PR #587, opened 2026-09-03
fix(wrc): reap NetworkPolicies from the live set, not blindly per workload (#582)

### LINKED ISSUE #582: bug(wrc): three *IfExists reap helpers never check existence — ~15.1k NOT_FOUND DELETE/hour, the delete-verb twin of #575
## Summary

The WRC recipe reconciler emits **~15 100 `DELETE` requests per hour** against NetworkPolicies on `clerum-dev`, and **100 % of them return `NOT_FOUND`**. Every one is a request the apiserver exists only to reject.

This is the same disease as #575, on the delete verb. `createOrReplace` did `POST` → 409 → `PUT`; the reap paths do `DELETE` → 404 → ignore. In both cases the controller states an intention and lets the apiserver decide whether it was already true, instead of knowing.

## Measured

Read-only, `clerum-dev` only, one hour of GKE audit logs from `system:serviceaccount:control-plane:workflow-recipes`.

| status | count |
|---|---:|
| 5 (`NOT_FOUND`) | **15 115** |
| anything else | **0** |

108 distinct policy names at ~206 deletes each per hour — matching the ~17.5 s reconcile cadence. By family:

| family | source | /h | share |
|---|---|---:|---:|
| `ui-ingress-*` | `deleteUiIngressIfExists` via `:4269`, `:4280` | 7 029 | 46.5 % |
| `wl-ingress-*` | `deleteWorkloadIngressIfExists` via `:4679`, `:4684` | 3 591 | 23.8 % |
| `wl-egress-*` | `deleteWorkloadEgressIfExists` via `:4551`, `:4645` | 2 004 | 13.3 % |
| `*-mcp-servers-egress-internet`, `*-mcp-host-to-codex-proxy` | run lane — **#584**, out of scope here | 1 436 | 9.5 % |
| `wf-*-oauth-broker-egress` | inline at `:6215` | 1 053 | 7.0 % |
| `ui-egress-*` | inline at `:4194` | 2 | ~0 |

**Genuine reaps do happen**: 21 successful deletes (status 0) in a six-hour window, clustered at recipe create/delete events. Removing the reap is therefore not the fix.

## The cause

Three helpers named `*IfExists` **do not check whether the object exists**. They issue the `DELETE` unconditionally and swallow the 404:

| helper | line | mechanism |
|---|---:|---|
| `deleteUiIngressIfExists` | `:4287` | `safeDelete` |
| `deleteWorkloadEgressIfExists` | `:5047` | `safeDelete` |
| `deleteWorkloadIngressIfExists` | `:5059` | `safeDelete` |

```ts
private async deleteWorkloadEgressIfExists(recipeName, workloadId, namespace) {
  const name = rb.workloadEgressPolicyName(recipeName, workloadId)
  await this.safeDelete(
    () => this.networkingApi.deleteNamespacedNetworkPolicy({ name, namespace }),
    `NetworkPolicy "${name}" in ${namespace}`
  )
}
```

The name states the desired semantics. The body does not implement them.

Two further inline deletes on the same per-pass paths have the same shape: `reconcileUiEgressPolicy:4194` and `reconcileOAuthBrokerEgressPolicy:6215`.

### Where they are called

Six helper call sites, inside the per-workload loops of `reconcileUiIngressPolicies:4254` and `reconcileWorkloadEgressPolicies:4533` (the latter holds both the egress and the ingress loop), on the branch where no policy is desired:

```
4269, 4280  deleteUiIngressIfExists        — workload declares no UI-reachable ports
4551, 4645  deleteWorkloadEgressIfExists   — no egressBindings; builder returned null
4679, 4684  deleteWorkloadIngressIfExists  — nobody points at this workload
```

### The intent is right; the implementation has no memory

```ts
for (const w of workloads) {
  const bindings = w.egressBindings ?? []
  if (bindings.length === 0) {
    // Reap any policy left over from a previous reconcile that DID
    // have bindings — keeps the policy set in lockstep with the spec.
    await this.deleteWorkloadEgressIfExists(recipe.metadata.name, w.id, wlNs)
    continue
  }
```

The comment at `:4547` is correct: a workload that *used to* declare bindings must have its leftover policy reaped. The defect is that the reap cannot distinguish "there is a leftover to remove" from "this workload has never had one" — which is the steady state for most of them.

### The arithmetic

- 19 recipes reconciling every ~17.5 s → ~206 passes/h each → **~3 900 recipe-passes/h**
- 15 115 ÷ 3 900 ≈ **3.9 `DELETE`s per recipe-pass**
- A workload with no UI ports, no `egressBindings` and no inbound siblings takes 3 per pass; `ui-ingress` dominates because its loop walks *every* workload of every recipe with `spec.ui`

The 100 % `NOT_FOUND` confirms the mechanism: if these were genuine reaps, the first pass after a spec change would return 200 and only subsequent passes 404. A sample that is entirely 404 means the objects were already absent every time.

## Why this matters more once #575 lands

| | before #580 | after #580 | after #581 PR-A+B |
|---|---:|---:|---:|
| no-op `PUT` | 12 450/h | ~0 | ~0 |
| `POST` → 409 | 12 450/h | 12 450/h | ~0 |
| **`DELETE` → 404** | **15 115/h** | **15 115/h** | **15 115/h** |

**#581 does not cover this**: its `ManagedApply` contract is create/update only. So the moment #580 merges, this becomes the **single largest block of rejected requests** against the apiserver, and stays that way through the entire #581 sequence.

#542 documents that this class of traffic — write-shaped requests over an unchanged fleet — saturated the production apiserver at 150 objects/min. The recipe-lane portion of this path runs at ~230/min on its own.

## The fix

**Reap from the desired set, not per workload.** Every NP lane records what it wants into a per-pass ledger; after all of them complete, **one `LIST` per namespace** selects by ownership — `clerum.io/managed-by=workflow-recipes,clerum.io/recipe=<name>` — and deletes only what is live, classifies into a recipe family, and is not wanted. Then the three helpers go.

`pruneStaleInternalDependencyPolicies:4481` already demonstrates this pattern and **already issues exactly those three LISTs every pass**, so the net request delta is zero while recipe-lane `DELETE` drops from ~13 700/h to ~0.

It also closes a gap the current shape structurally cannot: a policy whose **workload disappeared from the spec entirely** is never reaped today, because the loop no longer iterates over that workload. #368 is the same gap in HCC.

Rejected alternative: read-before-delete via `readNetworkPolicyOrNull:4714`. It trades 15 100 `DELETE`/h for 15 100 `GET`/h — cheaper per request, identical count — and fixes neither the missing-workload gap nor finalization.

### Classification by name, and why it is safe

Five of the six families carry no distinguishing label (#524), so the family is decided by **name prefix** while selection stays by **ownership**. That is sound for one specific reason: **neither `composePolicyName` implementation can touch the prefix** — `resourceBuilder`'s truncates only the recipe segment, the internal-dependency lane's shortens both segments and appends a hash, and both interpolate the prefix whole.

The oauth-broker name puts the recipe in the **middle** (`wf-<recipe>-oauth-broker-egress`), so it needs prefix **and** suffix: `wf-` alone names ten other resources in this package.

Never touched, because the classifier does not recognise them: the webhook gateway's policies (no `clerum.io/recipe` label), the run lane's, `coordinator-to-gfs`, and anything created by hand.

**Rename the helpers regardless**: `deleteXIfExists` should either check existence or be called `deleteX`. The name is what let this survive review.

### Failure semantics

`safeDelete:4023` logs and swallows everything that is not a 404, so a 403 on a stale allow is silent today. The replacement splits it:

| code | signal |
|---|---|
| 404 | tolerated — the normal case |
| 5xx / timeout | `RetryableReconcileError`, reusing the existing contract at `:1907` (*"Transient infra failure … Degrade instead of failing so the next reconcile retries"*) |
| **403** | error log + additive condition `NetworkPolicyReapFailed` + Prometheus counter by family. **Phase untouched** |

The phase is deliberately not touched on a 403. A latched `failed` would be **invisible** — `control-ui` has zero references to a recipe's `status.phase`, `conditions` or `message`, and `monitoring/` has no recipe-phase alert — **and sticky**, since a `failed` recipe self-heals only when its message matches `isRetryableInfraError` (`:1536`). The counter also answers #578's criticism that a persistent failure raises no metric.

## Separate finding, same root cause

**47 of 162** WRC-managed NetworkPolicies on `clerum-dev` name a recipe that no longer exists — 34 `wl-egress`, 7 `wl-ingress`, 6 `wr-intdep`, oldest 2026-06-01. `wl-egress`/`wl-ingress` carry **no ownerReference** and the finalizer never deletes them by name, so recipe deletion leaks them permanently.

**Zero are stale for a live recipe** — steady-state reaping works; the leak is at deletion. Their podSelectors match nothing today, but a recipe re-created with the same name would inherit the old allow for one pass (~17 s).

This is the mirror image of the storm: the reap fires eternal `DELETE`s for objects that do not exist, **and** leaves alive objects that do exist and should not. The same desired-set prune closes both halves, but the finalizer needs its own failure contract — one that throws can wedge a recipe in `Terminating` — so it is a separate PR.

## Verification

Read-only, same 60-minute window before and after:

```bash
gcloud logging read '
  resource.type="k8s_cluster" resource.labels.cluster_name="clerum-dev"
  protoPayload.authenticationInfo.principalEmail="system:serviceaccount:control-plane:workflow-recipes"
  protoPayload.methodName="io.k8s.networking.v1.networkpolicies.delete"' \
  --project eventfire-491421 --freshness=1h \
  --format='value(protoPayload.status.code)' | sort | uniq -c
```

Today: ~15 115 lines, all status 5. Target: **≈1 400 status 5** (run-lane residue, #584) plus a handful of status 0 when a spec actually changes.

`LIST` volume is **not observable** in GKE audit logs without data-access logging; it is unchanged by construction (the same three namespaces as `:4485-4490`).

A test must pin that a spec change still reaps: dropping `egressBindings` from a workload that had them must delete the policy on the next pass. Removing the reap is not the fix.

## Scope

**In**: `workflow-recipes/src/reconciler/workflowRecipeReconciler.ts` — the three helpers, their six call sites, and the two inline deletes.

**Out**:

- **The run lane** (`workflow/workflowReconciler.ts:2803`, `:2979`, 1 436/h) — different file, owned by #581 PR-G. → **#584**
- **The finalizer leak / the 47 orphans** — different path with its own failure contract. → PR 2
- **`wl-*` of workloads that gained `transport` or became the UI workload** — those names are written at only four sites, all after the skip, so nothing writes them for those workloads and they are genuinely stale. But the producer is a **spec edit**, not the reconcile loop: not continuous, and zero contribution to the storm. Reaping them here would widen this change into a family it does not own, so the prune carries an explicit carve-out. → **#583**

**Related**: #522 (external-egress GC deletes an unclaimed class silently) is the same conversation about deletion safety. #575/#580 fixed the write half of this pattern in the same file.

## Note on `deleteRequiredIfExists`

`:7018` shares the naming pattern but is **not** part of this: it deletes legacy StatefulSets, Deployments and Services, and is gated by `mayHaveLegacyRawWorkloads:6833` plus ownership reads. Left alone.

https://claude.ai/code/session_016eFKFk67BF9Gdqy5PtsSUV

alfredolopez80:
## (draft) — Trabajo diferido

Trabajo diferido por solicitud del autor (2026-09-09). Quedan pendientes los E2E específicos del PR y el cierre de la revisión de blast radius; no está aprobado para merge. No reanudar las validaciones hasta nueva indicación.

---

Closes #582.

Three helpers named `*IfExists` never checked existence. They issued the `DELETE` and swallowed the 404, from the branch of each per-workload loop where no policy is desired. A workload that never had a policy takes that branch too, so the delete fired on every pass, forever.

## Evidence, before the change

Measured on `clerum-dev`, read-only, one hour of GKE audit logs from `system:serviceaccount:control-plane:workflow-recipes`:

| status | count |
|---|---:|
| 5 (`NOT_FOUND`) | **15 115** |
| anything else | **0** |

108 distinct names at ~206 deletes each per hour — the ~17.5 s reconcile cadence. By family:

| family | source | /h | share |
|---|---|---:|---:|
| `ui-ingress-*` | `deleteUiIngressIfExists` | 7 029 | 46.5 % |
| `wl-ingress-*` | `deleteWorkloadIngressIfExists` | 3 591 | 23.8 % |
| `wl-egress-*` | `deleteWorkloadEgressIfExists` | 2 004 | 13.3 % |
| run lane — **out of scope**, #584 | `workflowReconciler.ts:2803`, `:2979` | 1 436 | 9.5 % |
| `wf-*-oauth-broker-egress` | inline | 1 053 | 7.0 % |
| `ui-egress-*` | inline | 2 | ~0 |

**Genuine reaps do happen**: 21 successful deletes in a six-hour window, at recipe create/delete events. Removing the reap is therefore not the fix.

This is the same disease as #575/#580 on the delete verb. `createOrReplace` did `POST` → 409 → `PUT`; the reap does `DELETE` → 404 → ignore. In both, the controller states an intention and lets the apiserver decide whether it was already true, instead of knowing.

## The intent was right; the implementation had no memory

```ts
const bindings = w.egressBindings ?? []
if (bindings.length === 0) {
  // Reap any policy left over from a previous reconcile that DID
  // have bindings — keeps the policy set in lockstep with the spec.
  await this.deleteWorkloadEgressIfExists(recipe.metadata.name, w.id, wlNs)
  continue
}
```

That comment is correct: a workload that *used to* declare bindings must have its leftover reaped. The defect is that the reap could not distinguish "there is a leftover to remove" from "this workload has never had one" — the steady state for most workloads.

## What changed

Give it memory. Every NP lane records what it wants into a per-pass ledger; after all of them complete, **one `LIST` per namespace** selects by ownership (`clerum.io/managed-by=workflow-recipes,clerum.io/recipe=<name>`) and deletes only what is live, classifies into a recipe family, and is not wanted.

**The `LIST` costs nothing new** — `pruneStaleInternalDependencyPolicies` already issued exactly these three every pass. Recipe-lane `DELETE` drops from ~13 700/h to ~0 **without adding a request**.

Rejected alternative: read-before-delete. It trades 15 100 `DELETE`/h for 15 100 `GET`/h — same count — and reaps neither a policy whose workload left the spec (the #368-shaped gap, now covered by T3) nor anything at finalization.

### Classification is by name, and that is safe for one specific reason

Five of the six families carry no distinguishing label (#524), so the selector is scoped by **ownership** and the family decided by **name prefix**. That is only sound because **neither `composePolicyName` implementation can touch the prefix**: `resourceBuilder`'s truncates only the recipe segment, the internal-dependency lane's shortens both and appends a hash, and both interpolate the prefix whole.

The oauth-broker name puts the recipe in the **middle** (`wf-<recipe>-oauth-broker-egress`), so it is matched on prefix **and** suffix — `wf-` alone names ten other resources in this package.

Anything unrecognised is never deleted: the webhook gateway's policies (no `clerum.io/recipe` label), the run lane's, `coordinator-to-gfs`, and anything hand-made. T4 pins that.

### `uiEgressPolicyName` extracted

Its name was an inline template literal in three places, and it is the only family that is per-recipe rather than per-workload and is not truncated. A name built in one place and classified in another needs exactly one definition.

### The ledger throws where its predecessor returned

`rememberDesired` returned silently on a manifest with no name or namespace. That was harmless when the prune only compared internal-dependency names; under the unified prune, **an unrecorded policy is a deleted policy**.

## Failure semantics — three classes

| code | signal | why |
|---|---|---|
| **404** | tolerated | the normal case |
| **5xx / timeout** | `RetryableReconcileError` | reuses the contract already at `:1907`, whose own comment reads *"Transient infra failure … Degrade instead of failing so the next reconcile retries"*. Nothing new is designed for this half |
| **403** | error log + `NetworkPolicyReapFailed` condition + Prometheus counter by family. **Phase untouched, no throw** | a real enforcement leak needing a durable signal |

**Why a 403 does not fail the recipe**, which is the judgement call worth reviewing. It was the original design and it was wrong on three counts, verified in code:

- The phase is a direct return value at `:1918`, not routed through the state machine — so "the transition is illegal" is **false**.
- `isTerminal('failed')` is **false**. The real latch is `:1536`: a `failed` recipe self-heals only if its message matches `isRetryableInfraError`, which a 403 message does not. So it would stick.
- **Decisive: nobody would see it.** `control-ui/src` has **zero** references to a recipe's `status.phase`, `conditions` or `message`, and never calls a recipes endpoint. `control-api`'s `RecipePodInfo.phase` is the **Pod** phase. `monitoring/` has no recipe-phase dashboard or alert. The only operator view is `kubectl`.

A latched `failed` is therefore invisible **and** sticky — worse on both axes than an error log, which at least reaches Loki. The counter also answers **#578**'s criticism that a persistent failure raises no metric.

A 403 must never reach the catch at `:1901`, whose default arm returns `phase: failed` **plus** `workflowPhase: failed`. Hence `continue`, not `throw`.

## The prune sits outside the internal-dependency `try`

Two reasons, and both are load-bearing:

- That catch maps everything to `OwnershipConflict`/`PolicyApplyFailed`, so a `RetryableReconcileError` from the prune would be swallowed and never reach the `degraded` path.
- The prune must run **only once every lane has completed**. Reaping against a half-built ledger would delete live policies.

## Tests

Eleven, each with the mutation that kills it. Every negative assertion carries a **liveness witness** — `not.toHaveBeenCalled` is satisfied identically by "correctly skipped" and by "the code never ran".

| test | pins | mutation that kills it |
|---|---|---|
| T1 | steady state issues no reap, and exactly one unconditional delete survives this PR | re-inserting any removed delete |
| T2 / T2b | dropping `egressBindings` still reaps; every family is reachable | prune removed; a classifier arm broken |
| T3 | a workload removed from the spec loses all three policies | the old per-workload loop structurally cannot pass this |
| T4 | foreign names survive | classifier defaulting to a family |
| **T6** | **a policy behind the #299 no-op gate counts as DESIRED** | **`remember()` moved after the gate** |
| T7 / T7b | a 403 reports on three channels and leaves the phase alone; the condition clears | metric removed; condition emitted only on failure |
| T7c | a 5xx yields **`degraded`**, not `failed` | a bare `Error` instead of `RetryableReconcileError` |
| T10 | the finalizer stays scoped to internal-dependency | families filter dropped |
| T11 | the `transport`/UI carve-out holds | carve-out dropped |

**T6 is the one that matters.** `egressWriteNeeded` returns without writing while the policy is alive and wanted; recording at the apply call instead would have the prune delete it every pass — an enforcement outage, not a wasted request. T6 feeds back the body the reconciler *actually created*, so the gate genuinely closes, and asserts the "egress set unchanged" line to prove it did. It is the only test in the repository that catches that mutation.

**The mutation run found five untested cases**, fixed in the second commit. The most instructive: T7c originally asserted `phase !== 'active'` plus a message substring — both true for a bare `Error` too, which falls to the terminal arm and returns `failed` with `String(error)` containing the same text. The assertion was true for the wrong reason. An assertion loose enough to hold for both the correct and the mutated behaviour is not a weaker test; it is not a test.

## Scope, and what was deliberately left out

| item | why | tracked |
|---|---|---|
| run lane, 1 436/h | different file, owned by #581 PR-G | **#584** |
| `wl-*` of `transport`/UI workloads | stale, but produced by a **spec edit** rather than the reconcile loop; zero contribution to the storm. Verified: those names appear at only four sites, all after the skip. Reaping them would widen this PR into a family it does not own | **#583** (explicit carve-out, T11) |
| finalizer leak / 47 orphan policies of deleted recipes | own failure contract — a finalizer that throws can wedge a recipe in `Terminating` | PR 2 |
| `reconcileUiEgressPolicy`'s own delete when `spec.ui` is unset | 2/h; T1 asserts it is the only unconditional delete left | — |
| `policy-type` on six families | the selector does not need it | #524 |

## Found while implementing, reported not fixed

**`composePolicyName` is duplicated with different guarantees.** `resourceBuilder.ts:1641` (from the initial import, `5d663a49`) truncates only the recipe and can exceed 63 chars when prefix plus workload already do; `internalDependencyNetworkPolicies.ts:33` truncates both and appends a hash, respecting 63 strictly. Both preserve the prefix, so neither blocks this change — but two functions with the same name and different contracts in the same package is a trap.

## Corrections made to issue #582 before implementing

The issue as originally filed named four helpers (three), eleven call sites in a method that does not exist, and asserted the run lane was finalization-only (it is not — 1 436/h). All four are corrected on the issue with a note saying what to discard.

## Verification after deploy

Read-only, always `resource.labels.cluster_name="clerum-dev"`:

```
gcloud logging read '
  resource.type="k8s_cluster" resource.labels.cluster_name="clerum-dev"
  protoPayload.authenticationInfo.principalEmail="system:serviceaccount:control-plane:workflow-recipes"
  protoPayload.methodName="io.k8s.networking.v1.networkpolicies.delete"' \
  --project eventfire-491421 --freshness=1h \
  --format='value(protoPayload.status.code)' | sort | uniq -c
```

Today ~15 115 status 5. Expected after this PR: **≈1 400 status 5** (run-lane residue, #584) plus a handful of status 0 on real spec changes. Dropping an `egressBindings` should produce exactly one status-0 delete and one `Deleted stale wl-egress` log line.

`LIST` volume is **not observable** in GKE audit logs without data-access logging. It is unchanged by construction — the same three namespaces the intdep prune already queried — and T1 counts the calls if a number is wanted.

## Gates

`workflow-recipes` **2594 passed** / 29 skipped (+11 over `dev`'s 2583) · `npx tsc --noEmit` exit 0 · `npm run build` exit 0.

All exit codes read from a captured log, never from a wrapper's notification. `npm run build` uses `tsconfig.build.json`, which excludes tests; CI's separate `tsc --noEmit` does not — both were run, because a sibling PR went red twice on exactly that gap.

The mutation harness backs up by file copy and verifies restoration by sha256; the tree was clean after every run.

https://claude.ai/code/session_016eFKFk67BF9Gdqy5PtsSUV


alfredolopez80:
@claude Full re-review and audit requested at `4c34e1fe`. CI is green, 37/37.

This PR has already been through a four-agent adversarial review that found **2 Critical and 5 High, all reproduced**, plus 15 findings below that. Everything is enumerated here with its disposition, so you can verify each against the code rather than rediscover it — and **please check my claims about what was fixed, not only the code**. Two of the fixes exist because my original reasoning was wrong, not because the code was.

---

# Part 1 — What this PR does

## The defect

Three helpers named `*IfExists` never checked existence. They issued the `DELETE` and swallowed the 404, from the branch of each per-workload loop where no policy is desired. A workload that never had a policy takes that branch too, so the delete fired on every pass, forever.

Measured on `clerum-dev`, read-only, one hour of GKE audit logs from `system:serviceaccount:control-plane:workflow-recipes`: **15 115 `DELETE`, 100 % `NOT_FOUND`, zero of any other status**. 108 distinct names at ~206 each — the ~17.5 s reconcile cadence.

| family | /h | share |
|---|---:|---:|
| `ui-ingress-*` | 7 029 | 46.5 % |
| `wl-ingress-*` | 3 591 | 23.8 % |
| `wl-egress-*` | 2 004 | 13.3 % |
| run lane — out of scope, #584 | 1 436 | 9.5 % |
| `wf-*-oauth-broker-egress` | 1 053 | 7.0 % |
| `ui-egress-*` | 2 | ~0 |

**Genuine reaps happen**: 21 successful deletes (status 0) in six hours, at recipe create/delete events. Removing the reap is not the fix.

Same disease as #575/#580 on the delete verb: `POST` → 409 → `PUT` there, `DELETE` → 404 → ignore here. The controller states an intention and lets the apiserver decide whether it was already true, instead of knowing.

## The fix

Every NP lane records what it wants into a per-pass ledger. After all lanes complete, **one `LIST` per namespace** selects by ownership (`clerum.io/managed-by=workflow-recipes,clerum.io/recipe=<name>`) and deletes only what is live, classifies into a recipe family, and is not wanted.

**The `LIST` adds no request** — the internal-dependency prune already issued one per namespace over the same three. What changed is the result set, not the count.

Classification is by **name prefix**, because five of the six families carry no distinguishing label (#524). That is sound only because neither `composePolicyName` implementation can touch the prefix — verified by property test over 1 254 generated names.

---

# Part 2 — Every finding from the review, and its disposition

## Critical — both fixed

### C-1 · Six of the seven `remember()` call sites had no test

Dropping `desired.remember(policy)` from ui-egress, ui-ingress, wl-ingress, oauth-broker, intdep-egress or intdep-ingress each killed **zero tests and zero type errors**. The class docstring calls this exact property *"the one mistake in this change that causes an enforcement outage rather than a wasted request"*. It was tested for one lane of seven.

**Fixed** with a table-driven test, one row per lane. Two passes each: the first authors the policy **and asserts it was authored** — without that witness a fixture that stopped producing it would make the second assertion pass for nothing — the second serves it back live and asserts the prune leaves it alone.

**Verify**: that the table covers all seven, and that removing any single `remember()` now fails.

### C-2 · The no-op-gate trap existed in two lanes and was tested in one

`reconcileUiEgressPolicy` has the same `egressWriteNeeded` short-circuit as `reconcileWorkloadEgressPolicies` and the same "record before the gate" comment. Moving its `remember()` after the gate killed nothing.

**Fixed**: ui-egress now gets the T6 treatment, including the `egress set unchanged` log assertion that proves the gate genuinely closed — without it the test also passes when the gate stayed open and the lane rewrote the policy, which is a different and uninteresting reason to see no deletion.

## High — all fixed

### H-1 · The condition was computed, merged, and then dropped

`NetworkPolicyReapFailed` never reached `status.conditions`. The three `??` chains downstream of the insertion point omitted the new link, and `mergeOwnedConditions` returns `undefined` for a group with nothing to merge, so in the ordinary case the fallback skipped past it entirely. One of the three channels this PR advertises did not exist.

The tests did not catch it because they asserted `result.networkPolicyReapConditions` — the in-memory object — and never the patch body.

**Fixed** in all three chains, with a test that asserts the **patch body**, following the neighbouring `InternalDependenciesReady` test which already did.

### H-2 · The `LIST` was outside the try

A denied `LIST` latched the recipe; on the workflow path it also set `workflowPhase`.

**It is unreachable today**, and I verified that rather than assuming it: `workflow-recipes` holds `list` and `delete` on `networkpolicies` in all three namespaces (`deploy/base/{sandbox-recipes,sandbox-ui,mcp-server}/rbac.yaml`), and the method queries no others because `resolveWorkloadNamespace` can return only those three. The pre-PR prune listed **exactly the same three** with the `LIST` equally outside its try — so this is inherited, not introduced.

**Fixed anyway**, because the PR changed its consequence: the prune is now the only reaper for all six families, where before a failed (intdep-only) `LIST` still left the per-workload deletes running. It now takes the same three-class treatment, skipping the namespace rather than failing the pass — without the `LIST` there is nothing to reason from.

`ReapDenial` became a discriminated union so the condition distinguishes `delete` (one named policy leaked) from `list` (a whole namespace unexamined). The counter gained an `operation` label for the same reason.

### H-3 · My reasoning for leaving the phase alone was wrong

The original rationale rested on *"nothing renders a recipe phase outside kubectl"*. **False**, and this is the correction most worth checking:

- `control-api/src/routes/internal/sandboxUi.ts` returns **409 `recipe_not_ready`** unless `phase === 'active'`
- `services/workflows/workflowRecipeAccessService.ts` blocks workflow invocation with `recipe_phase_not_ready`
- the phase is mapped into the `RuntimeWorkflowDto`

And the inversion that decides it: **`POST /recipes/:name/retry` accepts exactly `failed` and refuses `degraded`**. `failed` is operator-recoverable through the product; `degraded` is not.

The conclusion is unchanged — a 403 still leaves the phase alone — but for the true reason: failing a working recipe over a residual object takes it out of service, and the transient-latch self-heal only clears messages that look like infrastructure faults, which an authorization failure never does.

### H-4 · The ordering invariant had no test

Moving the prune to before `reconcileInternalDependencyPolicies` killed nothing. Every desired intdep policy would be reaped and immediately recreated on every pass. **Fixed** with a test that pins the order.

### H-5 · The fail-loud throw had no test

Making `remember` return silently on a nameless manifest — the exact predecessor behaviour the PR says it deliberately replaced — killed nothing.

**Fixed**, and worth noting how: my first attempt re-stated the guard inside the test and asserted on the re-statement, which proves only that the test agrees with itself. It now drives the real path, mocking a builder to yield a nameless manifest and asserting the pass fails loudly with the message naming the defect.

## Medium — fixed

| # | finding | fix |
|---|---|---|
| M-1 | Two of six families were never classified or reaped in any test; T2b was named *"reaps every recipe family"* and asserted 4 of 6. Reducing the oauth arm to a prefix-only `wf-` test killed nothing | T4 now feeds a `wf-*` name that is **not** the oauth policy — `wf-` alone names ten other resources in this package |
| M-2 | T1 and T6 survived "the prune LISTs but never deletes". Their witness proved the prune *looked*, not that it can delete | the new per-lane tests pair a negative with a positive delete in the same run |
| M-3 | The 404 swallow had no test; the transient arm was wider than documented | test added; 401 split out (see M-4) |
| M-4 | **401 was treated as transient** — an expired SA token parked the recipe in `degraded` forever, under a message claiming transience, in the one phase `POST /retry` refuses | 401 now takes the same route as 403: permanent, three channels, phase untouched |
| M-5 | The ui-egress inline delete was **redundant, not out of scope** — it targeted the same object the prune removes on the same pass | removed. T1 now asserts **no** unconditional delete survives, a stronger claim than the "exactly one" it made before |
| M-6 | The doc said *"the LIST costs nothing new: the prune already issued exactly these"* — the request count is unchanged, the result set is wider | corrected |
| M-7 | `composePolicyName` never truncated the workload segment: names up to **223 characters** against the 63-char limit | both segments now clamped. Nothing orphaned — every name it alters is one the apiserver rejected with 422, so no such object exists |
| M-8 | Two recipes differing only past the cut collide on one policy name: **199 collisions in 200** | **not fixed here** → **#589**. It is a rename with a migration, not a hunk |

## Low — fixed

- **L-1** A comment cited **"T9"**, a test that does not exist. Replaced with the real reference.
- **L-2** Three line-number cross-references pointed at unrelated code (`:1907` at a catch that maps to `failed`, `:1901` at an argument literal, `:1536` at an unrelated field). Replaced with the **names** of the sites, which do not rot on the next edit.
- **L-6** `managed-by: 'wrc'` vs `'workflow-recipes'` is now a **boundary between two lanes**, and nothing said so. A recipe legitimately named `wl-egress` or `ui-ingress` would make the run lane's 131 policy names classify into a reaped family; that label is the only reason the prune never sees them. Comment added at the source.
- **L-7** The gateway-exclusion comment claimed those policies are safe *"because they carry no `clerum.io/recipe` label"*. True, but they **do** carry `managed-by: workflow-recipes` — the exclusion rests on the second selector term, and the wording read as "not ours".

## Deliberately not fixed, each with an issue

| item | why | issue |
|---|---|---|
| name collisions | a rename with a deploy plan, not a hunk | **#589** |
| `wl-*` of `transport`/UI workloads | produced by a **spec edit**, not the reconcile loop; zero contribution to the storm. Explicit carve-out, pinned by T11 | **#583** |
| the run lane, 1 436/h | different file, owned by #581 PR-G | **#584** |
| `deleteRequiredIfExists` | shares the naming pattern but deletes StatefulSets/Deployments/Services, gated by `mayHaveLegacyRawWorkloads` plus ownership reads | — |

## One more, after the review: CodeQL

The first push failed CodeQL with **13 high-severity `js/insufficient-password-hash`** alerts — all reported as *"Password from a call to `recipeWithOauth` is hashed insecurely"*, all landing in production files this change does not touch.

All 13 are false positives, and the taint source was a **test fixture I added**. The rule treats the return of a function whose name reads like a credential source as tainted, then follows it into every `createHash('sha256')` in the package. I checked each sink: `specHash` hashes the manifest spec; `internalDependencyNetworkPolicies` hashes a name for disambiguation; `resourceBuilder` hashes `identity:workloadId` for resource naming. **None hashes a password.**

What identifies the *name* rather than the shape as the trigger: the sibling fixture `recipeWithBackgroundAccess`, in the same file, carries an identical `clientSecretRef` and raises nothing. And `clientSecretRef` is a **reference** (`name` + `key`) — the CRD never holds a secret value.

Renamed to `recipeWithBrokerOptIn`, with the reason documented so it is not renamed back. **No scanner rule or baseline was touched** — this removes a false-positive source, it does not suppress a finding. Please verify that reading; if any of those 13 sinks does reach a credential, the rename is the wrong fix.

---

# Part 3 — What the review found CLEAN

Verified independently, not assumed. Worth stating because it bounds where not to look:

1. **The lane inventory is complete.** Every exit leaving a policy alive-and-desired calls `remember()` first. Reproduced.
2. **The two call sites are identical.** `reconcileInternal` and `deployWorkflowWorkloads` run the same five lanes, same order, same family set, same carve-out. No asymmetry.
3. **No reachable false positive in the classifier.** Property test over 1 254 generated names (60- and 200-char recipes and workload ids, hyphens, unicode, empties) × 7 builders: zero misclassifications. All 96 static policies in `deploy/` parsed: none claimed.
4. **The prefix survives every truncation path**, in both `composePolicyName` implementations. Proven, not read.
5. **Deletion is fail-CLOSED.** All six families are **allow** policies; none is a deny. The baseline denies carry neither label and classify `null`. A wrongful reap closes traffic, never opens it.
6. **No secret exposure** in the new logs or the condition message.
7. **No vacuous tests**: making the prune a total no-op kills every test in the block.

---

# Part 4 — Mutations

**15, each killing exactly its target and nothing else.** This includes all 13 that survived the review, plus two found while closing them (the oauth prefix-only classifier and the merge chain). Restoration verified by `sha256` after every run.

| mutation | kills |
|---|---|
| `remember` dropped, each of six lanes | that lane's row only |
| `remember` moved after the ui-egress gate | C-2 only |
| `remember` moved after the wl-egress gate | T6 only |
| ledger returns instead of throwing | H-5 only |
| classifier drops the `ui-egress` arm | that family's test only |
| classifier reduced to prefix-only `wf-` | T4 only |
| 404 no longer swallowed | M-3 only |
| 401 back in the transient arm | M-4 only |
| `LIST` 403 throws raw | H-2 only |
| merge chain drops the link | H-1 only |
| prune disabled | all of them |

**The one that matters**: moving `remember()` after the no-op gate. It dies on `expected [ 'wl-egress-test-recipe-worker' ] to not include 'wl-egress-test-recipe-worker'` — the production failure verbatim, not a collateral crash.

---

# Part 5 — Gates

| gate | exit | result |
|---|---|---|
| `npm test` | **0** | **2608 passed**, 0 failed, 29 skipped (+25 over `dev`'s 2583) |
| `npx tsc --noEmit` | **0** | clean, includes tests |
| `npm run build` | **0** | clean (`tsconfig.build.json`, excludes tests — both were run) |
| `npm run lint` | **2** | **cannot run** — no `eslint.config.*` exists anywhere in the repo or on `dev`. Pre-existing, reported rather than called green |

All exit codes read from a captured log, never from a wrapper's notification.

---

# Part 6 — What I would like you to attack

1. **The lane table (C-1).** Is it actually complete? A lane added without a row would be silently uncovered again — that is the failure this table exists to prevent, and it only works if the list is right.
2. **H-3.** I reached the same conclusion twice by different reasoning, and the first reasoning was false. Is the second one true? Check `control-api`'s phase gates and the `retry` endpoint yourself.
3. **H-2.** I argue the denied-`LIST` path is unreachable with the current manifests. Verify the RBAC and the namespace set independently — I was wrong once already about what was reachable here.
4. **M-5.** I removed the ui-egress inline delete as redundant. Confirm the prune really does cover that object on the same pass, in every reconcile shape, including a recipe that never had `spec.ui`.
5. **M-7.** Truncating the workload segment changes names. I claim nothing is orphaned because every altered name exceeded 63 characters and so never existed. Check that claim.
6. **The mutations.** Run them rather than trusting the table. That is how the two Criticals were found — the eight originally declared all held, and the gap was in what had not been mutated.
7. **Anything a fifth pass would find.** Four independent agents each found something the others did not, including a defect inside a fix for a previous finding. Assume there is more.

https://claude.ai/code/session_016eFKFk67BF9Gdqy5PtsSUV

zach88:
Took a deep pass through this one. The core reap mechanism is genuinely solid — the per-pass ledger, the ownership selector, the `wrc` vs `workflow-recipes` label boundary, and the 404/403/5xx semantics all hold up, and I verified the "a `failed` recipe would be invisible" argument (control-ui really does have zero refs to recipe status). Where it comes apart is the *replacement* signal the PR introduces for that invisible-failed case: one of the three advertised failure channels never actually reaches the CRD.

## Verdict

🔴 **Not ready to approve**
- **Must resolve before approval: 2** (2 High)
- **Optional / non-blocking: 2** (1 Medium, 1 Low)

1. **[R1-H1] `NetworkPolicyReapFailed` never persists to the CRD, and can stick at `True`** — `k8sClient.ts:118` (`shouldPatchRecipeStatus`). The condition gets computed and merged into the body correctly, but `shouldPatchRecipeStatus` only fires a patch on a change to `phase`, `message`, `workloads`, the SDK projection, `clearWorkflowExecution`, or `ownedConditionsChanged(..., ['InternalDependenciesReady'])` — `NetworkPolicyReapFailed` isn't in that list. So on a steady `active` recipe whose Role gets revoked (reap → 403), nothing else changes, the patch is never emitted, and `kubectl describe` shows nothing. The reverse bites too: if a `True` did get persisted during the denial and RBAC is then restored with no other change, the recovery `False` isn't persisted either — the condition stays pinned at `True` forever, which is exactly the "sticky" the PR body says it avoids. This also contradicts the body claim that the condition is "written in both directions, so it clears on the next clean pass".
   - Repro: unit test on the pure function — feeding a fresh `networkPolicyReapConditions: [True]` with nothing else changed, `shouldPatchRecipeStatus` returns `false`.
   - Why it went green: T7/T7b/T7c assert `result.networkPolicyReapConditions` (the in-memory object), never the persisted patch through the watcher gate — so the observable (`kubectl describe`) is never checked.
   - Fix: add `NetworkPolicyReapFailed` to the `ownedConditionsChanged` set in `shouldPatchRecipeStatus`, same as `InternalDependenciesReady`. `sameConditionSignal` ignores `lastTransitionTime`, so the stable `False` won't churn patches. And assert the condition on the persisted patch, not the in-memory object.
   - (High rather than Blocker because the primary channel — the `reap_denied_total` counter + the log — does fire every pass; it's the condition channel the PR advertises that's broken.)

2. **[R1-H2] The classifier that drives every destructive delete has no round-trip test against the real producer, and a comment claims it does** — `resourceBuilder.ts:1707`. The comment says *"resourceBuilder.test.ts pins it with a round-trip over every builder, including names long enough to force truncation in each segment"* — there's no such test; `classifyRecipeNetworkPolicyName` isn't referenced in `resourceBuilder.test.ts` at all. The reap tests feed it hand-written canonical names. The invariant the whole reap safety rests on — *"`composePolicyName` never touches the prefix"* — isn't tested for `wl-*`/`ui-*`, and this PR modified `composePolicyName` to also truncate the workload segment. The code today is correct (I confirmed the prefix is interpolated whole), so this isn't a live bug — but it's a missing regression guard on a destructive decision plus a false coverage claim: a future change that truncates into the prefix would make `classify()→null` silently, the reap would stop deleting real policies, and nothing goes red. Fix: for each name builder, with recipe/workload long enough to force truncation, assert `classifyRecipeNetworkPolicyName(builderX(...)) === '<family>'` and `name.length <= 63` (all six families, incl. oauth mid-name and intdep hash-suffix) — or at minimum delete the false claim from the comment. Flagging it here since it guards a destructive op; if you'd rather fast-follow the test, the comment correction is the part worth doing now.

## Optional (non-blocking)

3. **[R1-M1] `composePolicyName` can emit 64-char names (over the 63-char DNS-1123 limit)** — `resourceBuilder.ts:1659`. The rewrite exists specifically to clamp to 63 and the comment says so, but the arithmetic hands the whole `budget` to the workload segment and then adds a 1-char floor for the recipe stem, overflowing by one: `budget = 63 - prefix.length - 2`, `workload = workloadId.slice(0, budget)`, `maxRecipe = max(1, budget - workloadStem.length) → 1`, total = 64.
   - Repro: `composePolicyName('wl-ingress', 'r'*40, 'a'*51)` → 64 chars. `workloads[].id` allows up to `maxLength: 63` (`charts/clerum-crds/crds/workflowrecipe.yaml:336`), so a 51-63 char id is valid input → apiserver rejects with `422 metadata.name: Invalid value` → the workload's NetworkPolicy doesn't apply. Doesn't affect the reap (prefix survives, classification is fine), only the apply. Not a strict regression — the base overflowed worse for the same inputs.
   - Fix: reserve one char for the recipe stem (`workloadId.slice(0, budget - 1)`) or clamp the final string to 63. Covering it with the H2 test (which forces segment truncation) would catch both.

4. **[R1-L1] Doc drift in `docs/deploy/workflow-recipes-guide.md:348`** — says "Two deliberate exclusions" but only one remains (the inline `reconcileUiEgressPolicy` delete was removed as redundant), and it attributes the webhook-gateway exclusion to the classifier when it's actually the selector (they stamp the recipe under `clerum.io/recipe-name`, not `clerum.io/recipe`). The code comment at `resourceBuilder.ts:1713-1717` has it right.

## What checked out clean

- Ledger completeness: all six lanes `remember()` before their early-returns / no-op gates, so the enforcement-outage trap doesn't fire; prune runs only after every lane completes, outside the intdep `try`.
- Foreign-policy protection: run-lane/coordinator (`managed-by=wrc`) and webhook-gateway (`clerum.io/recipe-name`, names classify `null`) are excluded — double-defense verified. The `wrc` vs `workflow-recipes` split is intentional and load-bearing.
- No labelSelector injection (`metadata.name` is DNS-1123 validated). The reap only deletes *allow* policies, so an errant delete restricts traffic (availability), it doesn't open isolation.
- Failure semantics: mutation-tested — 5xx→`degraded` not `failed`, and T3/T6/T7c each go red when the logic they pin is inverted.
- Suites green: `reconciler.test.ts` 335/335, `resourceBuilder.test.ts` 142/142, `workloadEgressPolicy.test.ts` 15/15.

alfredolopez80:
@claude Please perform a fresh, end-to-end re-review of PR #587 at exact head `90b6578585c0e2ce977015e6b6e4fea341089d3c`.

This PR closes #582: replace the WRC recipe lanes' ~15.1k/hour unconditional `DELETE -> 404` loop with an ownership-scoped desired/live-set reap, while preserving genuine revocation when a workload or binding leaves the spec. The immutable comparison base is `dev@c28fa3b54c691292f97c1a9128738399d501c1ec`; the previously reviewed PR head was `6d31a4c8d848a10b1aa06e3fe1666f48ad15b324`.

## Prior review findings and their disposition

Please re-check the earlier in-PR reviews, including [the prior Claude audit](https://github.com/evenfire-ai/evenfire/pull/587#issuecomment-5529652661) and [the later reviewer findings](https://github.com/evenfire-ai/evenfire/pull/587#issuecomment-5535414911), against the current code rather than trusting these dispositions:

1. **Condition persistence / recovery:** `NetworkPolicyReapFailed` now participates in the real watcher patch gate, uses a tri-state projection (`not evaluated` preserves; only a complete clean pass writes `False/Reaped`), persists failure and recovery, and does not churn on timestamp-only changes.
2. **Classifier/producer coverage:** every managed family is round-tripped through the real name producer, including long recipe/workload IDs; names are DNS-1123, `<=63`, prefix-preserving and collision-resistant.
3. **64-character name overflow:** corrected with compatibility-preserving hashed overflow names. Existing short names remain byte-identical.
4. **Documentation and rationale drift:** comments/docs now use the actual Control API phase gates and resolvable metric semantics; the stale exclusion count and selector explanation were corrected.
5. **Early-return cleanup observation:** reconciliation now records explicit per-family completeness and safely reaps completed families even when a different family cannot complete; workflow fast paths and zero-workload workflows execute the live-set pass without re-driving coordinator work.

## Exhaustive follow-up review (`RP-587-001` through `RP-587-010`)

All ten independently reproduced findings were addressed:

- **RP-587-001 — family/early-return convergence:** completed-family pruning, partial-clean authority rules, workflow/no-workflow coverage and zero-workload coverage.
- **RP-587-002 — finalizer strictness:** cleanup must be confirmed before finalizer removal; failures retain the UID-bound recipe and schedule retry.
- **RP-587-003 — condition lifecycle:** one tri-state projection drives comparison, patching and preservation; only a fully clean pass may clear the condition.
- **RP-587-004 — deterministic retry/restart:** incomplete reaps remain level-triggered, use bounded backoff and are reconstructed from durable status/finalizer state.
- **RP-587-005 — stale-generation deletion race:** fences before planning and before deletion, generation-stamped desired objects, ownership revalidation, and UID/resourceVersion delete preconditions.
- **RP-587-006 — error taxonomy:** 404 idempotence is operation-aware; authorization, retryable infra and permanent failures have distinct outcomes.
- **RP-587-007 — naming collision:** overflow names use a stable hash rather than lossy truncation.
- **RP-587-008 — logging:** new production paths use the service's structured/redacting logger; no raw error payloads or direct `console.*` were added.
- **RP-587-009 — metrics/alerting:** the counter remains an event signal with bounded labels; alert guidance uses `increase(...)` and a durable pending state rather than an absolute non-zero counter.
- **RP-587-010 — DNS-1123 bounds:** all admitted long inputs now produce valid `<=63` names.

## Findings caught during iterative independent audits

The independent audit was rerun after every correction and additionally found these paths; please attack them again:

- workflow awaiting-trigger/terminal/in-progress/active/runtime-scope fast paths previously rearmed retries without executing the live-set pass;
- egress no-op updates previously left an old generation/resourceVersion window;
- a clean reap projection could be lost after downstream SDK, binding, inner-workflow or coordinator-GFS failures;
- a raw Kubernetes `503` in a workflow fast path could become terminal;
- the second `ui-egress` read after a conflict did not originally revalidate ownership/generation;
- a snippet-Secret `503` normalized to a plain `Error` could be terminalized after the policy pass.

Current behavior is that intentional product-terminal outcomes are returned as `ReconcileResult`; any exception after a successful policy pass preserves recipe phase/message/workload status and all condition projections, publishes `False/Reaped` when needed, and requeues without deriving `workflowExecution=failed`. Subsequent passes do not cause status churn.

## Evidence on the exact final head

- Independent GPT-5.6 SOL/max audit verdict: **CLEAN** on all four requested dimensions (all findings closed, no new regressions, no adverse cross-service blast radius found, full `review-pr` audit clean).
- Critical slice: **496/496**.
- Full WRC: **136 files passed, 4 skipped; 2644 passed, 29 intentionally skipped**.
- `tsc --noEmit`, production build, Docker build, Prettier and `git diff --check`: PASS.
- Branch-owned Minikube exact-head gate: **T0 PASS; T1 163/163 across 35/35 Real PostgreSQL files; T2 PASS; NP-08 PASS; read-only Control API -> WRC health PASS; all required deployments Ready; PUBLIC_BOUNDARY_PASS**.
- GitHub CI/CD for this exact SHA: **37/37 checks PASS**, including WRC, Real PostgreSQL, cross-service builds/tests, E2E node-unit, Desktop, shell contracts, leak guards and CodeQL.

## Blast-radius boundaries to revalidate

- No RBAC, CRD, Deployment strategy, public HTTP contract or data migration changed.
- Destructive selection remains tenant/recipe ownership-scoped, limited to recognized managed families, and guarded by generation plus UID/resourceVersion CAS.
- Short existing policy names do not rename; hashing applies only to legacy-overflow identities that Kubernetes could not have accepted.
- Active-reconcile authorization failures preserve a working recipe while remaining observable/retryable; finalization remains fail-closed.
- CI and the exact-head runtime gate exercised Control API, HCC/NP-08, MCP Host, GFS, RPC, external REST, UIs, Desktop and adjacent packages without an adverse service regression.

## Explicit residual scope

The existing follow-ups remain separate: #583 (transport/UI workload transition residue), #584 (run-lane delete storm), cleanup expansion for other finalizer families, and a future identity-v2 migration. Runtime RBAC fault injection was not executed because the destructive mutation was not authorized; deterministic watcher/finalizer/RBAC tests and the read-only exact-head runtime gate are the available evidence.

Please independently re-run the prior attack list and the current concurrency/error-boundary invariants, then report either a clean verdict or any remaining concrete finding with exact file/line, reproduction and expected behavior.

alfredolopez80:
@claude — please perform a fresh, independent revalidation of PR #587 at exact head `1578e72b791a91d6debdc41c4106899050739963`.

This is the final coverage-closure delta on top of the comprehensive prior review and remediation summary in https://github.com/evenfire-ai/evenfire/pull/587#issuecomment-5542519985. Please re-check the valid findings and dispositions recorded there, plus the additional availability/finalizer work below, against the current diff rather than relying on the earlier head.

## Problem and scope

PR #587 addresses #582: WRC previously inferred policy names and issued roughly 15.1k `DELETE -> 404` requests/hour. The redesign reaps from the live NetworkPolicy set, bounded by WRC/recipe ownership, recognized families, generation fencing and UID/resourceVersion preconditions, while preserving real revocation after spec changes.

The coverage audit found one additional in-scope production risk: failures affecting only stale-policy reap had inconsistent lifecycle effects. A 403 preserved `active`, but a 503 changed the recipe to `degraded` and a 422 to `failed`. That could block Sandbox UI or new workflow triggers for an otherwise correctly configured and running service, without actually fencing existing workloads or removing the residual allow.

## Final behavior

- Reap-only authorization, transient and permanent outcomes are now orthogonal overlays: preserve the base phase/message/workload/workflow state, publish the durable `NetworkPolicyReapFailed` condition and schedule the class-appropriate retry.
- A clean reap publishes `False/Reaped` and cancels retry.
- Desired-policy `laneError` remains a separate boundary and retains its failure/degradation contract.
- During CR deletion, any incomplete reap remains fail-closed and retains `clerum.io/workload-cleanup`.
- An accepted NetworkPolicy DELETE is followed by a targeted GET. Only 404 proves absence; a live/Terminating object produces an incomplete reap and cannot release the recipe finalizer.

## Coverage added/updated

- Reconciler matrix: authorization/transient/permanent/clean, LIST and DELETE, mixed failures, multiple namespaces, stale generation, desired-policy controls, classic active, SDK-only, workflow awaiting/running/completed/first-deploy and terminal workflow states.
- Watcher lifecycle: condition-only status changes, level-triggered retry, retry recovery after restart, status-only MODIFIED events, UID recreation, patch failure and semantic `lastTransitionTime` stability.
- Control API contracts: active Sandbox UI and workflow trigger/readiness remain usable with pending reap condition; actual readiness/desired-policy failures still block.
- Kubernetes E2E: 47/47 checks using a synthetic source with KEEP and DROP dependencies. A legitimate spec update removes only DROP; DROP policy disappears and three packet probes are denied, KEEP policy/connectivity remains, recipe stays active, and `False/Reaped` is durable. A test-only finalizer barrier then proves WRC requests child deletion and retains the recipe finalizer until the policy is actually absent. The success path never directly deletes NetworkPolicies.
- Static shell contract prevents inferred-name shortcuts, success-path compensating policy deletion and kubectl calls without explicit context.

For realistic shape comparison only, read-only non-sensitive metadata from `clerum-dev` was consulted for `evenfire-prompt-notify-app`, `eventasks`, and `helpdesk-v2`; the E2E fixture is synthetic and copied no env values, Secrets, payloads or manual adjustments. `helpdesk-v2` was used only as a structural example because its observed webhook conditions were not wholly green.

## Exact-head evidence

- WRC: 136 passing files, 4 skipped; **2664 passed / 29 skipped**.
- Reconciler: 370 tests; watcher/Kubernetes client: 122 tests; independent critical slice: **516/516**.
- Control API: 402 passing files, 34 skipped; **5091 passed / 157 skipped**; focused routes **147/147**.
- Real PostgreSQL T1: **163/163 across 35/35 files**.
- Kubernetes E2E: **47/47** on the final head.
- Typecheck, production builds, Docker build, Prettier, diff check, bash syntax, ShellCheck and leak guard: PASS.
- Exact-head T0 PASS; T2 PASS; NP-08 PASS; Control API -> WRC health PASS; required deployments Ready; `PUBLIC_BOUNDARY_PASS`.
- Full-gate evidence: `.local-notes/infra/runs/20260904T163049Z-1578e72b791a-242099c8b0c04d69baf71d5fba9dacb2/evidence.json`.
- Post-E2E runtime evidence: `.local-notes/infra/runs/20260904T163916Z-1578e72b791a-189ed2d1b1624c93862eb18510e5c813/evidence.json`.
- GitHub CI/CD: **37/37 PASS, pending=0, fail=0**, including CodeQL and every adjacent service/package job. Runs: https://github.com/evenfire-ai/evenfire/actions/runs/33896580765 and https://github.com/evenfire-ai/evenfire/actions/runs/33896580903.

An independent GPT-5.6 SOL/max audit on this exact head returned `CLEAN`: all in-scope findings fixed, no new regression, no adverse cross-service blast radius, and the full `review-pr` audit clean. The final one-line delta from the certified functional commit only moves the E2E finalizer to the repository-sanctioned `e2e.clerum.io` domain; leak guard and the full runtime journey pass after that change.

## Blast-radius boundary

No CRD, RBAC, Deployment strategy, public HTTP contract or data migration changes. The destructive selector is ownership/family/generation fenced; DELETE is CAS-bound and absence-confirmed. CI and exact-head runtime gates cover Control API, HCC, MCP Host, RPC Proxy, external REST, GFS, UIs, Desktop and other packages/services. The E2E simultaneously proves stale authorization is revoked and the unrelated KEEP service remains reachable.

Known separate follow-ups remain #583, #584, cleanup expansion for other finalizer families and a future identity-v2 migration; none is introduced or worsened by this head.

Please revalidate the current head for correctness, regression risk, deletion/finalizer safety, availability semantics, retry/restart behavior, test adequacy and cross-service blast radius, and report any remaining actionable finding with a concrete code path and reproducer.

alfredolopez80:
@claude Please revalidate the complete PR on **8520047ec2fbd0e43bacf5918b617ce22f665441**, against **dev c28fa3b54c691292f97c1a9128738399d501c1ec**. This includes the valid earlier findings discussed in [the comprehensive review](https://github.com/evenfire-ai/evenfire/pull/587#issuecomment-5542519985), [the coverage review](https://github.com/evenfire-ai/evenfire/pull/587#issuecomment-5543748146), and [your prior response](https://github.com/evenfire-ai/evenfire/pull/587#issuecomment-5543751418), plus the independent Astra findings below. The earlier clean verdict on 1578e72b was superseded by reproduced regressions; please review the final code and contracts, not only the passing check list.

## Intent and scope

Issue #582 replaces recurring blind WRC NetworkPolicy DELETE/404 traffic with an owned live-set reap while retaining real revocation when recipe declarations change. The intended invariant is that cleanup-only failure never cuts availability of a correctly configured, otherwise healthy recipe. Desired enforcement failure remains a separate error class. No CRD, RBAC, DB schema, public HTTP contract, or blanket finalizer-family expansion is introduced.

## Prior valid review findings retained and rechecked

- Per-family desired ledger, recording before no-op returns, independent family progress, incomplete-family safety and coverage of workflow fast paths.
- A single reap projection across reconciler/status patch gate/watcher: durable True/Unknown/False condition, semantic patching, stable transition times, retry/backoff and restart reconstruction. Reap 403/503/422 does not by itself terminalize or degrade a healthy recipe.
- Fresh recipe identity/generation fences, definitive ownership rechecks after create conflicts, and UID/resourceVersion preconditions for destructive operations. No-op generation advancement persists the stamp without changing rules.
- Explicit API failure classification; non-404 read failures never establish absence. Accepted DELETE is not proof of removal: pending child finalizers retain the reap condition and block recipe-finalizer completion.
- Policy naming/classifier round-trips, collision-safe new names, bounded metrics, structured/redacted new logs, alerts and operational documentation.

## Astra findings and final corrections

1. **Fast paths received raw templates.** Common effective-recipe preparation now resolves inputs/computed values and actual workload names/namespaces before every policy path, using a detached copy. A failed/incomplete family cannot authorize an empty desired set.
2. **Generation was compared across recreated recipe UIDs.** Policies carry recipe UID and generation. Ordering is only meaningful within one UID; a current recipe can reclaim an owned previous-epoch policy with fresh identity checks/CAS. Earlier-epoch DNS history is not inherited.
3. **Legacy physical names and DNS state were lost during upgrade.** WRC probes the exact valid legacy forms emitted by the base, including API-valid names above the stricter new-name budget. It preserves/adopts compatible legacy identity, merges every verified same-epoch DNS window, and persists the merged window before removing a duplicate—even when the IP/spec signatures are unchanged. Migration does not refresh an existing expiry. Carveout and already-scoped cleanup account for aliases.
4. **Readiness recovery was overridden by cleanup state.** A readiness observer can publish active while preserving an outstanding reap condition and retry. Control API availability contracts cover healthy recipes with pending cleanup.
5. **Finalizer/absence E2E allowed false greens.** Structured snapshots require real calendar-valid RFC3339 timestamps and bound UID/resourceVersion. Only explicit successful absence counts; GET errors do not. Barrier installation/removal is CAS and only an installation owned by this test is released. A newly rejected WRC cleanup cycle must occur between two valid parent/child snapshots before release.
6. **HTTP stalls could resemble network denial.** Negative probes distinguish TCP connection timeout from an already-connected socket with a stalled HTTP response. DNS, ready Service endpoints and backend loopback HTTP are checked before and after the negative probe. KEEP/DROP environment observation is atomic, and failures of exec/probe/observation cannot certify enforcement.
7. **UID-stamping PUT could erase another controller's finalizer.** Both replacement paths now preserve finalizers from the same definitive GET whose resourceVersion fences the write. Six regressions cover finalizers preserved, added and removed between reads for workload/UI.
8. **The real watch adapter discarded the timestamp required by legacy continuity.** This additional final review found the producer/consumer gap despite reconciler-only fixtures being green. The mapper now carries the optional API creationTimestamp. Tests go through the actual watch adapter into reconciliation for rotating/transient DNS, with/without a coexisting hash alias; absent/ambiguous timestamps remain untrusted.
9. **A duplicate Terminating alias could degrade a healthy survivor.** Authorization to read history is now separate from eligibility to write. Every alias still receives ownership/UID/generation checks; only a live compatible alias can be selected. Authorized same-epoch history from the retiring alias is persisted in the survivor. Both alias orders, no live survivor, future generation and old UID are covered. A two-pass workload regression verifies active with Unknown/retry while the duplicate is retained.
10. **Real-client and status-timing E2E hardening.** [BusyBox NC_110_COMPAT](https://raw.githubusercontent.com/mirror/busybox/1_36_1/networking/nc_bloaty.c) can return a silent connect failure without verbosity. The final probe uses numeric/zero-I/O/verbose/bounded connection options, rejects silent failure, and first proves positive TCP connections to both backends. The E2E no longer polls for persistence of InternalDependenciesReady: the existing base/head health-refresh contract can clear it. A successful GET yielding absent/True is acceptable only alongside the positive active/Reaped/policy/packet proofs; False/Unknown/read errors fail. No production behavior was changed to satisfy this test.

## Evidence to revalidate

- WRC final full suite: **2702 passed, 29 intentional skips**; Control API: **5091 passed, 157 intentional skips**.
- Focused reconciler/watcher/naming/accumulator: **682/682** using Node 24.18, serial, no cache. Independent reconciler run: **403/403**.
- Python evidence **9/9**, static-audit regressions **4/4**, Bash, ShellCheck, E2E static audit, TypeScript builds, formatting and diff checks passed. The specific executable shell contract is also exercised by CI.
- Final CI: **37/37 checks passed**, zero pending/failure on 8520047e: [CI](https://github.com/evenfire-ai/evenfire/actions/runs/33956115643), [CodeQL](https://github.com/evenfire-ai/evenfire/actions/runs/33956115622).
- Exact-head **T0/T1/T2 passed**. T1 ran **163/163 real PostgreSQL tests across 35/35 physical files, zero skips**. NP-08 deployed authorization and bounded Control API/WRC health passed; required deployments were Ready.
- Real selective-convergence/finalizer E2E on the final head: **76/76 passed**. It demonstrated positive HTTP/TCP to both backends, denial for an unrelated caller, generation/rollout after removing DROP, selective policy/packet revocation while DROP itself remained healthy, KEEP availability, durable Reaped state, a newly rejected cleanup cycle while the held child and parent retained identity/finalizers, and confirmed child/parent absence after barrier release. The success path did not invoke emergency cleanup.
- A separate post-E2E exact-head runtime check selected **already-synced** and passed T2, NP-08 and health again. It validated the existing same-head T0/T1 attestations rather than skipping unexecuted lanes.
- Independent Astra core/security/architecture/cross-service and E2E reviewers returned CLEAN for the final SHA, with their runtime claims explicitly deferred to executed evidence.

## Blast radius and explicit limits

Production code in Control API, HCC, MCP Host, RPC Proxy, GFS, shared packages and deploy manifests is unchanged. The destructive set stays constrained by owned namespace/family/fresh identity; uncertainty cannot remove the recipe finalizer. Unit contracts and CI cover consumers, while the isolated Kubernetes journey proves actual selective connectivity and cleanup ordering. No production environment was mutated or deployed.

Structural reference recipes inspected read-only were **evenfire-prompt-notify-app**, **eventasks** and **helpdesk-v2**. They are not certified healthy fixtures, and no environment values, customer payloads or manual adjustments were copied. The E2E recipe is synthetic.

Existing follow-ups #583/#584, broader finalizer cleanup and a future identity-v2 migration remain separate. Runtime RBAC fault injection was not executed; deterministic failure matrices and the authorized runtime lanes are separate evidence. Browser/Playwright is not claimed for this controller/network journey.
The container install also reported existing dependency advisories; the base/head package and lockfile diff only changes WRC's own version, not dependencies. These are not new PR dependency changes and are not represented as a globally vulnerability-free result.

Please confirm on this exact head: (1) all prior valid and Astra findings are corrected; (2) no new bug/regression is introduced; (3) no adverse cross-service blast radius is found; and (4) the complete review is clean. For any remaining finding, include a concrete failing path, base/head attribution and the violated contract.

alfredolopez80:
Closing without merging. The problem it targets (#582) is resolved by #580: the three `*IfExists` helpers now read before deleting, and their NetworkPolicy families show 0 NOT_FOUND DELETE/h on clerum-dev at `2502a3cfe`. See the measurement in #582.

The PR has been a deferred draft since 2026-09-09 and conflicts with `dev` (#610, #618 and #633 landed since). The remaining NOT_FOUND NetworkPolicy deletes are tracked in #584 and are outside this PR's scope.

PR #588, opened 2026-09-03
feat(gfs): expose updatedAt on GFS read and listing APIs (#586)

### LINKED ISSUE #586: feat(gfs): expose resource updatedAt metadata in read and listing APIs
## Context

Both Control UI and Desktop need a “Modified” column in the GFS browser.

The `gfs_resources` table already stores:

- `created_at`
- `updated_at`

However, the public GFS resource/listing contracts currently omit these fields, so neither frontend can render the modification date.

No database migration should be required.

## Requested change

Expose `updatedAt` as an additive field on GFS resource views and all listing responses used by Control UI and Desktop.

Contract:

ts
updatedAt: string // ISO 8601 UTC timestamp
The value must come from gfs_resources.updated_at.

Required backend paths

Update the relevant types, SQL projections, row mappers, and response serializers for:

Control UI

- GET /api/v1/gfs/tree
- GET /api/v1/gfs/resources/:id/children

Relevant implementation:

- control-api/src/gfs/tree.ts

Desktop user plane

- GET /api/v1/me/gfs/resources
- GET /api/v1/me/gfs/resources/:id/children
- GFS resolve/resource views where the same canonical resource type is returned

Relevant implementation areas:

- control-api/src/routes/external/gfs.ts
- gfs-controller/src/api/read.ts
- gfs-controller/src/db/resourceStore.ts
- Desktop GFS transport/type declarations

Semantics

For this initial change, “Modified” means the timestamp at which the resource row itself was last changed, including operations such as:

- Initial creation/upload
- File replacement
- Rename
- Move

A folder’s updatedAt is not required to change merely because one of its children was created, modified, moved, or deleted. Recursive folder activity timestamps are out of scope.

Compatibility

This must be an additive API change.

Clients should temporarily treat updatedAt as optional during mixed-version deployments and render — when an older backend omits it.

Acceptance criteria

- [ ] Control API tree listings return updatedAt.
- [ ] Control API shared/accessible-resource listings return updatedAt.
- [ ] GFS Controller child listings return updatedAt.
- [ ] Resolve/resource views expose the same field consistently.
- [ ] Values are serialized as valid ISO 8601 UTC strings.
- [ ] Existing pagination and (name, resourceId) ordering remain unchanged.
- [ ] Existing authorization behavior remains unchanged.
- [ ] Tests verify timestamp projection and serialization for each listing path.
- [ ] Tests verify updatedAt changes after replace, rename, and move operations.
- [ ] No database migration is introduced because both timestamp columns already exist.

Follow-up

Once this contract is available, Control UI and Desktop will add matching “Modified” columns formatted in the user’s local timezone

alfredolopez80:
## Summary

- Close #586 by emitting `updatedAt` (ISO 8601 UTC from `gfs_resources.updated_at`) on GFS read and listing JSON: gfsc `toView` / accessible listing, control-api tree / resolve / Desktop Shared-with-me SQL, plus write/rename/copy `RETURNING` so post-mutation views stay complete.
- Desktop accepts `updatedAt?` on the wire for mixed-version deploys. No Modified column in Control UI or Desktop (follow-up). Folder timestamps stay per-row; they are not `max(children)`.
- Copy snapshots and pagination cursors do not include `updatedAt`. If-Match remains `version`.

## Testing

- [x] `npm test` passes for changed services
  - gfs-controller unit (excluding Real PG): 609 passed
  - control-api GFS unit: 170 passed
  - desktop-app `gfsUriHandler.test.ts`: 34 passed
- [x] `npx tsc --noEmit` clean in `gfs-controller` and `desktop-app`
- [ ] Real PG mutation cases (seed `2000-01-01T00:00:00Z`, then replace/rename/move) are written and skip-gated; this environment had no `CONTROL_API_REAL_PG_ADMIN_URL`

## Checklist

- [x] This is not a new third-party MCP server / WorkflowRecipe (those go to the registry)


alfredolopez80:
@claude please do a complete re-review of the current HEAD (`d22967eba6f67bda376d9a5804b92260114f9caa`).

CI/CodeQL are all green on this SHA. Please re-review the full GFS `updatedAt` change set, including the follow-up that reads copy `updatedAt` from published `gfs_resources.updated_at` (no wall-clock fallback).

Contract (#586):
- `updatedAt` is that row's `gfs_resources.updated_at`, emitted as ISO 8601 UTC with `Z` via `Date.toISOString()`.
- Server always emits the field; Desktop treats it as optional for mixed-version clients.
- Folder `updatedAt` is not `max(children)`. Child create/replace/delete must not stamp the parent; rename/move may stamp the subtree via `path_cache`.
- Do not expand the copy HTTP DTO. Cursor / `ORDER BY` / `If-Match=version` stay unchanged. No UI "Modified" column, Playwright, `createdAt`, or OpenAPI work.

Please review correctness, provenance, and scope against that contract.

zach88:
Took a deep pass through this one across all three services (gfs-controller, control-api, desktop). It's a clean, well-scoped additive change — the `updatedAt` field is threaded correctly through the read/listing paths, and the risky-looking bits (the `copyPublication` ambiguous-commit refactor, the "required + throw" timestamp coercion) all hold up under real code. Nothing here blocks a merge; what's left is quality/DRY follow-up.

## 🟡 Approve with nits
- **Optional / non-blocking: 4** (1 Medium, 3 Low)

_(This is my read from the review, not a formal GH approval.)_

## Optional (non-blocking)
The PR can merge without these — they're follow-up or author's call.

1. **[R1-M1] `toIsoTimestamp` is duplicated byte-for-byte in 7 places** _(Medium)_ — same coercion-with-throw helper copy-pasted across `control-api/src/gfs/resolve.ts:104`, `control-api/src/gfs/tree.ts:69`, `control-api/src/routes/external/gfs.ts:182`, `gfs-controller/src/db/writeStore.ts:165`, `gfs-controller/src/db/resourceStore.ts:50`, `gfs-controller/src/db/copyPublication.ts:94`, `gfs-controller/src/authz/accessibleStore.ts:67`. No behavior impact today (pg always hands back a `Date`, so every copy takes the same branch), but 7 divergent copies of a parsing/throwing helper will drift. Worth extracting a shared `timestamps.ts` per service — could easily be its own cleanup PR.

2. **[R1-L1] The cursor-exclusion tests are tautological** _(Low)_ — `gfs-controller/src/api/read.test.ts:356` only asserts `nextCursor === null` on a single-page result, so it never actually exercises a non-null cursor; `control-api/test/gfs.tree.test.ts:317` round-trips `encodeCursor(name, rid)`, whose signature can't carry `updatedAt`, so `.not.toContain('updatedAt')` is structurally impossible to fail. Both would pass unchanged against the parent commit — they document the "cursors exclude updatedAt" intent but give zero regression protection. If you want them to guard the invariant, force a multi-page result and assert the decoded cursor payload.

3. **[R1-L2] Silent non-ISO string fallback** _(Low)_ — the `toIsoTimestamp` copies return the raw string when `new Date(value)` fails to parse, silently violating the "ISO 8601 UTC" contract instead of throwing; `gfs-controller/src/db/renamePublication.ts:59` uses a separate `iso()` that does `String(value)` with no re-parse and a different guard/message. Dead branch in prod (TIMESTAMPTZ never emits a string), but worth settling on one fallback semantics (throwing is the defensible one) when you unify #1.

4. **[R1-L3] The mutation invariants only run in skip-gated real-PG tests** _(Low)_ — "advances updatedAt, per-row, not max(children)" is only verified against real Postgres in `gfs-controller/src/db/renamePublication.realPostgres.integration.test.ts:145` and `control-api/test/gfsStructuralIntegrity.realPostgres.integration.test.ts:312`, both gated behind `CONTROL_API_REAL_PG_ADMIN_URL` (didn't run in CI per the PR body). copy/write have no real-PG check on their new `RETURNING`. Acceptable — the SQL change is asserted as a substring and the column is DDL-guaranteed NOT NULL — but the write-semantics invariant runs nowhere by default. Consider wiring the real-PG suites into the pipeline.

## What checked out clean
- **The `gfs_resources.updated_at` column already exists** (`NOT NULL DEFAULT now()`, present at the merge-base) — no migration needed, no runtime blow-up on any `SELECT/RETURNING updated_at`, and the `throw` branch is unreachable for live rows.
- **Mixed-version compat is coherent both ways** — control-api isn't a required/throw HTTP client of gfsc (its user-plane routes proxy the response verbatim; its own required-emitting routes read the DB directly), and desktop treats `updatedAt?` as optional. New-server→old-desktop and old-server→new-desktop both hold.
- **The `copyPublication` refactor** (`"full"|"zero"|"unsafe"` → tagged union + `switch`) preserves the base behavior; the added `if (!root)` / `if (!recoveredRoot)` guards turn a possible `undefined` deref into an explicit throw, and the ambiguous-"full" path can't spuriously throw (any `toIsoTimestamp(undefined)` is caught and degrades to `unsafe`).
- **Security**: no leak/injection surface — `updatedAt` only projects over already-authorized rows, all new SQL is static column identifiers, nothing new interpolated.
- **Body invariants all hold in code**: copy snapshots and pagination cursors exclude `updatedAt`, folder timestamps stay per-row (nothing computes `max(children)`), If-Match stays on `version`.
- **Suites/typecheck**: gfs-controller 609 pass, control-api 117 pass, desktop 34 pass; `tsc --noEmit` clean in both TS services.

zach88 (approved):

PR #600, opened 2026-09-09
feat(desktop): unify titlebar header actions

josue-tech-94:
## Summary

- Move shared search and notification controls into the custom titlebar on authenticated routes.
- Preserve embedded-app notification drawers and keep titlebar overlays above an open chat drawer.
- Use native-style custom window controls with a verified preload-to-renderer bridge and tested
  desktop utility-gutter geometry.

## Validation

- `npm test -- --run test/browser/responsiveUtilityGutter.browser.test.ts`
- `npm test -- --run ui/src/__tests__/chatDrawerHeaderStacking.test.tsx`
- `npm test -- --run ui/src/components/WindowTitleBar/__tests__/WindowTitleBar.test.tsx`
- `node --version && npm run verify:electron`
- `npm run build`
- `npm run style-rules -- --strict`


josue-tech-94:
@claude Please perform one final, independent review of this entire PR against the current `dev` branch.

Do not limit the review to the latest fixes or previous findings. Review the complete final diff and current behavior as if this were the last merge-readiness pass.

Please focus on:

* correctness and regressions across the full PR;
* Electron main ↔ preload ↔ renderer contracts and runtime behavior;
* custom window controls and platform-specific behavior;
* titlebar, search, notifications, portals, chat drawer, and embedded-app interactions;
* CSS stacking contexts, responsive/layout behavior, and accessibility;
* tests for false-greens, unrealistic fixtures, weak assertions, or missing edge cases;
* security implications of IPC/preload exposure, trusted-sender checks, context isolation, navigation, and renderer privileges;
* anything that may have become incompatible because `dev` advanced while this PR was open;
* whether the PR description accurately reflects the final implementation and validation.

Please inspect the current code and tests rather than relying only on the PR description or prior review comments.

If you find an issue, include:

* severity;
* exact `file:line`;
* concrete failure scenario;
* impact;
* recommended fix;
* test or validation needed.

Please distinguish actual PR-introduced issues from unrelated pre-existing problems.

If the PR is clean, say so explicitly rather than inventing findings.

Final output should clearly state whether you see any remaining merge blockers or objections.

zach88:
Took a deep pass through the whole diff, not just the latest fixes. The Electron side is genuinely solid: every new `window:*` handler is trusted-sender gated, the preload bridge is a fixed allowlist with nothing leaked, the main→preload→renderer contract lines up end to end, and the big AppHeader→HeaderActions move is a faithful port with no lost logic. Where it gets shaky is the layout/CSS layer: there's one real visual regression at tablet widths, plus the stacking test that's supposed to guard the titlebar's headline behavior doesn't actually verify it.

**🟠 Not ready to approve**
- **Must resolve before approval: 1** (1 High)
- **Optional / non-blocking: 8** (2 Medium, 6 Low)

1. **[R1-H1] Tablet utility-gutter is still on the pre-titlebar model** — `ui/src/styles.css:13163-13184`. The diff moved search/bell out of the page and into the titlebar (AppHeader now renders only via `TitlebarActionsPortal`, `ui/src/App.tsx:2072-2084`), and the desktop base rule was correctly updated to reserve just the notification rail (`--app-notification-drawer-width + --space-4` = 438). But the tablet media query (901–1220px) wasn't migrated — it still reserves `clamp(160px,32vw,420px) + --space-2 + 36px` (398 closed / 466 open) for an in-page search that no longer exists there. At a reachable window width (~1100px) the apps/embedded content ends up over-indented ~398px with no header in that space, and opening the notification drawer reflows the content (398→466). Verified the geometry analytically against the shipped CSS. Fix: migrate the tablet override to the drawer-rail model like the desktop base rule (or drop the `+36px` search component).

## Optional (non-blocking)

1. **[R1-M1] Two nested `banner` landmarks** — `ui/src/components/WindowTitleBar/index.tsx:117` + `ui/src/components/HeaderActions/index.tsx:459`. `WindowTitleBar` renders `<header class="window-titlebar">` as a direct child of `.app-frame`, and App portals `HeaderActions`' `<header class="top-bar">` *inside* it, so the DOM ends up with `header > … > header` — two `banner` regions, one nested. Before this PR `top-bar` lived inside `.content-panel`, so it wasn't a banner. Screen-reader users navigating by landmarks now hit two "banner" regions. Fix: when `placement === 'titlebar'`, have `HeaderActions` render a `<div>` (or `role="presentation"`) instead of `<header>`; keep `<header>` for `placement === 'default'`.
2. **[R1-M2] The stacking test asserts the wrong mechanism (false green)** — `ui/src/__tests__/chatDrawerHeaderStacking.test.tsx:113-129`. The runtime behavior is actually correct — I traced the ancestor chain and the titlebar paints above the drawer because `.content-panel` traps the drawer at root z=1, not because `--layer-dropdown` (120) > `--layer-chat-overlay` (90). But the test asserts exactly that 120>90 comparison, which is irrelevant to the outcome (the two elements are in different stacking contexts). So it stays green even if the real invariant breaks (e.g. someone raises `.content-panel`'s z-index and the drawer starts painting over the titlebar), and goes red for a harmless change. It gives false confidence about the PR's central claim. jsdom can't compute stacking — this needs a real-browser assertion like the Playwright gutter test.
3. **[R1-L1] `aria-label` on a `<div>` with no role** — `ui/src/components/WindowTitleBar/index.tsx:118`. `aria-label="Window controls"` on a bare div isn't exposed by most screen readers, so the label is dropped. Add `role="group"` (or `role="toolbar"`) to `.window-titlebar__controls`.
4. **[R1-L2] Inconsistent destroyed-window guard across handlers** — `src/main.ts:170/175/193`. `window:minimize`/`close`/`getControlsState` use optional chaining only, while `toggleMaximize` (`:182`) also checks `isDestroyed()`. Optional chaining doesn't protect a destroyed-but-non-null `BrowserWindow` (where `minimize()`/`close()` throw). Unreachable today (trusted-sender + `'closed'` nulls `mainWindow`), just an inconsistency worth unifying.
5. **[R1-L3] Redundant synchronous `emitWindowControlsState` in `toggleMaximize`** — `src/main.ts:190`. Every branch already fires a wired event that emits; on platforms where the mutation is async, the synchronous read can emit a transient stale value before the event corrects it. Cosmetic — could rely on the event-driven emit alone.
6. **[R1-L4] Window-controls test hand-builds the bridge mock** — `ui/src/components/WindowTitleBar/__tests__/WindowTitleBar.test.tsx:20-58`. `installClerumWindowControls` reconstructs the shape of `clerum.window` instead of deriving it from `preload.ts`. The shape matches today so it isn't hiding a bug, but if preload changed a method name/shape this test would stay green while runtime throws `undefined is not a function`. Known limitation of the pattern (no harness derives the bridge from real preload), not specific to this PR.
7. **[R1-L5] `:has(.chat-drawer)` elevation is effectively redundant** — `ui/src/styles.css:60-65`. The titlebar at its default `--layer-header` (40) already paints above the whole `.content-panel` subtree (z=1) that holds the drawer, so the bump to `--layer-dropdown` isn't load-bearing. Harmless/defensive, but the comment implies it's what makes overlays usable above the drawer, when the real guarantee comes from content-panel's stacking context. Fix the comment to reflect the actual mechanism, or drop the rule.
8. **[R1-L6] Weak assertions in the stacking test file** — `ui/src/__tests__/chatDrawerHeaderStacking.test.tsx:147-163`. The last three cases are regex matches against the CSS source text (`stylesCss.toMatch(...)`), not behavioral assertions, and the `PortalStackingHarness` rebuilds the `app-frame > window-titlebar + app-root …` topology by hand rather than rendering `<App/>` — correct today, but if App's hierarchy changed the `:has()` selector would stop applying in prod while the test stayed green. Worth a comment tying the harness to the real structure.

## What checked out clean

- **Security (nothing introduced):** all four `window:*` handlers call `assertTrustedSender` as their first statement; `assertTrustedSender` only admits `file://`/dev-URL senders, so rpc-proxy embeds/nested iframes can't reach them; preload is a fixed allowlist with no channel-as-parameter and no leaked `ipcRenderer`; `webPreferences` is untouched (`contextIsolation:true`, `sandbox:true`, `nodeIntegration:false`) — only `frame:false` changed; `setWindowOpenHandler(deny)` and the `will-navigate` restriction are intact.
- **Bridge contract aligned end to end:** invoke channels and the `window:controlsState` event channel match across main/preload/renderer.d.ts/consumer; the `{maximized, fullscreen}` shape is identical in all layers; `wireWindowControls(window)` is actually invoked at window creation (`src/main.ts:400`) so the event pipeline is live; non-Electron path is guarded (`window.clerum?.window` with `?.` + `.catch()` fallback).
- **AppHeader→HeaderActions is a faithful move:** notification handlers (approve/deny/open/remove/clear + keyboard), embedded-app drawers (`notificationTrayMode='drawer'`), search memos, list keys, effect cleanups, and the overlay `useLayoutEffect` are all preserved identically.
- **Stacking fix itself is correct** — traced the full ancestor chain; the titlebar reliably paints above the chat drawer. (It's just the test that doesn't prove it — see R1-M2.)
- **Listener lifecycle:** `wireWindowControls` listeners are bound to the per-window `BrowserWindow` and GC'd with it on `'closed'`, no accumulation.
- **Suites/typecheck:** `build:main` / `tsc --noEmit` exit 0; 66 tests pass across the 5 diff test files; gutter geometry verified analytically against the shipped CSS.

josue-tech-94:
@claude Please perform a final independent merge-readiness review of the **current full PR** against the latest `dev`:

https://github.com/evenfire-ai/evenfire/pull/600

Please refresh the PR and `dev` before reviewing. Do not rely on an older reviewed SHA or previous comments as evidence that something is fixed.

This PR has gone through several review/fix cycles, so review both:

1. the complete PR diff and final behavior; and
2. whether the latest corrective commits introduced regressions or false-green tests.

Please specifically re-verify, without assuming they are correct:

* responsive utility-gutter behavior across desktop and tablet widths, including notification drawer closed/open states;
* real browser paint order between titlebar search/notification overlays and the chat drawer;
* whether the stacking regression coverage actually proves effective paint order rather than comparing unrelated z-index values;
* titlebar/HeaderActions accessibility landmarks and window-control grouping semantics;
* the real preload → renderer window-control contract and whether its tests are producer-backed;
* custom window controls, maximize/fullscreen state, event/listener lifecycle, and main ↔ preload ↔ renderer consistency;
* search, notifications, portals, embedded-app/WebContentsView interactions, and keyboard behavior;
* styling/token compliance, including the recently corrected titlebar shadow;
* accessibility and responsive/layout regressions introduced by the latest fixes;
* whether the complete branch remains compatible with the current `dev`;
* whether the PR description and listed validation accurately match the final implementation.

One intentional product decision to preserve when judging behavior:

* Search uses persistent Spotlight-style behavior. Tabbing away does not by itself close the search surface; Escape and the established click-outside path close it. Do not report that behavior merely because it differs from a conventional input blur pattern unless you find a concrete accessibility or correctness problem caused by it.

Also distinguish intentional choices from defects. In particular, Windows/Linux controls being on the left is intentional for this design; only report it if there is an actual functional/accessibility issue beyond platform convention.

For tests, be skeptical of false greens. Prefer observable behavior and real producer/runtime contracts over source-string checks, hand-built DOM assumptions, or mocks that merely duplicate the expected interface.

For every issue you believe remains, please provide:

* severity and whether it should block merge;
* exact `file:line`;
* concrete reproducible scenario;
* user/runtime/accessibility/security impact;
* why the issue is caused or exposed by this PR;
* recommended correction;
* test or validation required.

Keep unrelated pre-existing issues separate and non-blocking.

Please do not manufacture polish findings just to produce feedback. If an earlier concern is genuinely resolved, treat it as resolved.

Finish with a clear verdict:

* **No objection / ready to merge**
* **Ready with non-blocking follow-ups**
* **Needs fixes before merge**
* **Needs owner decision**

If there are no remaining merge-blocking issues, say exactly:

**No remaining merge blockers found.**

zach88:
Took another deep pass over the whole diff now that the fix round landed. Good news first: the six findings from the last round are genuinely closed — the tablet gutter is unified, the stacking test now asserts real paint order in a browser, the landmark/a11y fixes are clean, and the preload contract test is derived from the real producer (verified by mutation). The one thing that got worse is that the last commit, which expanded scope to align the notification drawer with the embedded-app rail, introduced a regression that defeats its own goal.

**🔴 Not ready to approve**
- **Must resolve before approval: 1** (1 High regression)
- **Optional / non-blocking: 4** (4 Low)

1. **[R2-H1] Notification drawer loses its rail alignment when you reopen an embedded app** — `ui/src/App.tsx:775,786,795` coupled to `ui/src/pages/SandboxUiPage.tsx:141-147`. **Regression from `6f628f89`.** The commit publishes the embed slot's right edge (`onEmbedSlotRightChange`) so the notification drawer fills the app rail, and `App.tsx` resets `notificationTrayLeft` to `null` in `handleSandboxUiOpening/Closed/Removed`. But `useEmbedBounds` dedupes the emission against `lastEmittedRightRef`, a `useRef` that persists across app opens (`SandboxUiPage` never remounts — no `key`). The slot's right edge is set by the chrome, so it's identical for every app. Sequence: first app `null→416`, emits → drawer aligned; reopen → `setNotificationTrayLeft(null)` but the ref is still `416` → `416 !== 416` is false → no re-emit → `notificationTrayLeft` stays `null` forever. So the drawer only fills the rail for the *first* app opened per `SandboxUiPage` mount; every reopen after that falls back to the default fixed-width, right-anchored geometry — exactly what this commit was supposed to fix.
   - Repro (ran it, two ways): open app A (aligns to 416) → "Back to apps" or open app B → reopen with the same 416 edge → `notificationTrayLeft = null`, drawer unaligned. Isolated log prints `1st open → 416 (aligned:true)` / `2nd open → null (aligned:false)`.
   - Why the `top` twin doesn't have this: `chatDrawerEmbedTop` is never reset to `null`, so its persistent ref keeps a value that's still correct. The asymmetry (resetting the state but not the ref) is the cause.
   - Fix: mirror the `top` path — don't reset `notificationTrayLeft` to `null` in `Opening` (the `notificationTrayUsesDrawer` gate at `App.tsx:2086` already prevents a stale value from leaking when there's no active app), or reset `lastEmittedRightRef.current = null` in the `useEmbedBounds` cleanup/disabled path. The first mirrors an already-correct path.
   - Test needed (T5): there's no test for the close→reopen cycle, which is exactly where it breaks. Add the reopen test alongside the fix — it should fail against `6f628f89`.

## Optional (non-blocking)

The PR can merge without these — follow-up or your call. If you're touching `main.ts` for the guard below anyway, R1-L2/L3 are a natural cleanup in the same pass.

1. **[R1-L2] Inconsistent destroyed-window guard across the `window:*` handlers** — `src/main.ts:170-196`. Still open since round 1. The guard (`!mainWindow || mainWindow.isDestroyed()`) is only on `toggleMaximize`; `getControlsState`/`minimize`/`close` use optional chaining only, which doesn't protect a destroyed-but-non-null `BrowserWindow`. On an `invoke` that lands in the race between `win.close()` and the `'closed'` event nulling `mainWindow`, those three throw "Object has been destroyed" and leave an unhandled rejection (`void controlsApi?.minimize()`). Narrow/unreachable today, but it's the whole class — the fix was applied to one instance and three identical siblings were left. Extract a `withLiveWindow(fn)` helper or repeat the guard in all four.
2. **[R1-L3] Redundant synchronous `emitWindowControlsState` in `toggleMaximize`** — `src/main.ts:190`. Still open since round 1. `wireWindowControls` already listens on `maximize`/`unmaximize`/`enter|leave-full-screen`, so state emits twice per toggle (a harmless extra re-render). Drop the line, or add a `why` comment if it's intentional immediate feedback.
3. **[R1-L5] `:has(.chat-drawer)` elevation is effectively redundant, and its comment is misleading** — `ui/src/styles.css:96-102`. Still open since round 1; reproduced (neutralized the rule and all three stacking browser tests stay green — it doesn't change the observable result). `.content-panel`'s stacking context (z=1) already traps the drawer below the titlebar (z=40). It's fine as defense-in-depth, but the comment claims it's what keeps overlays above the drawer. Fix the comment to say it's a safety net, or drop the rule.
4. **[R2-L1] No main-side coverage of the channel names** — `src/main.ts` (`window:*` handlers). Only `preload.windowControls.test.ts` pins the channel strings (renderer→invoke side). Nothing asserts `main.ts` registers `window:getControlsState/minimize/toggleMaximize/close` with those same strings — the main↔preload match is convention only, so a rename in `ipcMain.handle(...)` wouldn't break any suite. It's the counterpart of the contract the round-1 preload test pinned. Not a blocker (main is awkward to test without Electron).

## What checked out clean

- **Round-1 findings all verified closed:** tablet gutter unified at 438 with no reflow (browser test 3/3, T3 valid); nested `banner` landmarks gone (`header→div` in titlebar placement, one banner, T3 checked); stacking test now asserts observable paint order via `elementFromPoint` and would actually catch a regression; window controls get `role="group"` with a test on the accessible role; the preload contract test derives the bridge from real `preload.js` and fails under a channel rename (mutation-verified).
- **Security: nothing introduced.** All four `window:*` handlers call `assertTrustedSender` as their first statement; the trusted-sender boundary wasn't widened even though the PR adds destructive capabilities (`close`/`minimize`/`toggleMaximize`) — reachability is unchanged (embeds load over rpc-proxy HTTP, which the guard rejects). `webPreferences` untouched (`contextIsolation:true`, `sandbox:true`, `nodeIntegration:false`), only `frame:false` changed; `setWindowOpenHandler(deny)` and `will-navigate` intact. Preload stays a fixed allowlist with no leaked `ipcRenderer`.
- **Contracts changed by the fix round were audited against every caller:** `onEmbedSlotRightChange`, `notificationTrayLeft` (via the `AppHeader = HeaderActions` reexport, no orphaned prop), the `header→div` swap (`.top-bar` CSS is class-based, nothing queries `getByRole('banner')` in the titlebar path), and `role="group"`. Only the reset↔dedupe coupling is defective (R2-H1).
- **Tests:** preload contract 1/1 (+ mutation red), WindowTitleBar 9/9, 7 renderer suites 69/69, both stacking + gutter browser tests green in real Chromium.

zach88:
Took another pass, this time focused on the round-3 delta since it's tiny: the `notificationTrayLeft` reset removal in `App.tsx` (the R2-H1 fix) plus the dev merge. The merge is clean against desktop-app and the fix holds up well — the reopen regression is genuinely gone, and the repro-test actually fails on the previous head. What's left is a small pile of low-priority stuff, none of it gating.

**🟡 Approve with nits**
- Optional / non-blocking: **5**

## Optional (non-blocking)

The PR can merge without these. They're follow-up or your call — most are one-liners you could sweep in a final light pass.

1. **[R3-L1] Reopen repro-test only exercises 1 of the 3 removed resets** — `desktop-app/ui/src/__tests__/App.chatDrawer.test.tsx:449-483`. The fix removed `setNotificationTrayLeft(null)` from all three handlers (`handleSandboxUiOpening`, `handleSandboxUiClosed`, `handleSandboxUiRemoved`), but the test simulates reopen via `onOpenSandboxUiApp`→`launchSandboxUiApp` (which never touched the value) and the reset it actually exercises is `handleSandboxUiClosed` (through `onEmbeddedAppBack`). It never fires `onEmbeddedAppOpening` or `onEmbeddedAppRemoved` — and in real production flow, reopening an app *does* fire `onEmbeddedAppOpening`. So if the reset got re-added to just `Opening` or `Removed`, R2-H1 would come back and this test would stay green. Suggestion: add an `act(() => sandboxUiPageHarness.props?.onEmbeddedAppOpening?.(...))` after the reopen so all three paths the `Class-check` claims are covered.

2. **[R1-L2] `isDestroyed()` guard only on `toggleMaximize`** — `desktop-app/src/main.ts:170/175/193`. Still open since round 1. `window:toggleMaximize` (line 182) guards with `if (!mainWindow || mainWindow.isDestroyed()) return`, but `getControlsState`/`minimize`/`close` use bare `mainWindow?.` and would throw "Object has been destroyed" on an invoke landing in the race between `win.close()` and the `'closed'` event. Narrow/unreachable today, but it's the classic pattern-vs-instance thing: the idiom got fixed in one spot, three identical siblings didn't. A `withLiveWindow(fn)` helper or repeating the guard across all four closes it.

3. **[R1-L3] Redundant synchronous `emitWindowControlsState` in `toggleMaximize`** — `desktop-app/src/main.ts:190`. Still open since round 1. `wireWindowControls` already listens to `maximize`/`unmaximize`/`enter|leave-full-screen`, so state gets emitted twice per toggle (one extra idempotent re-render). Harmless; either drop the line or add a `why` comment if it's intentional.

4. **[R1-L5] Misleading comment on the `:has(.chat-drawer)` elevation** — `desktop-app/ui/src/styles.css:100`. Still open since round 1, and reproduced: neutralizing the rule leaves all three stacking browser tests green, so it's not what actually keeps overlays above the drawer (`.content-panel` z=1 already traps the drawer under the titlebar). Fine as defense-in-depth, but the comment claims it's the active mechanism. Fix is just the comment — say it's a safety net.

5. **[R2-L1] No test pins the main-side channel names** — `desktop-app/src/main.ts` (`window:*` handlers). Still open since round 2. Only `preload.windowControls.test.ts` pins the renderer→invoke channel strings; nothing checks that `main.ts` registers `window:getControlsState/minimize/toggleMaximize/close` with those same strings. A rename in `ipcMain.handle(...)` wouldn't break any suite. Counterpart to the contract the R1-L4 fix left unpinned; main.ts is awkward to test without Electron, so low priority.

## What checked out clean

- **R2-H1 is fixed and verified.** Removing the three `setNotificationTrayLeft(null)` calls makes the `left` path symmetric with `top` (which never reset). The reopen now keeps the rail alignment, and no stale-value leak on the opposite side — the `notificationTrayUsesDrawer` gate (`App.tsx:2082`) returns `null` whenever the embedded app is closed, regardless of stale state. **Reproduced**: the new test fails on the previous head (`expected null to be 716` at `App.chatDrawer.test.tsx:479`) and passes now; `App.chatDrawer.test.tsx` → 14/14.
- **No regressions from the fix.** Both the stale-leak-on-close and the app-A→app-B switch scenarios were reproduced and ruled out — the slot right-edge is shell layout geometry (same for every app), so the persisted ref value is the correct one, and real layout changes go through the `ResizeObserver` re-emit, not the reset.
- **Security: clean.** The delta is pure renderer UI state + a test — no IPC, preload, `webPreferences`, navigation, senders or tokens touched. The round-2 SAFE holds (`window:*` handlers behind `assertTrustedSender`, preload allowlist, `contextIsolation`/`sandbox` intact).
- **Dev merge is clean against desktop-app** — no semantic conflict in the service.

One tiny thing: the fix commit subject tags this as `r4-h1` but it's the same finding tracked as R2-H1 in our review — worth aligning the ID if you reference it again.

zach88 (approved):

Issue #598, opened 2026-09-09
bug(hcc): every reconcile pass probes for existence by POSTing — 14k-25k futile 409/hour, 90% of all HCC writes


alfredolopez80:
HCC decides whether an object exists by trying to create it. Every reconcile pass POSTs first and reads the 409 as the answer. The read, the no-op gate and the ownership guard that #469, #471 and #527 installed all sit **inside** the `catch (409)`, so they suppress the PUT and can never suppress the POST.

Measured on `clerum-dev` from apiserver audit logs, five 5-minute windows on 2026-09-09: **14,052–25,284 futile writes/hour**, which is **90.2%–93.9% of every write HCC makes** and **12.9%–19.6% of all writes in the cluster**.

## The measurement

`resource.type="k8s_cluster"`, `principalEmail` = the HCC service account, grouped by
`methodName` × `status.code` (`google.rpc.Code`: `0` = OK, `10` = 409). `base1`/`base2` predate the
#567/#568 deploy; `reg1`–`reg3` follow it. Rates are the 5-minute count × 12. Every percentage is
computed **within** a window.

| window | HCC writes/h | `create`→409 /h | share of HCC writes | `update`/h | PUT share | `patch`/h | share of cluster |
|---|---:|---:|---:|---:|---:|---:|---:|
| `base1` | 21,528 | 20,208 | **93.9%** | 756 | 3.5% | 96 | 16.1% |
| `base2` | 23,556 | 21,996 | **93.4%** | 744 | 3.2% | 96 | 16.9% |
| `reg1`  | 22,164 | 20,028 | **90.4%** | 540 | 2.4% | 120 | 17.8% |
| `reg2`  | 15,192 | 14,052 | **92.5%** | 336 | 2.2% | 72 | 12.9% |
| `reg3`  | 28,020 | 25,284 | **90.2%** | 444 | 1.6% | 168 | 19.6% |

`patch` is listed separately on purpose: it is `mcpservers.status.patch` and `deployments.patch`,
all status 0 — legitimate status writes, not candidates for suppression.

Absolute volume swings by a factor of 1.8 across windows. The **share does not**: in all five,
including both pre-deploy baselines, 90–94% of HCC's writes are a `create` that returns 409. This is
steady-state behaviour, not a spike. (Bounded on n=5; a base→regime step in the NetworkPolicy lane
remains unexplained — NP `create` OK tracks NP `delete` OK almost one-to-one: 62/61, 30/31, 89/88.)

[...]

PR #601, opened 2026-09-09
fix(hcc): read before creating existing resources (#598 PR 2)

alfredolopez80:
## Problem and result
HCC repeatedly attempts creates for resources that already exist, producing avoidable Kubernetes POST/409 traffic. This implements the approved second delivery of #598: read before create, then preserve each caller's convergence/preservation policy.

Converts 22 create sites and retains two documented exceptions. The inventory contains 24 observed create expressions and 26 observed read expressions. Presence is validated; create conflicts require fresh reads; admission fences and caller-specific preservation remain intact. `skipped` means create suppressed after successful presence-based convergence, not necessarily a no-op reconcile.

Following adversarial review of Claude and Zach88's feedback, Role convergence now uses the existing three-attempt conflict retry with fresh resourceVersions. Unexpected RBAC errors and exhausted conflicts still propagate. The correction also removes an unused eager Deployment build and dead create-side guards, and strengthens gate/fixture regression tests. Operators should expect previously silent RBAC errors to appear in reconciliation error telemetry.

The single SFS browser spec remains scoped to its lifecycle; production UI navigation is unchanged. The PR is consolidated as approved because the 22 conversions share one read/create contract; most added scope is regression coverage. Package version metadata follows repository hooks.

## Review decisions and rollout contract
RBAC propagation was explicitly approved on 2026-09-09 in canonical plan Addendum C.3, superseding the earlier pending B.5 decision. C.4 also supersedes B.7: measure each converted lane in stable five-minute windows and require zero redundant creates for resources proven present, with positive reconciliation witnesses. Separate cold start, legitimate new resources and races. An ext-egress prefix alone never exempts a conflict; residual conflicts require path evidence.

The forecast of 27 read expressions became 26 because the channel-reader retry now shares the validated read callback: 21 original +6 new −1 duplicate. No retry path was removed.

Rollback is PR 2 as a whole by restoring the previous HCC image through the normal deployment process; A–D are not independent rollback units. The SFS spec travels here because the approved plan required that lifecycle regression over HCC-managed resources. It is indirect coverage, not proof of read-first, and reverts with the PR in a source rollback. No production UI behavior changed.

Fable F1 now requires the correct POST/PUT path and read witnesses for the no-op case; F3 is clarified and F4 was already documented. The wider named-options cleanup is deferred to #602.

## Final local validation
Candidate `1061866b65c5f9f01ecb149fa8b617dd66d8f325`, base `c7741481a7794209102a392f589cdf904cda43a1`.

- Node24; HCC 2263/2263 tests plus build/typecheck. Control UI 1934/1934 plus build.
- Canonical branch-owned Minikube orchestrator: exit0, T0 PASS, T1 PASS, T2 PASS, deployed NP-08 authorization PASS, user-facing health PASS.
- T1 used actual PostgreSQL serially: 164/164 tests across all 35 selected files, zero pending.
- Direct compiled `ensureHostRole` integration against the real Kubernetes API: four cases passed (GET404→POST; existing/no POST or PUT; actual stale-RV409→fresh GET→successful PUT; three actual409s propagate without false skipped). UID-preconditioned cleanup completed and namespace GET404 was verified. This is direct method integration, not deployed-controller/watch E2E.
- Actual authorization checks for the deployed HCC ServiceAccount allow Role get/create/update in mcp-host.
- Independent GPT-6 Astra final correction review: no remaining confirmed findings. Historical Codex Security scans cover the pre-correction candidate, not a new scan of this SHA.

The modified SFS Playwright spec was subsequently executed on this same final SHA: 2/2 passed in 52.8 seconds, one Chromium worker, zero retries/skips/flaky results. It verified visible creation, folder rename/persistence, cleanup and the unauthenticated route/API guard. The exact-head runtime closure passed NP-08, health and Playwright, reusing the existing T0/T1 attestations for this SHA. Static E2E audit: one file scanned, zero findings. This accredits the changed browser spec; it remains separate from direct HCC read/create behavior evidence. Local database, runtime, direct-method tests and GitHub CI are separate lanes. No production rollout, post-deployment savings or universal blast-radius guarantee is claimed. The unhealthy prior local profile was preserved; a fresh exclusively owned profile validated the identical commit. Local artifacts remain ignored and uncommitted.

Refs #598. Rollout and measured comparison remain follow-up gates; this PR does not close them.


alfredolopez80:
@claude Please perform an exhaustive, detailed closing re-review and blast-radius audit of this entire PR at `523acce196923c0892ef2b7da99bf7ef9706171e` against `dev` (`c7741481a7794209102a392f589cdf904cda43a1`). Verify alignment with issue #598 and assess regressions in existing production reconciliation: read/create races, retries, resourceVersion handling, admission fences, RBAC error propagation, metadata preservation, metrics and test coverage. Examine the SFS journey and negative-test sensitivity. Local T0/T1/T2, browser and readiness checks passed as detailed in the PR; CI is separate and still running. Independently verify claims, distinguish introduced defects from pre-existing issues, and report concrete evidence, severity, remaining gaps and anything you did not execute. Please confirm whether any blockers remain before merge.

zach88:
Took a deep pass through this one across the helper, all the migrated call sites, security, contract consistency and test quality. Really clean delivery: `ensureResource` is well designed, the 22 call sites each preserve their caller's convergence policy faithfully, and every verifiable claim in the PR body checked out against the code. Full service suite is green (2257/2257). Only thing worth a decision is one metric/error-handling asymmetry on the Role path, plus a few nits.

**🟡 Approve with nits**
- Optional / non-blocking: 5 (1 Medium, 4 Low)

## Optional (non-blocking)

The PR can merge without these. They're follow-up or author's call.

1. **[R1-M1] Role converge propagates a no-retry 409 as `reconcile_exception`** (`host-context-controller/src/hostReconciler.ts:1108-1131`). The old code wrapped the Role read+replace in a try/catch that swallowed any replace error (including a stale-`resourceVersion` 409) and returned. The new `converge` rethrows everything non-404, and since `reconcileCore` doesn't wrap `ensureHostRole`, it bubbles to `reconcile()` and enqueues `controller_error`. It's deliberate and tested (`hostReconciler.rbacPvcReadFirst.test.ts:226`), but it's inconsistent with the siblings: the Host PVC still swallows a replace 409 (`ensurePvc` converge returns false) and Service/Deployment retry it via `replaceWithConflictRetry`. Only the Role escalates. Repro: green test at HEAD confirms the propagation; the old swallow was confirmed reading the removed catch block. Realistically low-probability (the per-Host serializer makes concurrent writes to the same Role rare), so it won't fire often, but a routine `secretRef` rotation racing another write to the same Role would now dirty `controller_error` instead of self-healing next reconcile. Suggestion: treat the Role's 409 as benign (return false) or give it conflict-retry, to line it up with PVC/Service/Deployment.

2. **[R1-L1] Wasted Deployment build in `ensureDeployment` converge** (`hostReconciler.ts:3472`). `body: this.buildDeployment(...)` is built on every converge but discarded, since `resolveBody: buildDesiredDeployment` always overrides it (`replaceWithConflictRetry` uses `resolveBody ? await resolveBody() : body`). Pointless work per converge call.

3. **[R1-L2] Gate coverage gap** (`src/__tests__/applyNetworkPolicyNoopGate.test.ts`). No test covers `mutationAllowed()===false` from entry producing zero GET. The code is correct (guard in `utils.ts:61` runs before any read), but a regression that moved the read ahead of the gate wouldn't go red. Suggestion: add a case where `mutationAllowed` returns false on first call and assert `readNamespacedNetworkPolicy` isn't called.

4. **[R1-L3] Dead guard in create wrappers** (`hostReconciler.ts:3372-3378`, `:3421-3426`). The `isBenignSupersessionError(error)` guard inside the `create` wrappers of `ensurePvc`/`ensureService` is effectively dead: only `observeCreate` (k8s API) errors flow there, never benign supersession. It's meaningful inside `converge`, just not in these two `create` bodies. Harmless, a bit confusing.

5. **[R1-L4] Hand-written fixture in NoopGate tests** (`llmHookNetworkPolicyNoopGate.test.ts:126-177`). The `desired` passed to `asApiserverNetworkPolicy` in NOOP-LLMNP-2/3 is hand-written rather than derived from the real builder. It's self-correcting (if the builder diverged the test would go red), but leaves the fixture free to drift. NOOP-LLMNP-3 already derives `rules` from `buildHostEgressRules`; worth doing the same for name/labels/podSelector.

A few behavior changes I looked at and confirmed are intentional and documented, so they're not findings: the extra GET on the POST-409 path in `applyPvc`/`applyService`, the sub-millisecond TOCTOU on the first replace (snapshot handed to attempt 1), external-egress staying fail-closed if the policy vanishes mid-reconcile, and the skip-count asymmetry between the exists path and the create-409 path (metric-only).

One process note, not a finding: this lands at 45 files / 2 services, over the soft budget. Most of it is tests and the production surface is really one service (HCC) plus a single e2e, so it's not a big PR in practice, but a one-liner in the body on why it isn't split would be nice.

## What checked out clean

- **Security: SAFE.** Every new read-first GET has `get` RBAC coverage in `deploy/base/{gfs,mcp-host}/rbac.yaml`. 403 vs 404 is handled correctly (only 404 means absent, so an RBAC error can't trigger a bad create). NetworkPolicy is never left unapplied (fail-closed). Credential rotation is preserved (it rides the pod-template annotation, and converge always runs for existing Deployments). No secrets in the new logs, no new deps.
- **Per-call-site policy preservation:** verified across all 22 (retry budget, admission/noop fence, `validateExisting`, `missingIsError`, ordering, the channel-reader double-404 policy). The post-read `mutationAllowed` recheck is a stricter fence, not a loosening.
- **PVC/SFS:** read-first correctly skips a bound PVC / existing Service without a destructive replace of immutable fields.
- **Contract consistency:** 22 call sites consistent, `kind` labels match between read/create/onSkipped at every site, and `skipped` semantics are coherent end to end (create-409 path doesn't count a skip, `skipped`+`writes` coexist by design).
- **PR body:** every verifiable claim is true (22 converts + 2 documented exceptions, 24 observed creates / 26 observed reads matching `CREATE_OBSERVABILITY.md`, RBAC propagation, the "no existing resource is lost" invariant backed by executable tests, e2e labels aligned with the real UI via the `/agent-files` rewrite).
- **Test quality (T1-T5):** fixtures derived from real producers, and T3 verified by running the new tests against the merge-base (`sharedFileSystemReconciler` 21 fail, `applyNetworkPolicyNoopGate` 13/13 fail), so they don't lock in the bug. Assertions target observable results (metrics + API calls).
- **Suites:** full service suite 2257/2257 green, no collateral.

alfredolopez80:
@claude Please perform an exhaustive closing re-review and blast-radius audit at `2c9897e68beaedb43915f18913f3c7e588cc4492`, especially Zach88’s R1-M1 and R1-L1–L4 corrections. Verify bounded Role retries, fresh resourceVersions, admission checks, error propagation, metrics and all affected reconciliation paths for regressions. Local T0/T1/T2 and deployed NP-08 passed; PostgreSQL ran 164/164 tests across 35 files. Four direct integration cases against the real Kubernetes API also passed, including actual 409 recovery and exhaustion. These are direct method tests, not deployed-controller E2E; optional Health/Playwright are explicitly NOT_RUN. Please independently verify issue #598 alignment, distinguish introduced defects from pre-existing risks, and state remaining blockers and checks you did not execute.

alfredolopez80:
## Adversarial review — PR #601 (stage 2 of #598, read-first)

**Head reviewed:** `2c9897e68beaedb43915f18913f3c7e588cc4492` · **merge-base:** `c7741481a7794209102a392f589cdf904cda43a1` · 7 commits · 45 files.
**Mode:** read-only. No edits, branches, commits or write operations. All mutations were applied to scratchpad copies.

### Verdict

**Approve with documentation conditions — no code blockers.** The PR implements Addendum B and does so correctly in everything mechanically verifiable. None of the 9 vacuity mutants survive; the full suite is green with a real exit 0. The conditions are three claims missing from the docs/PR body (B.5 explicit approval, B.7 per-lane criterion, B.8 rollback unit), plus one scope item the reviewer should decide deliberately rather than discover later.

---

### Should be addressed before merge

**S1. B.5 — GET≠404 propagation in the RBAC trio was resolved without evidence of explicit approval (process, not code).**

The plan states: *"pending explicit user approval before stage D… Do not resolve it implicitly during implementation."* At head, `host-context-controller/src/hostReconciler.ts:1032-1046` (ServiceAccount), `:1118-1137` (Role) and `:1160-1185` (RoleBinding) rethrow everything except 404. `reconcileCore` `:4410-4415` does not wrap the trio, so the error reaches `:4307-4321` and is queued as `reconcile_exception`.

Concrete scenario: a transient 403 or a transport failure on `GET roles/<host>` aborts the **entire** Host pass — Deployment, Service and NetworkPolicies are not reconciled in that pass — and marks `controller_error`. On base it was logged and the pass continued.

The behaviour is declared in the PR body and in `CREATE_OBSERVABILITY.md`, and the `get` verbs are granted (`deploy/base/mcp-host/rbac.yaml`), so operational risk is low. But the plan required explicit approval and I could not find it: not in issue #598 (0 comments), not in `.local-notes/reviews/*`, not in `.local-notes/infra/runs/issue598-pr2/*.json`. The two author comments on the PR ask reviewers to verify "RBAC error propagation" as a settled fact, which is implicit ratification, not the approval the plan asked for.

Minimal fix: confirm option (a) in a PR comment, or add a line to the plan's Addendum recording the decision. No code change.

**S2. B.7 — the per-lane / 5-minute-window acceptance criterion appears neither in the PR nor in `CREATE_OBSERVABILITY.md`.**

B.7 requires documenting: 0 creates in steady state across 8 lanes; in the NetworkPolicy lane only `ext-egress-*` may show `conflict`; cold start measured separately. The doc's "Read-first contract" section says *"verify that POST is suppressed while convergence… Separate legitimate new resources, races and cold start. A lower create count alone does not establish causality"* — correct but generic. It does not fix the window or the `ext-egress-` exception.

Minimal fix: one paragraph in `CREATE_OBSERVABILITY.md` with the literal B.7 criterion.

**S3. B.8 — the rollback unit (previous image; PR 2 as a whole) is not declared** in the PR body or the doc. Minimal fix: one line.

**S4. Scope creep: full rewrite of the Playwright spec `control-ui/e2e/qa-recorder-shared-fs-folders.spec.ts`** (+203/−85, across `b08523790`, `1ec1b7aa8`, `42385869a`, `523acce19`) and `control-ui/package.json` bumps 0.1.564→0.1.568.

Unrelated to read-first or to HCC. The PR justifies it in one sentence. @zach88 also flagged it as a process note. Risk: if PR 2 has to be reverted per B.8, an unrelated UI spec is reverted with it. Not blocking — the Playwright lane is NOT_RUN and the spec does not touch production. Minimal fix: split it out, or state explicitly in the PR body why it travels here.

**S5. Deviation 27→26 reads vs Addendum B.3/B.9, unexplained in the PR.**

The plan projected 27 (`existenceReadMetrics.test.ts` 21→27). Head has `:253` → `toBe(26)` and `:254` excluded → `toBe(43)`. Cause, verified from the diff: in `reconcileChannelReaderDeployment` (`hostReconciler.ts:2597-2670`) the outer read and the retry read were merged into a single expression `const existing = observed ?? (await read())` at `:2612`; base had two `observeExistenceRead('Deployment'` at `:2579` and `:2611`. So 21 − 1 + 6 = 26. Correct, but the PR states 26 without saying why it is not 27. Minimal fix: one sentence.

### Follow-up (non-blocking)

- **F1.** `src/__tests__/reconciler.credentialsRotation.test.ts:~105-115` — `createdDeployment` now accepts `replace ?? create`. It loses specificity: a rotation test would pass even if rotation arrived via POST instead of PUT. The `beforeEach` (`:126-155`) installs a stateful `live` mock, so in practice pass 1 creates and pass 2 replaces; better for each test to assert which path it expects.
- **F2.** `applyNetworkPolicy` (`utils.ts:564-602`) now takes 9 positional parameters (`existing` is the 9th). Ergonomics; candidate for an options object.
- **F3.** `latestReadSucceeded` in `ensureResource` (`utils.ts:51-120`) is invocation state, not snapshot state. Correct today; worth noting in the helper comment that a failed `readFresh` after POST-409 leaves `skipped` un-incremented by design.
- **F4.** `skipped` is incremented even when `converge` emitted a PUT (`utils.ts:105-108`). Consistent with B.6 and the `metrics.ts` help text, but dashboards must not read `skipped` as "no write occurred".

---

### Mechanical verification performed

**Full suite at head** (`git archive 2c9897e68`, whole tree, not just HCC), `vitest run --reporter=json`:
**428 files · 2263 total · 2263 passed · 0 failed · 0 pending · `EXIT=0`**, captured as `cmd > log 2>&1; echo EXIT=$?` with no pipe. `tsc --noEmit` → `TSC_EXIT=0`. Matches the 2263/2263 claimed in the PR. ESLint not reported (pre-existing exit 2 from the pending flat-config migration).

**Vacuity mutations** (isolated copies, HCC-only suite, 2237 tests). The HCC-only baseline has 7 ENOENT failures from files outside the package; those are subtracted. Net figures are tests failing *because of* the mutation:

| Mutant | What it disables | New failures | Files |
|---|---|---:|---:|
| M1 | Read-first helper neutered at its entry (`existing = null`, back to POST-first) | **248** | 26 |
| M2 | Post-GET `mutationAllowed` recheck removed (`utils.ts:93`) | **13** | 7 |
| M3 | Snapshot never consumed (reused after POST-409) | **21** | 13 |
| M4 | Rule 5 (`validate`: `metadata.name` required) removed | **6** | 2 |
| M5 | `gfsK8sApi.ts` PVC/Service: converge emits PUT | **6** | 1 |
| M6 | `sharedFileSystemReconciler.ts` PVC/Service: converge emits PUT | **7** | 1 |
| M7 | `result !== false` gate on `onSkipped` removed | **5** | 3 |
| M8 | `onSkipped()` added after converge-by-POST-409 | **19** | 5 |
| M9 | `onSkipped()` called unconditionally before converge | **14** | 4 |

All EXIT=1. **No mutant survives.** The finest ones (M4, M5, M7) are caught by dedicated tests (`ensureResource.test.ts`, `hostReconciler.readFirst.test.ts`, `gfsK8sApi.readFirst.test.ts`, `hostReconciler.rbacPvcReadFirst.test.ts`), not only by the AST inventory.

**Repository integrity:** `git status --porcelain` → 0 lines in both checkouts. `diff -rq` between a freshly extracted `git archive 2c9897e68 host-context-controller` and the working copy (excluding `node_modules/dist/coverage`) → **0 differences, EXIT=0**. Mutations lived only in the scratchpad. No branches or stashes created; `gh` used read-only.

**CI:** `gh pr checks 601` → 30 checks, all `pass`.

---

### Addendum B conformance

| Item | Status | Evidence |
|---|---|---|
| B.2 fixtures not flattened to 404 | PASS | `testMocks` keeps present objects with labels/name/uid; tests inject 404 explicitly per case. `egressObservation.test.ts` moved injection to PUT-409. No new `.skip/.only/.todo`. |
| B.3 wrappers in place; inventories | PASS (26, see S5) | `createMetrics.test.ts` → 24; `existenceReadMetrics.test.ts:253-254` → 26/43. All four POST-first tests rewritten with a read witness. |
| B.4 rule 5 + test | PASS | `utils.ts:63-68`; `ensureResource.test.ts` parameterised (`undefined`, `null`, `{}`, `{metadata:{}}`, `{metadata:{name:''}}`). M4 confirms. |
| B.5 GET≠404 across 6 sites | **See S1** | Implemented (option a) and declared; explicit approval not evidenced. |
| B.6 `skipped` once per invocation with no POST, never per retry | PASS | `utils.ts:105-108`; M7/M8/M9 caught. |
| B.7 per-lane / 5-min criterion | **FAIL — see S2** | Not documented. |
| B.8 rollback unit | **FAIL — see S3** | Not documented. |
| B.9 inventory 22+2=24 | PASS | 22 `ensureResource` (utils 1, hostReconciler 8, reconciler 3, llmHook 3, SFS 3, gfs 4) + runtime Secret + `createSafetyPolicy`. |

### Behavioural contracts (read + mutants)

- **Helper** `ensureResource` `utils.ts:51-120`: entry fence (`:61`), validate (`:63-68`), initial read (`:82-92`), post-GET fence (`:93`, M2), consume-once snapshot (`:96-104`, M3), fence before converge-on-409 (`:116`).
- **D9 with liveness witness**: `applyNetworkPolicyNoopGate.test.ts:189` GATE-NP-0 (denied at entry → `read` not called, with `mutationAllowed` as the witness) and `:210` GATE-NP-1 (expires during the GET → no POST, no PUT, with GET called exactly once).
- **`createSafetyPolicy`** `networkPolicyReconciler.ts:3505-3566`: untouched, POST-first with `deleteSafetyPolicySnapshot` compensation; does not reuse the compensation branch.
- **ext-egress**: `applyOwnedPolicy` `:3334-3361` passes `fresh.policy` as `existing`; `assertExternalEgressSnapshot` `:3363-3376` re-asserts uid+RV on every read via `validateExisting`; fail-closed if it disappears.
- **PVC/Service in gfs and SFS**: converge = `ignoreNotFound(read)` / `try read catch 404`, no PUT (`gfsK8sApi.ts:46-62`, `:154-170`; `sharedFileSystemReconciler.ts:275-304`, `:395-424`). M5/M6 caught.
- **SFS Deployment** keeps `preserveDeploymentAnnotations`/`deploymentMatchesDesired` (`:362`); llmHook NP keeps `mergeExisting: preserveObjectAnnotations`.
- **channel-reader** `:2597-2670`: owner mismatch → `false`; `validateExisting` throws on ownership change; `mergeExisting` rebuilds from `fresh.spec?.replicas`.
- **`reconciler.ts`**: the post-build `if (!isCurrent()) return` is replaced by `mutationAllowed: isCurrent` at entry and in `replaceWithConflictRetry` — equivalent or stricter.
- **Role with bounded retry** `:1118-1137` (@zach88's R1-M1 resolved in `2c9897e68`); `role-direct.json` documents 4 cases against the real API, including 3×409 propagated.
- **@zach88's R1-L1..L4 nits**: all addressed at head (L1: `ensureDeployment` no longer builds a discarded body, only `resolveBody` `:3499`; L2: GATE-NP-0; L3: dead guard removed at `:3374-3380`; L4: fixtures derived from the builder at `llmHookNetworkPolicyNoopGate.test.ts:93,185` or captured from the create at `:124,147`).

### Negative-assertion audit

Every `not.toHaveBeenCalled` added across the 20 touched test files carries a liveness witness (`read.toHaveBeenCalledTimes(n)`, `toHaveBeenCalledExactlyOnceWith({name,namespace})`, `replace.toHaveBeenCalledOnce`, captured structured logs). The only weakening is F1. Reduced read counts (`test/hostReconciler.test.ts` retries 3→2) are a real consequence of removing the redundant GET, not weakening.

### Process

- **Version bumps**: HCC 303→309 and control-ui 564→568, one per commit touching the package — consistent with `.githooks/pre-commit` → `npm run version:staged`.
- **T0/T1/T2 lanes** at exact head: `.local-notes/infra/runs/pr601-focused-20260910/t0-t1-t2-evidence.json` (mergeBase `c7741481a`, T0 PASS 358 s, T1 PASS 164/35/0 pending, T2 PASS, NP-08 PASS, orchestrator exit 0); `status.json` confirms head `2c9897e68`. Health/Playwright NOT_RUN, declared.

### Not verified

- Cluster figures (T0/T1/T2, NP-08, `role-direct.json`): the orchestrator was not run and no cluster was touched; the evidence JSONs are accepted as declared by the author.
- Control UI 1934/1934: not re-run locally; CI `Build (control-ui)` passes.
- Explicit user approval for B.5 (S1): observable only by the user.
- Real behaviour under load of the sub-millisecond TOCTOU window on the first replace (snapshot handed to attempt 1): confirmed as intentional by design and by the external reviewer, not measured.

---

**Merge summary:** code correct, suite green with a real exit 0, 9/9 mutants caught, CI green, repository intact. Conditions before merging: record the B.5 approval, add B.7/B.8 to the doc/PR, decide explicitly on the control-ui spec (S4), and explain 26 vs 27 (S5). None requires a logic change.

alfredolopez80:
@claude Please re-review `1061866b65c5f9f01ecb149fa8b617dd66d8f325` against the Fable review and audit remaining regressions/blast radius. The docs now record the approved C.3 RBAC decision, C.4 acceptance criteria with five-minute windows and no automatic prefix exemption, whole-PR2 rollback, intentional SFS test scope, and 26=21+6−1 reads. F1 now requires the correct POST/PUT path and read witnesses for no-op; F3 is clarified, F4 was already documented, and F2 is tracked in #602. No production logic changed. Local T0/T1/T2, NP-08 and health passed; T1 was 164/164 across 35 files, and four real-Kubernetes Role cases passed on this SHA. Playwright remains explicitly NOT_RUN. Reporter clarification: 428 suites correspond to 105 physical files, not 428 files. Please verify these changes and identify any remaining blockers or checks you did not execute.

alfredolopez80:
Current-head Playwright evidence: the modified `control-ui/e2e/qa-recorder-shared-fs-folders.spec.ts` was executed against `1061866b65c5f9f01ecb149fa8b617dd66d8f325` in the exclusively owned Minikube lane. **2/2 passed in 52.8s**, Chromium, one worker, zero retries, skips or flaky results. The visible SFS creation/folder-rename/persistence journey and the unauthenticated route/API guard passed, including the spec's cleanup assertions. Static E2E audit: one file, zero findings.

The canonical runtime-only closure exited0 with T2, NP-08, Health and Playwright PASS. T0/T1 reused their already-green attestations for this exact SHA. This supersedes the earlier Playwright NOT_RUN statement; no source change or new commit was needed. This browser evidence remains lifecycle regression coverage, separate from the direct real-Kubernetes HCC tests.

alfredolopez80:
### S4 — resolution: declare the browser spec a manual lane

The repository owner has decided how S4 closes: **option (b)** — the Playwright spec stays as it is
and runs manually; it is not wired into CI. No code change is required. One statement is missing from
the PR body.

The body already establishes that the spec was executed on the final SHA (2/2 passed, 52.8s, one
Chromium worker, zero retries/skips/flaky) and that it is indirect coverage rather than proof of the
read-first contract. That part is accurate and closes the "was it run" question.

What the body does not say is that **nothing will run it again**. Verified on this head:

- `control-ui/package.json` defines only `test: vitest run` and `test:watch: vitest`. There is no
  `e2e` or `playwright` script.
- `control-ui/vitest.config.ts` `include` covers `lib/**/__tests__/`, `lib/__tests__/`,
  `components/__tests__/` and `app/**/__tests__/`. The `e2e/` directory is in none of them.
- `.github/workflows/ci-public.yml:123` installs Playwright only under
  `if: matrix.service == 'desktop-app'`, so control-ui specs are never installed for, nor invoked by,
  CI.

A reader who sees "2/2 passed" and "exact-head runtime closure" will reasonably assume this is a lane
that re-runs on future commits. It is not. Please add a sentence to the PR body making that explicit,
for example:

> This spec is a manual lane. It is not included in `vitest run` and CI never installs Playwright for
> `control-ui`, so it does not re-run on future commits and must not be counted as automated
> regression coverage.

Placing the same note as a header comment in the spec file itself would help the next person who
edits it, but the PR body is the part that matters for merge.

No other findings from the previous review remain open: S1, S2, S3, S5, F1 and F3 are resolved, and
F2 is deferred to #602 as agreed.

zach88:
Did a round-2 pass focused on the fix delta since the last reviewed head (`523acce1..1061866b`) — the Role conflict-retry rework plus the test/doc changes. Ran it through a code pass, a security pass, and a dedicated "what did this fix break?" collateral pass. Short version: all five findings from round 1 are closed and nothing regressed. This is a clean follow-up.

## Verdict
🟢 **Ready to approve**

## Optional (non-blocking)
The PR is mergeable as-is; this is just a body-accuracy sync you already decided on.

1. **[S4] Spell out that the control-ui Playwright spec is a manual lane in the PR body.** You resolved S4 as option (b): the spec stays and runs manually, not wired into CI. The body says it was run 2/2 on the final SHA and that it's indirect coverage, but it doesn't yet say the part that matters for the next reader — that *nothing re-runs it*. `control-ui/package.json` only defines `test: vitest run` (no `e2e`/`playwright` script), `control-ui/vitest.config.ts` `include` doesn't cover `e2e/`, and `.github/workflows/ci-public.yml:123` installs Playwright only for `desktop-app`. One sentence like "this spec is a manual lane, not part of `vitest run`, and CI never installs Playwright for control-ui, so it doesn't re-run on future commits" keeps someone from reading '2/2 passed' as automated regression coverage. No code change.

## What checked out clean
All five round-1 findings are closed in `2c9897e`, verified against the head:

- **[R1-M1] Role 409 handling now matches PVC/Service/Deployment.** `ensureHostRole` converge routes the replace through `replaceWithConflictRetry` (bounded fresh-RV retry; a terminal 409 only propagates after the budget is exhausted). The admission fence (`mutationAllowed`/`revalidate`) is re-checked *before every write attempt* (`utils.ts:201,205`), not once — a boundary lost mid-retry aborts without writing stale permissions. Disappearance policy is intact (only 404 is swallowed; auth/transport/terminal-conflict propagate, fail-closed). Covered by 4 new retry/exhaustion/admission-recheck tests, and I confirmed they **fail against the previous head** (`523acce1`) — they don't golden the bug.
- **[R1-L1]** Dropped the wasted eager `body: buildDeployment(...)` on the Deployment converge; `resolveBody` already won, so it's a pure no-op removal.
- **[R1-L2]** Added `GATE-NP-0`: mutation denied at entry ⇒ zero GET, no create/replace.
- **[R1-L3]** Removed the dead `isBenignSupersessionError` guard from the PVC/Service *create* catches (unreachable there — create never runs `revalidate`); kept it on the *converge* catches where the fence actually runs.
- **[R1-L4]** The NP no-op fixture is now derived from the real producer instead of hand-built (and it asserts the read args too).

Other checks:
- **Collateral (fix delta): 0 regressions.** `revalidate` didn't get lost (it moved into `mutationAllowed`, still called on both the skip and pre-replace paths); the `body` XOR `resolveBody` union change compiles and behaves identically across all call sites; no existing test was weakened to fit the new code — the test changes are strictly tighter (the rotation test now splits create-only vs replace-only witnesses, and the key-order no-op asserts neither-create-nor-replace instead of a trivially-true body compare).
- **Security: safe.** Fence re-checked per attempt, RBAC propagation fail-closed, no new secrets in logs, no new deps (just the version bump).
- **Suite:** typecheck exit 0, full service suite 2263/2263 green.

zach88 (approved):

PR #605, opened 2026-09-10
feat: add openai-compatible LLM provider with brokered LAN egress

zach88:
## What

Add a generic `openai-compatible` LLM provider that lets an operator point a Host at a self-hosted / local OpenAI-compatible endpoint on their private LAN (e.g. `http://192.168.1.50:8000/v1`), served end-to-end through a dedicated in-cluster egress broker. The change spans the full vertical:

- **Shared contract** (`packages/llm-providers`): new `openai-compatible` provider id (23 → 24) with an **optional** credential slot (`OPENAI_COMPATIBLE_API_KEY`) — the first provider whose api key is optional, since local servers often need no auth. Derives to `static-credentials` / `static`, empty non-secret env.
- **CRD** (`charts/clerum-crds/crds/host.yaml`): `openai-compatible` added to `spec.model.provider` and `spec.llmPolicy.fallbacks[].provider` enums, plus an additive `baseURL` field (string, maxLength 2048) on both slots. Shape-only CEL guards: a biconditional `provider == 'openai-compatible' <=> has(baseURL)` and a case-insensitive `!baseURL.contains('.svc')` reject of cluster-internal service names.
- **control-api**: admission-time validation of `baseURL` (IP-literal RFC1918 only; DNS names rejected outright) via the new shared `@clerum/egress-policy`, plus a read-after-write `host_crd_outdated` 409 guard that catches a stale CRD silently pruning `baseURL`.
- **host-context-controller**: per-`(Host, slot)` egress broker reconciler — a hardened nginx reverse-proxy Deployment/Service/ConfigMap, an optional api-key mirror Secret, and three NetworkPolicies (a `/32` broker→LAN egress as the sole LAN route, broker ingress limited to the owning Host's mcp-host pods, and an mcp-host→broker egress lane).
- **mcp-host**: serves an `openai-compatible` model by dialing its dedicated broker's internal Service URL (derived deterministically from `(hostName, slotId)`), **never** the operator's LAN endpoint directly. Fail-closed if the broker URL can't be derived.
- **control-ui**: LAN `baseURL` input on the primary model and every fallback row, with client-side validation reusing the shared classifier (UX only; the authoritative gate is control-api admission + the broker `/32`).
- **deploy**: `llm-egress` namespace scaffolding (PSA baseline, static default-deny NetworkPolicy, HCC RBAC), broker namespace/port envs wired on both sides, opt-in cluster-CIDR overlay patch, and docs.

## Why

Operators want to run models on their own hardware (LAN GPU box, self-hosted inference server) instead of a public provider. The hard constraint is that giving the mcp-host a route to a private LAN address is an SSRF and data-exfiltration surface. The design keeps the mcp-host with **no LAN route at all**: the only component that can reach the operator's endpoint is a per-endpoint broker pinned by a `/32` NetworkPolicy to the exact classified IPv4 literal and port, with no DNS resolution anywhere in the path.

## How (key decisions)

- **Single source of truth for the egress rule.** `@clerum/egress-policy` owns the non-public CIDR set, the LAN classifier (`classifyLanBaseURL`), and the deterministic broker-naming derivation (`brokerNameFor` = `oai-egress-<sha256(host.name + slotId)[:16]>`). control-api, HCC, mcp-host and control-ui all import it — no duplicated copies of the rule across the seam, so HCC and mcp-host can never disagree on a broker name without a handshake.
- **Defense in depth, not a single gate.** DNS names are rejected at admission (closes SSRF-by-name and DNS-rebinding); the authoritative per-IP block is the broker's runtime `/32` NetworkPolicy. HCC re-classifies defensively and is fail-closed — it drops any slot that isn't a clean RFC1918 literal.
- **Additive field, rollout-order-sensitive.** `baseURL` is additive; an outdated CRD prunes it and the apiserver returns 200. control-api's read-after-write guard converts that silent loss into a `409 host_crd_outdated` pointing at the field and the `kubectl apply` command.
- **nginx config built from parsed URL components**, never the raw string; strict path reject-class; api key rides an env-fed conditional `Authorization` map (never enters the ConfigMap); `access_log off` so prompts aren't logged.
- **Whitespace-hardening**: the provider match uses `.trim()` like the rest of the file so a padded `'openai-compatible '` can't route as local at runtime while slipping past the baseURL gate.

## Testing

Coverage lives in the branch (not re-run in this PR step):

- `packages/egress-policy/index.test.cjs` + `control-api/test/egressPolicy.property.test.ts` — classifier and CIDR set, property-based.
- `control-api/test/routes.hostSpecValidation.test.ts` — admission gate, incl. the whitespace-padded-provider repro (fails against the strict `===` version; verified).
- `control-api/test/routes.resources.hostBaseUrlPruning.test.ts` — the `host_crd_outdated` read-after-write guard.
- `control-api/test/crd.llmProviderEnums.test.ts`, `llmProviders.test.ts`, `llmProviderMirrors.test.ts` — contract/enum invariants.
- `host-context-controller/src/openaiEgressBrokerReconciler.test.ts` — broker reconcile, teardown, orphan sweep; deployment-writer inventory.
- `mcp-host/src/llm/__tests__/registry.test.ts`, `failover/__tests__/policy.test.ts`, `hostLlmBinding.test.ts` — broker-URL derivation, fail-closed, failover slot identity, baseURL-only hot-reload.
- `control-ui/components/__tests__/LlmProviderConfig.test.tsx`, `lib/__tests__/llmLanBaseUrl.test.ts` — baseURL reset-on-provider-change / preserve-on-model-change, client validation.

## Risks / Follow-ups

- **CRD is structural — apply to the dev cluster before merge.** `spec.model.baseURL` / `fallbacks[].baseURL` are additive fields; an outdated CRD prunes them silently. Rollout order: **CRD → control-api → control-ui**. The control-api guard only protects requests that pass through an updated control-api.
- **Large PR (80 files, ~8 services) by design.** This is a single additive field threaded through a tightly coupled vertical (contract → CRD → control-api → HCC → mcp-host → control-ui → deploy); splitting by service would ship a half-wired feature where one layer references a field the layer below doesn't yet accept. The layers are individually reviewable via the one-commit-per-layer history.
- **Deferred:** functional unification of `isAllowedExternalEgressCidr` between control-api and HCC (they differ intentionally — control-api enforces canonical CIDR, HCC does not); broker dedup (one broker per slot for now); `workflowrecipe.yaml` not touched.
- The cluster-CIDR overlay patch is opt-in and not wired into the default overlay; a deploy that never renders it runs the always-on RFC1918 + `/32` tier.

https://claude.ai/code/session_01FA1UCpeczsDfTE9VrnjQXg


alfredolopez80:
@zach88 — full review of this PR at head `1c4a7f73d`. Recommendation: **do not merge yet**. There is 1 Blocker (no service image can build) and 4 High findings. Everything else can follow up.

**Findings:** Blocker 1 · Critical 0 · High 4 · Medium 8 · Low 16 (1 candidate finding discarded after verification). "REPRODUCED" means executed (script, test, render or simulated build); "READ" means traced in code only.

**CI:** 38/38 checks pass, but CI builds no service images (see B1). Full suites pass locally on Node 24 (details at the end).

---

## Blocker

### B1: `packages/egress-policy` is missing from every service's Docker build context
- **Where:**
  - `control-api/Dockerfile:9,41`
  - `host-context-controller/Dockerfile:14,42`
  - `mcp-host/Dockerfile:16,39` (and `Dockerfile.slim/.full/.desktop`)
  - `control-ui/Dockerfile:13,23-24`
  - the allowlists in `control-api/Dockerfile.dockerignore:9-10`, `control-ui/Dockerfile.dockerignore:16-17` and `host-context-controller/Dockerfile.dockerignore`
- **Defect:** all four services add `"@clerum/egress-policy": "file:../packages/egress-policy"` and import it at runtime. No Dockerfile copies it: `grep -c egress-policy` returns 0 in all four, while `llm-providers` appears in each.
- **Failure:** in the builder layout, `npm ci` exits 0 but leaves a dangling symlink. `require('@clerum/egress-policy')` then throws `MODULE_NOT_FOUND` and `tsc` fails with `TS2307`, so none of the feature's images can be built.
- **Why CI misses it:**
  - `ci-public.yml` builds no images, and `build-publish.yml` only runs on push to `dev`.
  - `scripts/tests/test-dockerfile-local-packages.sh:185-195` checks a hand-written package list per Dockerfile that this PR didn't extend, so it passes without checking the new package.
- **Status:** REPRODUCED by simulating the builder layout.
- **Fix:**
  - Add `COPY packages/egress-policy` before each `npm ci`, in both builder and runtime stages.
  - Add it to the `.dockerignore` allowlists.
  - In control-ui, materialize it into `node_modules` the same way as `llm-providers`.
  - Extend the lists in `test-dockerfile-local-packages.sh` so the test fails without it.
  - Add `packages/egress-policy/**` to `deploy/images.json` `source_paths`.

---

## High

### H1: The broker can reach node IPs, the real apiserver endpoint and VPC ranges, on any port (CWE-918)
- **Where:**
  - `packages/egress-policy/index.cjs:112-119`
  - `host-context-controller/src/openaiEgressBrokerReconciler.ts:174-185, 365-369, 562, 741-742`
  - `deploy/scripts/minikube-detect-cluster-cidrs.sh:40-60`
  - `deploy/overlays/minikube/patches/llm-egress-cluster-cidrs.yaml.template:31-34`
- **Defect:** the cluster-internal guard covers only these ranges:
  - the pod CIDR
  - the Service CIDR
  - `KUBERNETES_SERVICE_HOST/32`
  - the optional `k8sApiCidrs`, which the template leaves unset

  It does not cover:
  - the node CIDR
  - `hostNetwork` listeners
  - the apiserver's actual endpoint IP
  - private control-plane ranges, such as GKE's `172.16.0.0/28`

  The LAN port can be anything from 1 to 65535, and https upstreams use `proxy_ssl_verify off`.
- **Scenario:** a Host admin sets `baseURL: https://192.168.49.2:10250/` (kubelet on minikube) or `https://192.168.49.2:8443/` (the apiserver via the node IP). HCC writes the `/32` to that node, and nginx forwards mcp-host traffic there with an attacker-chosen `Authorization: Bearer`. The SSRF is mostly blind because the `/chat/completions` suffix is fixed, but it breaks the mcp-host egress boundary, under which direct access to private pod and node IPs is blocked.
- **Status:** REPRODUCED at the classifier and guard level with the minikube CIDRs (`10.244.0.0/16,10.96.0.0/12,10.96.0.1/32`):
  - `https://192.168.49.2:10250/` → `ok`
  - `http://10.128.0.5:10255/v1` → `ok`
  - `http://10.96.0.1/v1` → `cluster_internal`
- **Fix:**
  - Require the node CIDR and the apiserver endpoint CIDRs in the fail-closed set.
  - Allowlist LAN ports, or at minimum deny 10250/10255/2379-2380/6443/8443.
  - Revisit `proxy_ssl_verify off`.

### H2: A Host admin can send any other provider key in the Host's Secret, including the shared `chatllm-api-keys`, to a LAN IP they control (CWE-441 / CWE-200)
- **Where:**
  - `host-context-controller/src/openaiEgressBrokerReconciler.ts:300-306`: `dataKey = fb.credentialSlot?.trim() || OPENAI_COMPATIBLE_API_KEY_SLOT`.
  - `:455-469`: `readHostCredential` reads `host.spec.secretRef` with HCC's own namespace-wide permissions.
  - `:489-507, 530-533`: the key is mirrored into `llm-egress` and sent upstream as Bearer.
  - `control-api/src/routes/admin/hostSpecValidation.ts:53-60, 481, 890-925`: `openai-compatible` has one non-multiline slot, so a fallback `credentialSlot` is allowed and only format-checked.
  - `control-api/src/routes/admin/hostSecrets.ts:50-66`: the `secretRef` guard checks the Secret's type, not its owner, and accepts `chatllm-api-keys`. Its own comment documents this as a residual.
- **Scenario:** the Host admin writes:
  ```yaml
  spec:
    secretRef: chatllm-api-keys
    llmPolicy:
      fallbacks:
        - provider: openai-compatible
          baseURL: http://<attacker LAN box>:8000/v1
          credentialSlot: anthropic-api-key
  ```
  HCC mirrors the shared Anthropic key into the broker Secret, and nginx sends it as `Authorization: Bearer …` to the attacker's box. Each fallback can leak one more key.

  Before this PR, the documented cross-Host residual only let one Host *use* another key against fixed provider endpoints. Now the key can be *read*.
- **Precondition:** the attacker controls, or can sniff, an RFC1918 address that the cluster can route to.
- **Status:** READ. I traced it end to end in code but did not run it in a cluster. Open question: whether an operator-set `AZURE_OPENAI_ENDPOINT` already allowed an equivalent leak to a public host. If it did, this PR widens an existing path rather than opening a new one.
- **Fix (either option):**
  - For `openai-compatible`, reject any `credentialSlot` other than `openai-compatible-api-key`.
  - Before mirroring, require that the Secret belongs to the Host (owner or label).

### H3: Editing an existing slot's `baseURL` never restarts the broker
- **Where:**
  - `host-context-controller/src/openaiEgressBrokerReconciler.ts:405-409`: the rollout revision hashes only the credential.
  - `:604-606`: that hash is the only pod-template annotation.
  - `:516-529, 637`: the image renders `.conf.template` with envsubst once, at startup.
- **Scenario:** changing a slot's IP, port or path keeps the same `slotId`, so it keeps the same broker. The ConfigMap is updated and the LAN NetworkPolicy moves to the new `/32`, but the pod template is unchanged. No rollout happens, so the running nginx keeps dialing the old IP, which the new `/32` now blocks. Every request fails until someone deletes the pod.

  Swapping two local fallbacks triggers the same failure: `fallback-0` and `fallback-1` keep their names while their upstreams swap.
- **Status:** REPRODUCED by driving the reconciler against a stubbed API: `proxy_pass` and the NetworkPolicy change, and the Deployment pod template does not.
- **Fix:** add a pod-template annotation with a hash of the rendered config, or of ip, port and path.

### H4: control-ui cannot create an `openai-compatible` Host with "New secret", and blocks saving when the linked Secret has keys
- **Where:** `control-ui/lib/llm.ts:275-289`, in `getLlmGroupCompleteness`, which computes `usable: required.length > 0 && …`. Only `oauth-broker` is exempted. It is used in:
  - `control-ui/components/HostWizard/index.tsx:70-71, 196-198, 562`
  - `control-ui/app/hosts/[name]/page.tsx:911-927`
- **Scenario:** `openai-compatible` has zero required slots, so `usable` is always `false`.
  - In the wizard with `secretMode: 'new'`, step 1 never advances ("Add the OpenAI-compatible (local) credential…"), whether the key is blank or filled in.
  - On the edit page, Save fails with "does not contain a usable … credential" whenever the linked Secret has any keys.

  The only workaround is "existing secret" mode in the wizard. The new `LlmProviderConfig` tests never pass `credentials`, so this path isn't covered.
- **Status:** REPRODUCED by running the extracted logic: blank key → `false`, filled key → `false`.
- **Fix:** treat a `static-credentials` provider with no required slots as `usable: true`, the same way `oauth-broker` is treated. Add wizard and edit-page tests for `openai-compatible`.

---

## Medium

- **M1: a malformed cluster-CIDR entry leaves the guard open (CWE-636).**
  - **Where:**
    - `host-context-controller/src/config.ts` splits on `,` with no validation.
    - `openaiEgressBrokerReconciler.ts:184`: `guardConfigured = length > 0`.
    - `packages/egress-policy/index.cjs:74-78`: `cidrOverlaps` returns `false` for an unparseable CIDR.
  - **Repro:** with any of `["10.96.0.0"]`, `["10.96.0.0/12;10.244.0.0/16"]`, `[" 10.96.0.0/12"]` or `["fd00::/108"]` configured, `https://10.96.0.1/` classifies as `ok`. **REPRODUCED.**
  - **Fix:** fail at startup, or refuse to provision, when any entry doesn't parse, and count only valid entries as "configured".
- **M2: the CIDR detector accepts finding only one of the two ranges.**
  - **Where:** `deploy/scripts/minikube-detect-cluster-cidrs.sh:43-60`. Every extraction ends in `|| true`, and the script exits only when **both** ranges are missing.
  - **Failure:** without `--cluster-cidr`, it renders `value: "10.96.0.0/12"` and exits 0. HCC then treats the guard as configured while pod IPs stay unguarded.
  - **Also:** the format is never validated, and a `&` in the value is expanded by `sed` (`:67`). **REPRODUCED** with a stubbed kubectl.
- **M3: binding a Codex subscription to a Host that was `openai-compatible` fails with a misleading 503.**
  - **Where:** `control-api/src/routes/admin/codexSubscription.ts:252-322`. The file isn't in the diff; this PR makes the path reachable.
  - **Cause:**
    - `writeHostConnectionRef` shallow-copies `spec.model`, so `baseURL` survives the switch.
    - `validateHostSpec` checks only `openai-compatible ⇒ baseURL`, never the reverse (`hostSpecValidation.ts:159-224`).
    - The apiserver's CEL rule (`host.yaml:113-115, 264-265`) rejects the write, and the generic catch turns that into `503 hosts_unavailable`.
  - **Status:** **REPRODUCED.**
  - **Fix:** enforce the reverse direction in control-api, and/or drop `baseURL` when the provider changes.
- **M4: `baseURL: ""` passes the CRD, provisions nothing, and reports success.**
  - **Cause:** CEL `has(self.baseURL)` accepts `""`. control-api rejects it, but kubectl and GitOps writes bypass control-api. HCC's `if (!baseURL) return` (`openaiEgressBrokerReconciler.ts:282`) skips the slot without recording a drop.
  - **Result:**
    - The Host condition reads `True / AllSlotsProvisioned / "0 broker(s) provisioned"`.
    - mcp-host builds no provider.
    - The UI shows nothing.
  - **Status:** **REPRODUCED.**
  - **Fix:** add `size(self.baseURL) > 0` to the CEL rule and record a `missing_base_url` drop in HCC.
- **M5: there is no limit on brokers per Host (CWE-770).**
  - **Cause:**
    - `fallbacks` has no `maxItems` (`host.yaml:192-194`).
    - control-api doesn't cap the fallback count.
    - `llm-egress` has no ResourceQuota or LimitRange.
  - **Impact:** each `openai-compatible` fallback creates a Deployment, a Service, a ConfigMap, a Secret and 3 NetworkPolicies. **READ.** This is about a cap, not the deferred per-endpoint dedup.
- **M6: `packages/egress-policy` tests never run in CI.** The package is missing from the `ci-public.yml:48-72` matrix and from `TEST_SERVICES` in `Makefile:56-77`. Its 13 tests, covering the classifier, the CIDR set and the third contract-vector check, never execute in CI.
- **M7: several fail-closed tests have no liveness witness and pass with the feature disabled.**
  - **HCC** (`clusterCidrGuard.test.ts:92-107`, G1/G2; `openaiEgressBrokerReconciler.test.ts:334-377`, T7/T7b/T7c): still green when `reconcileForHost` returns immediately.
  - **mcp-host** (`src/llm/__tests__/registry.test.ts`, the three "fail-closed: … builds NO provider" tests): still green when the `openai-compatible` branch returns `null`. They mock `console.error` but never assert the "cannot derive egress-broker URL" message.
  - The targeted mutations do fail these tests, so they check the right condition. What they don't prove is that the path ran.
  - **Status:** **REPRODUCED** with vacuity mutations.
  - **Fix:** add a witness, such as the condition reason, the emitted log line, or a read count.
- **M8: the null-`slotIndex` guard in mcp-host has no test.** `mcp-host/src/main.ts:532`, in `buildFallbackProvider`, is the only thing stopping a local fallback without a `slotIndex` from dialing the primary's broker, and no test calls it. **READ.**

---

## Low

- **L1: the API key is substituted raw into nginx config syntax (CWE-94).**
  - **Where:** `openaiEgressBrokerReconciler.ts:530-533`, together with the image's envsubst entrypoint.
  - **Effect:**
    - `"`, `;` or `}` in the key can inject config blocks.
    - `$` breaks the config.
    - A trailing newline ends up inside the `Bearer` header.
  - Only the key's owner can trigger it, so the practical risk is a broker crash on unusual keys. **REPRODUCED** by rendering with envsubst.
  - **Fix:** validate the key's charset.
- **L2: the classifier accepts non-http(s) schemes and userinfo.** `gopher://`, `file://`, `ws://` and `http://user:pw@192.168.1.1/v1` all return `ok`. control-api stores them, which leaves the credential visible in the spec (CWE-312), while HCC drops them at runtime. **REPRODUCED.**
- **L3: the broker ingress podSelector checks only `clerum.io/host`** (`openaiEgressBrokerReconciler.ts:707-713`). The mcp-host-side policy (`:768-770`) and the repo convention also require `MANAGED_BY_LABEL`.
- **L4: rotating or revoking a key takes up to one 300 s resync** to reach the broker, because nothing watches the Secret.
- **L5: the path limits disagree across layers.**
  - The CRD allows `maxLength: 2048`, but HCC drops paths over 1024 (`:377`). That drop shows up only in the status condition.
  - A `%20` path passes every layer, yet nginx matches `location` against the decoded URI, so requests 404.
- **L6: the 409 pruning guard isn't wired into the codex-bind write** (`codexSubscription.ts:301`). During the outdated-CRD window, a fallback `baseURL` can be silently pruned while the write returns 200.
- **L7: an old control-ui can rewrite the Host during rollout.** Its `normalizeProvider` maps the unknown id to `openai`, and its save drops `baseURL` (`dev: control-ui/lib/llm.ts:467-469`, `page.tsx:887-896`).
- **L8: whitespace in `baseURL` is handled inconsistently.**
  - The UI trims the primary `baseURL` but not the fallbacks' (`HostWizard/index.tsx:722`, `page.tsx:963`).
  - `hostLlmBinding.ts:38` compares raw strings, so a whitespace-only edit triggers an unnecessary rebind.
- **L9: 6 new `console.*` calls in `host-context-controller/src/k8sClient.ts`.** The file already uses `console.*` throughout; the new reconciler correctly uses `hccLogger`.
- **L10: `llm-egress` enforces PSA `baseline` although the broker pod already meets `restricted`** (`deploy/base/namespaces.yaml:147`).
- **L11: `deploy/images.json` `source_paths` omit `packages/egress-policy/**`,** the same gap as `llm-providers`.
- **L12: `getEnvNumber` turns `8080-typo` into `8080` without a warning** (`mcp-host/src/config.ts:314-328`). The broker port now depends on it. `config.brokerPort.test.ts:27-36` describes `parseInt` truncation as a fallback.
- **L13: test gaps.**
  - Nothing asserts the nginx `proxy_pass` upstream or the https branch.
  - The pruning test has no update-with-fallback case.
  - No test for fallback reordering.
  - No test for a failed rollback on create.
  - `test-lint-networkpolicies.sh` doesn't render the `llm-egress` default-deny.
- **L14: dead and duplicated code.**
  - The `reserved` branch in `classifyLanBaseURL` is unreachable (`index.cjs:117-127`).
  - `seenSlotIds` (`:273,283`) never triggers.
  - `export { brokerNameFor }` has no importers (`:145`).
  - The key slot constants are duplicated in HCC (`:88-89`), even though `PROVIDER_CREDENTIAL_SLOTS` exists.
  - `LAN_BASE_URL_REASON_MESSAGE` is copied between control-api and control-ui.
  - `validateSlot` and `buildNginxConfigMap` are pure and could move out of the 1169-line reconciler, which would also make H3 testable.
- **L15: comments that don't match the code.**
  - `deploy/base/llm-egress/networkpolicies.yaml:6-7` mentions a "scoped CoreDNS egress" that HCC never emits.
  - The `.gitignore` comment mentions "apiserver CIDRs", which the template doesn't set.
  - The lazy `require('crypto')` in `egress-policy/index.cjs:169-173` does not keep the polyfill out of the bundle.
  - The HCC yaml says namespace and port are "pinned on both sides", but mcp-host is pinned only in the minikube overlay.
- **L16: the PR description and docs have drifted from the code.**
  - The description calls the cluster-CIDR patch opt-in. At head it's applied by default (`deploy/overlays/minikube/kustomization.yaml:195`, `Makefile:67,1024`, `full-setup.sh:269,277`), and base leaves the feature inert (fail-closed). Without the generated file, `kustomize build deploy/overlays/minikube` exits 1.
  - The description says 80 files; the diff has 99.
  - `docs/deploy/llm-providers.md` §8 says one broker per "(Host, endpoint)"; it is one per slot.
  - The stated rollout order (CRD → control-api → control-ui) leaves out HCC, mcp-host and the `llm-egress` namespace and RBAC. An order that avoids silent states: CRD → `llm-egress` (namespace, RBAC, CIDR patch) → HCC → mcp-host → control-api → control-ui.

---

## Verified clean
- **Classifier canonicalization.** Decimal, hex, octal, short, trailing-dot and percent-encoded IPv4 all resolve to the same dotted quad, and that value is what the `/32` and the nginx upstream use. The special cases classify as:
  - `[::ffff:192.168.1.1]` → `not_ip`
  - `169.254.169.254` → `link_local`
  - `100.64.0.1` → `cgnat`
  - `127.0.0.1` → `not_private_lan`
- **`brokerNameFor`.** It separates host and slot with `\x1f`, so there is no concatenation collision. All 3 contract vectors match in HCC, mcp-host and egress-policy.
- **Slot identity.** Both sides use the raw array index, so a malformed fallback doesn't shift the indices after it. The `slotIndex == null` check fails closed and doesn't treat index 0 as absent.
- **mcp-host never dials the LAN.** It uses only the `baseURL` pathname, and env autodetection excludes this provider.
- **nginx config is built safely from the URL.** The path rejects `[\r\n\t "'\\;{}$#\`\0]` and anything over 1024 characters. The port is range-checked, the host is the classified literal, and the query string is dropped. The key never enters the ConfigMap.
- **NetworkPolicies.** `llm-egress` has a default-deny for ingress and egress. Brokers get no DNS egress. The egress rule is a `/32` plus port, and each peer combines its namespace and pod selectors, so both must match.
- **Broker pod.** It runs as non-root (UID 101), drops ALL capabilities, disallows privilege escalation, uses RuntimeDefault seccomp, and has no service-account token.
- **RBAC.** It is a namespaced Role with exactly the verbs used; no new cluster-wide grants.
- **No secrets or private URLs leak.** Neither the key nor the `baseURL` appears in logs, the status condition or the 409 body.
- **409 read-after-write guard.** It compares against the object returned by the same write, so there is no TOCTOU window.
- **CRD CEL.** The rules are nested correctly, and `.svc` matching is case-insensitive.
- **CIDR list moved to the shared package.** `BLOCKED_EGRESS_CIDRS` → `NON_PUBLIC_EGRESS_CIDRS` keeps the same 18 entries in the same order.
- **Commit `1c4a7f73d`.** It only widens the existing test-file exemption and removes concrete LAN IPs from shipped files. It does not narrow what the guard scans.
- **Lockfiles.** Only `file:` link entries and version bumps; no new registry packages.
- **control-ui form state.** `baseURL` resets when the provider changes and survives a model change, on the primary and on each fallback row. Reordering moves whole rows. The broker condition renders as plain text.

## Local test run (Node 24, `npm ci` without `--ignore-scripts`)

| Suite | EXIT | Passed / failed / skipped |
|---|---|---|
| packages/egress-policy | 0 | 13 / 0 / 0 |
| control-api (full) | 0 | 5118 / 0 / 158 (skips are the `realPostgres` suites) |
| host-context-controller (full) | 0 | 2305 / 0 / 0 |
| mcp-host (full) | 0 | 3782 / 0 / 0 |
| control-ui (full) | 0 | 1967 / 0 / 0 |
| Changed shell tests (hermetic) | 0 | 104 assertions across 5 scripts |

- **Typecheck:** `tsc` is clean for control-api, HCC and mcp-host. control-ui's 7 `tsc` errors are all in files this PR doesn't touch.
- **Lint:** HCC has no ESLint setup on `dev` either, so its lint doesn't run.
- **Not run:** `next build` and a real `docker build`.

## Open questions for the maintainers
1. **Host-admin threat model.** If a Host admin can be an untrusted tenant, H1 and H2 become Critical. If Host admins are trusted platform operators, they should at least be documented as accepted residuals.
2. **Broker cap (M5).** Add `maxItems` on `fallbacks`, a ResourceQuota on `llm-egress`, or both?
3. **Rollout order.** Settle the official order (L16) before deploying.

alfredolopez80:
@zach88 — re-review of this PR at head `1f17b67ed`, after the `R4-*` fix commits. The prior review was at `1c4a7f73d`.

**Recommendation:** no Blocker, Critical or High findings remain.
- The prior Blocker (B1) and H1, H3 and H4 are fixed and verified.
- H2 is fixed for the cross-provider leak. A same-provider residual is still open (M1 below).
- Fix **M2** and **M3** before merging:
  - M2 is a reproduced availability bug that leaves a false `True` condition.
  - M3 breaks editing of existing Hosts that don't use this feature.
- Whether M1 and M4 block merging depends on the Host-admin threat model (see the open questions at the end).
- Everything else can be a follow-up.

**Findings:** Blocker 0 · Critical 0 · High 0 · Medium 9 · Low 30.
- **REPRODUCED:** we executed it (script, targeted test, mutation, render or build).
- **READ:** we traced it in the code only.

**CI:** 39/39 checks pass. All local suites pass on Node 24 (details at the end).

---

## Status of the prior review's findings

| ID | Status | Evidence |
|---|---|---|
| B1 egress-policy missing from images | **FIXED** | `85935c6b9`: COPY is in both stages of every Dockerfile, and `test-dockerfile-local-packages.sh` checks it. Removing the COPY makes the test fail (REPRODUCED). |
| H1 broker reaches node/apiserver | **FIXED** | `9fcaa35f7`, `52c868aa2`: node-CIDR guard is required and fail-closed, plus a port denylist. `:10250`/`:8443`/`:10255`/`:010250` → `port_denied`; node IP → `cluster_internal` (REPRODUCED). Residual: M4. |
| H2 other providers' keys leak to LAN | **PARTIAL** | `b2c889179`, `f3a430079`: control-api, CEL and HCC all reject a foreign `credentialSlot` (`claude-api-key` → `credential_slot_not_owned`). Residual: M1. |
| H3 baseURL edit doesn't roll pod | **FIXED** | `830fa747f`: `config-revision` pod-template annotation. The replace path works (REPRODUCED) but has no test: M7. |
| H4 UI blocks openai-compatible | **FIXED** | `7833c48f9`. Wizard and edit-page tests fail when the fix is reverted (REPRODUCED). |
| M1 malformed CIDR leaves guard open | **FIXED** | `2cad0ba75`: the config loader throws at startup. Side effect: Low L3. |
| M2 detector accepts one range | **FIXED** | `9f5b0505b`: the detector aborts when a range is missing and splices values literally with awk. |
| M3 Codex bind 503 | **FIXED** | `736925f7c`: reverse check added, and `baseURL` is dropped on bind. |
| M4 `baseURL: ""` silent | **FIXED** | `666f744ff` adds CEL `size>0`; `e1c16b329` adds a `missing_base_url` drop. |
| M5 no broker cap | **FIXED (cap)** | `MAX_LLM_FALLBACKS=8` enforced by the CRD (`maxItems`), control-api and the UI. There is still no ResourceQuota on `llm-egress`. |
| M6 egress-policy tests not in CI | **FIXED** | Added to the `ci-public.yml` matrix and to `Makefile` `TEST_SERVICES`. 16/16 pass. |
| M7 fail-closed tests had no witness | **FIXED** for the tests named | The same vacuity mutations now make them fail. The same pattern remains in other tests: L23. |
| M8 null-slotIndex guard untested | **FIXED** | `2a9de33b7` adds the `buildFallbackProvider` seam with tests. Those tests have no witness: L23. |
| L11 images.json source_paths | **FIXED** | |
| L1, L2, L3, L4, L5, L6, L7, L8, L9, L10, L12, L13, L14, L15, L16 | **OPEN / PARTIAL** | Folded into the Medium and Low findings below. |

---

## Medium

### M1: A Host can send another Host's `openai-compatible` key to its own LAN IP (H2 residual; CWE-200 / CWE-639)
- **Where:** `host-context-controller/src/openaiEgressBrokerReconciler.ts:298-311`, `oaiEgress/slotAdmission.ts:89,117`, `control-api/src/routes/admin/hostSecrets.ts:61-65` (the Secret check is a type gate, not an owner gate).
- **Scenario:** Host A sets `spec.secretRef: <Host B's LLM Secret>` and `baseURL: http://<attacker LAN box>:8000/v1`. HCC copies B's `openai-compatible-api-key` into A's broker Secret, and nginx sends it as `Authorization: Bearer`. The same holds for `chatllm-api-keys` if it ever stores that key. On the kubectl path, no layer checks Secret ownership.
- **Status:** REPRODUCED (the mirror Secret held Host B's key). The upstream send is READ. Reported independently by 4 agents.
- **Fix:** before mirroring, require the referenced Secret to carry an owner label matching `host.name`, or refuse shared or foreign Secrets for `openai-compatible` slots. Otherwise, document it as an accepted residual.

### M2: The orphan sweep can delete a broker that a concurrent reconcile just created
- **Where:** `host-context-controller/src/openaiEgressBrokerReconciler.ts:968-995`.
- **Defect:** `sweepOrphans` builds its `desired` snapshot before its three awaited LIST calls. It then runs `gcBroker` under a `gc:<broker>` key, which does not wait for the `host:<name>` key a live reconcile holds. The comment at `:989-993` says this cannot race, but it can: a Host added while the LISTs are in flight is missing from the snapshot.
- **Scenario:** a Host is created, or gains a local slot, during a resync sweep. The reconcile creates the broker and writes `True/AllSlotsProvisioned`. The sweep's LIST then returns that broker, and the sweep deletes the Deployment, Service, ConfigMap, Secret and all 3 NetworkPolicies. The local LLM stays down until the next Host event or the 300 s resync, and the condition still says `True`.
- **Status:** REPRODUCED by forcing the interleaving.
- **Fix:** serialize the sweep under `host:<name>` (brokers carry the host label), or re-check the live Host cache for each broker right before `gcBroker`. Fix the comment too.

### M3: The new `credentialSlot` ownership rule applies to every provider in control-api, which breaks existing Hosts
- **Where:** `control-api/src/routes/admin/hostSpecValidation.ts:996`, `control-ui/lib/llm.ts:838`.
- **Defect:** on `dev`, a fallback `credentialSlot` was only format-checked. control-api now requires `isCredentialSlotOwnedByProvider` for **all** providers. The CRD CEL (`host.yaml:277`) and HCC restrict only `openai-compatible`, and there is no tolerance for already-stored specs.
- **Scenario:** a stored Host has fallback `{provider: claude, credentialSlot: backup-claude-key}`. Every later request now returns 422 even when the incoming spec is byte-identical: PUTs that don't touch the LLM, and Codex binds (`codexSubscription.ts:292`).
- **Related UI mismatch:** the UI offers Secret keys matched with `key.startsWith(prefix)` and no `-`, so `openai-api-key2` is offered and then rejected on save.
- **Status:** REPRODUCED (the stored spec is identical and the response is still 422). Reported by 3 agents.
- **Fix:**
  - Either limit the rule to `openai-compatible`, matching the CRD and HCC, or allow an unchanged stored slot and document the migration.
  - Change the UI filter to `key === p || key.startsWith(p + '-')`.
  - Update the stale comments at `hostSpecValidation.ts:39-43,528` and `host.yaml:256`, which still say "format-checked only".

### M4: The node deny set is a startup snapshot, and the port denylist has gaps (CWE-918)
- **Where:** `host-context-controller/src/oaiEgress/clusterDenySet.ts:34-50`, `deploy/scripts/minikube-detect-cluster-cidrs.sh:172-211`, `packages/egress-policy/index.cjs:55-58`.
- **Defect:** HCC reads `CONTEXT_MAPPER_CLUSTER_NODE_CIDRS` once at startup and never compares it with the live Node list. The detector writes each node as a `/32`. It adds the node subnet only when `docker network inspect` answers, so not on a non-docker driver, and not when an operator declares `/32`s.
- **Scenario:** node `192.168.49.3` joins after the render. On the HCC path these are provisioned:
  - `http://192.168.49.3:30080/v1` (NodePort)
  - `:10249` (kube-proxy metrics)
  - `:9100` (node-exporter)
  - `:2381` (etcd metrics)

  On GKE, if `masterIpv4CidrBlock` is left out, `https://172.16.0.2/` is also accepted. Ports 10250, 6443, 2379-2380 and 8443 stay blocked.
- **Status:** REPRODUCED at the classifier and `deriveDesiredBrokers` level. That there is no live re-discovery is READ.
- **Fix:**
  - Render node subnets instead of `/32`s, or have HCC watch Nodes and fail closed when a node InternalIP is not covered.
  - Add 10249, 2381, 9099-9100 and 30000-32767 to the port denylist.

### M5: control-api and the UI accept `baseURL`s that HCC then drops, so writes return 2xx with no broker (prior L2 + L5)
- **Where:** `control-api/src/routes/admin/hostSpecValidation.ts:258-266`, `control-ui/lib/llm.ts:115`, `packages/egress-policy/index.cjs:105-161`, and the stricter HCC checks in `host-context-controller/src/oaiEgress/slotAdmission.ts:179-205`.
- **Scenario:** each of these passes the UI and control-api (201/200), and then HCC drops it. The only signal is the status condition.

  | `baseURL` | HCC drop reason |
  |---|---|
  | `gopher://`, `file://` or `ftp://` on a LAN IP | `scheme_unsupported` |
  | `http://192.168.1.10:0/v1` | `port_invalid` |
  | a path containing `;`, `$`, `'` | `path_unsafe` |
  | a path longer than 1024 chars (the CRD allows 2048) | `path_unsafe` |

  `http://user:secret@192.168.1.10/v1` is accepted by every layer. The credential stays in the Host spec (CWE-312), and nginx discards it.
- **Status:** REPRODUCED. Reported by 4 agents.
- **Fix:** move the http(s)-only scheme check, the empty-userinfo check, the 1-65535 port range and the path-safety regex with its 1024 limit into `classifyLanBaseURL`. Then control-api, the UI and HCC share one decision. Align the CRD `maxLength`.

### M6: The mcp-host startup log prints the LAN `baseURL`, including any userinfo (CWE-532)
- **Where:** `mcp-host/src/main.ts:702`: `console.log('[Main] Model config:', host.spec.model)`.
- **Defect:** the log line predates this PR. The PR adds `baseURL` to `spec.model`, and M5 lets `http://user:pw@192.168.1.50:8000/v1` through, so every boot and every Host change logs a private URL and possibly a credential. That breaks the repo logging standard. It also makes the prior review's "verified clean: no private URLs in logs" wrong for mcp-host.
- **Status:** READ.
- **Fix:** log only `{provider, name}`, and reject userinfo (M5).

### M7: The H3 fix is tested only on the create path
- **Where:** `host-context-controller/src/openaiEgressBrokerReconciler.ts:709-737` (`ensureDeployment`, the 409-then-replace path) and `openaiEgressBrokerReconciler.test.ts:651-697` (T16, T16c).
- **Defect:** no test references `replaceNamespacedDeployment` or `replaceNamespacedConfigMap`. With the replace path mutated into a no-op, all 2434 HCC tests still pass. A regression would silently bring H3 back.
- **Status:** REPRODUCED (mutation stays green). A throwaway update-path test confirmed the code works today.
- **Fix:** stub create to return 409 with an existing Deployment carrying the old `config-revision`, and assert the replace call receives the new value. Add the same test for the ConfigMap.

### M8: No test checks the rendered nginx upstream or the https branch
- **Where:** `openaiEgressBrokerReconciler.ts:362-363, 389, 411-413`.
- **Defect:** three mutations together leave all 2434 HCC tests green: `proxy_pass` pointed at `10.255.255.1`, the path dropped, and the `proxy_ssl_*` directives removed.
- **Status:** REPRODUCED.
- **Fix:** assert on the created ConfigMap:
  - `proxy_pass http://192.168.1.50:8000/v1;` for http
  - the `proxy_ssl_*` lines for https

### M9: control-ui bundles `crypto-browserify` into its client chunks
- **Where:** `packages/egress-policy/index.cjs:199-206` (the lazy `require('crypto')` in `brokerNameFor`), imported by `control-ui/lib/llm.ts`.
- **Defect:** `next build` puts crypto-browserify (about 327 KB raw, about 100 KB gzip, including secp256k1, ed25519 and aes) into a chunk that 22 route manifests reference, for a function the UI never calls. The comment says the lazy require prevents this; it doesn't.
- **Status:** REPRODUCED (Turbopack build output).
- **Fix:** do one of the following, and correct the comment:
  - add a crypto-free subpath export (classifier plus reason constants)
  - add `"browser": {"crypto": false}`
  - move `brokerNameFor` into a server-only module

---

## Low

**Egress broker / HCC**
- **L1** (prior L1, open): the API key is substituted raw into nginx config through envsubst (`openaiEgressBrokerReconciler.ts:372-375`). A `"`, `;` or `}` breaks out of the quoted string, and the broker crash-loops while the condition still says `True`. REPRODUCED (render). Fix: validate the charset in control-api and HCC.
- **L2** (CodeQL alert 1366 is valid): `clerum.io/credentials-revision` on the broker pod template (`:459`, via `oaiEgress/brokerRevision.ts:30-31`) is an unsalted `sha256(base64(key))`. Anyone who can read pods or Deployments in `llm-egress` can check key guesses offline. Local keys like `ollama` fell to a 6-word dictionary, and the empty-key hash shows publicly that an endpoint has no auth. REPRODUCED. Fix: use an HMAC with a key only the controller holds, or hash the source Secret's `resourceVersion`. The exported `credentialsRevision()` helper (`brokerRevision.ts:12`) has no callers and can be deleted.
- **L3** (regression from the M1 fix): `parseK8sApiCidrs` accepts IPv6 entries, but they are fed to the IPv4-only classifier. One `fd00::1/128` in `CONTEXT_MAPPER_K8S_API_CIDRS` drops every broker with `cluster_cidr_invalid`, so dual-stack clusters get no brokers (`clusterDenySet.ts:38-44`). REPRODUCED. Fix: filter to IPv4 first, and correct the comments that say the list is validated.
- **L4**: the detector's CIDR check is looser than HCC's. `CLUSTER_NODE_CIDRS=10.128.0.5/20` renders with exit 0, and then HCC throws `non-canonical` at startup and crash-loops. A dual-stack node InternalIP aborts the whole render (`minikube-detect-cluster-cidrs.sh:64-88`). REPRODUCED.
- **L5**: every `list*` helper turns an API error into `[]`. As a result, teardown, GC and the sweep succeed without doing anything, leaving the LAN `/32` policy and the key Secret of a deleted Host in place (`:833-846, :843, :858, :998-1011, :1023`). REPRODUCED. Fix: rethrow so the reconcile is retried.
- **L6**: `OpenAiEgressBrokersReady=True` means the API writes succeeded, not that the pod rolled out. A crash-looping broker shows `AllSlotsProvisioned`. If `gcHostBrokers` throws, the condition write is skipped (`:241-274`). The True-on-bad-key case is REPRODUCED; the rest is READ.
- **L7**: `fullReconcile` sends 3 label-selected LISTs per Host every 300 s, including Hosts with no `openai-compatible` slot, and repeats slot admission for each (`:211-213, :821-826`). REPRODUCED (5 Hosts → 6 LISTs of each kind).
- **L8**: mcp-host builds a provider for a slot that HCC dropped, and dials a Service that doesn't exist (`mcp-host/src/llm/index.ts:141-151`). The primary counts as constructible, so boot fallback is skipped. Each task makes 3 attempts with a 600 s timeout and gets a generic "Connection error." The only diagnosis is the HCC condition. REPRODUCED.
- **L9**: the `proxy_ssl_verify off` rationale comment (`:400-411`) says an attacker must "already control that exact LAN host". ARP or DHCP spoofing on the segment is enough to capture the `/32` and the Bearer key. READ. Fix: correct the comment and track CA pinning.
- **L10**: the `hostReconciler` orphan-cleanup inventory (`hostReconciler.ts:5044-5049`) counts `oai-egress-*-src` NetworkPolicies but deletes only a fixed name list (`:4224`). A leftover `-src` policy keeps re-triggering cleanup until the broker sweep removes it. READ.
- **L11** (prior L12 and L15): mcp-host gets the broker namespace and port only from the minikube overlay ConfigMap; the base comment (`host-context-controller.yaml:142`) says they are "pinned on both sides". The port parsers also differ:

  | Input | HCC | mcp-host |
  |---|---|---|
  | `0`, `-1` | accepted | replaced with 3000 |
  | `8080-typo` | — | 8080, no warning |
  | `70000` | — | accepted |

  REPRODUCED on the mcp-host side. Fix: HCC injects both env vars into the mcp-host pods it builds, and both sides accept only `^\d+$` in 1..65535.

**control-api / CRD**
- **L12**: the CEL `.svc` rule (`host.yaml:65,270`) scans the whole URL string, while the classifier checks only the host. `http://192.168.1.10:8000/v1.svc/` passes control-api, and the apiserver then rejects it with a misleading "cluster-internal service" message. REPRODUCED on the control-api side.
- **L13**: non-canonical IPv4 is silently rewritten. `http://192.168.001.010/v1` is read as octal and becomes `192.168.1.8` for the `/32` and the upstream, while the spec and UI still show the raw string. REPRODUCED. Fix: reject when the parsed host differs from the literal host.
- **L14**: on a fresh outdated CRD, the provider enum rejects the write with 422 before the `host_crd_outdated` 409 guard can fire. The guard only fires after a CRD rollback with ratcheting, and its tests stub an apiserver that accepts the enum (`resources.ts:239-266`). READ.
- **L15** (prior L6, open): the Codex bind write (`codexSubscription.ts:306`) has no prune guard. It is reachable only in the rollback-with-ratcheting window. READ.

**control-ui**
- **L16**: the wizard's step 1 never calls `validateLlmPolicy`. An `openai-compatible` fallback with a blank `baseURL` shows an inline error but Next stays enabled. The flow then POSTs the Secret and context, gets a 422 on the Host, and rolls back. REPRODUCED on the UI side.
- **L17**: `llmChainRequiresSecret('openai-compatible')` is true, but control-api doesn't require one for a key-less chain. The wizard creates an empty labeled Secret, and the edit page blocks Save for a standalone local provider with no `secretRef`. REPRODUCED.
- **L18**: after the H4 fix, the credential chip says "credentials present" when no key is set. `getProvidersWithCompleteCredentials` never recognizes `openai-compatible`, so a Secret holding only its key is labeled "No recognized provider credentials". REPRODUCED.
- **L19**: picking the provider that is already selected wipes the typed `baseURL` (`LlmProviderConfig/index.tsx:263-277`). REPRODUCED.
- **L20**: `LanBaseUrlField` doesn't set `aria-invalid` or `aria-describedby`, so screen readers never announce the error. REPRODUCED.
- **L21**: the client doesn't check the fallback cap when a Host loads with more than 8 fallbacks (only the server rejects it). `validateLlmPolicy` is skipped when the catalog is empty (`page.tsx:860-935`). REPRODUCED and READ.
- **L22** (prior L8): fallback `baseURL` is never trimmed in the UI, so a padded value loaded from the server goes back unchanged. `hostLlmBinding.ts:38` compares raw strings, so a whitespace-only change triggers a rebind and `clearCooldown()`. REPRODUCED.

**Tests**
- **L23**: these negative tests still pass with their feature disabled:
  - HCC T10 (`openaiEgressBrokerReconciler.test.ts:454-483`) stays green when `reconcileForHost` returns immediately.
  - mcp-host `buildFallbackProvider.test.ts:21-35, 77-89` stays green when the builder returns `null` for everything.

  REPRODUCED. Fix: add a positive assertion to each, such as a call count, a read of the store, or constructing the same entry with `slotIndex: 0`.
- **L24**: control-api has no admission test for `port_denied`; with the classifier's port deny disabled, 72/72 still pass. The "accepts exactly 8" and "accepts owned credentialSlot" tests assert only `res === null` and pass with `validateLlmPolicy` disabled. REPRODUCED.
- **L25** (prior L13, rest still open): no test for pruning with a fallback update, fallback reordering, or a failed rollback. `test-lint-networkpolicies.sh` doesn't render the `llm-egress` default-deny. READ.

**Docs, comments, maintainability**
- **L26** (prior L16, open): the PR body still says:
  - the cluster-CIDR patch is "opt-in"; it is applied by default, and base provisions nothing
  - "80 files"; the PR now has 128
  - "per-endpoint broker"; it is one broker per slot
  - the rollout order is CRD → control-api → control-ui

  `docs/deploy/llm-providers.md:204,222,231` and `docs/llm-providers/README.md:132` repeat the last two. The docs also leave out the port denylist, the fallback cap of 8, the ownership rule and the node-guard drop reasons. The dependency order the code actually needs is: CRD → `llm-egress` (namespace, RBAC, CIDR patch) → HCC → mcp-host → control-api → control-ui.
- **L27**: comments that no longer match the code:
  - `deploy/base/llm-egress/networkpolicies.yaml:7` mentions "scoped CoreDNS egress", which HCC never emits.
  - The CIDR template (`:45-50`) says "no over-broad rejection", but HCC now rejects those inputs.
  - `index.cjs:50` says kube-proxy uses "10250-10259"; its metrics port is 10249.
  - The lazy-`require('crypto')` comment is wrong (see M9).
  - `hostSpecValidation.ts:39-43,528` and `host.yaml:256` still say "format-checked only".
- **L28** (prior L14, partial): leftover dead or duplicated code:
  - The `reserved` branch in `classifyLanBaseURL` is unreachable.
  - `seenSlotIds` (`slotAdmission.ts:63,72`) never triggers.
  - Several exports are used only by tests.
  - The key slot constant is duplicated (`slotAdmission.ts:21`).
  - `LAN_BASE_URL_REASON_MESSAGE` is copied between control-api and control-ui, and the two copies already differ.
  - `buildNginxConfigMap` is still inside the 1027-line reconciler.
- **L29** (prior L9, open): 4 new `console.*` calls in `host-context-controller/src/k8sClient.ts` (2930, 2933, 4599, 4978), plus a `console.error` in `mcp-host/src/llm/index.ts:146`. The repo logging standard asks for the existing structured loggers.
- **L30**: prior findings that haven't changed:
  - L3: the broker ingress `podSelector` checks only `clerum.io/host` (`:567`).
  - L4: key rotation waits for the 300 s resync.
  - L7: an older control-ui can rewrite the Host during rollout.
  - L10: `llm-egress` enforces PSA `baseline` although the pod meets `restricted`.

---

## Verified clean
- **Slot identity and broker naming:** HCC, mcp-host and `egress-policy` agree on all 3 contract vectors. Both sides use the raw array index for slots. mcp-host uses only `new URL(baseURL).pathname`, and the pathname matches HCC's path across edge cases (trailing `/`, `..`, query, fragment, whitespace, `%20`).
- **`buildFallbackProvider` extraction and dev merges:** `git diff 1c4a7f73d..HEAD -- mcp-host/src/main.ts` only moves code. The logic is identical, nothing was lost or duplicated, and `main.ts:501,717` call the new seam.
- **Classifier:**
  - IPv4-mapped IPv6 and bracketed IPv6 → `not_ip`.
  - `192.168.1.10@evil.com` → `not_ip`.
  - `169.254.169.254` → `link_local`.
  - Port 65536 is rejected.
  - Trailing dot, `%2e`, backslashes, tabs and uppercase scheme all canonicalize, and denied ports still match after canonicalization.
- **Malformed CIDR lists** (`10.96.0.0`, `…/12;…`, `fd00::/108`, `10.96.0.5/12`) throw at startup.
- **Kubectl bypass:**
  - The provider enum stops a padded `"openai-compatible "`.
  - CRD `maxItems: 8` binds direct writes.
  - HCC re-checks ownership after `trim()` and drops a foreign slot before reading the Secret (T15).
- **Config revision:** the hash covers only the placeholder render, never the key. It is deterministic. The ConfigMap is written before the Deployment, and the desired annotation wins the merge.
- **Detector:** a missing range aborts it, and values are spliced literally (no `&` expansion).
- **NetworkPolicies:**
  - `llm-egress` has a default-deny for ingress and egress; brokers get no DNS egress.
  - `-src` policies are classified "unrelated" by `networkPolicyReconciler`, so no policy has two owners.
  - Name collision would need a 64-bit second preimage.
- **Broker pod and RBAC:** the pod runs non-root, drops ALL capabilities, disallows privilege escalation, uses RuntimeDefault seccomp and mounts no SA token. The Role grants only the verbs used, with no new cluster-wide grants and no EndpointSlice reads.
- **No key or `baseURL`** appears in HCC logs, status messages or the 409 bodies. (The mcp-host startup log is the exception: M6.)
- **Other providers stay unusable:** the H4 fix makes no other provider "usable" by mistake; all 24 `PROVIDER_IDS` were checked.
- **UI rendering:** the broker condition renders as text, so an XSS payload produces 0 elements. Removing fallback rows keeps `baseURL` with the correct row.
- **Docker contract test:** it fails when the egress-policy COPY is removed.

## Local test run (Node 24.18, `npm ci` without `--ignore-scripts`, real exit codes)

| Suite | EXIT | Passed / failed / skipped |
|---|---|---|
| packages/egress-policy | 0 | 16 / 0 / 0 |
| control-api (full, with workflow-runtime-core built as CI does) | 0 | 5127 / 0 / 158 (155 are `realPostgres`, plus 3 network integration tests) |
| host-context-controller (full) | 0 | 2434 / 0 / 0 |
| mcp-host (full) | 0 | 3786 / 0 / 0 (1 todo) |
| control-ui (full) | 0 | 1972 / 0 / 0 |
| Changed shell tests (8 scripts, all hermetic) | 0 | 114 checks |

- **Typecheck:** HCC and mcp-host are clean.
  - control-api is clean once `workflow-runtime-core` is built.
  - control-ui has 3 errors, all in `packages/frontend-components/src/index.tsx`. `dev` has the same 3, and this PR doesn't touch that file.
- **Build:** `next build` compiles; its type step fails on the same 3 pre-existing errors.
- **Vacuity mutations:**
  - Every test named in prior M7 now fails when its feature is disabled.
  - The new exceptions are listed in L23 and L24.
  - The three coverage gaps are M7, M8 and L24.
- **Not run:** a real `docker build`, `nginx -t`, and a server-side CEL dry-run.

## Open questions for the maintainers
1. **Host-admin threat model:** still not stated.
   - If Host admins can be untrusted tenants, fix M1 and M4 before this provider is offered to them.
   - If Host admins are trusted operators, document M1, M4 and `proxy_ssl_verify off` as accepted residuals.
2. **Ownership rule scope (M3):** should the rule apply only to `openai-compatible`, like the CRD and HCC, or to every provider? If every provider, it needs a migration path for stored Hosts.
3. **Broker capacity:** the per-Host cap of 8 is in place. Add a ResourceQuota on `llm-egress` too?
4. **Rollout order:** settle the official order (L26) and update the docs and PR body before deploying.

zach88:
Re-review follow-up (head `1f17b67ed` → `6c5f30a9b`). Here's where each item that gated merge landed.

## Fixed before merge

- **M3** (ownership rule breaks existing Hosts) — fixed. Scoped the `credentialSlot` owner-gate in control-api back to `openai-compatible` only, matching the CRD CEL and HCC (`1edbccd72`), and tightened the control-ui dropdown filter to `<canonical>` / `<canonical>-<suffix>` (`452fbf340`). For every other provider control-api is back to dev behavior (format-only), so a stored `{provider: claude, credentialSlot: …}` Host writes cleanly again — including the Codex bind path. This does **not** reopen H2: HCC only builds brokers for `openai-compatible` (`slotAdmission.ts`), its runtime owner-gate is untouched, and `baseURL` stays rejected for other providers, so a foreign key can only ever reach the vendor's official endpoint — never an admin-chosen LAN IP.
- **M2** (orphan sweep can delete a broker a concurrent reconcile just created) — fixed in `6c5f30a9b`. The sweep now serializes each orphan gc under the owner Host key (`host:<owner>`, read from the `clerum.io/host` label) and re-checks the live Host cache **inside** the serialized section before deleting — the same pattern `llmHookReconciler.sweepOrphanedWorkloads` already uses. Objects without an owner label fall back to `gc:<broker>` as before. Comment at `:989` corrected.
- **L3** (an IPv6 `K8S_API_CIDRS` entry drops every broker on dual-stack) — fixed in `e8eb604d0`. `resolveClusterInternalCidrs` now filters the k8s-API CIDR list to IPv4 before it reaches the IPv4-only classifier (warn-once per non-IPv4 entry). `parseK8sApiCidrs` is untouched, so IPv6 still feeds the `allow-k8s-api-egress` NetworkPolicies.

Each fix ships with a regression test that fails against `1f17b67ed` on the assertion (not setup). Verified by an independent pass — plan-conformance, collateral ("what did the fix break?" → clean), code review (0 blockers) and security (confirmed H2 stays closed). Full HCC suite 2437/2437; control-api and control-ui targeted suites green.

## Accepted residuals — answering open question #1 (threat model)

Host admins here are **trusted operators, not untrusted tenants**. This provider is only meaningful for self-hosted, single-org deployments where the operator runs the cluster, the LAN and the local model. A SaaS offering is structurally impossible — the cluster can't reach a tenant's LAN. So the Host-admin role isn't a trust boundary, and we're accepting these as documented residuals:

- **M1** (a Host mirrors another Host's key to its LAN IP): whoever can edit Hosts already has cluster access (`kubectl get secret`), and there's no tenant boundary to cross — the broker adds no new surface.
- **M4** (broker reaches node IP / control-plane ports): the operator already reaches those internals directly; the broker doesn't widen their real privilege.
- **`proxy_ssl_verify off`**: the operator owns the LAN, so on-LAN MITM is their own risk to accept.

If this is ever exposed to mutually-distrusting tenants (not possible today), we'd reopen M1 (owner-label the referenced Secret) and M4 (watch Nodes + finish the port denylist).

## Other open questions

- **#2 (ownership rule scope):** scoped to `openai-compatible` only — no migration path needed, since for other providers it's a revert to dev behavior.
- **#3 (ResourceQuota on `llm-egress`):** the per-Host cap of 8 stays; a namespace ResourceQuota is a follow-up, not a merge blocker.
- **#4 (rollout order):** CRD → control-api → control-ui. The CRD gets applied to the dev cluster before merge (the structural `host.yaml` changes are from the H2/M3/M4 round; the R5 edit to it is description-only).

Everything else (M5–M9 and the L-series) is tracked as post-merge follow-up.

alfredolopez80:
## Re-review of #605 at `73f8b58df` (follow-up to the 2026-09-14 review at `1f17b67ed`)

@zach88 this pass checks the 4 fix commits (`1edbccd72`, `452fbf340`, `e8eb604d0`, `6c5f30a9b`), the dev merge, and the factual claims behind the accepted residuals. All work was read-only: five parallel reviewers, each fix mutation-tested in a separate worktree. Every item is marked **REPRODUCED** (executed) or **READ** (traced in code).

### TL;DR
- **M2 (orphan sweep race): FIXED.** REPRODUCED against the real reconciler.
- **M3 (credentialSlot ownership broke stored Hosts): FIXED** in both control-api and control-ui. REPRODUCED through the real PUT and Codex-bind routes.
- **H2 is not reopened.** No provider other than `openai-compatible` lets a Host admin choose where a fallback key is sent.
- **L3 (IPv6 in `K8S_API_CIDRS`): FIXED**, but the new filter fails open on zero-padded IPv4 prefixes (N1).
- **Merge with dev: clean.** No PR hunk was lost or duplicated, and no package version was downgraded. CI is 39/39 green.
- **Two author statements do not hold, and both need a change before merge:**
  1. The residual rationale "whoever can edit Hosts already has `kubectl get secret`" is false for the product's Host-edit path, and the residuals are not documented anywhere.
  2. The rollout order CRD → control-api → control-ui is incomplete (same point as L26).

---

### 1. Fix validation

| Prior finding | Verdict | Evidence |
|---|---|---|
| **M2** orphan sweep race | **FIXED** | Independent tests on the real reconciler with an in-memory apiserver: (R1) Host added while the sweep LIST is in flight, all 7 broker objects survive and a real orphan is still deleted as the liveness witness; (R2) Host re-created mid-`gcBroker`, broker ends complete. No deadlock: `gcBroker` never re-enters `runSerialized` and `reconcileForHost` never calls the sweep. A Host deleted during the sweep gets exactly one delete per object (404 tolerated). `isBrokerDesiredNow` uses the same `buildDesiredBrokers` as reconcile. REPRODUCED |
| **M3** control-api | **FIXED** | PUT with a byte-identical stored spec returns 200 for fallbacks `{claude, backup-claude-key}`, `{openai, anthropic-api-key}` and `{azure, claude-api-key}`. Codex bind on stored Hosts returns 200. `openai-compatible` with a foreign slot still returns 422 on create, stored PUT and bind, including padded provider, uppercase slot, `…-api-key2` and `…-api-key.claude-api-key`. Comments at `hostSpecValidation.ts:39-44`, `:530` and `host.yaml:256` now match the code. REPRODUCED |
| **M3** `host.yaml` "description-only" | **CONFIRMED** | Schema before and after, with `description` keys stripped, is identical. No CEL rule, enum or `maxItems` changed. REPRODUCED |
| **M3** control-ui dropdown | **FIXED** | Offers `<canonical>` and `<canonical>-<suffix>`. No collision among the 24 canonical `dataKey`s. A stored slot that is no longer offered is still preserved on an unrelated save. REPRODUCED |
| **H2 not reopened** | **CONFIRMED** | For a fallback key, the destination is set at deploy level for every provider. `openai` and `claude` use SDK env (`OPENAI_BASE_URL`, `ANTHROPIC_BASE_URL`). Azure uses `AZURE_OPENAI_ENDPOINT`. mcp-host's `process.env` comes only from the `mcp-host-config` ConfigMap: per-Host env goes to ConfigStore, nothing assigns to `process.env`, and control-api never writes that ConfigMap. The other 17 providers use a fixed `baseURL` from `registryCore.ts`. vertex, bedrock and codex reject `credentialSlot`. `baseURL` on other providers is rejected by control-api, by CEL, and ignored by mcp-host. READ |
| **L3** IPv6 `K8S_API_CIDRS` | **FIXED, with N1** | `fd00::1/128` plus an IPv4 entry: brokers provision and the IPv4 entry still denies. Warn-once: 100 resolves emit 2 warnings, payload `{cidr}` only. `networkPolicyReconciler.ts:771` still receives the IPv6 entries. Dropping IPv6 is not fail-open in itself, because the classifier admits only IPv4 literals. REPRODUCED |

**Mutation checks on the new tests.** Each production change was reverted with its tests kept:
- `hostSpecValidation.ts` reverted: the new R5-M2 test fails on its assertion.
- The gate disabled entirely: the R4-H2 guard test fails.
- `llm.ts` reverted: both dropdown tests fail.
- `clusterDenySet.ts` reverted: G9 and the resolver test fail.
- M2 fully reverted: `sweepRace` S1 fails. It **still passes when only the serialization is removed** (N3).

### 2. Accepted residuals (M1, M4, `proxy_ssl_verify off`): fact check

We are not re-arguing the decision. These are the factual premises it rests on:

| Claim | Verdict | Evidence |
|---|---|---|
| "Whoever can edit Hosts already has cluster access (`kubectl get secret`)" | **FALSE** for the product path | Every Host write goes through control-api `/admin/*`: `routes/admin/resources.ts:588,843`, `codexSubscription.ts:306` and `personalization.ts:158`. Those routes are guarded by a Control UI session cookie checked against Postgres (`middleware/controlUIAuth.ts:14-31`, `services/adminSessionAuth.ts:12-37`). The single app role `admin` has no Kubernetes identity, and admins can invite other admins (`controlAdmins.ts:248`). control-api deliberately never returns Secret values: `secretService.ts:5-8,21-27` enforces summary shapes at compile time, and `hostSecrets.ts:28-38` returns key names only. `hostSecrets.ts:59-65` already records cross-Host `secretRef` as a residual, which was harmless while the key could only go to vendor endpoints. The Kubernetes principal with `get secrets` is the control-api ServiceAccount, not the human admin. READ |
| "A SaaS offering is structurally impossible" | **FALSE as written** | `docs/concepts/open-core-and-hosted.md:5-6,37-43` describes "evenfire hosted", a managed multi-tenant service. `deploy/base/kustomization.yaml:19-21` references GKE overlays, and this PR's own guide tells GKE operators how to enable the feature (`docs/deploy/llm-providers.md:267-271`). On GKE the broker's "LAN" is the VPC: other workloads, private-IP managed databases, peered networks. READ |
| M1 / M4 / TLS are "documented residuals" | **FALSE** | `docs/deploy/llm-providers.md` §8, `docs/llm-providers/README.md:114-140` and the CRD `baseURL` description say nothing about the cross-Host key mirror, unverified broker TLS, the node-IP startup snapshot, or the port gaps. §8 (`:239-241`) says the `/32` NetworkPolicy constrains the broker "to exactly that destination". The only mention of `ssl_verify off` is a code comment (`openaiEgressBrokerReconciler.ts:411-421`), and it still claims an attacker must control that exact LAN host; ARP spoofing on the segment is enough. READ |

**Resulting risk, stated plainly.** Any Control UI admin, including an invited one without kubectl, can exfiltrate any other Host's `openai-compatible` API key:
1. Point `spec.secretRef` at that Host's Secret.
2. Set `baseURL` to a machine the admin controls.

If the team accepts this, it should be written down as the trust model, not justified by kubectl access. Suggested wording and placement:
- A "Trust model and accepted residuals" subsection in `docs/deploy/llm-providers.md` §8, a pointer in the CRD `baseURL` description, and an entry in `SECURITY.md`. It should state: *adding a Control UI admin is equivalent to granting read access to every `openai-compatible` LLM key in `mcp-host`.*
- List the other residuals:
  - broker TLS to the LAN endpoint is unverified;
  - node IPs are snapshotted at HCC startup;
  - only ports 10250, 6443, 2379-2380 and 8443 are denied on node IPs (NodePorts, 10249, 9100 and 2381 stay reachable).
- For hosted/GKE, either state that the feature must stay disabled until M1 (Secret owner label) and M4 (watch Nodes, complete the port denylist) are fixed, or gate it behind a deploy flag.
- Correct the `ssl_verify` comment.

### 3. Rollout order: the author's order is incomplete (**Medium, READ**)

CRD → control-api → control-ui covers only the write path. The dependency order in the code is:

**CRD → `llm-egress` (namespace, RBAC, valid CIDR env) → HCC → mcp-host → control-api → control-ui**

HCC and mcp-host must both be live before control-api admits `openai-compatible` Hosts. What goes wrong otherwise:
- **Old HCC:** no broker is created and no `oaiEgress` condition is written, so the UI shows a healthy Host.
- **Old mcp-host:** `createLLMProvider` logs "[LLM] Unknown provider" and returns null, so chats fail with an unclear error.
- **New mcp-host without new HCC:** it dials a `<broker>.llm-egress.svc` that doesn't exist.
- **HCC before its RBAC:** every broker list gets a 403, which is logged and treated as empty. Each resync logs 3 errors per Host.
- **HCC with a malformed CIDR env:** `parseClusterCidrList` throws at module load (`k8sApiCidrs.ts:132-136`), and HCC crash-loops for the **whole fleet**.

`docs/deploy/llm-providers.md:231` and PR body line 39 still give the old order.

Related (**Low, READ**): the "silent prune" risk described in the docs (`:231-235`) and PR body line 21 cannot happen. The old CRD already has a `provider` enum (dev `host.yaml:63,191`), so a write with `openai-compatible` gets a loud 422. The `host_crd_outdated` guard is harmless but unreachable.

### 4. New findings

| ID | Sev | Location | Defect / scenario | Status | Fix |
|---|---|---|---|---|---|
| N1 | Low | `host-context-controller/src/oaiEgress/clusterDenySet.ts:61-72` | **Fail-open introduced by the L3 fix.** The filter uses `parseCidr` (at most a 2-digit prefix) to decide "not IPv4", but startup validation (`k8sApiCidrs.ts:23-27`) accepts zero-padded prefixes. With `CONTEXT_MAPPER_K8S_API_CIDRS=172.16.0.2/032`, the entry is dropped with a misleading "not an IPv4 CIDR" warning, and `https://172.16.0.2/v1` is admitted. Before the fix this config failed closed with `cluster_cidr_invalid`; a canonical `/32` still denies. Found independently by two reviewers. | REPRODUCED | Filter by address family (`isIP(addr) === 6`), or reject non-canonical prefixes in `parseCidrEntry`. |
| N2 | Low | `host-context-controller/src/openaiEgressBrokerReconciler.ts:1051-1056,1072-1076` | The in-lock re-check uses the Host named in the `clerum.io/host` label, not every Host that could own the broker name. If the Deployment's label is edited to point at Host A, which exists but wants no broker, while the real owner B is created between the LIST and the gc, all 7 of B's objects are deleted without waiting on B's queue. This needs a manually edited label, and B's next reconcile recreates the objects. | REPRODUCED | Before deleting, check whether any cached Host desires `brokerName`. |
| N3 | Low | `host-context-controller/src/openaiEgressBrokerReconciler.sweepRace.test.ts:158` | S1 covers only the cache re-check. With the lock key mutated back to `gc:${brokerName}` and the re-check kept, S1 still passes. A test where the gc deletes objects that a concurrent reconcile just created fails under that mutation (all 7 objects missing). | REPRODUCED | Add a serialization test, e.g. a Host re-created during `gcBroker`. |
| N4 | Low | `control-ui/components/LlmPolicyEditor/index.tsx:369,404` | A stored slot that is no longer offered is labelled "(not found in Secret)" even when the key exists in the Secret. With `secretKeys=['claude-api-key','openai-api-key','claude-api-key2']`, the options show `claude-api-key2 (not found in Secret)`. The value is still preserved on save. | REPRODUCED | Compute the label from `secretKeys`, not `slotOptions`, or use "(not a provider-owned key)". |
| N5 | Low | `control-ui/lib/llm.ts:823-828` | The comment says control-api returns 422 for a bare-prefix key. After `1edbccd72` that is only true for `openai-compatible` fallbacks and the prompt-bridge path. | READ | Reword the comment. |
| N6 | Low | PR body lines 15, 40, 42; `docs/deploy/llm-providers.md:204,222`; `docs/llm-providers/README.md:132` | Stale text. The CIDR patch is called "opt-in", but it is rendered by default and HCC provisions no broker without it; the body's claim that a deploy without it runs the RFC1918 + /32 tier is wrong. The body says "80 files" (the PR has 131). "per-endpoint" / `(Host, endpoint)` should be `(Host, slot)`. | READ | Edit the body and the docs. |
| N7 | Info | `docs/deploy/llm-providers.md:27-39` (pre-existing) | Tells operators to put `AZURE_OPENAI_ENDPOINT` in Host → Environment, but runtime reads `process.env`, which per-Host env never reaches. Azure fails closed with "AZURE_OPENAI_ENDPOINT is required". | READ | Point to the `mcp-host-config` ConfigMap. |
| N8 | Info | `deploy/images.json` (pre-existing) | `workflow-recipes`, `workflow-coordinator`, `workflow-snippet-runner` and `clerum-workflow-base` depend on `packages/llm-providers` but don't list `packages/llm-providers/**` in `source_paths`. This PR changes `index.cjs`, and those images won't be marked for rebuild. | READ | Add the path. |

**Still open from the prior review, deferred by the author:** M1 and M4 (see §2), M5–M9, and L1–L30 except L3. Please track them in a follow-up issue so they are not lost after merge.

### 5. Merge integrity (`73f8b58df`)
- **Attribution:** every PR-effective change since `1f17b67ed` comes from one of the 4 fix commits.
- **Conflicts:** all 8 merge conflicts were per-commit `package.json` / `package-lock.json` version counters (`bump-staged-package-versions.mjs`). No package is downgraded.
- **Shared files:** for the 11 non-package files both the PR and dev touched (`ci-public.yml`, `Makefile`, HostWizard, `images.json`, `mcp-host-config.yaml`, mcp-host `config.ts` and `main.ts`, `test-minikube-pre-gate-shadow.sh`), a hash of the PR hunks is identical before and after the merge.
- **CI at head:** 39 pass, 0 pending, 0 fail.

### 6. Local verification (Node 24.18.0, real exit codes, CI setup replicated)

| Suite | EXIT | Passed / failed / skipped | Δ vs `1f17b67ed` |
|---|---|---|---|
| packages/egress-policy | 0 | 16 / 0 / 0 | 0 |
| control-api | 0 | 5134 / 0 / 158 (real-PG and GKE-gated) | +7 |
| host-context-controller | 0 | 2464 / 0 / 0 | +30 |
| mcp-host | 0 | 4096 / 0 / 0 (1 todo) | +310 |
| control-ui | 0 | 1994 / 0 / 0 | +22 |

- **Test count drops:** none caused by the PR. The 13 cases the dev merge removed were also removed on dev.
- **`tsc --noEmit`:** 0 errors in control-api, HCC, mcp-host and control-ui.
- **Shell tests:** 6 of 8 scripts pass (103 checks). `test-lint-networkpolicies.sh` and `test-minikube-gfs-upload-v2-profile.sh` were not run locally because a local kubectl guard blocked them; they are covered by the green CI run.

### Suggested before merge
1. Replace the kubectl rationale with the real trust model, and document the residuals and the hosted/GKE stance (§2).
2. Fix the rollout order in the docs and the PR body (§3).
3. N1: filter by address family.
4. Optional, cheap: N3 serialization test, N2 cache-wide re-check, N4–N6 text fixes.

PR #608, opened 2026-09-10
fix(web-search): prevent SSRF and bound fetch_page resources

### LINKED ISSUE #198: security(web-search): restrict fetch_page egress and bound response bodies
## Priority
P1

## Source
Private issues #669 and #674, merged here because they describe one request path. Revalidated against private and public origin/dev.

## Problem
The fetch_page tool accepts a caller-controlled URL, follows redirects, and reads the whole upstream response before truncating returned content. It has no destination policy for loopback, link-local, metadata, private, or cluster addresses, and no byte limit on the upstream response.

## Impact
The tool can become an SSRF primitive in deployments where its network path reaches sensitive services. Independently, a large or streaming upstream response can exhaust memory before maxChars is applied.

## Acceptance criteria
- Apply an explicit destination policy before every connection and after redirects, rejecting private, link-local, loopback, metadata, and cluster destinations.
- Bound response bytes while reading; maxChars must not be the only limit.
- Preserve the request timeout and return clear errors for rejected destinations and oversized responses.
- Add tests for redirect-to-private-IP, IPv4/IPv6 private ranges, metadata addresses, and oversized/chunked responses.

## Additional validation and acceptance criteria (2026-09-10)

Source inspection confirms that the 15-second timer is cleared after response headers, before body consumption. `maxChars` limits returned text only after full buffering. Both destination authorization and resource limits must be fixed.

Network policy can reduce SSRF reachability, but does not replace application checks or stop a permitted public server from sending slow/oversized bodies. Tests must use this first-party connector, not a different upstream connector sharing its name. Production exposure was not verified.

- Validate normalized HTTP(S) URLs and all resolved IPv4/IPv6 addresses; reject non-public/special-purpose destinations and URL userinfo.
- Bind the actual connection to the validated IP while preserving Host, SNI and TLS hostname verification. No second unvalidated lookup, ambient proxy or unsafe fallback.
- Validate every redirect before connecting. Bound hops and loops; use one absolute deadline and cumulative byte budgets across DNS, connection, redirects and body consumption.
- Limit transferred and decoded bytes while streaming, including compression, chunked bodies, misleading Content-Length and HTTP errors. Cancel and release resources immediately on rejection.
- Bound concurrent calls without an unlimited queue. Release timers, DNS work, sockets, listeners and capacity on all exit paths. Verify successful calls still work after errors.
- Preserve MCP success and maxChars contracts. Return stable sanitized errors without exposing upstream or operational data.
- Add unit tests for address normalization/classification, mixed DNS responses, connection pinning, redirect validation, absolute deadlines, byte/decompression limits, cleanup and concurrency.
- Add real HTTP/TLS and MCP integration tests; fake transports alone cannot prove connection binding or cancellation.
- Add direct connector E2E through real MCP initialization, discovery and fetch_page invocation. Assert protocol state, actual upstream I/O and tool output. Desktop is an optional indirect integration check, not a closure gate for this connector-only fix.
- Cover blocked destination/redirect, slow/oversized body, recovery and access-denied cases with isolated local fixtures. Do not weaken production validation to permit fixtures.
- Apply e2e-test-guardian responsibility separation: direct MCP/image E2E executes the real connector and network path; mocks are confined to unit collaborators and controlled external upstreams. Browser-specific rules and the static Playwright audit apply only if browser specs change.
- Wire connector tests into local and CI entry points. Verify the image corresponds to the reviewed commit. Missing fixtures, skipped required tests and zero executed tests must not count as success.
- Revalidate both paths with Codex Security after implementation. Record unit, integration, image and MCP E2E results separately. A static scan or design review does not close this issue.

Implementation and direct connector validation are available in [PR #608](https://github.com/evenfire-ai/evenfire/pull/608). This issue remains open pending merge. No production verification is claimed.

alfredolopez80:
## Problem and change

`fetch_page` accepted arbitrary destinations, followed redirects without revalidation, and buffered the upstream body after clearing its timeout at headers. This change confines requests to validated public HTTP(S) destinations, pins DNS while preserving TLS hostname verification, and validates every redirect.

The operation now has an absolute 15-second deadline, independent 1 MiB transferred/decoded limits, bounded redirects and four active calls per process. HTTP disconnects, cancellation and unsupported protocol switches release upstream work. Errors are sanitized; ordinary MCP success output is preserved. The connector is added to the unit/CI matrix, including isolated tests against its built image.

## Validation

On `21bcb43e`, including `origin/dev` at `4d631144`:

- 95 unit tests passed; none skipped, including Unicode/title-prefix, argument-validation and hermetic runner-cancellation regressions.
- 10 real DNS/TLS/HTTP/MCP E2E tests passed against the built image, without mounting application code or dependencies over it.
- Regressions failed before the HTTP protocol-switch/disconnect fixes and passed after them.
- Temporary mutation checks detected removal of destination validation, DNS pinning, transferred/decoded byte limits and deadline.
- Additional bounded workload: 96 dense 1 MiB page requests in batches of four completed inside a 256 MiB isolated container; the connector peak stabilized at approximately 164 MiB for this workload.
- Formatting, diff hygiene and the repository public-boundary check passed.
- Internal parallel review covered runtime/security, contracts, tests/CI and architecture. It identified and corrected Unicode title offsets, bounded subprocess termination/inspection and the unused import reported by CodeQL. The protocol and title regressions were demonstrated before and after correction.

The tests use synthetic isolated endpoints. No production exposure or full-platform security certification is claimed. Desktop is an indirect integration path and is outside this connector-only PR.

Claude’s exhaustive re-review reported no blocking/high/medium findings. Its two low findings were validated and corrected: continue scanning past invalid title prefixes and use a precise safe error for invalid `maxChars`. Independent adversarial review of the final correction found no remaining actionable defects. Pre-existing follow-ups: #612 (CI action runtime migration) and #613 (HTML-context-aware title extraction).

Closes #198


alfredolopez80:
@claude Please perform a complete, exhaustive re-review and audit of this PR at HEAD `8691c62bfd555a099098f474d1c60a28dcb94d49` against `dev`.

Review the entire diff independently: runtime correctness and edge cases, destination/DNS/TLS/redirect controls, resource limits and cancellation, compatibility, maintainability, tests and false-positive risks, CI/CD, documentation, and GitHub security/CodeQL annotations. Include all remaining actionable findings, including non-blocking and Low items; do not suppress a finding just because CI passes.

For each finding give severity, exact file/line, actual versus expected behavior, evidence or a reproducible case, whether it is introduced by this PR or pre-existing, and the smallest appropriate remedy. Distinguish confirmed defects from assumptions and optional improvements. Keep the scope grounded in this connector PR.

Review only: do not modify code, create issues, or merge. State the reviewed SHA and a clear final verdict, including any remaining gaps if no actionable findings remain.

alfredolopez80:
Review follow-up completed at `21bcb43e`, including current `origin/dev` (`4d631144`).

- Fixed the confirmed invalid-title-prefix regression with linear scanning and the misleading helper error for invalid `maxChars`. Added regressions; 95 unit tests pass with no skips.
- 10 real DNS/TLS/HTTP/MCP tests pass against the image built from this exact commit.
- Independent adversarial review of the final correction found no remaining actionable defects.
- CI [34522472833](https://github.com/evenfire-ai/evenfire/actions/runs/34522472833) and CodeQL [34522473086](https://github.com/evenfire-ai/evenfire/actions/runs/34522473086) completed successfully. CodeQL reports zero results on the current PR merge analysis; alert 1364 is fixed.

The comment-context title limitation predates this PR and is tracked in #613. Partial comment skipping can regress valid quoted attributes, so contextual parsing needs separate coverage. Existing action-runtime deprecation is tracked in #612. Global concurrency, single validated-IP selection and per-fetch logging were informational observations, with no demonstrated regression requiring a change.

One attribution correction to the Claude review: `24c96d47` is an ancestor of `4d631144`; the latter is newer, and its Makefile already includes the GFS package. The merge preserves that job and the connector job.

The PR remains open and requires repository review approval; no merge was performed.

alfredolopez80:
@claude Review follow-up completed at `21bcb43e`, including current `origin/dev` (`4d631144`).

- Fixed the confirmed invalid-title-prefix regression with linear scanning and the misleading helper error for invalid `maxChars`. Added regressions; 95 unit tests pass with no skips.
- 10 real DNS/TLS/HTTP/MCP tests pass against the image built from this exact commit.
- Independent adversarial review of the final correction found no remaining actionable defects.
- CI [34522472833](https://github.com/evenfire-ai/evenfire/actions/runs/34522472833) and CodeQL [34522473086](https://github.com/evenfire-ai/evenfire/actions/runs/34522473086) completed successfully. CodeQL reports zero results on the current PR merge analysis; alert 1364 is fixed.

The comment-context title limitation predates this PR and is tracked in #613. Partial comment skipping can regress valid quoted attributes, so contextual parsing needs separate coverage. Existing action-runtime deprecation is tracked in #612. Global concurrency, single validated-IP selection and per-fetch logging were informational observations, with no demonstrated regression requiring a change.

One attribution correction to the Claude review: `24c96d47` is an ancestor of `4d631144`; the latter is newer, and its Makefile already includes the GFS package. The merge preserves that job and the connector job.

The PR remains open and requires repository review approval; no merge was performed.

jozer-rami (commented):
## Evenfire Reviewer — full review of #608

Reviewed head `2e053375c4f4263f9822bcd749e1038181139560` against base `dev` @ `7c7e34c47bdb5d65b7823cbbbe63ec64c23f2979`.
Scope is the connector diff (16 files, +1487/−216). I re-read the fetch, destination, and HTTP wiring at this SHA and executed the Node 24.16.0 behaviors the security paths depend on (URL normalization, header parsing, request event order, DNS timeout units, block-list boundaries). I did not re-run the Docker image suite.

**Verdict: do not merge yet.** The SSRF design is real and most of the controls hold, but one runtime bug can pin a `fetch_page` slot until process restart, and the DNS timeout does not mean what the code and the README say. Both are introduced by this PR.

---

### What holds

These are the controls I re-checked and consider sound:

- **Literal and disguised private destinations are blocked before connect.** WHATWG `URL` folds decimal, hex, and octal forms (`http://2130706433`, `http://0x7f000001`, `http://0177.0.0.1`, `http://0xA000001`) to canonical IPs, and `isPublicAddress` then rejects them. IPv4-mapped IPv6 (`::ffff:7f00:1`) is outside `2000::/3`, so it is rejected. Zone identifiers fail URL parsing. Userinfo, non-HTTP schemes, NULs, spaces, and port `0` are rejected.
- **The IANA special-purpose samples I checked do not leak.** No gap on the usual private, link-local, CGNAT, documentation, multicast, and reserved ranges, including the cloud link-local range inside `169.254.0.0/16` and `100.100.100.0/24` inside `100.64.0.0/10`. IPv6 ULA, link-local, multicast, NAT64 well-known prefixes, 6to4 (`2002::/16`), and ISATAP (`…:0:5efe:` / `…:200:5efe:`) are rejected. Ordinary public addresses I sampled stay allowed, including `2001:200::/23` just outside the blocked `2001::/23`.
- **DNS rebinding is closed for the usual case.** `resolve4` and `resolve6` run together; any non-public answer fails the name; the socket is opened to the chosen literal with `agent: false`, so there is no second lookup. `ENODATA` is absence; `ENOTFOUND` / `ESERVFAIL` / timeout are not. I confirmed a local resolver returns `ENODATA` for NOERROR/empty and `ENOTFOUND` for NXDOMAIN, matching the code's assumption. A delayed AAAA of `::1` beside a public A is still observed before return, so the mixed-record check is not skipped.
- **Redirects are re-parsed and re-resolved.** HTTPS→HTTP is rejected. Credentials in `Location` fail `parsePageUrl`. `javascript:` and port `0` fail. `res.destroy()` runs on `continue`, so redirect bodies are not accumulated. Five redirects then a 200 is allowed; a sixth redirect is not. Duplicate `Content-Length` and `Transfer-Encoding`+`Content-Length` are rejected by the Node parser (`HPE_UNEXPECTED_CONTENT_LENGTH`, `HPE_INVALID_CONTENT_LENGTH`) rather than smuggled.
- **Byte limits are independent and the exact 1 MiB ceiling is the one the tests describe.** Declared length, wire bytes, and decoded bytes each fail closed. `Content-Length` values that are not a digit string are rejected. CRLF in `path` or `Host` throws before the request is sent.
- **TLS pinning is the right shape.** The dial target is the validated IP; `servername` and `checkServerIdentity` use the URL host, so the certificate is not checked against the IP. IP literals skip SNI. Downgrade is a separate check.
- **The concurrency counter itself is not racy.** The check and `active++` happen with no `await` between them. `invalid_max_chars` and `busy` return before the increment. `finally` always decrements on the normal and thrown paths.
- **Caller-visible `fetch_page` failures are codes, not upstream text.** The MCP 500 path no longer interpolates the raw exception. `web_search` / `search_news` still return `err.message`; that behavior is pre-existing and unchanged.
- **Cancellation is wired.** The `/mcp` middleware stores a per-request `AbortSignal`, and Node's `AsyncLocalStorage` still sees it after `await` and `setImmediate`. The outer 15s `AbortController` does cancel a hung resolver (`ECANCELLED` at ~400ms in a local probe). Tests cover deadline-during-body, HTTP disconnect, protocol switch, and cross-session `busy`.
- **CI is consistent and green.** `mcp-servers/web-search` is in both `TEST_SERVICES` and the workflow matrix. The image job builds the image and runs `run-network.mjs --image`, mounting only `test/`. All 42 checks on this head are `success`, including `Build / test / typecheck (mcp-servers/web-search)` and CodeQL. The earlier unused-import thread is resolved and outdated.

---

### Findings

#### 1. High — `connect` rejects twice on failure, and the second rejection escapes `fetchPage`

`mcp-servers/web-search/src/fetchPage.ts`, `connect` (the `error`, `upgrade`, and `close` listeners).

`http.ClientRequest` emits `error` and then `close` when the connect fails or the socket is reset, and it emits `upgrade` and then `close` for a 101 with no `response` callback. Both listeners call `reject`. The first reject settles `connect`. The second reject is not handled by the `await`, because the promise is already settled.

I reproduced this on Node 24.16.0 with the same listener shape:

- Upgrade: `upgrade` listener runs, then `close` listener runs, and the awaited promise rejects with the first reason.
- Connection reset: `error` listener runs, the await rejects, then `close` listener runs.

A second `reject` on an already-settled promise does not, by itself, become an `unhandledRejection`. It becomes one as soon as the only consumer has finished, which is exactly what `fetchPage` does. `active++` happens before `await connect(...)`. The `await` throws into `catch`/`finally`, `finally` runs `active--`, and that is an await point (the `finally` of an async function yields). The `close` listener then rejects with nothing left attached. Node's default `unhandledRejection` behavior is to crash the process. Until that crash is fully handled, the in-flight state is already wrong: the slot was released, but a late rejection still escapes the operation the slot was guarding.

The existing network test only asserts that the tool returns `upstream_failure` quickly. It never installs an `unhandledRejection` hook, so it stays green while the second reject is still fired. The unit double is an `EventEmitter` that never emits `close` or `error`, so it cannot catch this either.

**Expected:** one terminal signal. `close` after `error`/`upgrade`/`response` must not reject again.
**Smallest fix:** a `settled` flag (or `once` plus removal of the other listeners) so only the first of `error`, `upgrade`, or a `close` that happened with no response can reject. Add a regression that listens for `unhandledRejection` while driving a real refused connection and a real 101.

#### 2. Medium — DNS `timeout: 15000` is not a 15 second cap

`mcp-servers/web-search/src/fetchPageDestination.ts`, `new Resolver({ timeout: 15000, tries: 1 })`.
README states the operation has a 15-second absolute deadline and the resolver is constructed as if it shared that budget.

Node's documentation says `Resolver` `timeout` is milliseconds. c-ares treats `timeout` as **seconds**, and this Node 24.16.0 build passes the number through. Measured against a blackhole on 24.16.0:

| option | elapsed to `ETIMEOUT` |
|---|---|
| `{ timeout: 1, tries: 1 }` | ~250ms (c-ares floor, not 1ms) |
| `{ timeout: 15000, tries: 1 }` | still running at the outer abort; the option is 15000 seconds |
| default timeout, `tries: 1` | ~2000ms |

So the option does not bound DNS to 15s. The fetch-level `AbortController` does cancel the resolver when the 15s deadline fires, which is why a single call still ends. The hole is everywhere that signal is absent or already raced:

- `resolvePageAddress` starts the lookup and only then attaches the abort listener. A lookup that completes or hangs in that gap is not covered by this call's timer. With a 15s c-ares timeout that gap would be short. With a 15000s c-ares timeout, a lookup that escapes the signal (a second in-flight query after the sibling already settled and the function returned, or a future caller that passes a signal that does not abort) stays in c-ares until that absurd timeout.
- `tries` is 1, so this is not a multi-try amplification today. It is still the wrong unit, and the next person who "tunes" `tries` will multiply a 4 hour timeout.

**Expected:** the resolver timeout is a small number of seconds, strictly inside the 15s operation deadline (2–5 is the right order), and a test fails if a blackhole lookup is still pending after that budget with the abort signal disconnected.
**Smallest fix:** `timeout: 5` (seconds, with a comment that says seconds and points at the c-ares unit, not the Node doc sentence). Do not write `15000` here.

#### 3. Low — `content-length` / `content-encoding` are not validated as a single header value before use

`readBody` in `fetchPage.ts`.

Node joins duplicate `Content-Encoding` into one string (`identity, gzip` in a local probe). The code then compares that string to `gzip` / `deflate` / `br` and correctly fails closed with `unsupported_encoding`. Duplicate `Content-Length` never reaches this function; the parser throws. So this is not an SSRF or a bypass today.

It is still a brittle contract: `Number('01048576')` is accepted because of the leading-zero digit rule, and a future relaxation of the encoding compare (for example treating any string that includes `gzip` as gzip) would decode only part of a stacked encoding. Reject a header that is an array or that contains `,` before the switch, with a stable error, so the fail-closed behavior is local instead of accidental.

#### 4. Low — title and body extraction are still not HTML-aware

`titleText` / `pageText` in `fetchPage.ts`.

Confirmed at this head:

- `<!-- <title>comment</title> --><title>real</title>` returns `comment`.
- `<textarea><title>nope</title></textarea><title>real</title>` returns `nope`.
- `<script>secret token</script><p>visible</p>` returns `secret token visible`.

The prefix-scan fix is present and works (`<titlex>` no longer hides a later `<title>`). The comment/textarea case is the one already tracked in #613 and called out in the PR body. Script/style text is the same behavior as the old `stripHtml`. Not a security control. Not a merge blocker. Do not treat #613 as fixed.

#### 5. Low — CI image step does not use the deadline wrapper the Makefile path uses

`.github/workflows/ci-public.yml` runs `docker build` and `node test/run-network.mjs --image …` directly, with `timeout-minutes: 10`. `scripts/tests/test-web-search-image.sh` wraps the same work in `run-with-deadline.mjs`. The harness itself has a 60s deadline and exact-name cleanup, so a stuck container is still reaped. The gap is only that a hung `docker build` waits for the job timeout instead of the 300s wrapper. Worth aligning, not blocking.

---

### Not findings

- **Process-wide cap of 4.** Deliberate, documented, and covered by the cross-session test. It is an availability limit: one session can hold every slot for 15s. NetworkPolicy / per-tenant quotas are the complementary control, and this directory still has no NetworkPolicy (pre-existing, called out in the README).
- **Single pinned address, no failover.** Deliberate. Failover would re-open a rebinding window.
- **No per-fetch log line.** Consistent with the sanitization goal.
- **Operator-specific NAT64 prefixes** inside global unicast are not in the IANA special registry, so they are treated as public. A CLAT that then translates them to a private address is outside what a static block list can see. Residual deployment risk, not a code defect in this PR.
- **Hairpin to the node's own public address.** Also outside a static block list. Same residual.
- **`web_search` / `search_news` error strings.** Pre-existing. This diff reformats them and does not make them worse.
- **Version `1.0.7` → `1.0.14`.** Lockfile matches `package.json`. No other consumer in this diff.

---

### Tests that are still missing

1. Real `http.request` (not the `EventEmitter` double) that refuses the connection and that returns 101, with `process.on('unhandledRejection')` asserting zero events, and a follow-up `fetchPage` that must not see `busy` or a dead process.
2. Blackhole DNS with the abort signal aborted, **and** with a signal that never aborts, asserting the resolver rejects in well under a minute. The second case is what locks the unit of `timeout`.
3. A `Content-Encoding` value of `identity, gzip` asserting `unsupported_encoding` rather than a decoded body.

The unit suite's transport double never emits `close`, so it cannot protect finding 1. The network 101 test checks latency and the error code only.

---

### Verdict

Request changes for findings 1 and 2. Finding 1 is a process-lifetime bug on a path the PR claims to have fixed (unsupported protocol switch and failed connects). Finding 2 is a wrong timeout unit sitting under a security deadline. Findings 3–5 can land in the same fix or as immediate follow-ups. The destination policy, pinning, redirect revalidation, body caps, and error sanitization are in good shape and I do not want those reworked.

Review was done by Evenfire Reviewer.

jozer-rami (commented):
## Re-review of `2d82efaf` (collateral pass against `2e053375`)

Reviewed the web-search delta independently of the PR body's disposition, and re-ran the two previously reported runtime claims on Node 24.16.0. The desktop and minikube files in this range come from the `dev` merges, not from this change.

### Previously reported

**Double `reject` in `connect` — withdrawn as a blocker.** On this Node, rejecting an already-rejected promise is a no-op, including when the second rejection is delivered later by `close`. A synchronous probe and a deferred-rejection probe both produced no `unhandledRejection`. The `close` handler is still redundant with `error`/`upgrade`, but it does not escape `fetchPage`.

**Resolver timeout — still open, and the previous unit was wrong.** `new Resolver({ timeout: 15000, tries: 1 })` in `resolvePageAddress` does not mean 15 seconds, and it is not the seconds interpretation I gave last round either. Against an unreachable resolver (`192.0.2.1`, one try) on Node 24.16.0:

| option | elapsed |
|---|---|
| `1`, `2`, `3`, `5`, `15` | ~250 ms (floor) |
| `1000` | 1942 ms |
| `2000` | 2996 ms |
| `3000` | 3003 ms |
| `15000` | 5004 ms |

The option tracks milliseconds, but this build clamps it: 15000 ms comes back in 5 seconds. The 15-second `AbortController` still bounds the whole `fetchPage` call, so a hung lookup cannot run away, but the value does not implement the deadline the name and the PR body describe, and nothing in the suite measures it. Set it from the documented millisecond unit and assert the elapsed time against a blackholed resolver.

### New, introduced by `6e9a420` / `b0e93a68`

**High — an XML-style close on a raw-text element swallows the rest of the page.** `scanPage` treats every start tag as paired and then searches for `</name`. For `<script>`, `<style>`, `<textarea>`, `<xmp>`, `<iframe>`, `<noembed>` and `<noframes>`, `/>` is not a close, so the search runs to EOF. I ran the built scanner:

| input | extracted prose |
|---|---|
| `<script src="app.js" defer /><title>Actual</title><p>Visible</p>` | empty |
| `<style media="all" /><title>Actual</title><p>Visible</p>` | empty |
| `<textarea /><title>Actual</title><p>Visible</p>` | empty |
| `<script src="app.js"></script><title>Actual</title><p>Visible</p>` | `Actual Visible` |
| `<script src=x /><p>hidden</p></script><title>Actual</title><p>Visible</p>` | `Actual Visible` (only because a later real close happened to exist) |

One self-closing raw-text tag with no later matching close removes the title and the body of an otherwise ordinary page. The new tests only cover paired tags (`<script>…</script>`, `<textarea>…</textarea>`), so they stay green. A tag that ends in `/>` needs to terminate that element the same way a paired close does, and the regression fixture has to be the no-later-close form above — the form with a following `</script>` hides the bug.

The rest of the scanner holds up on the cases I executed: comments and `<textarea>` no longer supply the title, quoted `>` stays inside the attribute, an unclosed comment discards its tail, `<plaintext>` keeps its text, and `<`.repeat(200000)` finishes in 180 ms. Two accepted gaps, not findings: `<noscript>` and `<template>` are still visible, and a declaration (`<!…>`) still ends at the first `>` even inside quotes.

### Still closed

Destination policy, DNS pinning, TLS `servername`/`checkServerIdentity`, redirect revalidation, the downgrade block, both 1 MiB ceilings, the single-layer encoding allow-list, the process-wide slot, and cancellation are unchanged and consistent. `Accept-Encoding: identity` makes the decoder path defensive. No secrets in the diff.

### Verdict

Not mergeable yet. The promise finding is closed. The resolver option still does not mean what the contract says, and the new HTML scanner drops the whole page on a self-closing raw-text tag.

---

*Review by Evenfire Reviewer*

Issue #609, opened 2026-09-10
test(e2e): distinguish unsupported shell journeys in the Playwright static auditor


alfredolopez80:
The repository's local E2E guidance requires `tools/e2e_static_audit.py` on changed E2E files, but the tool documents itself as a Playwright-spec audit and accepts arbitrary file paths. Passing a shell infrastructure E2E script always reports a missing `waitForResponse`/`waitForRequest`, even when the script uses bounded Kubernetes predicates and verifies a real MCP result.

Reproduction:

```sh
python3 tools/e2e_static_audit.py scripts/e2e/e2e-hcc-networkpolicy-lifecycle.sh
```

The reported missing browser wait is not applicable to this shell journey. Adding a fake browser-wait string or IPC marker would create misleading green evidence.

Suggested scope:

- Define and enforce the auditor's supported file types, with an explicit unsupported/not-applicable result for other languages.
- Align changed-file routing and local guidance with that scope; unsupported files must not count as audited or silently pass.
- Keep shell infrastructure checks separate: bounded predicates, explicit Kubernetes context, fixture ownership, cleanup on failure, and business-result assertions.
- Add tests for valid Playwright inputs, unsupported shell inputs, and mixed selections without weakening existing browser checks.

This is a pre-existing tooling follow-up discovered while reviewing #604. It is separate from the JavaScript lexical/reachability limitations in #454 and #551.

Issue #604, opened 2026-09-10
HCC deletes and recreates 31 NetworkPolicies every ~5 minutes after each watch reconnect


alfredolopez80:
## Summary

Every time an HCC watch dies, HCC deletes 29–31 NetworkPolicies belonging to disabled or
secret-failing McpServers, and two other reconciler lanes recreate them 17–20 seconds later. The
watches die on a fixed cadence, so this repeats indefinitely.

Current rate on `clerum-dev`: **463 accepted NetworkPolicy deletes per hour**, matched by 463
creates. Over 24 hours that is 10,719 real deletes — writes that reach etcd, unlike the rejected
404/409 traffic tracked by #582 and #584.

The mechanism predates the write-amplification campaign. **The rate does not**: it was 136/h until
2026-09-08 and tripled after #568 merged. HCC's accepted writes rose from ~900/h to ~1,500/h (+65 %)
as a direct result, while its total request volume fell.

Measured on `clerum-dev` from the GKE Admin Activity audit log; code verified against
`origin/dev` = `c7741481a`, read with `git show c7741481a:<path>`.

## The chain

1. **Every HCC watch dies with `Error: Premature close`** (`ERR_STREAM_PREMATURE_CLOSE`). Idle
   watches (CommunicationChannel, SharedFileSystem, GFS, Host) die every 5:05 minutes almost exactly;
   watches carrying traffic die roughly every 10.5 minutes. HCC opens **7** watches in
   `host-context-controller/src/k8sClient.ts` and **none of them passes `timeoutSeconds`**:
   `:3477` (SharedFileSystem), `:3547` (GFS), `:3718` (CommunicationChannel), `:4144` (McpServer),
   `:4273` (LlmHook), `:4517`, `:4650` (Host). The connection lifetime is therefore decided by
   whatever idle timeout sits in front of the apiserver, not by HCC.

2. **Each death triggers a full re-convergence.** `McpServer watch ended; recovering authoritative
   inventory...` (`k8sClient.ts:4140`) → `runInitialNetworkPolicyConvergence` (`:1500`, `:1535`) →
   `runInitialMcpServerConvergence` (`:3294`) → `Running full reconciliation for 111 McpServer(s)`.

3. **The full reconcile deletes.** For every McpServer with `spec.enabled === false`
   (`reconciler.ts:2148`) or a fail-closed secret validation
[...]

PR #610, opened 2026-09-10
fix(hcc): converge network policies with runtime eligibility

alfredolopez80:
## Summary

Disabled McpServers and fail-closed Secret verdicts could make the runtime reconciler delete policies that the Context and external-egress lanes immediately recreated. Give those policy owners a shared runtime-eligibility verdict and remove NetworkPolicy deletion from the runtime owner. Published eligibility changes propagate through WATCH and recovered LIST snapshots so recovery also restores the four HCC-owned allow families.

Preserve enabled WRC-owned runtimes and their HCC policies on Secret failures; disabling still revokes HCC policies. Retain the existing presence-based orphan classification, with regression tests ensuring eligibility revocation does not consume its safety cap. UID/resourceVersion and authoritative-absence guards remain in force.

Add stateful convergence/regression tests and a real Minikube policy-lifecycle gate with MCP business-result checks and observed watch cuts. Give that gate bounded, restore-first cleanup that retains its fault lock if restoration or resource removal cannot be verified.

Refs #604. This implements the policy-convergence portion; watch-lifetime/bookmark optimization is separate. It retains the read-first behavior merged in #601. The Playwright-auditor applicability follow-up is tracked in #609.

## Testing

Current follow-up head: `0aff52f8ce60a7813133e940f962d3963d66bf86`, based on `origin/dev` at `4d6311440`.

- [x] 334 focused HCC safety, watcher, integration and logger tests; strict TypeScript check with Node 24.
- [x] Full cleanup contracts on Bash 3.2.57 and 5.3.15, including real TERM injection at both parent/subshell boundaries. No version skip.
- [x] Shell syntax, `t2.sh` ShellCheck, Make help visibility and public-boundary checks.
- [x] GitHub CI: all 38 checks pass, including the full HCC service suite/build.

The latest full local Minikube certification belongs to `ce2fdd83a8cf2402f6d8220c8171538ae9309818`: T0/T1/T2, NP08 and Health PASS; PostgreSQL 164/164 tests across 35/35 files with no skips; policy lifecycle with 10 Context-watch cuts, MCP result `42`, verified restoration/cleanup and exit 0. The subsequent changes add documentation, a structured debug breadcrumb, shell-test portability and lint/help polish; no new Minikube certification is claimed for them.

Playwright is not part of this infrastructure change. Browser-only static-audit rules are not counted as green evidence for shell files; the existing applicability follow-up is #609.

## Checklist


alfredolopez80:
@claude Please perform a fresh, complete and exhaustive adversarial review/audit of the entire PR at head `ad779bf413d3773ae2f58efdf1d4e7afd51df728`, against base `24c96d4791db52ceb43cc00cdfb5652fedd3c689`. Review the full diff and relevant callers, not only the latest commit or the PR summary.

Independently examine correctness, security/ownership, concurrency and recovery, maintainability, and whether the tests can falsely pass. In particular, trace published Secret eligibility through WATCH and recovered LIST, the four policy families, actual queued effects, WRC ownership, UID/resourceVersion and authoritative-absence fences, orphan safety caps, and the fault-injection gate's timeout/TERM restoration, resource absence and lock release.

Report every substantiated unresolved finding, including Low severity, with file/line, concrete trigger, impact, evidence and a minimal remedy. Distinguish introduced regressions from pre-existing or out-of-scope issues. Verify claims against code/tests; do not assume the internal review established correctness.

The intended contract preserves enabled WRC-owned runtimes/HCC policies on Secret failures and revokes HCC policies when disabled. This PR addresses policy convergence; watch transport/bookmark optimization is separately scoped. Review only; do not modify the branch. Please give an explicit final verdict and identify any validation limitations.

alfredolopez80:
Addressed all seven observations from the [audit](https://github.com/evenfire-ai/evenfire/pull/610#issuecomment-5624221380) in `ce2fdd83a8cf2402f6d8220c8171538ae9309818`, after merging `origin/dev` at `4d6311440`:

1. T2 exports its actual runner grace; lifecycle refuses entry unless it is 300 seconds, covering cleanup plus margin.
2. Four ineligible-with-bindings cases assert `ExternalEgressReady=False/RuntimeNotDesired`.
3. The real scoped Context helper/queue is tested for incomplete current work and stale verdict/watch cases; dispatch arguments are asserted.
4. Documented initial optimistic policy intent under the approved published-verdict contract; runtime validation remains authoritative.
5. Clarified that the final PR preserves the baseline presence-based orphan classifier. The filter regression was caught and removed during internal review before publication.
6. Removed the unreachable private wait; shared-budget cleanup owns absence verification.
7. Protected the parent/subshell signal handoff, with separate TERM witnesses preserving the original exit status.

Mutation probes detect both the incorrect status reason and the missing scoped retry. Grace propagation, early rejection, timeout/TERM, resource absence and lock-retention contracts pass.

Validation on this exact head: CI 38/38; HCC 2,298 tests/106 files plus build/typecheck; T0/T1/T2, NP08 and Health PASS. PostgreSQL ran 164/164 tests across 35/35 files without skips. The real policy lifecycle observed 10 Context-watch cuts and MCP result `42`, then verified restoration and resource removal with exit 0. Playwright was not run for this infrastructure change. The separate static-auditor applicability issue remains tracked in #609.


@claude Please verify and certify the resolution of all seven findings on current head `ce2fdd83a8cf2402f6d8220c8171538ae9309818`. Independently inspect each correction and its regression tests, check the current CI results, and reassess the complete PR for regressions introduced by these fixes or the merge of `origin/dev`. Give an explicit final verdict: all seven findings resolved and ready for approval, or list every remaining finding with severity and file/line. Distinguish checks you execute or verify directly from the reported local Minikube evidence; state any limitations. Review only; do not modify or merge the branch.

alfredolopez80:
@claude Please execute the certification request in the [updated resolution note](https://github.com/evenfire-ai/evenfire/pull/610#issuecomment-5624805261) for head `ce2fdd83a8cf2402f6d8220c8171538ae9309818`. Verify all seven fixes, reassess the PR, and give an explicit verdict with any remaining findings and validation limitations.

zach88:
Took a deep pass through this one across the whole HCC change: the core convergence logic (`reconciler.ts`/`mcpServerSafety.ts`/`networkPolicyReconciler.ts`), the WATCH/LIST propagation in `k8sClient.ts`, the e2e/CI shell gates, and a dedicated security pass on the NetworkPolicy revocation move. The redesign holds up really well: dropping the imperative NetworkPolicy delete from the runtime owner and letting the policy owners revoke stale allows through desired-set convergence is the right call, and it's backed by tests that actually fail at the merge-base. Nothing structural to push back on, just a handful of small polish items.

## Verdict

**🟡 Approve with nits**
- Optional / non-blocking: 6

## Optional (non-blocking)

This PR can merge without any of these; they're follow-up or author's-call polish.

1. **[R1-L1] `runtimeDesired` ignores `condition.status`, leans on producer coupling** (`host-context-controller/src/mcpServerSafety.ts:110-115`). The fail-closed decision reads only `condition.reason`, not `condition.status`. That's safe today because the sole producer (`reconciler.ts:2139-2145,2273-2275`) sets `reason` and `status` in lockstep, but if any future writer ever published `SecretResolved` with `status: 'True'` and a failure `reason`, the server would be treated as not-desired and its allows revoked for a healthy server. Cheap to harden: check `condition?.status === 'False'` before looking at the reason, or drop a comment pinning the coupling as a contract. This is the one I'd actually pick up.

2. **[R1-L2] Comment reads like the runtime owner is never re-entered** (`host-context-controller/src/k8sClient.ts:3937-3944`). The first sentence says the edge propagates "without re-entering the runtime owner merely because it published status", but `scheduleRetry('MODIFIED', server)` defers a `replayIntent → replay → reconcileMcpServerWatchEvent` that does re-run `reconciler.reconcile` at `:4087`. Behavior is fine (only fires on a genuine verdict change, and reconcile is idempotent); the second comment already acknowledges the replay. Just reword the first line to "without re-entering the runtime owner inline/synchronously" so it doesn't mislead later.

3. **[R1-L3] Lost causal log breadcrumb in the refactor** (`host-context-controller/src/k8sClient.ts:3984-4004`). The old ADDED loop logged `[K8s] Re-reconciling context "X" after McpServer "Y" cached`, a per-context causal breadcrumb, and it's gone with no replacement. Per-context visibility survives via `reconcileContext`'s own log, but the "why" (after cache / after verdict edge) is lost. If you bring it back, use `hccLogger` (already imported) rather than `console.log`.

4. **[R1-L4] Cleanup test needs bash ≥4 with no skip guard** (`scripts/tests/test-hcc-lifecycle-cleanup.sh:143-153`). The `SIGNAL_AT_CLEANUP_HANDOFF` probe uses `BASHPID`, `set -T` and `trap … DEBUG`, so it aborts with `BASHPID: unbound variable` under the documented local macOS bash 3.2. CI (ubuntu bash 5) is unaffected, and there are sibling tests in the repo with the same assumption, so it's a documented-shell inconsistency rather than a defect. Optional: guard with a bash-version check + skip, or note the bash-4 requirement in a comment.

5. **[R1-L5] New Makefile target has no `## help` doc** (`Makefile`, `minikube-t2-hcc-networkpolicy-lifecycle`). Unlike its neighbours it lacks a `## …` description, so it won't show up in `make help`. Add one for consistency.

6. **[R1-L6] Recurring shellcheck SC2097/SC2098 noise** (`scripts/minikube/t2.sh:398-401`). Not a bug: the prefix assignment `T2_HEALTHCHECK_KILL_GRACE_SECONDS=$T2_HEALTHCHECK_KILL_GRACE_SECONDS node …` has the same value as the already-set shell var and exists to export it to the child (the e2e guard asserts this at `:42-44`). Worth a `# shellcheck disable=SC2097,SC2098` with a one-line why so it stops resurfacing in review.

## What checked out clean

- **No allow-policy fail-open (reproduced).** When a server flips to `runtimeDesired === false`, all four allow families (context ingress, host egress, rpc-proxy egress, external egress) get revoked through desired-set convergence: `revokeStaleContextAllows` (`networkPolicyReconciler.ts:1146-1153`, uid/resourceVersion-fenced, outside the orphan cap) and `reconcileExternalEgressSafety`/`reconcileExternalEgress` (`:2192`, `:2363-2396`), driven by three redundant triggers (the status verdict edge, the certified per-context pass, and LIST recovery). Test "revokes once and restores all four families" walks `store.size` 4→0→4.
- **Security: safe.** Any allow that lingers through a watch cut points at a pod the runtime owner already tore down (`reconciler.ts:2090,2124`), so it's inert and not reusable by another workload. HCC is the sole writer of the `SecretResolved` condition, so the verdict isn't spoofable within the trust model. New logs carry `serverName`/`contextId`/`err` where `err` comes from NetworkPolicy API calls, no Secret values.
- **Orphan cap intact (reproduced).** `orphanCandidates`/`isExternalLaneOrphan` key on `presentServerNames`, so a present-but-ineligible server doesn't consume the cap; eligibility revocation routes through `recordSafetyPassRevocation`, never `recordOrphanDelete`. Test "keeps the real orphan cap while revoking a present ineligible owner" confirms it (cap=1, 2 real orphans survive, the 4 policies of the disabled server get deleted).
- **`enabled === false` strict is correct**: the CRD `spec.enabled` defaults to `true` (`charts/clerum-crds/crds/mcpserver.yaml:333-336`), so undefined/absent is legitimately enabled.
- **Refactor into `reconcileContextsForMcpServer` is a hardening, not a regression.** The new `isCurrent` adds `runtimeDesired(server) === runtimeDesired(current)`; since a status-only verdict flip doesn't bump the desired revision, the old path would have kept reconciling with a stale verdict. No coverage gap, and the `completed === false` retry rides the coordinator's capped backoff (not a tight loop).
- **Test quality (T1–T5).** Fixtures come from the real producer (the integration test stands up the actual `McpServerWatcher` against an in-memory apiserver with real uid/resourceVersion preconditions), asserts are on observable state (policy store size + keys), and T3 was verified: the new integration test fails at the merge-base (`expected 4 to be 0`) on both the WATCH and LIST arms. `reconciler.test.ts` also asserts `list/deleteNamespacedNetworkPolicy` are never called, guarding against reintroducing the old deleter.
- **Suites (run during review):** `tsc --noEmit` clean; vitest 305 passed across the touched files; `bash -n` + `node --check` clean on all shell/mjs; shellcheck clean on the new lib scripts; fault-lock/restore-first cleanup logic verified, and the MCP business-result (`42`) and watch-cut asserts are real, not cosmetic.

zach88 (approved):

PR #611, opened 2026-09-10
refactor(desktop-app): unify the renderer on visual language

zach88:
## What

Restyles the desktop-app renderer to a single VS Code-style visual language, delivered in phases F0–F5 (plus follow-up fixes from visual and real-app review). CSS/token-first: no changes to component public APIs, navigation, state, or the resize hook. Touches `desktop-app/ui` and `desktop-app/test` only (21 files, one service).

- **F0 — token foundation:** control tokens (`--radius-xs/-control`, `--control-height-sm/-/-lg`, numbered `--surface-0..4` ramp), flat per-zone solids replacing the three zone gradients, `--control-border(-hover)`, and `--status-dot-bg` (fixes the light-mode invisible dot). Migrates the legacy dark-only `ProgressStepper.css` into the central stylesheet (theme-aware tokens) and deletes it. Zero functional change.
- **F1 — Common primitives:** collapses the duplicated button/select/chip/tab families onto one control contract (flat fills over the `linear-gradient(160deg)` accents, single `:focus-visible` ring, `--radius-control`). Fixes the `Badge tone="accent"` bug (was painted with success green) and completes the tone enum. `types.ts` changes are additive; exports unchanged.
- **F2 / F2b — shell:** flattens the app frame into crisp VS Code zones with 1px seams and an activity-bar accent, then mounts the command-center (search + notifications) in the native title bar on **every** authenticated route (previously Chat only).
- **F3 — chat drawer:** docks the floating rounded drawer flush as a secondary side bar with a persistent 1px seam; resize hook untouched.
- **F4 — chat content:** flattens turns to text-in-bands (a background is reserved for code/JSON, tool-calls, and errors); the user turn is the only turn with a neutral band.
- **F5 — mounted-app actions:** moves the Back / Refresh / Copy / chat-toggle row into the native title bar's leading slot, icon-only, recovering the band. "Back to apps" is removed; returning to the catalog stays available via the Apps sidebar nav and the `app.backToApps` command.

## Why

The renderer carried three parallel button/select/chip/tab families, hardcoded radii and gradient fills, dark-only component CSS that broke in light mode, and a floating-card shell that read as frosted overlays rather than structural zones. Unifying on one control contract removes the duplication and the light-mode defects, and gives the shell a consistent, flatter VS Code language across every route.

## How

- Token-first: new geometry and surface tokens land in `tokens.css` (F0) so later phases restyle by referencing them, keeping the diff CSS-heavy and logic-light.
- The command-center unification (F2b) is the one deliberate behavioral change (mount placement): a single `TitlebarActionsPortal` + `AppHeader placement="titlebar"` replaces the chat-vs-other mount ternary; the portal target is set at `.app-frame` level outside the auth gate, so only authenticated routes mount it. Same component, same props.
- Title-bar interactivity: Chromium unions all `-webkit-app-region` rects and the last in document order wins on overlap, so the leading actions are made clickable by collapsing to a single root `drag` region with each control carving itself out as a `no-drag` descendant (the wrapper is `pointer-events: none`, on which `-webkit-app-region` is ignored). The mounted-app actions track the app content's left edge via a published `--sidebar-width`, clamped to clear the window controls (which sit on the left on every platform).
- The sandbox app renders in a native `WebContentsView` that always composites above renderer DOM, so the DOM tooltip flyout was dropped in favor of the native `title` (the only tooltip that paints over the app content).
- Sidebar hover/active legibility: the global `button:hover:enabled` accent wash was overriding bare `<button>` rows; scoped guards restore a neutral edge overlay with measured label contrast ≥6:1 in both themes.

## Testing

- `style-rules` lint: exit 0.
- `vite build` / `build:renderer`: compiles.
- Vitest: **2180/2180 green**; unit specs (SandboxUiPage + focus) re-anchored to a stable `sandbox-ui-mounted` testid and the native-tooltip model, injecting a portal container so the real portal path is exercised. The `responsiveUtilityGutter` browser test keeps the embed-slot right-edge and base-gutter invariants (base gutter 0 at every width; apps drawer reserves its rail only while open).
- Code + security + collateral reviews: clean across phases.
- Visual verification: components rendered against the real stylesheet and screenshotted at rest/hover/active in dark and light (Playwright); the title-bar drag/`-webkit-app-region` fixes were verified in the real Electron app (they cannot be exercised in a browser).

## Risks / Follow-ups

- **e2e (Playwright) lane not run here** — it needs a cluster. Five e2e specs were re-anchored off the removed "Back to apps" button (to the testid + Apps sidebar nav); they should run in CI before merge.
- Deferred out of these visual phases: unifying the `StatusBanner`/`Toast` `Tone` type rename (`error→danger`, `warn→warning`) is a breaking change across 23 call sites and was left for a separate change.


josue-tech-94:
Found one merge-blocking issue in the updated E2E test:

**`R1-H1` — `desktop-app/test/e2e-playwright/responsive-layout-producer-backed.spec.ts:190`**

The test currently waits for a DOM `[role="tooltip"]`, but the implementation intentionally no longer renders a DOM tooltip here. `TitlebarLeadingAction` uses the native `title` attribute instead, and the updated unit test explicitly confirms that no `[role="tooltip"]` exists.

As written, the cluster-backed Playwright scenario will deterministically time out when it reaches this assertion.

**Required fix:** update the E2E assertion to verify the native `title` attribute and accessible button name, while keeping the existing geometry/non-overlap check. Please do not restore a DOM tooltip just to satisfy the test, since it would be occluded by the mounted `WebContentsView` and contradict the intended implementation.

josue-tech-94 (approved):

PR #615, opened 2026-09-11
fix(profile-ui): update vulnerable dependencies

josue-tech-94:
## Summary

Updates the vulnerable Profile UI dependencies: Next.js pinned exactly at `16.3.5`, Vitest to `^4.1.11`, and the transitive Sharp override to patched `0.35.4`.

## Validation

- `npm install` in Profile UI and its local frontend-components package
- `npm audit --package-lock-only` — 0 vulnerabilities
- `npm test` — 75 Node tests and 25 Vitest tests passed
- `npm run build` — passed

### DIFF
--- a/profile-ui/next-env.d.ts
+++ b/profile-ui/next-env.d.ts
@@ -1,6 +1,7 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
+import "./.next/types/routes.d.ts";
+import "./.next/types/root-params.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/profile-ui/package-lock.json
+++ b/profile-ui/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "clerum-profile-ui",
-  "version": "0.1.136",
+  "version": "0.1.137",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "clerum-profile-ui",
-      "version": "0.1.136",
+      "version": "0.1.137",
       "dependencies": {
         "@clerum/desktop-app-links": "file:../packages/desktop-app-links",
         "@clerum/frontend-components": "file:../packages/frontend-components",
-        "next": "16.2.11",
+        "next": "^16.3.5",
         "react": "^18.3.1",
         "react-dom": "^18.3.1"
       },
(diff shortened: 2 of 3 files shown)


zach88:
Took a pass through this one focused on where it actually matters for a deps-only bump: does the lockfile really resolve the patched versions with no vulnerable copies left behind, and is the provenance clean. It does, and it is. This is a clean, deterministic bump.

**🟡 Approve with nits**
- Optional / non-blocking: 1

## Optional (non-blocking)

The PR can merge without this — it's a policy call, your judgment.

1. **[R1-L1] `next` loses its exact pin (pin → caret)** — `profile-ui/package.json:14`. `next` goes from an exact pin `16.2.11` to a caret range `^16.3.5`. The repo's convention is to pin `next` exactly (`control-ui` is still on `16.2.11` exact, and this manifest was exact before too). The committed lock does pin the concrete `16.3.5` with integrity, so `npm ci`/CI stay deterministic — the only risk is a future lock regeneration (`npm install`/`npm update`) floating `next` to an un-audited 16.x minor, which cuts against the point of a hardening PR. Counterpoint: the caret actually makes `next` consistent with every other dep in this manifest (`react`, `react-dom`, `vitest` and the devDeps all use caret already). So either re-pin to exact `16.3.5` to match `control-ui`, or keep the caret as a deliberate policy change — both are defensible. My lean is re-pin, for parity with `control-ui`.

## What checked out clean

- **Single, patched copies.** The lock resolves exactly one node each: `next` 16.3.5, `sharp` 0.35.4 (the override took — no residual copy), `vitest` 4.1.11. No vulnerable transitive copy of any of the three anywhere in the tree.
- **Supply chain.** All 194 `resolved` URLs point to `registry.npmjs.org`, every one with a `sha512-` integrity hash. No git URLs, no odd tarball hosts, no host swaps. The only non-registry entries are the pre-existing `file:../packages/*` workspace links, untouched.
- **`npm audit --package-lock-only` → 0 vulnerabilities**, which backs the PR body's claim.
- **Lock ↔ manifest in sync**; `next-env.d.ts` is a legit Next 16.3 regen (type-path move + `root-params.d.ts`, generated file); `frontend-components` doesn't depend on next/sharp so its lock rightly didn't change; app code only uses stable Next APIs (`next/image`, `next/navigation`) and this is a minor bump, so breaking risk is low.

One thing I didn't run locally: a build/typecheck against 16.3.5 (would've needed installing the head's node_modules). The body says it passed; low risk given it's a minor with stable APIs, just flagging it as not independently verified here.

zach88 (approved):

PR #616, opened 2026-09-11
fix(control-ui): update vulnerable dependencies

josue-tech-94:
## Summary

Updates the vulnerable Control UI dependencies: Next.js to `^16.3.5`, Vitest to `^4.1.11`, and the transitive Sharp override to patched `0.35.4`.

Aligns test fixtures with the current GFS upload and profile-admin API contracts so the Next.js production type-check completes.

## Validation

- `npm install` in Control UI and its local frontend-components package
- `npm audit` — 0 vulnerabilities
- `npm test` — 183 files and 1,955 tests passed
- `npm run build` — passed

### DIFF
--- a/control-ui/components/__tests__/GfsBrowser.uploadConflict.t3.test.tsx
+++ b/control-ui/components/__tests__/GfsBrowser.uploadConflict.t3.test.tsx
@@ -98,20 +98,16 @@ describe('GfsBrowser Upload v2 conflict retry — T3 historical reproduction', (
     vi.stubGlobal('fetch', fetchMock)
     mockCreateGfsUploadJob.mockImplementation(input => {
       const job = actualUpload.createGfsUploadJob(input)
-      return {
-        start: async () => {
-          try {
-            return await job.start()
-          } catch (error) {
-            producerErrors.push(error)
-            throw error
-          }
-        },
-        pause: job.pause.bind(job),
-        resume: job.resume.bind(job),
-        cancel: job.cancel.bind(job),
-        snapshot: job.snapshot.bind(job),
+      const start = job.start.bind(job)
+      job.start = async () => {
+        try {
+          return await start()
+        } catch (error) {
+          producerErrors.push(error)
+          throw error
+        }
       }
+      return job
     })
 
     render(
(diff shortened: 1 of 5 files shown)


zach88:
Took a careful pass through this one, both on the supply-chain side and the two test edits. It's a clean, well-scoped security bump — I actually ran the audit, the full suite, and the production build against the PR head rather than taking the body's word for it, and everything lines up.

**🟢 Ready to approve**
- Optional / non-blocking: **0**

## What checked out clean

Everything here was verified by running it in a throwaway worktree at the PR head, not assumed:

- **Vulns are actually gone.** Baseline (merge-base) had 7 vulnerabilities (1 critical + 3 high + 3 moderate); the PR head reports `0`. The `sharp` override to `0.35.4` resolves across the *entire* `@img/sharp-*` tree (all 26 platform binaries) — no branch left on the vulnerable `0.35.3`. `next` resolves to `16.3.5` with no open CVE, and the transitive bumps (`undici` 7.29.1, `nanoid` 3.3.19) land too. No security pin was loosened and no override was removed.
- **The T3 test still reproduces.** In `GfsBrowser.uploadConflict.t3.test.tsx` the mock switched from returning a partial plain-object wrapper (which no longer satisfies the real `GfsUploadJob` return type — that's what broke the prod type-check) to returning the real instance and only shadowing `start` to capture producer errors. `pause/resume/cancel/snapshot` stay on the prototype, correctly bound. Ran it: the conflict/retry path still fires (`uploadNames === ['report.txt','report (1).txt']`) and `producerErrors[0]` still captures the structured 409. Fidelity went up, not down.
- **The fixture alignment matches the real producer types.** In `ProfileAdminAgentCompatibility.test.tsx` the added fields match `lib/api.ts` exactly — `ContextSpec.mcpServers` (required `string[]`, api.ts:874), `AdminUser.picture` (required `string|null`, :1006), and `AdminUserContext` requiring `id`+`picture` (:1048,1051). This file uses `vi.mocked(...)` so the compiler enforces the real signature, which is why it needed the update and the untyped-cast sibling tests didn't.
- **Body claims all true:** `npm audit` → 0, `npm test` → 183 files / 1955 tests pass, `next build` → passes with a clean production type-check.

One small heads-up, not a blocker and no action needed: `next` moved from an exact pin (`16.2.11`) to a caret (`^16.3.5`) in package.json:71. The lockfile still pins `16.3.5` so `npm ci` stays reproducible, and it lines up with the rest of the deps in the file (`react`, `recharts`, `yaml` are already carets). Just flagging the pin-strictness change since the sibling `sharp` override stays exact.

Nice tidy security PR — good to merge as-is from my side.

zach88 (approved):

PR #617, opened 2026-09-11
fix(desktop-app): update vulnerable dependencies

josue-tech-94:
## Summary

Updates Desktop dependencies with security findings: Electron to `^41.10.7`, Electron Forge packages to `^7.11.2`, js-yaml to `^4.3.2`, and Vitest to `^4.1.11`.

Adapts test doubles and file-system spy cleanup to Vitest 4 so the test and TypeScript build paths remain compatible.

## Validation

- Node `v24.3.0` and `npm run verify:electron` passed
- `npm test` — 234 files and 2,229 tests passed
- `npm run build` — passed

## Remaining audit note

`npm audit` still reports 29 high-severity findings in the Electron Forge dependency chain. The only automated remediation is a downgrade to Forge 6.4.2, which is not a safe upgrade path and is intentionally excluded.


zach88:
Took a careful pass through this one, focused on the two things a security dependency bump can quietly get wrong: whether the lockfile actually resolves to versions that close the CVEs (and doesn't drag in anything sketchy), and whether the Vitest 4 test-double adaptations left any test hollow. Verified against the real npm registry/OSV and by running the suite under real Vitest 4.1.11 — not just reading the diff. It's a clean, well-scoped change.

**🟢 Ready to approve**
- Optional / non-blocking: 0

## What checked out clean

**Dependencies / lockfile**
- package.json ↔ lock consistent: every resolved version satisfies the new ranges — `js-yaml 4.3.2`, `electron 41.10.7`, `vitest 4.1.11`, and the whole forge chain at `7.11.2`.
- Supply-chain integrity verified: all registry packages carry `integrity`, no `resolved` points anywhere unexpected (just the workspace `file:` links and the pre-existing `@electron/node-gyp` git pin, byte-identical base↔head). The −1371/+443 net reduction is the Vitest 3 tree dropping out plus forge dedup — coherent. The two non-trivial additions (`obug`, `@electron-internal/extract-zip`) are legit, not typosquats, integrity matches the registry.
- The runtime bumps close real CVEs, not cosmetic: js-yaml (merge-key + `!!omap` DoS) and electron (sandbox/CORS/session-cache advisories). `npm audit --omit=dev` = 0 in production runtime.

**Test doubles / Vitest 4**
- `tsc --noEmit` clean. The mock types now match the real exported signatures (`makeArgs(): StartSandboxUiRefreshArgs`, `onError: (error: Error) => void`), and no assert capability was lost — the one test asserting on `onError` uses a local `vi.fn()`, not the typed field.
- `vi.restoreAllMocks()` in `chatStoreBinding.test.ts` runs in `afterEach` after the assertions, so it can't silence anything — it's the right mitigation for re-spying an already-spied method under v4.
- `createMockBrowserWindow` in `desktopWindow.test.ts` is a pure refactor — new object per call, implementation re-applied in `beforeEach`.
- Ran it green under real Vitest 4.1.11: the 3 touched files 42/42, and the full desktop-app suite 234 files / 2229 tests pass. Matches the validation note in the PR body.

Nice tight PR. Nothing blocking from my side.

zach88 (approved):

PR #618, opened 2026-09-11
perf(hcc): skip completed watch recovery passes without losing repairs

alfredolopez80:
## Summary

HCC omits a redundant full policy pass after unchanged McpServer or Context watch recovery only when inventory authority is usable, policy and runtime effects are complete, no repair remains pending, and the existing periodic timer is active. Failed or retired effects retain their repair obligations. BOOKMARK events perform no reconciliation work; only McpServer and Context watches request them.

This builds on PR B (#610), preserving its ownership and eligibility behavior. The Minikube overlay enables the observed 3600-second policy resync interval; base configuration is unchanged. The development fixture proves actual repair, revocation, protected-API recovery and MCP results, then restores its configuration.

Refs: #604

## Previous complete validation — ef854c1b9027

- HCC: 2,372 tests across 110 files; build/typecheck passed.
- Fixture: 52 Node tests and nine shell contract groups passed. The invalid-command regression fails on the prior implementation and passes after the fix.
- Exact-head owned Minikube certification: T0=PASS, T1=PASS (164 tests, 35 files, zero pending), T2=PASS, NP08_HCC_AUTHORIZATION=PASS, Health=PASS; process exit 0. Playwright=NOT_RUN.
- Real journey: three controlled omissions per watch without recovery admission, drift repair, interrupted runtime repair, exact retained egress policy identity, protected API 503/recovery and MCP result 42 passed.
- Fixture cleanup and restoration passed; independent public-state checks confirmed no run resources, HCC Ready/Available and policy resync restored to 3600 seconds.
- CI and CodeQL: all 38 checks successful. Advanced Security: zero annotations and zero open PR alerts; no suppressions or manual dismissals.

## Review resolution

Internal review corrected new-UID ownership recreation, nested interruption diagnostics, environment dependency-order restoration and a vacuous policy assertion. TLS now uses one direct OpenSSL generator without a shell and configures CA trust in a reusable TLS context. The exhaustive Claude audit found no blocking correctness defects. Its two minor observations are resolved: invalid commands are acknowledged once with safe identity handling, and the observation literal is consistently NOT_DEMONSTRATED. Shared pre-existing harness input validation is tracked separately in #619.

The browser-oriented static auditor still emits the known non-applicable missing-network-wait diagnostic on infrastructure scripts (#609); this is not claimed as a passing browser audit. Operational Q6-A remains separate and unmeasured. BOOKMARK receipt does not prove callback processing or reduced disconnect frequency. No GKE mutation, production deployment or merge is included.

## Checklist

## Zach88 review follow-up

At `b4d0733cc`, M1 is fixed by requiring OpenSSL in both the helper and preflight before generation; LibreSSL and unknown implementations fail explicitly without creating files. The real system LibreSSL was tested safely: rejected before generation, empty stdout, zero artifacts. M2 now asserts pending repair and its discharge after a successful retry, without manufacturing a readiness verdict through its mock.

L1 is clarified in the overlay comment; L2 uses the existing structured logger while preserving the consumed failure marker. L3's BOOKMARK guards and L4's UID-local ownership reset are required by the recovery contract and remain. L5 identifies no current fixture defect; L6 was already corrected to NOT_DEMONSTRATED. No additional issue is needed for these intentional or already-resolved items.

The branch includes current dev (`309dc02bd`); updated UI/Desktop dependencies and Electron were verified locally. Local checks: 54 fixture tests, 2,372 HCC tests, build and recovery log-marker contracts passed. Final CI on `b4d0733cc`: all 38 checks passed; Advanced Security has zero annotations and zero open PR alerts. T0 passed locally. The fresh Minikube bootstrap stopped before T1/T2 because the unchanged Control UI Docker build omits shared typecheck fixtures. This base issue is tracked in #623; no current-head T2 certification is claimed. The failure predates these HCC adjustments in the current dev inputs.


alfredolopez80:
@claude Please perform a complete, exhaustive re-review and audit of PR #618 against its current `dev` base and HEAD, including every changed production, test, fixture, deployment and CI file. Report all actionable findings, including medium and inexpensive low-severity issues, with exact locations, concrete failure scenarios and whether the PR introduced them. Distinguish confirmed defects from pre-existing issues and unsupported risks.

Please independently challenge correctness rather than relying on green tests. Pay particular attention to:

- Identical McpServer/Context LIST→WATCH recovery, usable policy authority, completed runtime/policy revisions, pending repair obligations, overlapping passes and scheduler admission/trailing behavior.
- Same-name CRD deletion/recreation with a new UID and changed `managed` ownership; same-UID ownership changes must remain forbidden and interrupted cleanup must not erase a replacement's state.
- BOOKMARK handling, real transport parameters, observation bounds and the distinction between receipt and demonstrated disconnect reduction.
- Nested inventory-interruption error classification, mixed real failures, and the eight production convergence causes.
- Real resource repair, policy revocation, `/ready` versus protected-API authorization, actual MCP invocation, and false-positive/vacuous test paths.
- Fixture restoration: exact environment dependency order, UID/resourceVersion checks, preservation of unrelated values, rejection of ambiguous concurrent structural edits, cleanup and finite deadlines.
- TLS generation through direct OpenSSL argv, no private-key persistence, malformed identities, timeout/reaping, upstream CA and hostname verification. Reassess GitHub Advanced Security alerts #1368 (`js/file-access-to-http`) and #1369 (`js/indirect-command-line-injection`) against the current implementation and report the exact dataflow if either remains concerning.

Internal review found and corrected ownership recreation, ordered restoration, an empty-policy-set oracle and nested interruption diagnostics. Do not assume those fixes are complete: attempt to falsify them. Shared pre-existing harness input validation is separately tracked in #619; flag distinct gaps rather than duplicating it.

Please end with remaining blockers, non-blocking findings, evidence gaps and an explicit audit verdict. This request does not authorize merging or changing production infrastructure. The operational Q6-A measurement remains separate and unperformed; do not infer a production SLA or performance benefit from local certification.

alfredolopez80:
Both non-blocking observations from the exhaustive review are addressed in `ef854c1b9`:

- The proxy records each command publication before validation, acknowledges a syntactically valid rejected ID correctly, and does not repeatedly write rejection records for unchanged invalid JSON or commands. Unsafe IDs are never echoed. Existing valid-ID idempotence is preserved. A deterministic timer test fails on the prior implementation and passes after the fix, including acceptance of a subsequent valid command.
- The development observation literal is consistently `NOT_DEMONSTRATED`; no performance claim or historical evidence was changed.

The earlier four internal findings remain fixed: new-UID ownership recreation, nested interruption classification, exact environment-order restoration, and an exact nonempty egress-policy identity witness. Shared pre-existing harness input validation is tracked separately in #619.

The review's evidence note was based on an older PR description: `51c9e9f19` had already passed exact-head T0/T1/T2, NP-08, Health and CI when the audit was requested. GHAS #1368/#1369 disappeared after the direct OpenSSL and TLS-context changes; no suppression or manual dismissal was required. Current final validation is recorded in the PR description.

The audit found no blocking correctness defects. Operational Q6-A remains separate and unmeasured, and Playwright is not part of this infrastructure lane. This closes the requested review remediation; it does not merge or close the PR.

zach88:
Took a deep pass through this one, tracing the skip decision and the `pendingReconciliations` obligation lifecycle end to end against the real code, with tests and typecheck run on a worktree. The core behavior holds up: the skip gate is genuinely conservative and I couldn't build a case where an unchanged-recovery skip drops a repair. What's left is all in the test/e2e scaffolding plus a couple of hygiene items, nothing that gates the merge.

### 🟡 Approve with nits
- **Optional / non-blocking: 8** (2 medium, 6 low)

None. No blockers, criticals, highs, or regressions.

## Optional (non-blocking)

The PR can merge without these. My call is to knock out the first one because it breaks the local suite, and treat the rest as author's-discretion / follow-up.

1. **[R1-M1] TLS fixture breaks on macOS/LibreSSL and drops a private key in the cwd** (`scripts/e2e/_lib/hcc-watch-tls.mjs:24`). The generator uses `openssl ... -keyout -` expecting GNU OpenSSL's stdout semantics. On the default macOS toolchain (LibreSSL 3.3.6), `-keyout -` writes the private key to a file literally named `-` in the cwd and emits only the CERTIFICATE on stdout, so `blocks.length !== 2` and it throws `fixture_certificate_generation_failed`. Reproduced (verified the command semantics directly): running `bash scripts/tests/test-hcc-watch-api-proxy.sh` on a dev laptop reds the four TLS-dependent suites and litters an ephemeral RSA key file named `-`. CI is unaffected (ubuntu-latest = GNU OpenSSL, confirmed at `.github/workflows/ci-public.yml:432/449`). Fix: detect LibreSSL in the preflight and fail with a clear message, or write the key to an explicit `mkdtemp` path and read it back instead of `-`.

2. **[R1-M2] A regression test doesn't actually cover the new behavior** (`host-context-controller/src/k8sClient.test.ts:5851`). Reproduced (read + ran). `'retains repair and invalidates authority after a post-revocation additive failure'` has no assertion on `networkPolicyRepairPending` (the new state the PR sets), and its `isReadinessInventoryAuthoritative()).toBe(false)` assertion (line 5879) is manufactured by the test's own mock (`hasCertifiedSafetyInventory.mockReturnValue(false)` at line 5864), so it passes regardless of the PR code. Worth noting: the retention invariant itself *is* genuinely covered elsewhere with real `failPolicyType` injection (`policyEligibility` A-T4 / A-T6), so this is a weak test rather than a coverage hole. Fix: assert `state.networkPolicyRepairPending === true` after the failure, and drop the artificial `hasCertifiedSafetyInventory=false` if the intent is that authority stays up with repair pending.

3. **[R1-L1] The minikube overlay enables 3600s resync for every HCC deployment** (`deploy/overlays/minikube/patches/hcc-policy-resync.yaml:1`). Base default is `0` (periodic resync disabled). PR A overrides it to 60s for its own path, so the 3600 only applies to non-PR-A runs, where a periodic pass could in theory fire mid-test during the sibling `E2E_HCC_POLICY_LIFECYCLE`/np604 gate. Very low probability (a fresh minikube never reaches the interval within the 900s budget), and the body's "base config unchanged" claim does hold. Consider scoping the enablement to the PR A path or noting the interaction in the patch comment.

4. **[R1-L4] `beginStatusTracking` clears `managedSnapshot` on a uid change** (`host-context-controller/src/reconciler.ts:441`). This is correct and covered by `reconciler.managedRecreate.integration.test.ts` (lets the immutable `managed` field change legitimately across delete+recreate), but it's tangential to the PR's stated intent (`pendingReconciliations`). Just flagging to confirm it belongs in this PR rather than being a stray hitchhiker.

5. **[R1-L2] Partial logging migration** (`host-context-controller/src/k8sClient.ts:2954, 4701`). The diff moved several recovery/convergence lines from `console.*` to `hccLogger` but left `console.*` on adjacent lines it also touched. No mandatory migration rule in HCC and the file already uses `console.*` in ~150 spots, so purely local consistency. Boy-scout the lines you already edited if you feel like it.

6. **[R1-L3] Defensive BOOKMARK guards on watches that never request bookmarks** (`host-context-controller/src/k8sClient.ts:3610, 3695, 3760`). `allowWatchBookmarks` is only set on the McpServer and Context watches, so SFS/GFS/CC will never receive a BOOKMARK. Reasonable safety belt, not a defect, noting only in case you'd rather not leave unreachable branches.

7. **[R1-L5] Hand-written apiserver fixture in the ownership test** (`host-context-controller/src/reconciler.managedRecreate.integration.test.ts:36`). Low risk here (the real reconciler runs against it and only reads `status`), but if this file grows to assert body content, derive the stored Deployment from the reconciler's own generated body like `runtimeRecovery` does.

8. **[R1-L6] Mixed-language literal** (`scripts/e2e/_lib/hcc-watch-bookmarks.mjs`): `disconnectBenefit: 'NO_DEMOSTRADO'` is Spanish in otherwise-English code. Intentional and consistent with its tests, just noting (`NOT_DEMONSTRATED`).

## What checked out clean

Verified, not assumed:

- **Skip gate** (`k8sClient.ts`): all four conditions required to skip (authoritative inventory, policy+runtime effects complete, no repair pending incl. `!reconciler.hasPendingReconciliation()`, active resync timer). `runtimeRepairPending` is set before the authority check and cleared only on real success; failed effects, interrupted readiness polls and failed status publications all retain the obligation. Timer inactive (`netPolResyncTimer === null`) forces a request, so the optimization can never strand work when no periodic pass will follow. The decision is synchronous, so no watch callback can interleave between the authority check and the skip.
- **`pendingReconciliations` lifecycle** (`reconciler.ts`): creation, identity-guarded clearing (`=== obligation`, which is a feature: a second in-flight reconcile overwrites the object so the earlier caller declines to clear work it no longer owns), and fresh-object replacement on failure via `recordIncompletePublication`. An unchanged-but-successful status write does *not* record an obligation (the short-circuit at 1979-1981 returns `false` without it), so no-op writes don't break the optimization. No under-retention on any `performReconcile` exit.
- **`deleteRuntimeResources` boolean** threaded through all three callers; the identically-named `hostReconciler.ts` method is a different class and untouched.
- **`networkPolicyReconciler`**: `reconcileContext` already returned boolean at base; the added `if (!completed) throw <sentinel>` lands in the per-context try/catch, gets wrapped in an `AggregateError`, and `isDesiredNetworkPolicyInventoryInterruption` already unwraps it recursively. No caller breaks.
- **Security** (full diff): TLS with `rejectUnauthorized:true` + strict `checkHost`/`checkIP`, `spawnSync(openssl, [...], {shell:false})` with the hostname validated against a DNS/IP regex before interpolation (no command/arg injection), key kept in memory. Proxy forwards the bearer token in-memory only to the pinned `kubernetes.default.svc` upstream with a pinned CA; logs only an allow-listed `code`, never the token/headers/body. Config snapshot excludes potentially-secret env values. No fail-open flip in `shouldFailClosedForSecretFailure`. No desired NetworkPolicy left unapplied by the skip. No findings.
- **Failure-path coverage** (the PR's core promise): well covered with observable assertions (resource op failure, interrupted readiness poll, failed status publication, BOOKMARK=zero-work with a positive control). The integration suites drive `reconcile()`/`recover...()` end to end against a fixture store and assert observable resource/CRD state rather than call counts.
- **CI green** on head; HCC build/test/typecheck pass, CodeQL pass. Locally: `tsc` exit 0 across the service, the touched vitest suites green (1000+ tests), `shellcheck -S warning` and `bash -n` clean, `test-hcc-watch-pr-a.sh` 9/9 pass.

alfredolopez80:
@zach88 Thanks for the detailed review. I checked each observation adversarially against the current code and addressed the applicable changes in `aed67365e` and `b4d0733cc`. The branch also incorporates current `dev` at `309dc02bd` through merge commit `e001826b4`.

Here is the disposition of all eight items:

1. **R1-M1 — Fixed.** The TLS helper now checks the OpenSSL implementation before generating anything. LibreSSL and unrecognized implementations fail explicitly with `fixture_openssl_required`; the same check runs in the fixture runner and PR A preflight. I chose rejection rather than writing a temporary private-key file, preserving the fixture's memory-only requirement. The actual system LibreSSL 3.3.6 was tested through this guarded path: exit 1, clear preflight error, empty stdout and zero created files. Negative tests also prove that unsupported implementations never reach generation. The supported OpenSSL path and signer-timeout regression remain green.

2. **R1-M2 — Fixed, with a clarification about authority.** The test no longer sets `hasCertifiedSafetyInventory` to false and then asserts the result of its own mock. It now checks the relevant lifecycle directly: `networkPolicyRepairPending` starts false, becomes true after the additive failure, a retry is scheduled, and successful retry clears both the pending state and timer. I did not replace the old assertion with “authority must remain true”: the real `NetworkPolicyReconciler.fullReconcile` catches failures and calls `recordSafetyPassOutcome('failed')`, which invalidates its safety certificate even after an earlier revocation callback. The real authority behavior remains covered by the integration suites; this unit test now focuses on pending repair and retry.

3. **R1-L1 — Clarified; intended behavior retained.** The overlay comment now explicitly states that 3600 seconds applies to every HCC using the Minikube overlay, including sibling tests, and that gates must account for periodic ticks. PR A temporarily uses 60 seconds to exercise the mechanism and restores 3600. Scoping the persistent timer exclusively to the PR A fixture would contradict the approved recovery plan: skipping an otherwise-complete recovery requires an active periodic repair path. The shared base and GKE configuration remain unchanged.

4. **R1-L4 — Reviewed and retained as a necessary regression fix, not an unrelated change.** Retaining an incomplete guarded-delete obligation means `reconcileDelete` can intentionally finish without `clearStatus`. If a same-name replacement with a new UID arrives, the old `managedSnapshot` would otherwise survive and incorrectly reject its legitimate ownership change. Resetting it only when the UID changes fixes that consequence of this PR. The barrier tests cover both ownership directions, verify that pending obligations drain, and retain rejection of a managed change with the same UID and a new generation. Removing this reset would reintroduce the reproduced regression.

5. **R1-L2 — Fixed locally.** The touched periodic-resync and Context-failure paths now use the existing `hccLogger`, with structured fields. The Context-failure message retains its existing text because sibling E2E gates consume that marker. CI caught the initial marker change; I restored the contract and ran the complete marker/recovery contract suite. No service-wide logging migration was introduced.

6. **R1-L3 — Reviewed and retained intentionally.** Only McpServer and Context request bookmarks, as intended. The approved plan nevertheless requires all seven callbacks to tolerate a metadata-only BOOKMARK without dereferencing `spec` or producing work. The early SFS/GFS/CC guards satisfy that defensive callback contract; they do not enable bookmarks on those watches. Tests also feed an ADDED event through the same callbacks as a positive control, so the no-work assertions cannot pass merely because the callback is inert. There is no defect to fix by removing those guards.

7. **R1-L5 — Dismissed as nonactionable for the current test.** The ownership fixture exercises the real reconciler and the delete/recreate race; it is not intended to verify Deployment rendering. Capturing generated request bodies would add a different assertion scope already covered elsewhere, without addressing a demonstrated defect here. No change or follow-up issue was created for this item.

8. **R1-L6 — Already fixed.** The code and tests consistently use `NOT_DEMONSTRATED` since `ef854c1b9`. Historical evidence was left intact; the wording change does not imply any measured disconnect benefit.

**Validation on the updated branch:** 2,372 HCC tests across 110 files, build/typecheck, 54 fixture tests, and the applicable shell/recovery-marker contracts passed. All 38 GitHub checks are green on `b4d0733cc`; Advanced Security has zero annotations and no open alerts for this PR.

**Local runtime limitation:** the previous Minikube container was missing, so I recreated the owned profile and attempted a fresh canonical certification. T0 passed, but bootstrap stopped before T1/T2 because Control UI's Docker build typechecks shared fixtures it does not copy. The Control UI tree and all relevant Docker inputs are identical to current `dev`; this separate build issue is tracked in #623. I am not claiming that an older T2 verdict certifies the new HEAD. Operational Q6-A remains separate and unmeasured.

This resolves the review dispositions above; it does not merge or close the PR.

jozer-rami (commented):
Review of `b4d0733cc221fe1a2a5e6dff422d24a31b959e9c` against `dev`. Three issues, in order of severity. I'm following up with fixes.

### 1. Medium: the `e2e-wrc-hcc-context-noop-resync` lane will fail every time

[`k8sClient.ts:2954`](https://github.com/evenfire-ai/evenfire/blob/b4d0733cc221fe1a2a5e6dff422d24a31b959e9c/host-context-controller/src/k8sClient.ts#L2954) moved the NetworkPolicy resync startup log to `hccLogger.info('[K8s] NetworkPolicy periodic resync enabled', { intervalSeconds })`, so the message no longer contains `(every Ns)`.

[`e2e-wrc-hcc-context-noop-resync.sh:566`](https://github.com/evenfire-ai/evenfire/blob/b4d0733cc221fe1a2a5e6dff422d24a31b959e9c/scripts/e2e/e2e-wrc-hcc-context-noop-resync.sh#L566) (`hcc_netpol_timer_arm_epoch`) still searches the HCC logs for `NetworkPolicy periodic resync enabled (every ${RESYNC_SECONDS}s)`. That text is never logged now, so the function returns 1 and the gate aborts with "could not timestamp the freshly armed NetworkPolicy periodic timer". The PR doesn't update that script.

### 2. Low: two negative assertions can no longer fail

[`readinessSafetyFence.integration.test.ts:627`](https://github.com/evenfire-ai/evenfire/blob/b4d0733cc221fe1a2a5e6dff422d24a31b959e9c/host-context-controller/src/readinessSafetyFence.integration.test.ts#L627) and [`k8sClient.test.ts:4641`](https://github.com/evenfire-ai/evenfire/blob/b4d0733cc221fe1a2a5e6dff422d24a31b959e9c/host-context-controller/src/k8sClient.test.ts#L4641) check that no `console.warn` call satisfies `String(call[0]) === '[K8s] pass ended without certifying: inventory authority lost'`.

That warning now goes through `hccLogger.warn`, which writes one JSON string to `console.warn` ([`logger.ts:70-79`](https://github.com/evenfire-ai/evenfire/blob/b4d0733cc221fe1a2a5e6dff422d24a31b959e9c/host-context-controller/src/logger.ts#L70-L79)). An exact match never happens, so the `toBe(false)` checks pass even if the certification bug they guard against comes back. The positive `.includes(...)` checks nearby still work.

### 3. Low: one persistently failing McpServer turns off the identical-recovery skip for the whole controller

- If `ensureService` or `ensureDeployment` keeps failing for one server (admission webhook, ResourceQuota, PodSecurity), `performReconcile` returns `false` ([`reconciler.ts:2275-2282`](https://github.com/evenfire-ai/evenfire/blob/b4d0733cc221fe1a2a5e6dff422d24a31b959e9c/host-context-controller/src/reconciler.ts#L2275-L2282)). That server's obligation then stays in `pendingReconciliations` indefinitely.
- `hasIncompleteReconciliation()` counts every server, so [`runInitialMcpServerConvergenceCore`](https://github.com/evenfire-ai/evenfire/blob/b4d0733cc221fe1a2a5e6dff422d24a31b959e9c/host-context-controller/src/k8sClient.ts#L3207-L3213) schedules a McpServer retry on every pass and re-reconciles the whole fleet at the backoff cap.
- `runtimeRepairPending` never clears and `completedRuntimeRevision` never catches up. [`shouldRequestNetworkPolicyRecovery`](https://github.com/evenfire-ai/evenfire/blob/b4d0733cc221fe1a2a5e6dff422d24a31b959e9c/host-context-controller/src/k8sClient.ts#L3069-L3104) therefore always returns `work-pending`, and every identical watch recovery starts a full NetworkPolicy pass again.

That is the cost this PR removes. Before the PR, the failure was reported through status and the lane finished. Retrying the failed server is reasonable, but it shouldn't block certification for unrelated servers or force a fleet-wide retry.

**Checked, no issues found:** revision-based skip decision, missing revocation callback treated as aborted, managed-snapshot removal on uid change, orphan and pending-name cleanup, `reconcileDelete` obligation retention, BOOKMARK handling, harness restore patch, Minikube overlay patch merge, proxy command/ack protocol, and the new `mcp-server:mcp-proxy` readiness entry.

jozer-rami:
I pushed fixes for the three review findings to this branch:

- `2f012beb` **fix(e2e):** `hcc_netpol_timer_arm_epoch` now matches the JSON marker (`"msg":"[K8s] NetworkPolicy periodic resync enabled"` plus an exact `"intervalSeconds":N`).
- `cfddc53c` **test(hcc):** the two negative "inventory authority lost" assertions now use `.includes(...)`, so they can fail against the structured log output again.
- `e3e32253` **fix(hcc):** when a pass runs every effect under its certificate but some servers are still incomplete, the McpServer retry now reconciles only those names, pinned to that inventory revision.
  - The retry widens to a full pass if the inventory moves or a full retry is requested.
  - `fullReconcile` gets an optional `scope`: it still receives the whole inventory, so out-of-scope servers are never treated as orphans.
  - Identical recovery still requests a pass while a failure is pending, so the A-T7 contract is unchanged.
  - New tests: a failing server next to a healthy peer retries alone, certifies once repaired, and identical recovery skips again. There are also widening tests and a scoped orphan-cleanup unit test.

**Tested:** the `host-context-controller` Vitest suite passes (110 files, 2376 tests), `tsc --noEmit` is clean, and Prettier is clean. The four new tests fail without the source change. Minikube and e2e lanes were not run.

jozer-rami (approved):

PR #624, opened 2026-09-14
chore: make session title server-authoritative across devices

zach88:
## What

Make a session's name server-authoritative so it is the same on any device. This covers both the auto-title derived from a session's first turn and a user's manual rename, and it removes the `Remote · <id>` / `Chat <id>` placeholder a fresh client with no local cache used to fabricate for every session.

The change spans four services (control-api, rpc-proxy, mcp-host, desktop-app) in two phases:

- Phase A (read path): mcp-host materializes the auto-title into the existing `sessions.title` column and projects it on the sessions-list wire; the desktop app reads it and merges it with local state.
- Phase B (write path): a user's rename propagates over RPC (desktop -> rpc-proxy -> mcp-host) behind a dedicated `host:session:write` scope, applied optimistically and reconciled across devices.

## Why

Today the custom name a user gives a session is local-only (rename by IPC, never over RPC), so it does not exist on another device; and mcp-host always writes `sessions.title` as null and never projects it, so a cache-less client shows a placeholder for every session. This makes the session name consistent across devices without a broadcast or SSE (the echo to other devices is the next `listSessions` poll).

## How

- Single column, no migration: `sessions.title` is reused. Turn 1 writes `title = COALESCE(title, @autoTitle)` (idempotent, a turn-1 retry keeps the first title, turn 2 never overwrites); an explicit rename overwrites unconditionally.
- Redaction order is fixed: the first user input is redacted in full before the title is derived, because truncating first would break literal secret matching. The list projection re-redacts as defense in depth.
- Title sanitization is unified (`normalizeTitleText`: NFC, whitespace fold, strip invisible control/format code points, collapse, trim) and shared by the auto-title helper and the rename validator (which adds a non-empty check and a 120 code-point cap). The client also sanitizes titles on read, since it does not trust a title coming from another device or an older host.
- The rename endpoint resolves ownership by the edge-provided user id, returns a uniform 404 for a missing or foreign session (no cross-user existence oracle), rejects channel sessions structurally, and never logs the raw title (only its length). It is no-op when the sanitized title is unchanged, so an idempotent rename never enqueues a write on the shared persist queue.
- Client precedence is a small pure decision module (`resolveSessionTitle`): the server wins for a cached chat when it has a title and there is no pending local rename; a pending local rename (in-flight or offline) wins over the server until it is confirmed. A pending-rename queue drives the RPC: 200 confirms, 404 keeps the optimistic title and retries after turn 1 (no rollback), a 4xx/403 rolls back and toasts, a network or 5xx failure queues offline and retries when connectivity returns.
- The rename token carries only `host:session:write` and is not wake-eligible, and is kept out of the finite-operation read tokens.

### Why this is not split

The spec planned this as staged PRs. It is shipped here as one branch because the two phases share the same column and projection and the read/write paths are one atomic feature; splitting would leave a half-feature (a wire field no client reads, or a write path with no read path). The required rollout order still applies at deploy time (see Risks).

## Testing

Per service, on this branch:

- control-api: scope suites green (default grant per role, the normalizer keeps the scope, a write-only token passes the host-access gate).
- rpc-proxy: typecheck clean; full suite green including a 14-case passthrough test (401 without auth, 403 without the scope, upstream 400/403/404/200 echoed, edge headers forwarded and client Authorization not, raw title not logged, timeout signal set).
- mcp-host: title suites green (auto-title truncation with a secret across the cut, unicode, COALESCE idempotency, dual-store parity asserting the projected title; rename validation, uniform 404, channel-not-renamable, no-op, no raw title logged). Two pre-existing tsc errors unrelated to this change (see Risks).
- desktop-app: typecheck (main) clean; full suite green (2287 tests). Property-based tests on the precedence resolver; pending-queue tests for the 200/404/4xx/offline transitions and the concurrency guards, asserting the shown title.

Invariants covered by executable tests on this branch:

- A pending manual rename (in-flight or offline) wins over the server title until confirmed, and a stale success does not drop a newer rename's protection.
- The auto-title never overwrites an existing title (COALESCE).
- A secret in the first input is redacted before the title is truncated.
- A missing and a foreign session return the same 404.
- A concurrent poll does not fire a duplicate rename PATCH.

## Risks / Follow-ups

- Required rollout order: control-api, then rpc-proxy, then mcp-host, then desktop. `normalizeRequestedScopes` silently drops unknown scopes, so a client ahead of control-api would get an opaque 403. The contract is additive and backward-compatible.
- **Resolved after review (this branch).** The desktop pending-rename queue is now cleared at the user-identity seam — on logout / user change, keyed on the authenticated user id — while a same-user team-switch preserves it so the rename still syncs (sessions are per-user server-side, so clearing on a team-switch would let the server title revert the user's own optimistic rename). The pending-rename retry is now bounded by a per-entry budget that counts only genuine network/5xx failures: a 404 awaiting server-side materialization no longer consumes it and is retried only when a poll lists the session, so a deterministic failure no longer re-issues the PATCH forever.
- **Out of scope — `CLERUM_SESSION_STORE=dual`.** The auto-title persists in every store mode, but an explicit rename is not persisted under `dual`: `DualConversationStore` forwards `persistTurnStart` / `persistSessionUsage` to both backends and does not implement `persistTitle`, so the rename stays in the memory store and is lost on cold-load or on a flip to `sqlite`. This is intentionally not addressed here: `dual` is a transitional shadow-write canary mode that is being deprecated, and the authoritative `sqlite` store persists the rename correctly. (The same-shaped gap already exists for `persistModelSelections` on `dual`.)
- Pre-existing broken builds inherited from the base branch, unrelated to this change: control-api (`modelsDevClient.ts`, TS2741) and mcp-host (`registryCore.ts` / `workflowService.ts`, plus a `registry.test.ts` provider-count drift). This change adds zero new type errors. These should be fixed separately so CI is green.
- Follow-ups (non-blocking): rename status is classified by parsing the status code out of the error message because the structured status does not cross the IPC boundary; the auto-title secret-pattern list does not cover a few personal-token shapes (JWT, `sk-proj-`, `sk-ant-api03-`, `github_pat_`) on the new list surface (accepted residual, mitigated by double redaction). A residual corner case of the bounded retry above: a session that IS listed but whose rename keeps returning 404 (a backend list-vs-rename inconsistency) re-PATCHes once per poll unbounded — pathological and self-resolving, accepted rather than capped, since capping the in-flight state is exactly what would drop a legitimate slow-to-materialize rename. A pre-existing debt, not touched here: `rpc:setHostModel` / `rpc:getHostModels` accept renderer-supplied host refs.

No database migration, no CRD.


josue-tech-94:
@zach88  Currently **blocked**. 

* **R1-H1 — Desktop pending rename crosses auth scope**

  * File: `desktop-app/ui/src/hooks/domain/useChatListController.ts`
  * Clear pending rename state when the authenticated user/team scope changes, and make sure an in-flight completion from the old scope cannot affect the new one.
  * Add a regression test covering: offline rename → logout/user or team switch → same session IDs → confirm the old title is neither rendered nor PATCHed with the new user's token.

* **R1-H2 — Dual conversation store loses explicit renames**

  * File: `mcp-host/src/core/conversation/persistence/dualConversationStore.ts`
  * Implement `persistTitle` fan-out for the dual store so an acknowledged rename is durably written to both backing stores.
  * Add a regression test that renames in dual mode, then cold-loads/recreates from SQLite and verifies the explicit title is preserved.

Please also update the PR description:

* remove the pending-rename scope leak and dual-store rename loss from the “accepted/non-blocking follow-up” list;
* describe them as resolved behavior once fixed;
* preserve the deployment order: `control-api → rpc-proxy → mcp-host → desktop`.

josue-tech-94 (approved):

PR #630, opened 2026-09-14
fix(control-ui): restore provider-grouped LLM model catalog

josue-tech-94:
## Summary

- Restore provider-family grouping in the `/llm-models` Catalog and Discovery Review.
- Add a domain-neutral grouped-table disclosure primitive to
  `@clerum/frontend-components`.
- Balance grouped provider summaries independently from their expanded child-table schemas.
- Split the shared table implementation into cohesive internal modules while keeping root imports
  stable.
- Record the owner-decision gate and implementation rules for grouped table datasets.

## Why

The standardized flat LLM model table repeated provider identity across a large catalog and became
difficult to scan. The approved presentation groups models by provider family without rolling back
the standardized table shell, filtering, sorting, states, warnings, or row actions.

## LLM Models behavior

- Catalog and Discovery Review provider rows are collapsed by default and expand independently to
  semantic model rows.
- Provider summaries show the provider identity plus total, enabled, stale, and matching counts.
- Both tabs balance Provider, Models, and Availability/Details across the available width while a
  fixed Actions lane remains right-aligned.
- Model-count text starts on the same vertical line as the Models header.
- Expanded groups render their exact model-level columns in independent nested tables, so adding or
  removing child columns cannot move the provider headers.
- Collapsed provider summaries use provider-level headers; expanded groups expose exact child-model
  headers and Catalog sorting controls.
- Search and concrete provider filters expose matching child models without stealing focus.
- Clearing search preserves the operator's expansion context.
- Child rows retain model sorting, credential details, warnings, states, and edit/delete actions.
- Discovery Review retains bulk selection and per-model enable controls within expanded groups.
- Model ascending is the visible initial child-row sort; provider groups remain deterministic.
- Catalog and Discovery Review are the owner-approved ordinary table row-expansion cases. All other
  existing tables remain flat.

## Shared frontend-components architecture

- `GroupedTableBody` owns controlled disclosure, semantic table rowgroups, unique accessible DOM
  relationships, and child-row visibility without knowing any LLM or provider concepts.
- Its `summaryCells` mode renders native table cells and validates that their spans cover the full
  grouped row, preventing silent header drift as grouped layouts gain or remove columns.
- Its opt-in nested-child-table mode separates provider-summary and child-row column schemas while
  retaining semantic table headers, rows, and ARIA relationships.
- The grouped `DataTable` variant shares remaining width across non-action columns and reserves a
  stable right-aligned Actions lane.
- The labelled first-cell button remains the sole disclosure control; non-disclosure cells do not
  add pointer-only interaction.
- The legacy full-width `summary` mode remains supported for existing consumers.
- `@clerum/frontend-components` keeps its generic package name and stable root import seam.
- The thin root barrel exports a table module tree split across shell, primitives, sorting,
  truncation, row actions, grouped rows, utilities, and public types.
- A compile-time public API contract guards the existing exports and grouped-table types.

## Table presentation rule

An owner/product presentation decision is required when a dataset is known to be unusually long,
or when a meaningful candidate column has at least three distinct values and every one of those
values occurs at least four times. The gate does not choose grouping automatically; the owner may
retain a flat table or select grouping, tabs, pagination, virtualization, or another suitable model.
Approved grouped summaries with column-shaped values must use the shared native summary-cell API.
When child rows have a different schema, they must use the shared nested-child-table mode instead of
forcing parent and child columns into one physical grid.

## Testing

- `packages/frontend-components`: `npm test`; `npm run build`
- `control-ui`: `npm test` (183 files, 1965 tests); `npm run build` (60 routes)
- Follow-up review nits: Discovery child-cell spacing, search-focus coverage, and no-op
  expansion updates are validated
- Follow-up grouped-table checks: 3 focused files, 47 tests passed; TypeScript passed
- `profile-ui`: `npm test` (100 tests); `npm run build` (12 routes)
- Style-rule contract tests: 8 passed
- Style rules: 0 errors; one pre-existing Desktop test warning
- Strict style rules: 0 errors and the same warning; exits nonzero because strict mode promotes it
- Changed-file Prettier check
- Docker local-package materialization contract
- `git diff --check`
- Historical fail-before proofs for restored grouping, review fixes, native summary-cell spans, and
  independent provider/child schemas
- Final code review SAFE, security review SAFE, and collateral review CLEAN
- Exact-head GitHub CI and CodeQL: all 38 checks passed

## UI
image

image


zach88:
Took a deep pass through this one — the big refactor (splitting the 616-line `index.tsx` into cohesive table modules + the new `GroupedTableBody` primitive) held up really well. Verified the split is faithful byte-for-byte, the public API contract lost nothing, and no consumer breaks. The few things I found are all in the control-ui consumer, and none of them block.

🟡 **Approve with nits**

## Optional (non-blocking)

This PR can merge without these — they're follow-ups / author's call.

1. **[R1-M1] Deleted CSS rule regresses padding in `LlmDiscoveryPanel`** — The diff drops `.cu-llm-model-row > .cu-llm-provider-cell { padding-left: var(--cu-space-2) }` from `control-ui/app/globals.css`. That rule wasn't LLM-catalog-specific: `LlmDiscoveryPanel/index.tsx:376` uses `cu-llm-model-row` on the `<tr>` and `:386` uses `cu-llm-provider-cell` on its **second** `<td>` (the first is the selection checkbox), so the `td:first-child` rules don't cover it. After this PR nothing gives that provider column its left padding anymore.
   - Repro: open `/llm-models` → Discovery Review; the Provider column sits flush against the checkbox, missing the `--cu-space-2` indent it had before. Doesn't repro on the base branch (the rule existed there), so it's from this change.
   - Suggested fix: keep the padding scoped to the panel that still needs it, e.g. `.cu-llm-review-table .cu-llm-model-row > .cu-llm-provider-cell { padding-left: var(--cu-space-2) }`, or move the rule next to the discovery panel styles. Worth double-checking the scoped selector only hits the discovery panel and doesn't reintroduce padding in the grouped catalog.

2. **[R1-L1] "without stealing focus" invariant has no test** — The PR body promises search/filter "expose matching child models without stealing focus". The code is actually correct here (I checked with a throwaway test — focus stays on the search input while groups auto-expand, since the effect only does `setState` and `GroupedTableBody` doesn't auto-focus), but no test in the diff asserts it. `LlmModelTable.tsx:207-216`. Cheap to lock in: in the search test, `input.focus()` + `expect(input).toHaveFocus()` after the `fireEvent.change`.

3. **[R1-L2] `setExpandedProviders` returns a fresh `Set` on every keystroke** — `LlmModelTable.tsx:209-215`: the updater always returns a new `Set` even when no provider was added, and the effect runs on every `normalizedSearch` change, so you get an extra render per keystroke when the expanded set didn't actually change. Purely wasted render work — correctness is fine. Optional: return `current` when nothing new was added.

## What checked out clean

- **Refactor fidelity** — every moved symbol (`DataTable`, `TableRow/Cell/HeaderCell/StateRow`, `DataViewHeader`, `TableSearch`, `TableViewport`, `RecordList/Row`, `RowActionMenu`, `TruncatedText`, `compareSortValues/stableSortRows/useTableSort`) is byte-for-byte identical to base. No export dropped or renamed. Package `tsc --noEmit` exits 0.
- **Public API contract** — `public-api.test.ts` is a real compile-time guard, not a vacuous one: dropping the `GroupedTableBody` export makes `tsc` fail with TS2724. Covers all 16 runtime exports + 9 types. No consumer (control-ui, profile-ui) imported anything this PR stopped exporting.
- **`GroupedTableBody` end-to-end** — all 7 required + 3 optional props passed correctly from `LlmModelTable.tsx:380`, no missing props, no reliance on defaults the primitive doesn't declare.
- **Accessibility** — unique ids via `useId()` + `groupId` (no collision across groups/tables), aria-controls/expanded/labelledby paired correctly, disclosure fully controlled, two `<tbody>` per table is valid rowgroups. Chevron rotation preserved via `[aria-expanded='true'] .eft-table-group__indicator`.
- **Security** — no raw-HTML sinks, model/provider data rendered as escaped React text children, the Credential column shows a presence label (`API key` · `Subscription`), never the actual secret. Version bump is the internal package only, no new third-party deps.
- **Suites** — control-ui `LlmModelTable.test.tsx` + `FrontendComponents.test.tsx` = 35/35 green; contract + props = 38/38.

zach88 (approved):

Issue #627, opened 2026-09-14
fix(codex): preserve access to approved connector and plugin tools


alfredolopez80:
## Problem

An Evenfire agent using `codex-subscription` successfully connects to approved MCP services and loads their tools, but the provider adapter removes every MCP tool before authorizing/sending the LLM request. The model therefore cannot use the approved connectors. This affects tools exposed by plugin services through the same MCP registry.

These services execute inside Evenfire. OpenAI receives function definitions and model-visible tool results; connector connections, credentials, authorization and execution remain in Evenfire. The fix must preserve that architecture.

## Confirmed diagnosis

Analyzed refreshed `origin/dev` at `98840fbfe675f5dca8fdf312c7267c1ffb42f65a` on 2026-09-14.

- `mcp-host/src/llm/codexSubscription.ts`: `selectCodexAdvertisedTools()` unconditionally keeps only names considered native and slices the result to 32. Even a single MCP tool is dropped.
- `packages/llm-provider-attempt-contract/index.cjs`: `LIMITS.maxTools = 32` rejects larger definition arrays. It also limits assistant tool-call arrays. This is an Evenfire contract bound; the reviewed evidence does not establish an upstream limit of 32.
- `control-api/src/services/llmProviderAttemptAuthorizer.ts` and `codex-llm-proxy/src/codexTransport.ts` validate that shared contract. Removing only the adapter filter therefore makes larger catalogs fail validation.
- `codex-llm-proxy/src/codexTransport.ts` already maps definitions to function tools and transports function calls/results. It does not need to connect to MCP services.
- Existing provider tests explicitly require MCP omission and must be replaced with tests for the intended product behavior.

History: `d67f89e7b` introduced the bounded contract; `01c25ab4b` introduced the native-only filter to let large catalogs authorize. The reviewed history/documentation supplies no specific justification for choosing 32 beyond bounded requests.

## Reproduction and evidence

Sanitized runtime comparison: the affected agent loaded 83 tools from one approved connector and 90 MCP tools overall; the adapter emitted `advertised 32 native tool(s) of 126 offered (MCP omitted, cap 32)`. A comparison agent using Z.ai selected tools from the same service.

Local execution of the actual functions from `origin/dev` confirmed:

| Input | Actual result |
| --- | --- |
| 1 native + 1 MCP definition | Only native survives |
| 1 MCP definition | Zero definitions survive |
| 36 native + 90 MCP definitions | First 32 native definitions survive |
[...]

PR #632, opened 2026-09-14
fix(codex): preserve approved connector access with selective discovery

alfredolopez80:
## Problem and change

Codex Subscription agents could connect approved MCP services, but the provider adapter discarded MCP definitions and truncated native tools to 32. This change preserves the final presentation through authorization and dispatch, removes the arbitrary 32-definition limit, and retains request-size limits and an independent bound on returned calls before execution.

Direct mode preserves compatible definitions. Automatic discovery keeps large MCP catalogs local and exposes bounded search, individual schema description, and invocation through the existing Evenfire controls. Related issue: #627.

## Live validation: real subscription, 83 MCP tools

Tested on September 14, 2026 using an isolated, branch-owned ARM64 Minikube and the built Evenfire Desktop 0.8.0 application.

| Item | Verified result |
| --- | --- |
| Agent and provider | `chatllm`, `codex-subscription`, `gpt-5.6-luna` |
| Deployed platform revision | `c345ef5dc6e86dbe8286532d0870b9f7cd0cc9b3` |
| Worktracker / EvenTasks | Real PostgreSQL, API, UI and MCP, built locally for ARM64; 75 tools |
| GitHub MCP | Real connector built locally for ARM64; 8 tools |
| Effective MCP catalog | 83 tools, successfully listed from the actual agent pod; no remaining pagination |
| Host presentation logs | `nativeCount=39`, `mcpCount=83`, `presentedCount=39`, `strategy=discovery` |
| Upstream | Real operator-authorized ChatGPT subscription through the normal Codex proxy; no simulated upstream |

The agent and both MCP services were Ready. Only Worktracker and GitHub were assigned to this test context. Wikipedia was excluded from the agent's catalog, and EvenSign was excluded from testing.

### End-to-end journey

1. In Desktop, the agent was asked to consult GitHub issue #627 and PR #632 and create a Worktracker task containing their problem summary, links and acceptance criteria. The request used ordinary language without discovery commands. It explicitly excluded GitHub writes.
2. The trace showed selected schema descriptions, GitHub `search_issues`, `list_repo_issues` and `get_pull_request`, followed by Worktracker operations. A real GitHub approval was presented in Desktop; approval resumed the task. Native HTTP and spillover tools also participated.
3. The agent created **INIT-1** and **TASK-1**. A subsequent user request assigned TASK-1 to the test user.
4. TASK-1 was opened in the embedded EvenTasks UI. The task, source links, acceptance criteria, Backlog state, assignee and creation/assignment activity were visible.
5. An independent, read-only MCP `workitem_get` confirmed both persisted records, TASK-1's parent relationship to INIT-1, its human assignee, and both exact GitHub links. This verification did not create or modify either record.

**Result: PASS for this bounded real-subscription journey with 83 MCP tools across two services.** Codex selected tools, Evenfire executed them, and the response referred to independently verified application state.

### Limitations observed

- The creation trace contained 21 tool entries; the assignment follow-up contained 4. These UI counts are not a separately measured upstream request count.
- One spillover read requested 12,000 bytes from an 11,417-byte result and was rejected. The agent recovered and completed the task. This was not an error-free first attempt.
- The model's earlier prose said seven GitHub tools while listing eight. The authoritative MCP results were 8 + 75 = 83.
- The run presented 39 native/bridge definitions while deferring the 83 MCP schemas. It does not establish an unlimited upstream maximum or validate direct presentation of all 83 MCP schemas.
- No subscription-quota saving or controlled latency improvement is claimed. The conversation also contained catalog enumeration, manual approvals and an assignment follow-up.
- Application actor labels alone do not certify every identity-isolation invariant.

## Other evidence and outstanding gates

- At `c345ef5dc`, Codex T0 passed 11 groups and PostgreSQL T1 passed 164/164 tests across 35/35 files. T2 stopped at NP08 because profile metadata retained its initial dirty state.
- At `8568d2b5a`, lifecycle integration passed 45/45 tests.
- At `1f1dbc4c9`, upstream-simulator protocol tests passed 15/15 and workflow-fixture tests passed 4/4. At `7493d4cf7`, the strengthened workflow-fixture suite passed 5/5.
- Automated Playwright, final exact-head T2, the wider authorization/regression matrix, and remaining direct-mode, scale and efficiency checks are still pending. The synthetic setup implementation is now present; its complete runtime execution remains pending.

The successful live run applies to the deployed revision stated above, not automatically to every subsequent PR commit. It must not be presented as final certification of the current head or as proof of every planned 150/250-tool scenario.

The PR was created as draft and marked ready by the user. No merge or production deployment is authorized by this validation. Deploy updated contract consumers before senders that advertise more than 32 definitions; roll back senders before validators.

## Review follow-up (September 15, 2026)

Commit `6fa2d0c83` addresses the confirmed local findings from [jozer-rami's review](https://github.com/evenfire-ai/evenfire/pull/632#pullrequestreview-5207814364):

- **M1, resource isolation:** selected MCP schema compilation and argument validation now run in bounded disposable Node workers. Validation remains non-coercing and rejects unsupported schemas. The Host rechecks the live schema and arguments after awaiting validation and before dispatch. The two-second deadline, input/depth limits and worker concurrency cap fail closed. Worker startup adds latency; V8 memory limits are not an OS-level total-RSS guarantee.
- **M3, transport evidence:** the proxy conformance test sends 250 definitions and asserts the complete mapped `body.tools` array, its ordering and `parallel_tool_calls === true`.
- **M5, certification ordering:** the orchestrator re-runs its final exact-head preflight after Health/Playwright, before publishing completion, since journeys can mutate their owned environment.

Focused validation: MCP Host typecheck passed; three selected suites passed **74/74 tests**; proxy conformance passed **22/22 tests**. An isolated execution of the orchestrator control flow confirmed that post-journey preflight failure prevents completion and successful post-journey verification permits it. This is not a live T2 verdict. CI for this commit was still running when this section was written.

**B1 remains a release blocker:** static inspection confirmed that the deployment pipeline applies consumer and sender-producing changes together before its readiness gates. Existing proxy rollout waits occur after that combined apply. Ordered rollout and inverse rollback are being handled separately in infrastructure; documenting the order does not enforce it. No deployment was performed to validate the transition.

**M2:** strict validation of explicitly invalid presentation configuration is retained; missing values already have defaults. A legacy truncation mode would reintroduce the original loss of approved tools and was not added.

**M4, transport compatibility (`299762dac`):** the proxy now preserves canonical names in authorization/hash/history while mapping names outside the conservative standard function-name envelope to reversible, collision-checked transport aliases. Active definitions and historical assistant calls use the same map; incoming calls are restored before Host dispatch and call IDs are unchanged. Existing compliant names remain unchanged. Focused name-map and transport roundtrip tests passed **29/29**, with typecheck on Node 24. This removes reliance on accepting punctuation or overlong function names without dropping any tools. It does not claim that rejection by the specific Codex endpoint was empirically reproduced; real-endpoint interoperability remains pending.

Final exact-head T1/T2, automated browser E2E preparation/execution and the remaining real-endpoint compatibility evidence are still pending. No closing audit or merge readiness is claimed.

## Latest review remediation status

Implementation updated through `09bb5f1ef719ec2513f9da3aac1a35ecbbab6038`, following review comment #5678404717.

- H1 / M-2 / M-3: strict selected-schema checks are scoped to Codex provider chains; supported dialect variants normalize; bounded queueing and distinct safe causes replace immediate fifth-call rejection; current identical inputs reuse successful validation.
- H2 / M-8: certifying preflight verifies live proxy and Control API production image baselines and rejects fixture images, environment/run markers and the OAuth activation file. Inherited journey leases and all four approved-tools Make targets have hermetic coverage.
- M-1 / M-9: opaque canonical names stay within request-byte bounds; transport tests cover unknown aliases, history-only names and registered canonical echoes.
- M-4 and cheap lows: accurate direct/discovery guidance; exact count/UTF-8 thresholds; positive dispatch witness; safe logger fallback; disabled login traces; rendered-state negative witnesses; original-error preservation; import-time presentation configuration tests; and updated runbook.
- M-5 implementation: the seed and isolated external OAuth provider are present. Seven-image acquisition records three exact derived-base bindings. The fixture imports the actual Control API and keeps application flows real. The full image/build/browser execution remains pending.
- M-6 implementation: Kubernetes resources use UID/version/run journals and dependency-ordered deletion. Fresh identity IDs and UI-created subscription IDs are recorded durably. Cleanup refuses ambiguous ownership, uses normal subscription revocation, retains audit/tombstones, deletes owned identities transactionally, and requires successful runtime allowlist publication. Failed publication and API restoration can be retried. Already-removed workflow grants are accepted without duplicate revoke events.
- M-7: the actual image builder is exercised through the existing fake-runtime harness across sequential partial builds, all three derived bindings, invalid/stale state and late inventory failures, preserving the previous valid manifest byte-for-byte.

Validation on Node 24: integrated fixture group 115/115; after the final workflow cleanup adjustment, affected preparation/restoration/connection/identity suites 51/51, including 17 identity tests. Counts overlap. Configuration-import tests 8/8; hermetic builder, mutation-boundary, proxy/API runtime-guard and T2 scenario tests passed. Playwright typecheck passed. The static browser audit passes for the changed journey/helper; its non-browser wait-rule findings and two preexisting positional selectors in the shared page object are not claimed as a global audit pass.

Low follow-up issues: #637 (HTTP envelope bytes), #638 (JSON depth), #639 (plugin build/installer wiring).

**Still pending:** closing audit and its surviving findings, CI on the new head, fixture image acquisition, automated browser execution and exact-head T0/T1/T2/current real-subscription certification. B1 remains a separate infrastructure release dependency. The agreed order is closing audit plus green CI, then the expensive local certification lanes. No merge readiness, completed runtime validation or measured quota savings are claimed.


jozer-rami (commented):
## Deep review: blocker and major findings (head `7a4f9496`)

The core fix looks right: bridged calls keep approval/guardrail/identity binding and live revocation checks, the request hash covers every definition, and the returned-call bound is enforced in the contract, proxy, and host. The items below should be resolved before merge. Each was verified against the code at this head.

### Blocker

**B1. Validator-before-sender rollout order is not enforced, and skew fails every Codex tool turn, not only large catalogs.**
- `mcp-host/src/llm/codexSubscription.ts` now sends every presented definition. In `auto`/`discovery` the host still presents all natives plus the bridge tools (this PR's live run logged `presentedCount=39`), so typical Hosts exceed 32 even without MCP.
- A contract consumer built before this PR (control-api authorizer or `codex-llm-proxy`) rejects with `tools exceed 32`. mcp-host surfaces only `authorize failed with 400` (`providerAttemptAuthorizer.ts:100`), `classifyError` makes it non-retryable (`codexSubscription.ts:169-175`), and the old `[Codex] advertised …` diagnostic is gone.
- control-api, codex-llm-proxy and the Host image are deployed and rolled independently (per-image promotion, HCC selects the Host image on its own schedule), so skew occurs during canaries, partial rollouts, and single-service rollbacks. The PR body sentence is the only safeguard.
- Suggested fix: ship in two releases (relax validators first, remove the sender cap next), or keep a configurable sender cap (default 32) for one release, or have control-api advertise the capability before the host sends >32.

### Major

**M1. New host-side MCP argument validation applies to all providers, unflagged, and fails closed** (`toolRegistryAdapter.ts:18-51, 154-179, 197-206`; `toolUseLoopToolBatch.ts:224-228`).
Reproduced with ajv 8.20 and the options in this PR:
- `$schema: http://json-schema.org/draft-04/schema#` → compile throws `no schema with key or ref`.
- No `$schema` + draft-07 tuple `items: [...]` → compiled as 2020-12 → `schema is invalid`.
- `{"n":"5"}` against `integer` → rejected (previously reached the server).

When compile fails, the tool returns "schema is invalid or unsupported" on every call, including on OpenAI/Anthropic/Bedrock Hosts where it works today. Validation is also synchronous on the event loop with Ajv's default regex engine; a connector `pattern` such as `^(a+)+$` shows exponential validation time with input length, so one vulnerable schema can stall the Host.
Suggested fix: split into its own PR or scope it to bridged calls; on compile failure let the server validate; use `code: { regExp: re2 }` or skip `pattern`.

**M2. The presentation setting cannot restore pre-PR behaviour and throws at import** (`toolPresentationPolicy.ts:4-17`, `config.ts:828-829`).
`CODEX_TOOL_PRESENTATION` / `CODEX_TOOL_DISCOVERY_BYTES` are read at module load and throw on invalid values, so a ConfigMap typo crash-loops agent pods on restart. They are not in any manifest, are cluster-wide only, and none of `auto|direct|discovery` reproduces the capped request, so B1 cannot be mitigated without a redeploy.
Suggested fix: log and fall back to the default instead of throwing; add a legacy/cap mode for the transition release.

**M3. No test asserts the proxy forwards tool definitions upstream** (`codex-llm-proxy/test/codexTransport.conformance.test.ts:613-667`).
The only proxy test sending `tools` checks `instructions`, `store`, `tool_choice`, `input` but never `body.tools` or `parallel_tool_calls`. Mutating `codexTransport.ts:282-289` to send `request.tools.slice(0, 32)` or drop `payload.tools` leaves the proxy suite green; only the manual Playwright simulator would catch it.
Suggested fix: conformance test with 250 definitions asserting `body.tools` equals the mapped `request.tools` (order and last entry) and `parallel_tool_calls === true`.

**M4. MCP tool names are forwarded verbatim upstream in direct presentation** (`manager.ts:823`, `deferrableToolController.ts:95-100`, `codexTransport.ts:283-288`).
In `auto`, catalogs at or below the thresholds are presented directly. The contract accepts `^[A-Za-z0-9._:/-]{1,128}$`, and MCP tool names may contain `.`; `serverName__toolName` can also exceed 64 characters with generated server names. OpenAI function APIs commonly require `^[a-zA-Z0-9_-]{1,64}$`. **Not verified** for `/codex/responses`, and the live run only exercised discovery. If the stricter rule applies, one such definition fails the whole turn, and the bridge's rewritten real name in tool-result `messages[].name` (`toolUseLoopToolBatch.ts:152-156`, `codexSubscription.ts:312`) keeps failing later turns.
Suggested fix: measure the upstream rule; if needed, apply a stable reversible alias (≤64 chars) at the provider boundary or route non-conforming names through discovery.

**M5. T2 lease is passed to journey commands after the final preflight** (`scripts/minikube/t2.sh:408, 452`; `run_final_preflight` at `:530`).
Health and Playwright children now inherit `T2_SKIP_LOCK` + token and can run lease-guarded mutations (e.g. swapping `codex-llm-proxy` to a fixture image, rewriting `.image-manifest.json`) after the final exact-head preflight, while `complete PASS` only re-runs `t2_process_check`. This weakens the rule that the T2 verdict comes from the final exact-head preflight.
Suggested fix: re-run the final preflight before PASS, or do not pass the lease to journey commands.

---
Minor items (partial-stream tool-call drop, 128-message cap, discovery guidance under direct strategy, Codex-as-fallback changing a non-Codex primary's presentation, 1 MiB envelope mismatch, test gaps, scope split) can follow separately.

alfredolopez80:
## Multi-agent review: PR #632 at head `299762dac`

| | |
|---|---|
| Repo / PR | `evenfire-ai/evenfire` #632 |
| Title | fix(codex): preserve approved connector access with selective discovery |
| Base ← head | `dev` ← `fix/627-codex-approved-tools` @ `299762dac09e50236ce68a99a8913e17137cca4b` |
| Author | @alfredolopez80 |
| Size | 97 files, +9058 / −469, 25 commits |
| Findings (deduplicated) | **Blocker 1 · High 2 · Medium 9 · Low 20+** |

**How this was produced:** eight independent read-only reviewers ran in parallel on a clean worktree at the PR head:
- mcp-host presentation/discovery
- mcp-host schema validation/dispatch
- proxy + contract + rollout
- minikube/CI/Makefile
- E2E scripts/fixtures/Playwright
- security (full diff)
- tests/coverage (suites + mutation testing)
- docs/simplicity/architecture/performance

Each finding was checked against the code at `299762dac`. **REPRODUCED** means a command or repro was run and its output observed. **READ** means it was confirmed by code tracing only, so confidence is lower. No files in the PR branch were modified, and all repro copies lived outside the repo and were deleted.

The overall assessment is that the core fix is sound:
- Bridged calls keep approval, guardrails, identity binding and the live revocation check.
- The 32-definition cap is gone end to end (mutation-verified in host, contract and proxy).
- The returned-call bound is enforced in three places.
- `ToolNameMap` aliases are collision-resistant.

What remains is rollout safety, a cross-provider validation regression, a T2 certification gap, and a large amount of harness scope that is not runnable yet.

---

### Blocker

**B1. Validator-before-sender rollout order is still not enforced** (prior B1: **NOT FIXED**; the author agrees it is a release blocker)
- **What the docs say:** `docs/architecture/codex-subscription-transport-contract.md:94-97` documents the required order, and `:135` states that the order is not enforced.
- **What the pipeline does:** `build-publish.yml` builds consumers and the mcp-host sender in one matrix. `pre-gate-incremental.sh` applies everything and then waits for rollout. Nothing makes validators roll out first.
- **Failure scenario:** a new Host talks to an old control-api authorizer or an old `codex-llm-proxy`.
  - The Host presents 39+ definitions even with no MCP servers (the live run logged `presentedCount=39`).
  - The old consumer rejects the request with `tools exceed 32`.
  - The Host classifies this as non-retryable, so almost every Codex tool turn fails during a canary, a partial rollout, or a single-service rollback.
- **Doc gap:** the compatibility matrix (`:89-92`) omits two cases:
  - new host + new authorizer + old proxy: the ticket is minted, then the proxy returns 400;
  - mixed-version replicas: rejections become nondeterministic.
- READ (pipeline trace) + REPRODUCED (contract parse).

### High

**H1. Host-side MCP argument validation makes previously working tools unusable on every provider, with no flag** (prior M1: **PARTIALLY FIXED**; flagged by 4 of 8 reviewers)
- **What changed:** `toolUseLoopToolBatch.ts:224-228` and `toolRegistryAdapter.ts:115-129,175-184` apply to OpenAI/Anthropic/Bedrock hosts too. Before this PR, `safety.validateToolParams` let every MCP call through.
- **Fixed part:** ReDoS no longer blocks the event loop. `^(a+)+$` with `'a'*40+'!'` returned `false` in 2002 ms while the loop kept ticking, and `terminate()` freed the slot.
- **Remaining regression** (REPRODUCED with Ajv 8.20.0 on Node 24.18). These return `false` or throw, so the tool fails on every call:
  - `$schema` draft-04 or draft-06;
  - **`https://json-schema.org/draft-07/schema#`** and **`http://json-schema.org/draft/2020-12/schema`**: these are only alternate spellings of dialects the PR says it supports;
  - draft-07 tuple `items: [...]` without `$schema`, and draft-04 `exclusiveMinimum: true`;
  - remote `$ref`, and PCRE-only patterns;
  - argument strings over 256 KiB (for example a file-write tool), depth > 64, arrays over 10k elements;
  - schemas over 10k nodes (800 small properties).
- **Why it matters:** the model cannot recover by retrying.
- **Suggestion:** normalize equivalent dialect URIs; scope strict validation to bridged/Codex calls, or put it behind a flag for one release. Keeping fail-closed as the default is the author's decision and is not disputed here.

**H2. T2 can print `MINIKUBE_T2_PASS` while `codex-llm-proxy` runs the test-only fixture image** (prior M5: **PARTIALLY FIXED**)
- **Fixed part:** `scripts/minikube/t2.sh:530-546` now re-runs `run_final_preflight` after Health/Playwright. A contract test protects that ordering, and a vacuity mutation confirmed the test fails without it.
- **Remaining gap:** the preflight (`t2-common.sh:431,495-560,655-760`) checks the marker ConfigMap, the manifest file and deployment readiness. It never compares live container images or env with the manifest.
- **What this PR adds:** `prepare-codex-approved-tools.mjs:1100-1130` patches the production proxy to `clerum/codex-approved-tools-proxy-e2e:test` with `NODE_ENV=test`. The `prepare` action exits 0 without restoring (`:1217-1225`), and there is no SIGINT/SIGTERM handler.
- **Failure scenario:**
  1. `make minikube-prepare-codex-approved-tools` runs, either on its own or as `T2_PLAYWRIGHT_COMMAND`, or an interrupted `run`.
  2. `make minikube-t2-runtime` runs next. The plan is `already-synced` and the proxy pod is Ready.
  3. Result: PASS on a profile whose proxy runs the fixture image. The comment "Certify the restored runtime" (`t2.sh:534`) does not hold.
- **Suggestion:** have the preflight compare live images/digests with the manifest, or fail on the `evenfire.ai/codex-tools-fixture-run` annotation and test env.
- READ (the lead confirmed that no live-image comparison exists and that `prepare` leaves the swap in place).

### Medium

**M-1. One out-of-pattern MCP tool name fails every Codex turn in `auto`/`direct`** (REPRODUCED; contract and security reviewers)
- **Where:** `packages/llm-provider-attempt-contract/index.cjs:25,239` (`^[A-Za-z0-9._:/-]{1,128}$`) and `codexSubscription.ts:330-333`, which forwards names unchanged.
- **Failure scenario:** a server exposes `list items`, `herramienta_ñ`, `tool@v2`, or a full name over 128 characters (server names can already be 63 characters plus `__`).
  - `parseCodexCompletionRequestV1` rejects the whole request with `tools[n].name is invalid`, a non-retryable 400.
  - Before this PR that definition was dropped and the turn worked.
- **Why the aliases don't help:** `ToolNameMap` (299762dac) runs only after this check passes.
- **Suggestion:** exclude or route non-conforming names through discovery at the host.

**M-2. The global 4-worker validation cap rejects valid calls, and every rejection has the same message** (REPRODUCED by 3 reviewers)
- **Cap behavior:** `boundedSchemaValidation.ts:7,88-93` rejects instead of queueing.
  - 5 concurrent valid validations → `[true,true,true,true,false]`.
  - 4 in-flight ReDoS validations block unrelated servers for about 2 s.
- **Exposure:** default `CLERUM_MAX_CONCURRENT_SESSIONS=3` stays under the cap; values of 5 or more reach it.
- **Message problems** (`toolRegistryAdapter.ts:119`):
  - One text covers four causes: bad arguments (no field named), a permanently unsupported schema, saturation, and timeout. None is logged.
  - The message says "describe the tool again", but `clerum__tool_describe` is only registered when `dynamicToolsEnabled` is on, which is off by default.
- **Result:** the model retries until the loop limit, and operators cannot see why.

**M-3. Two worker spawns per MCP call** (REPRODUCED)
- **Where:** `toolUseLoopToolBatch.ts:225-227` validates, then `McpToolAdapter.execute` validates again (`toolRegistryAdapter.ts:175`).
- **Cost:** each run spawns a worker, loads Ajv and compiles the schema, measured at about 41 ms for a trivial schema and 128 ms for a 38 KB schema. That adds roughly 85–250 ms to every MCP call on every provider, and each run also copies the full catalog twice through `liveTool()`.

**M-4. Discovery guidance is wrong when `auto` resolves to direct** (prior minor: **NOT FIXED**; READ)
- **Where:** `taskExecutor.ts:1652,1668` enables `TOOL_DISCOVERY_TEXT` whenever `clerum__tool_search` is present.
- **Failure scenario:** under `auto`, below 60 tools or 32 KB, every MCP definition is already presented directly (`deferrableToolController.ts:93-118`), yet the prompt says tools are "not all listed directly". No test covers the auto→direct case.

**M-5. The synthetic approved-tools E2E lane cannot run at head** (REPRODUCED; 3 reviewers)
- **Evidence:** `tests/e2e/fixtures/codex-subscription/approved-tools-setup/seed.mjs` is missing. The README says it is "deliberately absent" because a local content-control hook blocked it. `prepare-codex-approved-tools.mjs:903-907` reads it and throws `ENOENT`.
- **Impact:**
  - The Playwright spec, runner, 3 fixture images, 4 Make targets and about 3k lines can never produce a verdict.
  - Green CI only covers the hermetic node tests.
  - The README records a detail of one developer's machine (the hook).
- **Note:** the PR body is honest about this. Merging an unrunnable lane is still a placeholder by absence.

**M-6. Cluster resources created per fixture run are never cleaned up** (READ)
- **Where:** `prepare-codex-approved-tools.mjs:687-750,1029-1033`. `restoreOwnedFixture` only reverts the proxy and stops port-forwards. `state.resources` is recorded but never used, and `validateKubectlArgs` has no delete path.
- **Impact:** each run leaves 4 Contexts, 4 Hosts, 4 McpServers, 1 WorkflowRecipe, 2 NetworkPolicies and seeded DB rows in the shared branch profile.

**M-7. The rewritten image-manifest writer has no test** (READ)
- **Where:** `scripts/minikube/build-images.sh:1413-1717`, about 250 lines on the path every setup, pre-gate and `--only` build uses.
- **Evidence:** `scripts/e2e/approved-tools-image-proof.test.mjs:8-11` itself says the hermetic builder test "is still owed".

**M-8. Journeys now mutate the profile under the lease, with no test or doc** (READ)
- **Where:** `t2.sh:407-410,452-455` forward `T2_SKIP_LOCK=true` to Health/Playwright children.
- **Gaps:** no test asserts that forwarding. `AGENTS.md:229-233` and `docs/testing/minikube-t2-runbook.md:295-304` still describe these journeys as opt-in evidence lanes.

**M-9. An unknown transport alias is only tested at unit level** (REPRODUCED by mutation)
- **Where:** `codex-llm-proxy/src/codexTransport.ts:342-348`.
- **Evidence:** removing the `canonicalName === undefined` rejection leaves the proxy suite at 70/70 green.
- **Consequence if it regresses:** a `tool_call` frame with `name: undefined` would reach the Host.

### Low

**Contract / proxy**
- **Model echoes a canonical name (REPRODUCED, proxy side).** The prompt shows canonical names (e.g. `**files.read**`, `promptBuilder.ts:299-303`) while upstream only sees `__codex_tool_<hash>`. If the model calls `files.read`, `fromWire` returns undefined and the whole attempt fails as retryable `provider_unavailable`, instead of the Host rejecting that one call.
- **1 MiB measured on three envelopes (REPRODUCED).** Authorize overhead is 434 B and proxy overhead is 1582 B (RS256 ticket). A request in that window passes authorize, mints a ticket and reserves budget, then gets 413 at the proxy. Doc `:76` says oversized direct requests fail "before authorization", but the Host has no pre-check.
- **Classification of bad outputs.** More than 32 returned calls maps to retryable `provider_unavailable` even though the output is deterministic. A stream with text and no `response.completed` drops buffered calls and returns `FinishReason.Unknown` with no error (`codexTransport.ts:400-404`, `codexSubscription.ts:82`).
- **SSE buffer (unverified against the real endpoint).** The 1 MiB buffer (`codexTransport.ts:356,362`) could be exceeded if upstream echoes `tools` in `response.created`/`completed`.

**Host**
- **Deep nesting.** A schema nested 10k levels deep makes `JSON.stringify` in `deferrableToolController.ts:100` throw `RangeError`, so every Codex turn fails closed. `assertFiniteTree` in the contract also recurses without a depth limit. REPRODUCED for the V8 limits.
- **Logger bug (REPRODUCED).** `logger.ts:126` spreads `redactUnknown(fields)`. When redaction fails, that value is the string `'[Unserializable]'`, so the log line gets one key per character and `taskId` is lost.
- **Logging migration dropped diagnostics.** `Loop failed` logs only `err.name`, and `Skipping server not ready` drops `status.message`.
- **Workflow steps skip validation.** They dispatch through `client.callTool` (`workflow/stepRouter.ts:268`) and bypass `McpToolAdapter`. Not a regression, but broader than the claim.
- **Freshness check.** It reads the representative client's schema, not the per-user `oauth-user` client (`toolRegistryAdapter.ts:109-113`). READ.
- **Stale comment.** `taskExecutor.ts:1729-1730` says presentation "follows the primary provider". In fact any Codex fallback switches a non-Codex primary to Codex presentation (`toolPresentationPolicy.ts:27-28`). This is intentional and tested; it should be documented for operators.
- **Architecture.** Provider-specific logic (`'codex-subscription'`, the `codexMode` branch) now lives in `core/orchestration`. Defaults are duplicated: `32_768` in two files, and `?? 'auto'` despite config always setting a value.

**Tests**
- **Two returned-call checks.** `pending.size > 32` and `toolFrames.length >= 32` each survive removal alone; only removing both fails the 33-call test.
- **Threshold boundaries.** `>` → `>=` in `deferrableToolController` survives (128/128); 60 vs 61 is never tested.
- **Past-call alias.** Removing past-call names from the `ToolNameMap` input leaves 44/44 green.
- **Negative assertions without a witness.**
  - `approvedToolsLifecycle.integration.test.ts:260-287` still passes when dispatch is deleted.
  - `codex-subscription-approved-tools.spec.ts:147-149` uses `toHaveCount(0)` before the chat view is shown to have rendered.
- **Weaker patterns.**
  - `assert.throws` without matchers.
  - A "replacement detected" test that passes because of `JSON.parse('')`.
  - An ordering test built on `source.indexOf`.
  - The spec's `finally` can replace the original error.
  - `approvedToolsUpstream.ts:144-147` has a bare `catch {}`.
- **Config wiring.** No test covers the import-time throw for invalid `CODEX_TOOL_PRESENTATION`.

**Harness / infra**
- **Passwords in traces.** `trace: 'retain-on-failure'` records `fill()` values, including `*_PASSWORD`, in `trace.zip` (CWE-532). Local only, mode 0700.
- **Restore can refuse its own prepare.** `prepare` records the original proxy image without validation, but `restore` refuses refs like `ghcr.io/...@sha256:` or `localhost:5000/...`, leaving the fixture image in place. REPRODUCED.
- **Issue-627 scripts.** `build-627-*.sh` build `docker.io/clerum/issue627-*`, but `install-627-real-tools.mjs` deploys `registry.evenfire.ai/...` refs, so the local builds are never used. The scripts are one-off, non-transactional, and exit silently on precondition failures under `set -e`.
- **Mutation-boundary test.** The 4 new approved-tools Make targets are not covered by `test-minikube-mutation-boundary.sh`, and they are not in the runbook.
- **Zero-test files.** The node:test group checks (T0 and `ci-public.yml`) verify only aggregate counts, so a file that registers zero tests would still pass.
- **Leftover processes.** Detached port-forwards outlive the verdict and the lease.

**Docs / PR body**
- **Stale body.** The body predates `299762dac`; its line "No aliasing or silent tool omission was introduced" is inaccurate at head.
- **Doc tables.** The contract doc compatibility table understates skew impact (see B1). The approved-tools README says `CATALOG_SIZE` must be set, but `server.mjs:145` defaults it to 83.
- **Unrelated scope.** The `external-rest-api` version bump and test flake fix (`b79c5dc35`) are unrelated to #627.

---

### Verdicts on prior review (jozer-rami, 7a4f94969)

| Item | Verdict at `299762dac` |
|---|---|
| B1 rollout order | **NOT FIXED** (see B1) |
| M1 host validation | **PARTIALLY FIXED**: isolation and ReDoS fixed; cross-provider regression remains (H1), plus new capacity and latency costs (M-2, M-3) |
| M2 strict config / no legacy mode | Author decision; unchanged. Not re-litigated |
| M3 proxy forwards `body.tools` | **FIXED**: `slice(0,32)`, dropping `payload.tools`, and dropping `parallel_tool_calls` all fail the suite |
| M4 names upstream | **FIXED at the wire** by `ToolNameMap` (collision and reserved-alias attacks tested). Residual: contract regex (M-1), canonical-name echo (Low) |
| M5 lease/final preflight | **PARTIALLY FIXED**: re-run added; preflight cannot see live image swaps (H2) |
| Minor: partial-stream tool-call drop | **FIXED**: calls only released on `success` |
| Minor: discovery guidance under direct | **NOT FIXED** (M-4) |
| Minor: Codex-as-fallback presentation | Intentional and tested; code comment contradicts it |
| Minor: 128-message cap | **NOT ADDRESSED** |
| Minor: 1 MiB envelope mismatch | **CONFIRMED** (Low, reproduced) |
| Minor: test gaps / scope split | **PARTIALLY FIXED** |

### CodeQL alerts

| Alert | Verdict |
|---|---|
| Shell command / indirect command line, `prepare-codex-approved-tools.mjs` :120, :603, :610, :617 | **False positive.** `spawnSync`/`spawn` with argument arrays, no shell; exact allowlists in `validateKubectlArgs`/`validateOwnerArgs`; profile regex-checked. Residual: `PATH` lookup and inconsistent env scrubbing on `spawn` :1051 (dev-only) |
| FS race, `prepare` :283, :388; `run` :149 | **True positive in a narrow sense, low risk, dev-only.** `O_NOFOLLOW` covers only the last component; exploiting it needs the same user writing inside their own 0700 directory. The `dev`/`ino` comparison at :321-323 always matches, since it fstats the same descriptor. |
| Network data written to file, `approved-tools-workflow/index.js` | **False positive.** Constant path; receipt reduced to 4 regex-checked fields; `wx`/0600; test-only image |

### Verified clean
- **Authorization of bridge tools.** search/describe/invoke use the same `mcpManager.getAllTools()` catalog.
  - `resolveBridgeCall` rewrites to the real tool before validation, guardrails and approval.
  - Native, bridge and non-catalog targets are rejected.
  - Identity stays bound to the caller.
  - A revoked server fails at `liveTool()`.
  - Approval resume re-checks schema and args with no `await` before dispatch.
  - A failed tool presentation now errors instead of falling back to the full catalog.
- **Name mapping.** Only the first `__` separates server from tool, in the manager and in the approval gate. Kubernetes server names cannot contain `_`. Alias squatting with `__codex_tool_<hash>` and `_0` resolved correctly regardless of input order.
- **Bounds that replace the 32 cap.**
  - 1 MiB body; `maxToolCalls=32` in contract, proxy and host.
  - Calls released only after success.
  - Search 32 KiB/page, 512 B descriptions.
  - Worker limits 256 KiB / 10k nodes / depth 64 / 64 MB.
  - No leftover `LIMITS.maxTools` consumer.
- **Hashes.** Every definition, including `parameters`, is hashed, and the canonical hash fixtures are unchanged.
- **Logging.** Error objects are reduced to name/code/status. Proxy URL and session key are no longer logged. The touched files moved from `console.*` to the service logger.
- **CI and supply chain.** `ci-public.yml` adds one `pull_request` step, with no new permissions, secrets or untrusted interpolation, and actions stay SHA-pinned. `package.json`/lock changes are only version bumps of the repo's own packages; no new dependencies.
- **Fixtures stay out of production.** They are excluded from the production proxy Dockerfile and from publish/pull/verify sets. `--only` substring selectors do not pull them in. The proxy fixture refuses to start without test markers.
- **`seed-e2e-data.sh` −69 lines.** A byte-identical extraction into `desktop-login-seed.sh`.
- **Secrets in harness.** Credentials go over stdin (never argv), under umask 077, with `wx`/0600 files.

### Suites and gates run by reviewers

| Run | EXIT | Result |
|---|---|---|
| mcp-host `npm ci` / `tsc --noEmit` / lint / build | 0/0/0/0 | lint 0 errors (32 warnings already present, none in touched files) |
| mcp-host touched test files (12) | 0 | 318/318 |
| mcp-host full suite (354 files) | 0 | 3927 passed, 0 failed, 1 todo (already present, not in diff) |
| `approvedToolsLifecycle.integration.test.ts` | 0 | 51/51 |
| codex-llm-proxy typecheck / build / full suite | 0/0/0 | 72/72 (10 files) |
| `llm-provider-attempt-contract` `index.test.cjs` | 0 | 21/21 |
| E2E node:test CI step (7 `.mjs` files) | 0 | 60/60 |
| Shell: t2-contract / images-manifest / mutation-boundary / build-section-headers / verify-images / pre-gate-shadow / docker-cli-env (stubbed) | 0 each | 217 / 24 / 5 / 9 / 24 / 35 / 31, 0 failures |
| GitHub checks at `299762dac` | — | 38/38 passing; merge state `BLOCKED` (review required) |

**Not run:**
- Playwright spec and prepare/run/restore: need a cluster, and cannot run anyway (M-5).
- `test-hcc-lifecycle-cleanup.sh`: blocked by a local kube-context guard.
- `test-codex-subscription-t0.sh`: needs Electron.
- Any live T2.

**Mutation testing** (14 mutants on copies): 10 killed. The survivors are the individual returned-call checks (M6, M7), the unknown-alias rejection (M8), and the threshold boundary (M9).

### Recommendation

- **Before merge:**
  - B1: enforce the order, or ship the validator relaxation in its own release first.
  - H1: normalize dialect URI spellings, and scope or flag cross-provider validation.
  - H2: make the final preflight detect live image/env drift.
- **Follow-ups:**
  - M-1 through M-4 and M-9 in the product code.
  - M-5 through M-8 in the harness.
- **Suggested split,** to reduce the review surface and the size of a rollback:
  1. Product fix + its tests + contract doc (about 39 files).
  2. The T2 lease / post-journey preflight change.
  3. The approved-tools fixtures, Playwright lane and Make targets, once `seed.mjs` exists.
  4. The `external-rest-api` flake fix.
  5. Optionally, the issue-627 operator scripts, or drop them.

alfredolopez80:
@claude Please perform a complete closing audit of PR #632 at `09bb5f1ef719ec2513f9da3aac1a35ecbbab6038`, against `dev`, following up on your review in https://github.com/evenfire-ai/evenfire/pull/632#issuecomment-5678404717. Review the integrated result, not only the latest commit. Do not merge or deploy.

Please return an explicit disposition for B1, H1–H2, M-1–M-9 and the actionable lows: fixed with file/line evidence, still reproducible, not applicable with reasons, or tracked follow-up. Prioritize blockers/high/medium; distinguish verified failures from assumptions. Recommend the smallest correction for every surviving finding. Check whether the new tests could pass with the behavior removed or broken.

Please audit these areas in particular:

1. Product behavior: removal of Evenfire's 32-definition restriction throughout host/authorizer/proxy, preservation of request-byte and returned-call bounds, exact 60/61 and UTF-8 selection thresholds, direct versus discovery guidance, all approved tools remaining reachable, canonical names/transport aliases/history/unknown aliases, and Codex fallback behavior. Confirm approval, guardrails, caller identity and revocation still apply to resolved calls. Do not infer an unlimited upstream limit or subscription-quota savings from local tests.
2. Selected MCP validation: scope to Codex provider chains, dialect aliases, bounded worker admission/deadlines, safe diagnostics, cache invalidation and the final live-schema check. Documentation now explicitly distinguishes interactive task adapters from direct workflow steps and per-user OAuth schemas.
3. M-5 setup: the missing seed now exists. A separate unpublished Control API fixture image installs an external-only OAuth simulator before importing the real API. Verify the actual device start/pending/poll/exchange contract, startup guards, packaging paths, and that unsupported auth-provider operations cannot contact the real provider. Application login, consent, persistence, catalog publication and approvals must remain real. Verify the seven-image acquisition and all three exact derived-base bindings.
4. M-6 restoration: inspect both the complete Kubernetes resource lifecycle and database/subscription lifecycle. Check UID/version/run preconditions; preallocated identity IDs; durable per-scenario connection intent before the UI Create action; capture of exact public response identity before subsequent assertions; strict reingestion; refusal to adopt unrecorded or replaced connections; normal revocation with retained audit/tombstones; transactional identity deletion; and successful post-commit runtime ConfigMap publication before `cleaned`. Verify publication retries after a committed deletion, repeated restore, partial failure, and restoration of the original API even when cleanup fails. WRC can delete the owned workflow grant first: absence is accepted, foreign grants are rejected, and duplicate revoke audit is not invented. Host/Context grant retention and ordinary Desktop login were inspected separately.
5. H2/M-8: the final preflight now checks both the original proxy and Control API live image identities, deployment/pod configuration and fixture markers. Confirm it cannot certify a residual test image or activation file. Verify the inherited mutation lease, final post-journey preflight, bounded commands, exact branch profile and all four public Make targets. Infrastructure rollout B1 remains an external release dependency owned by the separate infrastructure task; do not treat these local guards as its resolution.
6. Test/CI quality: actual builder provenance tests, per-file nonzero Node-test registration, import-time presentation-config tests, callback durability through later UI failure, negative assertions with positive witnesses, unknown-alias transport rejection, disabled password-bearing traces, and preservation of the original failure during cleanup. Check production publish/default-image exclusion and no new dependencies. Version-only package/lock bumps came from the repository commit hook.

Evidence available before this request:

- The integrated Node24 fixture group passed 115/115; after the final controller-cleanup correction, its affected identity/connection/preparation/restoration suites passed 51/51 (including 17 identity cases). These counts overlap and should not be summed.
- HTTP MCP fixture tests exercised actual local HTTP servers with 83, 150 and 250 definitions and selected business calls; this is deterministic local evidence, not a real Codex subscription run.
- Import-time configuration: 8/8. Hermetic image provenance and mutation-boundary suites passed using exact-source isolated Git fixtures with unchanged-checkout guards. Proxy and Control API runtime-guard suites and existing T2 scenario tests passed.
- Playwright typecheck passed. Static audit passed for the changed journey/helper; it reports a browser-wait rule on non-browser fixture modules and two preexisting unrelated positional selectors in the shared page object. Those were not reported as a global audit pass or hidden with markers.
- Previous published head `438763b89` had 38 successful GitHub checks. CI for this new head must be evaluated separately.

Important remaining evidence boundary: the new fixture image has not yet been built/deployed and the complete automated browser journey, exact-head T0/T1/T2 and current-head real-subscription run have NOT been certified. The user explicitly requested this audit and green CI before those expensive lanes. Please distinguish implementation defects that prevent executing that next validation from runtime claims that still require evidence. Earlier real-subscription results documented in the PR remain historical and revision-bound.

Expensive low follow-ups already filed: #637 (canonical versus HTTP envelope byte accounting), #638 (JSON depth handling), #639 (local plugin build/installer wiring). Please identify any additional justified follow-up rather than silently treating it as resolved.

alfredolopez80:
CI follow-up: `066f3aa56` adds the new Control API OAuth fixture to the strict unpublished-image contract and marks its hermetic runtime-guard script executable, as required by repository hygiene. The manifest contract passed 24/24 checks locally on Node 24.18.0; the indexed shell executable-mode check also passed. These are the only changes after the implementation commit `09bb5f1ef` currently under closing audit. CI must be confirmed on this updated head; no new runtime certification is claimed.

alfredolopez80:
@claude Please recheck the M-4 and M-7 dispositions against current head `90fd50bc56ae73e6162f44f70876b0de61399f22`. This is a focused follow-up to the completed audit, not a request to repeat all suites or change code.

**M-4:** the current runtime text in `mcp-host/src/core/reasoning/promptBuilder.ts` begins “Use directly listed tools when available. For additional approved tools, …”. It no longer says MCP schemas are hidden or mandates discovery before direct invocation. The comment at `taskExecutor.ts:1643-1649` concerns the separate `MCP_SERVER_SELECTION_TEXT` heuristic (`includes('__')`), not deferring the discovery wording fix. An independent adversarial trace confirmed that tiered parts are built before per-turn presentation and cached while `auto` can change on subsequent catalog refreshes; gating that cached instruction on one selected strategy would need additional invalidation logic. The shared direct-first instruction remains accurate on both sides of the threshold.

The new integration regression in `approvedToolsLifecycle.integration.test.ts` uses the real manager/registry, presentation controller and both prompt builders. For each of 83/150/250 tools it proves: an initially small catalog is fully direct; adding the live catalog defers MCP definitions while retaining the bridge; removing the added server restores direct definitions; both prompt paths retain the direct-first/conditional discovery wording; stable/context hashes remain compatible. The three selected cases passed, and MCP Host typecheck passed. The explanatory prompt comment was corrected; no new limit or mandatory discovery step was introduced.

**M-7:** the generator test already gates CI through this unconditional call chain:

- `.github/workflows/ci-public.yml:588` runs `scripts/tests/test-minikube-t2-contract.sh`.
- `scripts/tests/test-minikube-t2-contract.sh:666` runs `scripts/tests/test-minikube-docker-cli-env.sh`.
- The [actual GitHub job log](https://github.com/evenfire-ai/evenfire/actions/runs/34972818133/job/104393057689) at `2026-09-15T13:10:52Z` reports `PASS: seven sequential --only builds preserve prior revisions and all exact derived-base bindings`; subsequent lines confirm replaced/missing/profile/changed-base/changed-workflow-base/malformed rejection cases.

I added a workflow comment documenting this indirect invocation. A second execution or follow-up issue would duplicate existing coverage.

The two CI failures in your audit were already fixed in `066f3aa56`. This latest commit also adds the cheap public-logger fallback test and import-time byte-budget tests (the actual variable is `CODEX_TOOL_DISCOVERY_BYTES`, not `CODEX_TOOL_PRESENTATION_BYTES`): 27/27 focused tests passed. B1 remains with the separate infrastructure task; no runtime certification is claimed. Please confirm or provide a concrete remaining M-4/M-7 counterexample so we can close the code audit before T0/T1/T2.

alfredolopez80:
@claude Please perform a focused read-only check of `aeca8d9e1495c71eeab7c2a0d0a1a97764caf5a4`, the two-file correction discovered during the authorized local certification. Do not repeat the full audit or change code.

On `90fd50bc5`, canonical T0 passed and real PostgreSQL T1 passed 164/164 tests across 35/35 files, with no pending tests. T2 then correctly refused to certify because the new Control API guard included the completed migration Job in the serving population: both share `app=control-api`. The actual serving Deployment/pod image matched the manifest; the extra Job pod had a `migrate` container and was not Ready. The migration log confirms successful completion, and the pod subsequently expired normally. The browser lane had not started.

The correction excludes a pod only when it is `Succeeded`, controlled by a `batch/v1` Job, has reported normal container statuses, and all reported normal/init containers are terminated with exit code zero. This happens before the required serving-count check. All remaining pods still undergo the existing image, readiness, environment and fixture-marker checks. No Job is deleted to make the verdict pass, and failed/running/unknown or malformed states remain fail-closed.

The regression reproduces the original failure, then covers healthy serving + completed migration, migration alone, insufficient serving replicas, wrong serving image, retained fixture markers, failed/running/unknown Jobs, wrong owner/controller, missing status and nonzero/running containers. Node24 hermetic API suite: 3 positive and 37 negative cases passed; existing proxy/lease suite, syntax, ShellCheck and public-boundary gate passed.

Please confirm the narrow filter does not weaken H2 or identify a concrete counterexample. CI and final exact-head T0/T1/T2 for this new commit remain pending; results from `90fd50bc5` are not being reused as certification of the new revision. B1 remains external to this repository's task.

jozer-rami (approved):
Approving code at `aeca8d9e`. Verified locally: mcp-host 3974 tests, proxy 77, contract 25, T2 runtime/contract shell tests green; CI 38/38. M1/H1, M3, M4, M5/H2 resolved and mutation-checked.

This approval is not release sign-off. Before this ships to any environment:
- **B1** rollout order must be enforced (infra #87 with a populated catalog, plus MCC tenant upgrades), or split into a validators-first release.
- Exact-head T2 and the approved-tools Playwright journey still need to run.

Small follow-up: no test covers non-Codex chains skipping strict MCP validation — forcing `strictValidation: true` at `taskExecutor.ts:1767` keeps the full mcp-host suite green.

Issue #10, opened 2026-07-14
profile-ui tab title still reads "Clerum Profile UI" (should be evenfire per code-names.md)


jozer-rami:
## Summary

The profile-ui browser tab title still reads **"Clerum Profile UI"**. Per the project's own naming policy in [`docs/concepts/code-names.md`](docs/concepts/code-names.md), user-visible product surfaces (product, docs, UI) use **evenfire**; `clerum` is retained only for internal surfaces (the `clerum.io` API group, `CLERUM_*` env vars, `clerum__*` tool names, chart names). A browser tab title is user-visible, so this is the last user-visible "Clerum" in a shipped UI.

## Where

`profile-ui/app/layout.tsx:8`

```ts
title: 'Clerum Profile UI',
```

## Suggested fix

Rename to an evenfire-branded title, e.g. `'evenfire Profile'` (or whatever matches the other UIs' title convention). One line.

## Notes

This was left as a product-code change rather than bundled into the `oss-wave0-scrub` docs branch. It is the only genuine user-visible naming inconsistency found in a repo-wide sweep — the other `clerum` occurrences in UI code (`clerum.io/*` labels, the `window.clerum` bridge global, `clerum__*` tool names) are internal API surface and are intentional per `code-names.md`.

PR #633, opened 2026-09-14
fix(hcc): retract runtime conditions when HCC does not run a server

### LINKED ISSUE #606: HCC writes NetworkReady=True without creating or verifying any NetworkPolicy, and never retracts it
## Summary

HCC writes `NetworkReady=True` with `reason: NetworkPoliciesApplied` and the message
`NetworkPolicies and Service created` without creating, reading or verifying a single
NetworkPolicy. The fail-closed branch then leaves that condition — and `DeploymentReady` —
in place after deleting the runtime. On `clerum-dev` three McpServers currently advertise
readiness for resources that have not existed for over two weeks.

Code verified against `origin/dev` = `c7741481a`, read with `git show c7741481a:<path>`.
Cluster state read on 2026-09-10.

## The condition is written without evidence

`reconciler.ts:1853-1862`:

```ts
const networkWrote = await this.writeStatusCondition(
  server,
  {
    type: 'NetworkReady',
    status: 'True',
    reason: 'NetworkPoliciesApplied',
    message: 'NetworkPolicies and Service created',
  },
  isCurrent
)
```

`reconciler.ts` contains **zero** occurrences of `createNamespacedNetworkPolicy`. The file
only ever LISTs (`:1735`) and DELETEs (`:1744`) NetworkPolicies. Every policy in the four
families is created by `networkPolicyReconciler.ts` or by
`ExternalEgressConvergenceCoordinator`, neither of which participates in this write.

The condition therefore asserts an outcome the writing lane has no knowledge of.

## The fail-closed branch does not retract it

`performReconcile` writes exactly two conditions when secret validation fails closed
(`reconciler.ts:2199-2219`): `SecretResolved=False` and `Ready=False`. It does not touch
`NetworkReady` or `DeploymentReady`, even though `deleteRuntimeResources` (`:2187`) has just
deleted the Deployment, Service and NetworkPolicies.

## Observed on clerum-dev

`kubectl -n mcp-server get mcpserver web-research`:

```
NetworkReady=True        reason=NetworkPoliciesApplied   since=2026-05-08T14:30:40Z
DeploymentReady=True     reason=ReplicasAvailable        since=2026-08-25T09:36:27Z
SecretResolved=False     reason=SecretNotFound           since=2026-08-25T21:15:53Z
Ready=False              reason=SecretValidationFailed   since=2026-08-25T21:15:53Z
```

The Deployment does not exist and there are no pods:

```
deployments.apps "web-research" not found     pods=0
```

`DeploymentReady=True reason=ReplicasAvailable` has been stale for 16 days. Same shape for
`evenfire-mcp-whois` and `evenfire-mcp-github-clerum`.

`Ready=False` does make the aggregate verdict correct today, which is why this has gone
unnoticed. The per-resource conditions are the ones that lie.

## Why it matters

It is latent rather than active, and #604 is what would activate it. That issue proposes
gating the creating lanes on server state; with that gate, a server whose secret is fixed
gets its Deployment back through the `SecretInformer` path while its NetworkPolicies wait
for the next full pass. During that window `Ready` would be `True` and `NetworkReady` would
be `True`, while every call to the server times out. An operator reading the status cannot
distinguish that from a bug inside the server.

The condition also cannot be made truthful where it currently lives: after #604 the lane
that knows whether the allows exist is the Context lane, not the McpServer lane.

## Suggested direction

Not prescriptive.

- Have the lane that actually creates the policies own `NetworkReady`, or
- drop the condition, or
- narrow its `reason`/`message` to what the McpServer lane can actually attest (the Service).

Independently, the fail-closed branch should retract `NetworkReady` and `DeploymentReady`
when it deletes the runtime they describe.

## Non-goals

- The delete/create loop itself (#604).
- The `SecretResolved` verdict semantics, including the deliberate `ReadError` exclusion in
  `shouldFailClosedForSecretFailure` (`reconciler.ts:2377-2385`), which is correct as written.

alfredolopez80:
## Summary
- Stop writing `NetworkReady` and strip it from every HCC status write so the condition can no longer claim a ready network without NetworkPolicies.
- On disabled and fail-closed managed servers, retire the readiness poll and publish `DeploymentReady=False/RuntimeNotDesired` after runtime delete so leftover `True` conditions cannot outlive a deleted Deployment.
- Keep WRC-owned paths unchanged: HCC still does not delete or retract `DeploymentReady` for runtimes it does not own.

Closes #606

## Test plan
- [x] HCC unit: T1–T15, T17 plus §6.2 regressions (`reconciler.test.ts`, `reconciler.credentialsRotation.test.ts`, `reconciler.remote.test.ts`)
- [x] Vacuity mutations M1–M12 each failed the intended tests and were reverted
- [x] `npx vitest run` in `host-context-controller` (2403 passed) plus public-boundary (`PUBLIC_BOUNDARY_PASS`)
- [ ] Cluster e2e `integration/mcp-runtime-conditions.test.ts` via `scripts/e2e/run-vitest-e2e.sh` — tracked in #636 (not a #633 merge gate)
- [ ] `make minikube-t2` on this exact HEAD (last T2 pass was an earlier commit)


alfredolopez80:
@claude re-review and closing audit on HEAD `ab7ed881`.

Local T2 certified: `MINIKUBE_T2_PASS` (`T0=PASS` `T1=PASS` `T2=PASS`, NP-08 PASS). CI is 38/38 green on this HEAD. Please re-check the residual C/D/F follow-up (poll decline-at-decline-time, retract incomplete on failed RuntimeNotDesired write, `waitForStatusCondition` docstring) and certify close if nothing new remains.

alfredolopez80:
@claude re-review and **closing audit** of every review already performed on this PR. Do not treat the short prior ping as the brief. Audit **HEAD `ab7ed881850c2145657e351b6d708295232c664f`** (`fix/issue-606-hcc-runtime-conditions` → `dev` @ `98840fbfe675f5dca8fdf312c7267c1ffb42f65a`). Issue **#606**.

This comment is the source of truth for prior local-only reviews (nothing below was previously published as a GitHub review). Please independently re-read the current producer + tests, then either **CERTIFY CLOSE** or name remaining defects with severity, file:line, and a failing scenario. Do not weaken public-boundary. Do not merge.

---

## Certification already on this HEAD (do not re-litigate as missing)

- **CI:** 38/38 `SUCCESS` on `ab7ed881` (including HCC unit/typecheck, Real Postgres 16, public-boundary / the job displayed as “Shell script syntax (bash -n)”, CodeQL no new alerts).
- **Local T2 (real verdict, not the planner):** `MINIKUBE_T2_PASS` with `T0=PASS` `T1=PASS` `T2=PASS`, `NP08_HCC_AUTHORIZATION=PASS`. Health/Playwright `NOT_RUN` (opt-in; not blockers). HCC package `0.1.326`.
- **Public-boundary vs the same base CI uses:** `T2_PUBLIC_BASE_REF=98840fbfe675f5dca8fdf312c7267c1ffb42f65a` → `PUBLIC_BOUNDARY_PASS` on `d618d10` and later. Do **not** add `cGFzcw==` as a plus-line. Do **not** teach the scanner or add test files to `safe_source_paths`. The BASE `validateSecret` fixture `password: 'cGFzcw=='` (~line 735) is pre-existing and must stay.
- **HCC unit gate on this HEAD:** `src/reconciler.test.ts` + `src/__tests__/reconciler.credentialsRotation.test.ts` + `src/__tests__/reconciler.remote.test.ts` plus full HCC suite **2398** passed after the C/D follow-up.

---

## Product promise (#606) and chosen design (must still hold)

HCC must:

1. Stop writing `NetworkReady`; strip it on **every** status write (`RETIRED_STATUS_CONDITION_TYPES = ['NetworkReady']`).
2. On **managed** disable / fail-closed Secret failure: delete the **HCC-owned** runtime and persist `DeploymentReady=False/RuntimeNotDesired` only after the Deployment is **confirmed-gone** (delete 2xx/404 or GET 404).
3. Leave WRC-owned (`managed:false`) alone: no delete, no `DeploymentReady`.
4. Keep the WRC handshake annotation `clerum.io/network-ready` (no WRC deploy required).
5. Keep the public signal `False/RuntimeNotDesired` (do **not** drop the type while retired).
6. Single helper `retractManagedRuntime(server, kind, isCurrent) → { retracted, cleanupComplete, foreignRuntime }` for both disabled and fail-closed.
7. Sibling Service/ConfigMap failure must **not** skip retract once the Deployment is gone.
8. If the Deployment is still present and delete is **denied**: no retract, incomplete, **zero** status patches (T6').
9. If the runtime is **foreign** (present but not HCC-owned): no `RuntimeNotDesired`, still publish `SecretResolved` / `Ready` / `Disabled`, `hasIncompleteReconciliation()===false` (RP-008).
10. `retireReadinessWindow` after the retract write; poll decline is silent **only when retired**.
11. Bump `lastTransitionTime` on `DeploymentReady` when **reason or message** changes (not all condition types; not `SecretResolved` message-only).
12. One `withdrawRuntime` boolean per fail-closed branch.
13. Distinguish Deployment delete outcomes `'gone' | 'denied' | 'foreign'`.

Messages:

- `RUNTIME_NOT_DESIRED_DISABLED_MESSAGE` = `McpServer is disabled; HCC does not run its runtime`
- `RUNTIME_NOT_DESIRED_FAIL_CLOSED_MESSAGE` = `Env Secret validation failed; HCC does not run its runtime`

Rejected (must stay rejected):

- Move `retireReadinessWindow` after delete so a poll 404-writes `False` (reopens the BASE stale-`True` race).
- Teach Control UI / e2e to accept stale LTT (reopens #223).
- Bump LTT on every reason change for all types (watch-loop risk on `SecretResolved`).
- Weaken public-boundary.
- Let an in-flight tick write `RuntimeNotDesired` (two writers).
- Evict `readinessRetirements` (recreated names must keep the epoch so stale ticks fail). Residual **E** is by design; do not implement; do not file an issue.

---

## Commit ledger (audit each; do not review only the tip)

| SHA | What it was for |
|---|---|
| `1d2b4ce7` | First cut: stop writing `NetworkReady`; write `RuntimeNotDesired` after cleanup; retire poll. **Introduced RP-001…RP-007.** CI public-boundary red (`password: 'cGFzcw=='` plus-line). |
| `d618d10f` | Redesign: persist `RuntimeNotDesired` only after Deployment **gone**; LTT bump; public-boundary fixture `storedSecret('fixture-key')`; e2e suite. **Fixed RP-001…RP-007. Introduced RP-008** (foreign abort). |
| `43a2d93f` | RP-008: `'foreign'` vs `'denied'`; foreign still publishes Secret/Ready/Disabled; no eternal obligation. |
| `ab7ed881` | Residual **C/D** + cheap **F**: decline factory at decline time; `StatusWriteOutcome`; retract incomplete on `'declined'`; docstring “at or after `sinceMs`”. HCC `0.1.325` → `0.1.326`. |

---

## Review ledger (local-only; this is what you are auditing)

Reviews were **not** posted to GitHub. Lanes used Grok 4.6, Fable 5.1, and Codex. Supermemory disabled. No merge, no approval.

### Pass 1 — initial review of `1d2b4ce7` (NO merge)

Roster: HCC reconciler + bugs (Grok); HCC unit tests (Grok); security full diff (Grok); regression/evidence (Grok); architecture/root cause (Fable); CRD status contract (Grok); CLAUDE.md + blame (Grok); refactoring (Fable); docs/product (Codex); redesign architect (Fable); compat planner (Grok); verification planner (Grok).

Counts then: Blocker 1, High 1, Medium 3, Low 2. Merge recommendation **NO**.

| ID | Sev | Defect (as found at `1d2b4ce7`) | Required fix | Status at `ab7ed881` |
|---|---|---|---|---|
| **RP-001** | Blocker | Recovery fixture `password: 'cGFzcw=='` trips public-boundary (CI job mislabeled `bash -n`). | Use `storedSecret('fixture-key')`. Do not weaken scanner. | **FIXED** at `d618d10`. Repro: `PUBLIC_BOUNDARY_PASS`. CI green on tip. |
| **RP-002** | High | Retract only if full-tuple cleanup succeeded. Deployment gone + sibling deny/throw → leftover `DeploymentReady=True`, poll already retired. | Persist retract when Deployment confirmed-gone; siblings must not skip it. | **FIXED** at `d618d10`. P1: Service deny and Service 403 still write `RuntimeNotDesired`. |
| **RP-003** | Medium | LTT kept when `status` stays `False`. Recovery `RuntimeNotDesired` → `WaitingForReplicas` reused `retiredAt`. UI/e2e `sinceMs` miss the recovery. | Bump LTT on `DeploymentReady` reason/message change only. Invert recovery test: `WaitingForReplicas.LTT > retiredAt`. | **FIXED** at `d618d10`. |
| **RP-004** | Medium | T6 vacuous: only forbade `RuntimeNotDesired` while `isCurrent` already false. | T6': deny `isCurrent` **once after Deployment GET**; 0 status patches; 0 deletes; obligation true. | **FIXED** at `d618d10`. |
| **RP-005** | Medium | T17 did not pin W2 timer cancelled. | After retract, `readNamespacedDeployment` count must stay flat across 24 ticks. | **FIXED** at `d618d10` (T17'). |
| **RP-006** | Low | Retired in-flight tick still `recordIncompletePublication`. Doc vs tests disagreed. | Silent decline when **retired**. T12/T17: `hasIncompleteReconciliation()===false`. | **FIXED** at `d618d10`; **refined** at `ab7ed881` (factory, see C). |
| **RP-007** | Low | Fail-closed re-evaluated `shouldFailClosedForSecretFailure` three times. | One `withdrawRuntime` per branch. | **FIXED** at `d618d10`. |

Verified clean at Pass 1 (still required): no remaining `type: 'NetworkReady'` writer; WRC-owned no delete / no `DeploymentReady`; §6.2 regressions not weakened; no new authz/injection/deps; handshake annotation unchanged; no control-api/WRC consumer of the **condition**.

### Pass 2 — Grok 4.6 revalidation of `d618d10` (NO close)

Verdict **NO**. RP-001…RP-007 **FIXED**. New **RP-008 (High)**: `deleteDeploymentIfHccOwned` returned `'present'` for both **not-owned** and **delete-denied**. `if (!retract.retracted) return false` aborted the whole reconcile. `managed:true` + WRC-labeled Deployment: no delete (correct) but also no `SecretResolved`/`Ready`/`Disabled` and **eternal** `hasIncompleteReconciliation()`. On `1d2b4ce7`, ownership skip was `complete=true` and those conditions **were** published.

Required fix (landed in `43a2d93f`): distinguish `'denied'` vs `'foreign'`. Denied → incomplete, zero writes (T6'). Foreign → no `RuntimeNotDesired`, continue to Secret/Ready/Disabled, obligation false.

Also noted then (later C/D): `retractManagedRuntime` ignored `wrote` and always `retracted: true` after `'gone'`.

### Pass 3 — Fable 5.1 closing audit of redesign at `d618d10` (**CERTIFY** design; same foreign gap)

Implementation matched the chosen helper order. Evidence matrix (mutations in `/tmp`, not the repo):

| ID | Then | Notes |
|---|---|---|
| P1 | MET | Service deny / 403 still persist `RuntimeNotDesired`. Gating write on `cleanupComplete` → red. |
| T6' | MET | Removing `!deploymentGone` guard → red. |
| P2 | **GAP (evidence only)** | Code records obligation if `isCurrent` flaps after gone; **no test** that flaps during retract write. Mutation “retract write with `'silent'`” stayed green. |
| T17' | MET | Flat Deployment reads after retract. |
| P3 / T-rec old | MET | `WaitingForReplicas.LTT > retiredAt`; old `LTT === retiredAt` pin gone. |
| T10 / T7 / T11/T12/T15 / P6 | MET | WRC-owned; Deployment 403; no late True; public-boundary. |

Deviations called out: foreign-owned `managed:true` (same as RP-008); `wrote` ignored after retract.

### Pass 4 — Fable 5.1 blast-radius at `43a2d93f` vs `98840fb` (**LIMITED**)

Promise delivered. **HCC-only prod rollout is OK** (WRC reads the annotation, not the retired condition). Defects **A** (foreign leak / eternal obligation) and **B** (`RuntimeNotDesired` on foreign at `1d2b4ce7`) **fixed** at `43a2d93f`.

Cross-layer:

1. Bounded window `DeploymentReady=True` while delete is in flight (two API RTTs). Pre-PR this window was **unbounded**. Control UI cannot false-green: stale True LTT is before rotation cutoff.
2. Control UI does **not** understand `RuntimeNotDesired`; rotation poll only preserves `RolloutIncomplete` diagnostics → generic 180s timeout. Same operator outcome as pre-PR (then stale True → timeout). **Optional UX later; not this PR; not a rollout blocker.**
3. At `43a2d93f`, `wrote===false` still set `retracted:true` (became residual **D**).
4. No remaining reader of the `NetworkReady` **condition**. Annotation handshake untouched. CRD `reason` is a free string. No CRD/WRC/Control UI/control-api deploy required for correctness.

Residuals at `43a2d93f` (user then ordered: fix all Medium; cheap Lows only; expensive Lows to issue only if they deserve it):

| ID | Sev | Finding | Disposition |
|---|---|---|---|
| **C** | Medium | Poll passed `onDecline: 'silent'` for **both** retirement **and** revision/authority fence. A tick whose fence flips mid-write without retirement lost `True` without an obligation. | **FIXED** at `ab7ed881`. `StatusWriteDeclineArg` may be a factory. `pollWriteDecline = () => retiredNow() ? 'silent' : 'record-obligation'`. Evaluated **inside** `writeStatusCondition` at decline time, not at `await` call time. Test: hold ready-tick status GET, flip **same** `isCurrent` closure to false **without** fail-closed/retire, release → no `True` write, `hasIncompleteReconciliation()===true`. T12/T17 must stay: after retract + silent retired tick, obligation **false**. |
| **D** | Medium | After Deployment `'gone'`, retract ignored `wrote`. Failed `RuntimeNotDesired` patch still `retracted: true` → `performReconcile` could write SecretResolved/Ready and look complete. | **FIXED** at `ab7ed881`. `writeStatusCondition` returns `'wrote' \| 'unchanged' \| 'declined'`. `'unchanged'` = retract success (already RuntimeNotDesired). `'declined'` / I/O → still `retireReadinessWindow` + siblings, return `{ retracted:false, cleanupComplete:false, foreignRuntime:false }` so Ready is not written. `updateStatusConditions` logs only on `'wrote'`. Test: fail-closed, delete OK, patch throws 500 **only** when body has `RuntimeNotDesired` → delete happened, no Ready/`SecretValidationFailed`, obligation true. (Attempted patch still appears in `mock.calls`; do not assert the call was absent.) |
| **E** | Low | `readinessRetirements` never evicts. | **NOT implemented. NOT filed.** By design (name reuse must fail stale ticks). Audit said “acceptable, note only.” Eviction is not cheap. |
| **F** | Low | `waitForStatusCondition` doc said “strictly after `sinceMs`”; code is `transitionMs < opts.sinceMs` (accepts equal). Control UI uses `>`. | **FIXED cheaply** at `ab7ed881`: docstring → “at or after `sinceMs`”. **Did not** change `<` to `<=` (same-ms flake). |

Race (poll tick vs retract) at blast-radius: **no defect**. Retract writes status then bumps epoch; patches `test` `resourceVersion`; tick re-evaluates `publishIfCurrent`; 409/422 loser retries then fails epoch and declines.

---

## Evidence matrix you must re-pin on `ab7ed881`

Re-open any ID that is green only because the assertion is vacuous.

- **P1** — Deployment delete OK + Service deny or 403 → `RuntimeNotDesired` persisted.
- **T6'** — Deployment still present / delete denied → 0 retract, 0 status patches, `isCurrent` true except the single deny; obligation true.
- **P2** — still the weakest witness: flap after Deployment gone during retract write. Code should leave obligation / no orphan True. If still no named test, say so (do not invent coverage).
- **T17'** — after retract, Deployment reads do not grow across 24 ticks; obligation false after silent retired tick.
- **C (new)** — fence flip **without** retirement → obligation true, no True write.
- **D (new)** — 500 on RuntimeNotDesired body → no Ready/`SecretValidationFailed`, obligation true, delete happened.
- **P3** — recovery `WaitingForReplicas.LTT > retiredAt` (and later True LTT > retiredAt). Recovery secret helper is `storedSecret('fixture-key')`, **not** `'rotated-key'`.
- **T10** — WRC-owned (`managed:false`): no delete, no `DeploymentReady`.
- **RP-008 / foreign** — `managed:true` + WRC-labeled Deployment: no delete, no `RuntimeNotDesired`, **does** write Secret/Disabled/Ready as appropriate, obligation **false**.
- **T7** — Deployment 403 → no retract.
- **T11/T12/T15** — no late True / RolloutIncomplete from window-1.
- **P6** — no plus-line `cGFzcw==`; public-boundary PASS vs `98840fb`.
- **F** — helper docs say at-or-after; comparison unchanged.

e2e (in this PR): `tests/e2e/integration/mcp-runtime-conditions.test.ts` + helpers `waitForStatusCondition`, `deleteMcpSecret`, `postMcpSecret`, `putMcpServer`, `requireHccDevModeUnset`. Listed in `scripts/e2e/run-vitest-e2e.sh` after `mcp-secret-rotation-remote.test.ts`. Mutations go through control-api, not kubectl Secret writes (except `kubectlApply` fixture in `beforeAll`).

---

## What you must not regress

- Exhaustive `switch` + `never` on `RuntimeWithdrawKind` and `DeploymentDeleteOutcome`.
- Do not bump LTT on `SecretResolved` message-only changes.
- Do not change `findFreshDeploymentReady` / teach UI to accept stale LTT.
- Do not require a WRC or Control UI deploy for this HCC-only fix.
- HCC-only rollout: one extra status patch per server that still has leftover `NetworkReady` (strip). LTT on that strip stays if reason/message unchanged → no false-fresh True during an in-flight UI rotation.
- `ReadError` / transient Secret read still **preserves** runtime (`shouldFailClosedForSecretFailure` unchanged).
- Pre-existing `console.log` in `performReconcile` is out of scope.
- Worktree-only GFS/desktop dirt is **not** this PR.

---

## Closing questions (answer each)

1. Do RP-001…RP-008 and residuals C/D/F still hold as **FIXED** on `ab7ed881`, with tests that would go red if reverted?
2. Is P2 still an evidence gap only, or is there now a real liveness hole?
3. Any **new** High/Medium on the C/D `StatusWriteOutcome` / decline-factory change (including T12/T17 silent-after-retire)?
4. Blast radius still **LIMITED** / HCC-only rollout still safe?
5. Anything that deserves an issue (not E)? If yes, write the issue text; do not open it unless it is clearly warranted.
6. Final verdict: **CERTIFY CLOSE** or **NO** with the remaining defect list.

Please walk the current `retractManagedRuntime`, `writeStatusCondition`, poll `pollWriteDecline`, and the new C/D tests by line, not by commit message.

host-context-controller/src/reconciler.ts
```diff
@@ -2196,13 +2409,12 @@ ${authHeaderLines ? '\n        # ── Credential auth headers (envsubst-resolv
...
-          !(await this.deleteRuntimeResources(server.name, server.namespace, async () =>
-            isCurrent()
-          ))
-        )
-          return false
+      const withdrawRuntime = shouldFailClosedForSecretFailure(secretResult.reason)
+      let cleanupComplete = true
+      if (withdrawRuntime) {
+        const retract = await this.retractManagedRuntime(server, 'fail-closed', isCurrent)
+        if (!retract.retracted && !retract.foreignRuntime) return false
```
jozer-rami: **Major: a declined retract write leaves the server reported as ready after its Deployment is deleted.**

When the `DeploymentReady=False/RuntimeNotDesired` write in `retractManagedRuntime` comes back `'declined'` while `isCurrent()` is still true, `retracted` is false (`:1885-1887`) and we return here. That happens on a non-404 status GET error (`:2168-2172`), a patch error, or 3× 409/422 conflicts. By this point the Deployment is already deleted, but we skip:
- `setStatus({ deployed: false, ready: false })`
- `SecretResolved=False`
- `Ready=False/SecretValidationFailed`

On `dev` these always ran after cleanup; only a denied delete returned early.

**Trigger:** delete the envSecret of a healthy managed server. The generation doesn't change, so `statusMap` keeps `ready:true`, and `k8sClient.ts:2726` keeps serving it to mcp-host as authoritative. The CRD also keeps `Ready=True` / `SecretResolved=True`. The obligation recorded by `recordIncompletePublication` is only picked up by the full/initial convergence retry (`k8sClient.ts:3253`), not by watch events, so the stale state can last a long time.

The test at `reconciler.test.ts:2052` pins "skips Ready" but never checks `getStatus()`.

**Suggested fix:** on a declined write while still current, still run `setStatus` and the `SecretResolved`/`Ready` writes, and return `false` only so the obligation is kept. Then extend that test to assert `getStatus()` is not ready.

host-context-controller/src/reconciler.ts
```diff
@@ -1657,37 +1748,148 @@ ${authHeaderLines ? '\n        # ── Credential auth headers (envsubst-resolv
...
+      server,
+      {
+        type: 'DeploymentReady',
+        status: 'False',
+        reason: 'RuntimeNotDesired',
+        message: runtimeNotDesiredMessage(kind),
+      },
+      isCurrent
+    )
+    this.retireReadinessWindow(name)
```
jozer-rami: **Minor: a poll tick can overwrite `RuntimeNotDesired`.**

`retireReadinessWindow(name)` runs after the retract write resolves. On the fail-closed path the revision doesn't change, so the poll's `state.isCurrent()` stays true. Interleaving:

1. A tick is in its write phase (ready branch or exhausted branch).
2. The retract PATCH commits.
3. Before this await resumes, the tick's re-read sees the new resourceVersion and passes `publishIfCurrent()`, because the epoch hasn't been bumped yet.
4. The tick's PATCH succeeds and stores `DeploymentReady=True/ReplicasAvailable` (or `RolloutIncomplete`) with no Deployment.

Nothing corrects it afterwards. Status-only MODIFIED events don't reconcile (`k8sClient.ts:4192`), and the retract returned `'wrote'`, so no obligation remains. The window is narrow, but it breaks the guarantee this PR adds.

**Suggested fix:** retire the window as soon as the Deployment outcome is `'gone'`, before the write. A tick PATCH already in flight then fails the resourceVersion test op, and its retry declines silently via `retiredNow()`. A unit test that holds a tick's status read across the retract write (like `holdNextStatusRead`) would pin it. Note that moving the retire call earlier (before the delete) currently goes undetected by the suite, so an ordering test is worth adding either way.

jozer-rami (commented):
Reviewed at `0765ccfc`. Overall this is solid. Teardown scope and ownership/fencing match `dev`, nothing outside HCC reads the `NetworkReady` condition, and the unit tests hold up under mutation. Two things I'd like addressed before merge (inline):

1. **Major:** a declined `RuntimeNotDesired` write on the fail-closed path skips `setStatus` and the `SecretResolved`/`Ready` writes, so discovery keeps serving `ready:true` for a server whose Deployment was just deleted.
2. **Minor:** the readiness window is retired *after* the retract write, so an in-flight poll tick can overwrite `RuntimeNotDesired` with a stale verdict, and nothing corrects it.

alfredolopez80:
## Independent multi-agent review — PR #633 at HEAD `0765ccfc`

**Repo:** `evenfire-ai/evenfire` · **PR:** #633 · **Title:** fix(hcc): retract runtime conditions when HCC does not run a server
**Base ← head:** `dev` ← `fix/issue-606-hcc-runtime-conditions` · **Author:** @alfredolopez80
**HEAD reviewed:** `0765ccfcec79e2cd7eb6abb0d275c9fd2112d375`. This is a merge of `dev` into `ab7ed881`, and it changes none of the 12 PR files relative to `ab7ed881`.

**Findings:** Blocker 0 · Critical 0 · High 0 · **Medium 7** · Low 20

**Scope and method:**
- Seven reviewers ran in parallel: correctness, unit tests and mutations, cluster e2e harness, security and silent failures, cross-service contract and blast radius, concurrency, and maintainability/docs/deps.
- Each reviewer worked in its own detached worktree and checked every finding against HEAD.
- **REPRODUCED** means a throwaway test or mutation showed the defect. **READ-ONLY** means the finding comes from reading the code.
- Nothing was committed or pushed. No cluster was used.

**Verdict:** **Not ready to certify close as-is.**
- The main #606 promise holds on the disable path.
- It can still be broken on the **fail-closed** path by a narrow race (M1). Three reviewers reproduced it independently. It contradicts the earlier "race: no defect" conclusion.
- The fix is small, but it touches decision #10, so the owner has to decide.

---

### Gates run on HEAD

| Gate | Result |
|---|---|
| `host-context-controller`: `npm ci` (Node 24.18.0) | EXIT 0 |
| `npx tsc --noEmit` / `npm run build` | EXIT 0 / EXIT 0 |
| `npx vitest run` (full HCC suite, JSON reporter) | EXIT 0 — **110 files, 2398 tests, 2398 passed, 0 failed, 0 skipped**. The PR body says 2391; that number is stale. |
| The 3 PR test files, default order | EXIT 0 — 214/214 |
| The 3 PR test files, `--sequence.shuffle`, seeds 1–13 | **6/13 seeds red** on HEAD; 0/13 on merge-base versions (see M2) |
| `npm run lint` | EXIT 2 — there is no `eslint.config.*` for ESLint 9. This predates the PR, so no lint gate can run here. |
| `T2_PUBLIC_BASE_REF=98f5bdae… scripts/tests/test-minikube-t2-public-boundary.sh` | EXIT 0 — `PUBLIC_BOUNDARY_PASS` |
| `tests/e2e`: `npx tsc --noEmit` | EXIT 2 — 242 errors, same set as the merge base, **0 in PR files** |
| `bash -n scripts/e2e/run-vitest-e2e.sh` | EXIT 0 |
| `validate-release-version-bumps.mjs --previous 98f5bdae5` / desktop manifest `--validate-only` | EXIT 0 / EXIT 0 |
| Cluster e2e `integration/mcp-runtime-conditions.test.ts` | **NOT RUN.** It is not run in this review, not in any CI job, and the PR body box is unchecked. |

---

## Medium

### M1 — A poll tick can overwrite `RuntimeNotDesired` with `DeploymentReady=True` after the Deployment is deleted (fail-closed path)
Found independently by 3 reviewers (correctness, concurrency, security). **REPRODUCED** each time.

**Where:** `host-context-controller/src/reconciler.ts:1866-1876`. The retract write is at `:1866`; `retireReadinessWindow` runs at `:1876`, after the write resolves. The tick fences are at `:700-702`, `:770`, `:781`, `:2140`, `:2174` and `:2258`.

**Defect:** The retirement epoch changes only when HCC *processes the reply* to the `RuntimeNotDesired` PATCH, not when the API server commits it. A readiness tick can finish its status GET and PATCH inside that window:
- `publishIfCurrent()` is still true: a Secret change leaves the McpServer revision and `generation` unchanged, and the epoch has not been bumped yet.
- The `resourceVersion` test passes, because the tick read the retract's resourceVersion.

**Failure scenario:**
1. A managed server has an open readiness window, and a tick has seen the rollout converge.
2. The Secret is deleted. The fail-closed reconcile deletes the Deployment, and its `RuntimeNotDesired` PATCH commits on the server (rv N→N+1), but the reply is still in flight.
3. The tick's status GET returns rv N+1. Its PATCH `DeploymentReady=True/ReplicasAvailable` (plus `Ready=True/ReconcileSuccess`) passes every `test` op and commits.
4. The retract reply arrives, and the window is retired too late. The reconcile writes `SecretResolved=False` / `Ready=False` and clears the obligation.

**End state:** the Deployment is gone, but the CRD says `DeploymentReady=True/ReplicasAvailable` and nothing will retry. That is the #606 symptom. Status-only updates do not trigger a reconcile (`k8sClient.ts:4192`, `isMcpServerStatusOnlyUpdate`).

A variant also fails: the tick's held GET returns the old rv, its PATCH gets 422, and the retry re-reads and wins, because the fence has still not flipped.

**Evidence:**
- Throwaway vitest with a stateful fake status subresource that enforces the JSON Patch `test` ops (422 on mismatch, rv bump on commit). One variant commits the retract PATCH but delays its reply.
- On pristine HEAD the final `DeploymentReady` is `True/ReplicasAvailable`; expected `False/RuntimeNotDesired`.
- Liveness witnesses: delete called once, `RuntimeNotDesired` committed, the tick logged "is now ready".

**Disable path:** not affected against a real API server. The `enabled:false` edit bumps `generation`, so the stale tick's `test /metadata/generation` (`:2233-2238`) gets 422, and the retry is then refused by `isCurrent()`, which reads the informer cache (`:2321`). A fake that skips the generation `test` op shows the race there as well; the real API server does not.

**Suggested fix (verified):** Bump `readinessRetirements` right after the Deployment is confirmed `'gone'` and **before** the `RuntimeNotDesired` write. Keep timer clearing, map deletion and the log after the write.
- With only that change, the repro passes and the 3 PR suites stay green: 216/216, and 218/218 together with L2.
- Decision #10 still holds as written ("retire the window after the retract write").
- This is different from the rejected option ("move retire after delete so a poll 404-writes False"), because no poll write is introduced.
- **Owner decision required.** Please add the repro as a named test; it also covers the P2 scenario.

**This contradicts the prior closing audits:** the "Race (poll tick vs retract) … no defect … 409/422 loser retries then fails epoch" conclusion, and the Claude certification on `ab7ed881`.

### M2 — The new unit tests leak mock implementations and make the existing suite order-dependent
REPRODUCED (tests reviewer).

**Where:** `host-context-controller/src/reconciler.test.ts:195` (`installLiveStatus`) and the patch mock that throws on `RuntimeNotDesired` (around `:2056-2080`). `beforeEach` → `vi.clearAllMocks()` at `:643`, `:1892`, `:2605`.

**Defect:** `vi.clearAllMocks()` clears recorded calls but not `mockImplementation`, so the stateful fake status store and the throwing patch mock carry over into later tests.

**Failure scenario:**
- Suppose "fail-closed writes RuntimeNotDesired when Service cleanup is denied…" runs just before "fails closed and removes managed runtime resources when secret validation fails" (`:746`). The stored conditions are already identical, every write returns `'unchanged'`, and `:786` fails with "Number of calls: 0".
- "skips cleanup when the existing runtime is WRC-owned" is broken the same way by the leaked throwing mock.

**Evidence:** `--sequence.shuffle`, seeds 1–13: 6 red on HEAD, 0 red with the merge-base test files. A probe showed the leaked conditions. CI passes only because it uses the declared order.

**Fix:** Build these mocks per test, or restore the default implementations in `afterEach`.

### M3 — Coverage gap: the disabled path can drop `cleanupComplete` and no test fails
REPRODUCED (mutation k1).

**Where:** `reconciler.ts:2398`, `return cleanupComplete && isCurrent()`.

**Failure scenario:** A disabled managed server's Deployment is deleted and `RuntimeNotDesired` is written, then the Service delete is refused once. If the line is changed to `return isCurrent()`, the obligation is cleared and the HCC-owned Service leaks with nothing left to retry.

**Evidence:**
- With k1 applied, the full suite still passes 2398/2398.
- A throwaway test passes on HEAD and fails under k1.
- The fail-closed twin (k2, `:2449`) is covered.

**Fix:** Add the disabled-path version of "fail-closed writes RuntimeNotDesired when Service cleanup is denied after Deployment delete", asserting `hasIncompleteReconciliation() === true`.

### M4 — Coverage gap: the foreign-Deployment branch can drop `cleanupComplete` and no test fails
REPRODUCED (mutation m).

**Where:** `reconciler.ts:~1855`, `cleanupComplete: cleanupComplete && isCurrent()` in the `'foreign'` case.

**Failure scenario:** A WRC-labelled Deployment sits next to an HCC-owned Service whose delete is refused. If the line is forced to `true`, the obligation is cleared and the Service leaks.

**Evidence:** With m applied, the full suite passes 2398/2398. A throwaway test fails under m.

**Fix:** Add that case to the `ownership-safe fail-closed cleanup` describe block.

### M5 — e2e: the "Deployment gone" check passes whenever kubectl fails, and does not prove ordering
REPRODUCED (e2e reviewer, without a cluster).

**Where:** `tests/e2e/integration/mcp-runtime-conditions.test.ts:120-133`.

**Defect:**
- `kubectlSafe('get deploy …') === null ? 'absent'` treats any kubectl error as "absent".
- The check runs in a 90 s `waitFor` *after* `RuntimeNotDesired` is observed, so it never proves "published only after the Deployment is gone".
- No absence check follows the disable step.

**Failure scenario:**
- One transient apiserver or RBAC error makes the step pass while the Deployment still exists.
- An HCC that publishes `RuntimeNotDesired` early and deletes later also passes.

**Evidence:** Probe run with `KUBECONFIG=<empty>` and a nonexistent context: `kubectlSafe=null verdict=absent`, exit 0. The same pattern already exists at `integration/channel-reader-via-api.test.ts:522-527`.

**Fix:** Use the throwing `kubectl(['get','deploy',X,'--ignore-not-found','-o','name'])`, require `''`, run it once right after `RuntimeNotDesired` is seen, and repeat it after the disable step.

### M6 — e2e: the disable PUT can get a 409 because HCC is still writing status
READ-ONLY.

**Where:** `tests/e2e/integration/mcp-runtime-conditions.test.ts:143-164`.

**Defect:**
- The test reads `resourceVersion` right after `DeploymentReady=True` and sends it as the PUT precondition.
- control-api does not retry when a reader `resourceVersion` is supplied (`control-api/src/services/resourceService.ts:322-325`, `maxAttempts = readerResourceVersion ? 1 : 3`).
- HCC writes `Ready=True/ReconcileSuccess` right after `DeploymentReady=True` (`reconciler.ts:2557-2586`).

**Failure scenario:** That follow-up write lands between the GET and the replace. The test fails with `PUT disable failed` for a reason unrelated to #606.

**Fix:** First wait for `Ready=True/ReconcileSuccess` and `SecretResolved=True` with `sinceMs: restoredAt`. This also adds the missing Ready/SecretResolved recovery assertion. Alternatively, drop the `resourceVersion` precondition.

### M7 — e2e: the new cluster suite has never run and no CI lane runs it
READ-ONLY.

**Where:** `scripts/e2e/run-vitest-e2e.sh:37`. `.github/workflows/ci-public.yml:323` runs only `E2E_VITEST_SUITE_GROUP=node-unit`. The suite runs only through `make test-e2e-vitest` / `test-e2e-all` / `validate-all` / `test-integration`, and no T0/T1/T2 script calls it.

**Failure scenario:** M5, M6 or an HCC regression shows up only when someone runs the suite by hand. The PR body's own checkbox is unchecked.

**Fix:** Before merge, run `bash scripts/e2e/run-vitest-e2e.sh integration/mcp-runtime-conditions.test.ts` on the exact HEAD and attach the log, test count and exit code.

---

## Low

| ID | Where | Defect / scenario | Evidence |
|---|---|---|---|
| L1 | `reconciler.ts:1851-1857` | **Pre-existing, not introduced by this PR.** Fail-closed with a *foreign* Deployment does not retire an earlier readiness window on the same generation. About 5 s later the tick overwrites `Ready=False/SecretValidationFailed` with `True/ReconcileSuccess` while `SecretResolved=False`. Fix: call `retireReadinessWindow(name)` in the `'foreign'` branch, with no `RuntimeNotDesired` write, so decision 9 still holds. | REPRODUCED |
| L2 | `reconciler.ts:2166, 2172, 2274, 2290, 2293` | A retired tick still records an obligation on status GET error, 404, a non-retryable PATCH error, or exhausted 409/422. These paths call `recordIncompletePublication` directly instead of `decline()`. Result: `hasIncompleteReconciliation()===true` for a server that was withdrawn on purpose, and the NetworkPolicy recovery decision reports `work-pending` (`k8sClient.ts:3112`). Fix: `return decline()` on those 5 paths; non-poll callers keep the `'record-obligation'` default. | REPRODUCED |
| L3 | `reconciler.ts:2174` + `:2332` | A tick that declines while the disable retract reply is in flight replaces the reconcile's obligation object, so the finished reconcile cannot clear it. Not a regression (it self-heals via `k8sClient.ts:3253`). The M1 fix also resolves it. | REPRODUCED |
| L4 | `reconciler.ts:2449`, test `reconciler.test.ts:1948` | No test asserts that a transient Secret `ReadError` keeps the obligation. The mutation `return cleanupComplete && isCurrent()` survives 214/214. Fix: add `expect(reconciler.hasIncompleteReconciliation()).toBe(true)`. | REPRODUCED |
| L5 | `reconciler.ts:2198` | Removing the `prior.message !== condition.message` clause survives the suite, because the lastTransitionTime bump test only changes `reason`. Fix: add a same-reason, message-only `DeploymentReady` case. | REPRODUCED |
| L6 | `reconciler.ts:1877-1882` | "Retracted managed runtime" is logged at info level even when `retractWrite === 'declined'` (the function returns `retracted:false`). `wrote:false` cannot distinguish unchanged from declined. The `?? 0` on `epoch` can never apply. Fix: log `outcome: retractWrite` and warn on `'declined'`. (2 reviewers) | READ-ONLY |
| L7 | `tests/e2e/integration/mcpCredentialRotation.helpers.ts:361-370` | `requireHccDevModeUnset` passes silently when it cannot check: `kubectlSafe` returns `null` on any error; it matches only `=== 'true'` while HCC also accepts `1`/`TRUE` (`config.ts:401-405`); and it reads only `containers[0].env`, not `envFrom`. Fix: use the throwing `kubectl`, select the container by name, match HCC's boolean rule. (4 reviewers) | REPRODUCED (kubectl-error case) |
| L8 | `reconciler.ts:1757, 1760, 1765, 1776, 1779, 2080` | CLAUDE.md logging standard: the PR rewrote `deleteDeploymentIfHccOwned` and `updateStatusConditions` but kept direct `console.*` calls, some passing a raw `ApiException` whose message includes the response body and headers. That is k8s Status data, not Secret data. Fix: use `hccLogger` in those two functions only. (2 reviewers) | READ-ONLY |
| L9 | `reconciler.ts:1784-1789`, `1723-1736`, `1795-1809`, `1831-1845` | The "Withdraw an HCC-owned runtime…" docstring sits on `deleteHccOwnedRuntimeSiblings` instead of `retractManagedRuntime`. The ConfigMap+Service cleanup loop now exists 3 times. The `return` at `:1845` can never run because `throwCleanupFailures` always throws there (blame `43a2d93f`). (2 reviewers) | READ-ONLY |
| L10 | `reconciler.ts:770, 781, 812` | The three `if (retiredNow()) return` guards are redundant with `publishIfCurrent` and the decline factory. Removing all three keeps 2398/2398 green. Keep one mechanism. (2 reviewers) | REPRODUCED |
| L11 | `reconciler.ts:153, 2050, 2109-2112`; `397-408` | `StatusWriteDeclineArg` accepts a value or a factory, but only the factory is ever passed, and a static `'silent'` would be evaluated too early. The `retireReadinessWindow` docstring still says ticks pass `onDecline: 'silent'`. Fix: type it as `() => StatusWriteDecline` and correct the docstring. | READ-ONLY |
| L12 | `reconciler.ts:142`; `docs/architecture/overview.md:901, 936` | "Strips `NetworkReady` from every status write" is true only of `McpServerReconciler.writeStatusCondition`. `networkPolicyReconciler.ts:3131-3137, 3185` carries it forward (harmless: the next McpServer write strips it). The docs say `RuntimeNotDesired` is written for "disabled or fail-closed", but it is written only for `managed:true` servers whose HCC-owned Deployment is confirmed gone. (2 reviewers) | READ-ONLY |
| L13 | `reconciler.ts:2085-2104` | The `writeStatusCondition` docstring omits retired-type stripping, the forced write when a retired type is present, `onDecline`, and the `'wrote'/'unchanged'/'declined'` return. It still says "Other errors → log + swallow". | READ-ONLY |
| L14 | `control-ui/lib/api.ts:902-908` (outside the PR files) | The `McpServerCondition` doc says `lastTransitionTime` advances only on a status change. For `DeploymentReady` the PR now also advances it on a reason or message change. No runtime break today (`findFreshDeploymentReady` needs a fresh `True`, and `True` has exactly one reason and message), but the comment is now wrong. | READ-ONLY |
| L15 | `mcpCredentialRotation.helpers.ts:399-401, 416, 440-447`; `mcp-runtime-conditions.test.ts:7-9, 92` | The `waitFor` label says "after" but the predicate accepts "at or after". The new `waitForRolloutCondition` docstring copies a **Spanish** reference to an external plan ("plan Fase 3, requisito 6 / hallazgo 1", first authored in `e2ea9050ec`), which breaks the English-only rule for repository comments. The test header explains `SecretMissingKey`, which the test never exercises. `ready.reason` is asserted twice. | READ-ONLY |
| L16 | `mcp-runtime-conditions.test.ts:94, 135, 155`; helpers `:424` | `sinceMs` uses the host clock, while `lastTransitionTime` comes from the HCC pod clock inside the minikube VM. If the VM clock is behind, a wait times out after 90 s (a flake, not a false pass). Fix: derive `sinceMs` from the previous condition's server timestamp. | READ-ONLY |
| L17 | `mcp-runtime-conditions.test.ts:46-51` | `(status?.conditions ?? [])` passes `not.toContain('NetworkReady')` when no status was read. Fix: in the same helper, also assert that `DeploymentReady` is present. | READ-ONLY |
| L18 | `mcp-runtime-conditions.test.ts:182` | The test timeout is `ROLLOUT_TIMEOUT_MS*4+30s` = 390 s, but the test runs 8 sequential waits of up to 90 s each. A slow legitimate run hits vitest's generic timeout, and the `waitFor` diagnostic dump is lost. | READ-ONLY |
| L19 | `src/__tests__/reconciler.credentialsRotation.test.ts:~814/834, ~879/898` | `vi.spyOn(console,'warn')` is restored at the end of the test body, not in `finally`, so a failing assertion leaks the spy into later tests. | READ-ONLY |
| L20 | PR body; `host-context-controller/package.json` | The PR body says 2391 tests; HEAD runs 2398. `npm run lint` cannot run (no ESLint 9 config; pre-existing). The version range against the merge base is HCC 0.1.322→0.1.326 and e2e 0.1.79→0.1.81, consistent with the per-commit pre-commit bump. | REPRODUCED |

---

## Verdicts on the prior review threads

**Claude closing audits on `ab7ed881`** ([comment 2](https://github.com/evenfire-ai/evenfire/pull/633#issuecomment-5671802170), [comment 4](https://github.com/evenfire-ai/evenfire/pull/633#issuecomment-5671816103)):

- **RP-001…RP-008, C, D, F — confirmed FIXED on HEAD, with tests that fail when the fix is reverted.**
  - a1/a2 (NetworkReady strip): 11 and 9 tests fail.
  - b (write before gone): 7 fail.
  - c (`pollWriteDecline` always `'silent'`): 1 fail, the C test.
  - d (declined treated as retracted): 1 fail, the D test.
  - e (`'foreign'`→`'gone'`): 2 fail.
  - f (LTT bump for all types): 1 fail.
  - g/g2 (no retire / epoch check always false): 4 and 3 fail.
  - RP-001: `PUBLIC_BOUNDARY_PASS` reproduced; no `cGFzcw==` plus-line.
- **P2 — behavior correct, still no named test.**
  - Throwaway tests flipped `isCurrent` at the status read, before the PATCH, and permanently, on both fail-closed and disabled, including one case with a live poll.
  - Result: 7/7 pass. Write declined, no `Ready`/`RuntimeNotDesired`, obligation kept, poll retired, and a later reconcile publishes and clears.
  - Today mutation d is caught by only one test (the 500 case). The M1 repro plus these cases would close P2.
- **"Race (poll tick vs retract): no defect" — refuted for the fail-closed path.** See M1. The 409/422 retry does not close the window, because the epoch fence has not flipped when the retry re-reads.
- **Blast radius LIMITED / HCC-only rollout safe — confirmed.**
  - No non-test reader of the `NetworkReady` *condition* remains; WRC reads the `clerum.io/network-ready` annotation (`workflow-recipes/src/reconciler/mcpDelegation.ts:1184-1194`).
  - The CRD `reason` is a free string (`charts/clerum-crds/crds/mcpserver.yaml:562-583`).
  - HCC runs `replicas: 1` with `strategy: Recreate`, so an old pod and a new pod never write at the same time.
  - `managed` is immutable (CEL rules `mcpserver.yaml:483-488`), so `DeploymentReady` has a single writer for the life of each object.
  - Control UI cannot show a false green from a stale `True` (its LTT precedes the cutoff). Under M1 the fail-closed window can produce a *fresh* `True` for a rollout that really converged moments before; that is not a regression.
- **"Anything that deserves an issue?"** Not as a new issue: M1 and M2 belong in this PR. L1 (pre-existing foreign-poll leak) could be a follow-up.
- **Decisions already taken were respected.** Residual E, `<` vs `<=`, public-boundary, the Control UI `RuntimeNotDesired` UX and the pre-existing `performReconcile` `console.log` were not re-reported. M1's fix is presented as an owner decision because it changes where the epoch is bumped relative to decision #10.

- **Ownership-safe delete** (`reconciler.ts:1739-1782`): GET 404 → `'gone'`; unowned → `'foreign'`; `deleteAllowed` false → `'denied'`; DELETE 2xx/404 → `'gone'`. Other GET/DELETE errors throw after sibling cleanup. Label spoofing requires Deployment write RBAC in `mcp-server`, which only HCC and workflow-recipes hold.
- **RBAC:** no new verbs needed.
- **Secret handling:** no Secret names or values in the new status messages or logs. E2E Secrets go to kubectl on stdin via `execFileSync`; the admin cookie is sent in a header and never logged.
- **Transient `ReadError`:** still preserves the runtime (`mcpServerSafety.ts:96-100`). `SecretNotFound`, `SecretMissingKey` and `SecretAccessDenied` withdraw it.
- **Exhaustive `switch` + `never`:** intact on `RuntimeWithdrawKind` (`:169-176`) and `DeploymentDeleteOutcome` (`:1860-1863`).
- **LTT rule:** bumps only for `DeploymentReady` on a reason or message change; a `SecretResolved` message-only change and a generation-only bump both keep the LTT.
- **Reconcile serialization:** per server via `inFlight` (`:2326-2336`) and the keyed queue in `k8sClient.ts:1250-1258`. A re-enable during a retract waits, and a new window snapshots the bumped epoch. Delete and recreate is fenced by the uid `test` op and `beginStatusTracking`.
- **Lockfiles** in `host-context-controller` and `tests/e2e`: only the root `version` fields changed; no dependency, `resolved` or `integrity` changes.
- **E2E runner:** fails on a missing listed file or zero executed tests; `passWithNoTests: false`; the new suite is registered. `afterAll` cleanup runs even if `beforeAll` throws.
- **Tests:** no `.skip`, `.only` or `.todo` in changed test files. Negative assertions have liveness witnesses or were shown non-vacuous by mutation. The log-skip tests fail when the log is made unconditional.

## Merge recommendation
- **Fix before merge:**
  - **M1**, owner decision on epoch placement.
  - **M2**, test isolation.
  - Run the cluster e2e once (**M7**) after fixing **M5**/**M6**, and attach the log.
- **Should fix, can be a follow-up:** M3 and M4 (add the two missing tests), L2, L4 and L5.
- **Optional:** L1 (pre-existing) and the rest of the Lows (comments, docs, logging, duplication).

<sub>This is an independent read-only review. Nothing was pushed, merged or approved, and no cluster was used. Throwaway repro tests are kept outside the repository.</sub>

alfredolopez80:
@claude closing **audit + re-review** of every finding already raised on this PR. Do not treat any earlier short ping as the brief. Independently re-read producer + tests at **exact HEAD `d7fea87bf83fd1638e0272601463e3ce03d77451`** (`fix/issue-606-hcc-runtime-conditions`, HCC `0.1.328`, e2e `0.1.82`). Issue **#606**. PR **#633**.

This comment is the source of truth for what was claimed fixed, adjusted, rejected, or deferred. Please either **CERTIFY CLOSE** or name remaining defects with severity, `file:line`, and a failing scenario. Do **not** weaken public-boundary. Do **not** merge. Do **not** treat `T2_PREFLIGHT_PASS` or an unchecked cluster-e2e box as a close blocker if the unit/contract promise holds.

---

## What this audit must do

Audit **all** of the following as one ledger, not a glance at the latest commit:

1. Product promise of #606 (still holds?).
2. Owner closing audit on `ab7ed881` ([comment 5671813968](https://github.com/evenfire-ai/evenfire/pull/633#issuecomment-5671813968)) and Claude replies ([5671802170](https://github.com/evenfire-ai/evenfire/pull/633#issuecomment-5671802170), [5671816103](https://github.com/evenfire-ai/evenfire/pull/633#issuecomment-5671816103)) — RP-001…RP-008, residual C/D/F.
3. @jozer-rami review [5207994554](https://github.com/evenfire-ai/evenfire/pull/633#pullrequestreview-5207994554) at `0765ccfc` — Major + Minor.
4. Independent multi-agent review at `0765ccfc` ([comment 5677823210](https://github.com/evenfire-ai/evenfire/pull/633#issuecomment-5677823210)) — M1–M7 and L1–L20.
5. Commits after those reviews: `998b01ac` (Jozer) and `d7fea87bf` (extra-review follow-up).

For each item, say **FIXED** (test would fail if reverted), **ADJUSTED** (intentional design change, still correct), **REJECTED-WITH-REASON** (must stay rejected), or **FOLLOW-UP** (issue filed, not a #633 merge gate). If you disagree with an owner disposition, say so explicitly.

---

## HEAD and commits to read

| SHA | What it did |
|---|---|
| `1d2b4ce7` | Initial retract of runtime conditions |
| `d618d10f` | Persist `RuntimeNotDesired` only after Deployment gone; public-boundary fixture (`storedSecret`) |
| `43a2d93f` | Foreign vs denied (`'foreign'`); keep Secret/Ready/Disabled |
| `ab7ed881` | Write/poll decline outcomes; residual C/D |
| `0765ccfc` | Merge `dev` (no PR-file change vs `ab7ed881`) |
| `998b01ac` | Jozer: keep discovery not-ready when retract write declines; retire **before** retract PATCH |
| **`d7fea87bf` (audit this)** | Extra-review: mock isolation, decline() on status errors, foreign retire, disabled/foreign obligation tests, e2e absence/Ready waits, cheap lows |

Producer: `host-context-controller/src/reconciler.ts`  
Tests: `host-context-controller/src/reconciler.test.ts`, `host-context-controller/src/__tests__/reconciler.credentialsRotation.test.ts`  
E2E (not a CI/T2 gate): `tests/e2e/integration/mcp-runtime-conditions.test.ts`, `tests/e2e/integration/mcpCredentialRotation.helpers.ts`  
Docs: `docs/architecture/overview.md` G11

---

## Product promise (#606) — must still hold

HCC must:

1. Stop writing `NetworkReady`; strip it on **every** `McpServerReconciler.writeStatusCondition` (`RETIRED_STATUS_CONDITION_TYPES = ['NetworkReady']`). `NetworkPolicyReconciler` may still echo a leftover until the next McpServer write (L12 — docs now say this).
2. On **managed** disable / fail-closed Secret failure: delete the **HCC-owned** runtime and persist `DeploymentReady=False/RuntimeNotDesired` only after the Deployment is **confirmed-gone** (`'gone'`).
3. Leave WRC-owned (`managed:false`) alone: no delete, no `DeploymentReady`.
4. Keep the WRC handshake annotation `clerum.io/network-ready`.
5. Keep the public signal `False/RuntimeNotDesired` (do **not** drop the type).
6. Single helper `retractManagedRuntime(server, kind, isCurrent) → { retracted, cleanupComplete, foreignRuntime, runtimeGone }`.
7. Sibling Service/ConfigMap failure must **not** skip retract once the Deployment is gone.
8. If the Deployment is still present and delete is **denied**: no retract, incomplete, **zero** status patches (T6').
9. If the runtime is **foreign** (present, not HCC-owned): **no** `RuntimeNotDesired`; still publish Secret/Ready/Disabled. After L1, the foreign branch **does** call `retireReadinessWindow` so a leftover poll cannot republish `Ready=True` on the same generation.
10. **UPDATED (Jozer Minor / extra-review M1) — do not apply the old decision #10 from `ab7ed881`.** `retireReadinessWindow` now runs **after** Deployment `'gone'` (or on `'foreign'`) and **before** the `RuntimeNotDesired` write. This is intentional. The rejected option remains: do **not** retire *before delete* (that reopens the BASE stale-`True` race).
11. Bump `lastTransitionTime` on `DeploymentReady` when **reason or message** changes (L5 now has a message-only test). Not all types; not `SecretResolved` message-only.
12. One `withdrawRuntime` boolean per fail-closed branch.
13. Distinguish `'gone' | 'denied' | 'foreign'`.
14. **Jozer Major / `998b01ac`:** a declined retract write after the runtime is gone must **not** skip discovery writes. `shouldStopAfterIncompleteRetract` returns false when `runtimeGone && isCurrent()` so `setStatus` + `SecretResolved`/`Ready`/`Disabled` still run; reconcile still returns `false` to keep the obligation (`retracted:false`, `runtimeGone:true`).

Messages (unchanged):

- disabled: `McpServer is disabled; HCC does not run its runtime`
- fail-closed: `Env Secret validation failed; HCC does not run its runtime`

---

## Ledger — please tick every row

### A. Residual / RP items from the `ab7ed881` closing audit

These were already certified FIXED on `ab7ed881`. Confirm they are **still** FIXED on `d7fea87bf` (no regression from Jozer + extra-review):

| ID | Claim | Expected owner status |
|---|---|---|
| RP-001 | No new `cGFzcw==` plus-line; scanner not taught; BASE fixture stays | FIXED |
| RP-002 / a1-a2 | `NetworkReady` stripped on every McpServer status write | FIXED |
| RP-003 / b | No `RuntimeNotDesired` before Deployment `'gone'` | FIXED |
| RP-004 / c | Poll `onDecline` factory: silent **only** when `retiredNow()` | FIXED (type is factory-only now, L11) |
| RP-005 / d | Declined retract is not treated as successful retract | FIXED, then **ADJUSTED** by Jozer: still `retracted:false`, but discovery not-ready is published when `runtimeGone` |
| RP-006 / e | `'foreign'` ≠ `'gone'` | FIXED |
| RP-007 / f | LTT bump is DeploymentReady reason **or** message, not all types | FIXED + L5 message-only test |
| RP-008 / g | Foreign still writes Secret/Ready/Disabled; no RuntimeNotDesired | FIXED |
| C | Silent decline only if retired at decline time | FIXED |
| D | Failed RuntimeNotDesired must not look like successful persist; after Jozer, discovery still not-ready | FIXED / ADJUSTED |
| F | Cheap: fail-closed D test asserts SecretResolved=False only (not merged True leftover) | FIXED |

### B. @jozer-rami at `0765ccfc` — landed in `998b01ac`

| ID | Finding | Disposition on `d7fea87bf` | Where to verify |
|---|---|---|---|
| Jozer Major | Declined `RuntimeNotDesired` skipped `setStatus` + Secret/Ready, so mcp-host could keep `ready:true` after delete | **FIXED** | `shouldStopAfterIncompleteRetract`; fail-closed / disabled D tests assert `getStatus()` not-ready **and** `hasIncompleteReconciliation()===true` |
| Jozer Minor | `retireReadinessWindow` ran *after* retract PATCH; tick could overwrite `RuntimeNotDesired` | **FIXED** | retire on `'gone'` **before** `writeStatusCondition`; named test `in-flight ready tick cannot overwrite RuntimeNotDesired while retract write is in flight` |

### C. Extra-review M1–M7 at `0765ccfc` — landed in `998b01ac` + `d7fea87bf`

| ID | Finding | Disposition | Notes |
|---|---|---|---|
| M1 | Tick can overwrite `RuntimeNotDesired` after delete (fail-closed) | **FIXED** in `998b01ac` (same as Jozer Minor). L3 (obligation object replaced mid-reply) is resolved by the same retire-before-write. |
| M2 | `vi.clearAllMocks()` leaked `installLiveStatus` / throwing patch mocks; shuffle red | **FIXED** in `d7fea87bf` | Recreate `appsApi`/`coreApi`/`customApi` in `beforeEach` of leaky describes. Shuffle seeds 1,2,3,7,11: 121/121 green on this HEAD. |
| M3 | Disabled path can drop `cleanupComplete`; no test | **FIXED** | `disabled writes RuntimeNotDesired when Service cleanup is denied after Deployment delete` asserts obligation. |
| M4 | Foreign + denied HCC Service can drop `cleanupComplete`; no test | **FIXED** | `keeps the obligation when a foreign Deployment sits next to a denied HCC Service`. |
| M5 | e2e treated any kubectl error as Deployment absent; no post-disable absence | **FIXED** | `requireDeploymentAbsent` uses throwing `kubectl … --ignore-not-found -o name` and requires `''`, once after fail-closed retract and again after disable. Not inside a 90s `waitFor`. |
| M6 | Disable PUT could 409 because Ready/SecretResolved writes were still in flight | **FIXED** | Wait `Ready=True/ReconcileSuccess` + `SecretResolved=True` (server-side `sinceMs`) before the disable PUT. |
| M7 | Cluster e2e never run; not in CI | **FOLLOW-UP** | [#636](https://github.com/evenfire-ai/evenfire/issues/636). **Not** a #633 merge gate. Do not fail the audit because the PR checkbox is unchecked. |

### D. Extra-review Lows

| ID | Disposition | What landed / why not |
|---|---|---|
| L1 | **FIXED** (cheap) | `retireReadinessWindow` in `'foreign'` branch; still no `RuntimeNotDesired`. |
| L2 | **FIXED** | Five status GET/PATCH error / exhausted-retry paths `return decline()` instead of always `recordIncompletePublication`. Retired ticks stay silent. |
| L3 | **FIXED** by M1 | No extra code. |
| L4 | **FIXED** | ReadError test asserts `hasIncompleteReconciliation()===true`. |
| L5 | **FIXED** | Same-status, message-only `DeploymentReady` LTT bump test. |
| L6 | **FIXED** | Log `outcome: retractWrite`; `warn` on `'declined'`; no `epoch ?? 0`. |
| L7 | **FIXED** | `requireHccDevModeUnset` uses throwing `kubectl`, container name `host-context-controller`, boolean rule matching `getEnvBool` (`true`/`TRUE`/`1`). |
| L8 | **FIXED** (scoped) | `hccLogger` only in `deleteDeploymentIfHccOwned` + `updateStatusConditions` catch. Do not demand a repo-wide console sweep. |
| L9 | **FIXED** (cheap) | Docstring is on `retractManagedRuntime`; siblings have a short delete-only doc; dead success-shaped `return` after `throwCleanupFailures` removed (now throws). Loop extract (3× sibling cleanup) **not** done — refactor risk, not a defect. |
| L10 | **REJECTED-WITH-REASON** | Keep the three `if (retiredNow()) return` guards **and** `publishIfCurrent` / decline factory. Redundant on purpose. Do not remove them in this PR. |
| L11 | **FIXED** | `StatusWriteDeclineArg = () => StatusWriteDecline` only. Defaults are factories. `retireReadinessWindow` docstring no longer says ticks pass a static `'silent'`. |
| L12 | **FIXED** (docs) | G11 now: managed + Deployment gone; strip is `writeStatusCondition`; NP reconciler may echo leftover. |
| L13 | **FIXED** | `writeStatusCondition` docstring covers retired-type strip, forced write, `onDecline`, `'wrote'/'unchanged'/'declined'`. |
| L14 | **REJECTED-WITH-REASON** | `control-ui/lib/api.ts` comment is **outside** this PR. No runtime break (`True` has one reason/message). Do not expand the PR. |
| L15 | **FIXED** | English-only comments; wait label “at or after”; header no longer explains unused `SecretMissingKey`; duplicate `ready.reason` assert removed. |
| L16 | **FIXED** | `sinceMs` derived from previous condition `lastTransitionTime` via `conditionTransitionMs` (HCC clock, not host `Date.now()`). |
| L17 | **FIXED** | `expectNoNetworkReady` requires `status.conditions` **and** `DeploymentReady` present. |
| L18 | **FIXED** | Test timeout `ROLLOUT_TIMEOUT_MS * 8 + 30_000` for the sequential waits. |
| L19 | **FIXED** | `console.warn` spies restored in `finally`. |
| L20 | **ADJUSTED** | PR body test count updated to **2403**. Lint-without-ESLint-9 is pre-existing, not a #633 defect. |

---

## Still rejected (must stay rejected)

- Retire the readiness window **before Deployment delete**.
- Teach Control UI / e2e to accept stale LTT (reopens #223).
- Bump LTT on every reason/message change for **all** condition types.
- Weaken public-boundary, add `cGFzcw==` plus-lines, teach the scanner, or add test files to `safe_source_paths`.
- Let an in-flight tick write `RuntimeNotDesired` (two writers).
- Evict `readinessRetirements` (recreated names must keep the epoch).
- Treat `'foreign'` as `'gone'`.
- Treat a declined retract as `retracted:true`.
- Skip discovery Secret/Ready/Disabled after a gone runtime just because the retract PATCH declined.
- Make cluster e2e / CI job #636 a merge blocker for this PR.
- Remove the three `retiredNow()` poll guards (L10).
- Edit `control-ui` comments in this PR (L14).

---

## Gates already run on `d7fea87bf` (do not re-litigate as missing)

- HCC `npx vitest run`: **110 files, 2403 passed, 0 failed, 0 skipped**.
- The two PR test files under `--sequence.shuffle` seeds **1, 2, 3, 7, 11**: **121/121** each (M2).
- `T2_PUBLIC_BASE_REF=98f5bdae571b35334e5b6d429bb59b410386bb5c` (`origin/dev` merge-base) → `PUBLIC_BOUNDARY_PASS`.
- Versions: HCC `0.1.327 → 0.1.328`, e2e `0.1.81 → 0.1.82` (pre-commit bump on `d7fea87bf`).
- Local T0/T1/T2 on **this** HEAD are being re-run after this comment; the last full `MINIKUBE_T2_PASS` was on an **earlier** SHA (`ab7ed881`). Do not CERTIFY CLOSE on the basis of that older T2. Do not fail the *code* audit solely because T2 on `d7fea87bf` is still in flight.

---

## Suggested verification (mutations / reads, not cluster)

Please actually break the claim, not just agree:

- Revert `shouldStopAfterIncompleteRetract` to “stop whenever `!retracted && isCurrent`” → Jozer Major / D tests must fail.
- Move `retireReadinessWindow` back to **after** the retract write → Jozer Minor / M1 tick-hold test must fail.
- Force foreign `cleanupComplete: true` → M4 test must fail.
- Force disabled `return isCurrent()` after sibling deny → M3 test must fail.
- Restore `vi.clearAllMocks()` + describe-scoped API objects in the leaky describes → shuffle should be able to fail again (M2).
- Replace the five `return decline()` error paths with `recordIncompletePublication` → a retired tick can record obligation (L2).
- Remove `prior.message !== condition.message` from the DeploymentReady LTT clause → L5 must fail.

---

## Required output shape

```text
VERDICT: CERTIFY CLOSE | NOT READY
HEAD: d7fea87bf83fd1638e0272601463e3ce03d77451

Ledger:
- RP-001 … F: FIXED|REGRESSED|…
- Jozer Major/Minor: …
- M1–M7: …
- L1–L20: …

Remaining defects (if any):
- SEV / file:line / scenario / why the owner disposition is wrong

Design notes:
- Confirm decision #10 update (retire after gone, before retract write) is correct and the “retire before delete” option stays rejected.
- Confirm declined retract + runtimeGone still publishes not-ready discovery and keeps the obligation.
```

If you need a single sentence of owner intent: **#606 is retract-after-gone, never lie that a deleted managed runtime is ready, never touch WRC-owned Deployments, and every review finding above is either pinned by a test, an intentional keep, or #636.**

jozer-rami:
## Review: major finding (fix before merge)

Reviewed at head `6b3d9d6b`. No blockers; the core retract/strip logic, poll-retirement fencing, and mutation coverage look solid. One major finding:

### `reconciler.ts` does not meet the repository logging standard

`CLAUDE.md` § Logging standard: *"When editing a file, replace its direct `console.*` calls with the existing service logger, preserve levels/context/redaction."* HCC has `hccLogger` for this.

This PR edits `host-context-controller/src/reconciler.ts` but converts only 8 of its direct `console.*` calls (the 6 in `deleteDeploymentIfHccOwned` and 2 in `updateStatusConditions`). **46 remain** (dev had 54), including several inside functions this PR changed:

- `writeStatusCondition`: L2200, L2205, L2308, L2323
- `performReconcile` disabled / fail-closed branches: L2392, L2413, L2420, L2448, L2459
- `pollReadiness` tick: L769, L813

The new retract flow also makes the logs inconsistent. `retractManagedRuntime` → `deleteHccOwnedRuntimeSiblings` calls `deleteConfigMapIfHccOwned` / `deleteServiceIfHccOwned` (L1934–L1992, 12 `console.*` calls). A single disabled / fail-closed withdraw therefore logs the Deployment step as structured `hccLogger` JSON and the ConfigMap/Service steps as plain `[Reconciler] …` text, so one withdraw can't be followed as a single structured trail. The `console.warn(..., error)` calls at L2205 / L2323 also print the raw k8s `ApiException`, which skips `hccLogger`'s error sanitization. That behaviour predates this PR, but these functions are now being edited.

**Suggested fix:** move the `console.*` calls in the touched functions (`writeStatusCondition`, the `performReconcile` disabled/fail-closed branches, the poll tick) and the two sibling-delete helpers (`deleteConfigMapIfHccOwned`, `deleteServiceIfHccOwned`) to `hccLogger` with `{ name, err }` context, keeping the current levels. Existing tests that spy on `console.warn` for these messages (e.g. `reconciler.credentialsRotation.test.ts:958`, which waits on `'not ready after 24 polls'`) would need to key off state instead.

zach88:
Took a deep pass through this one — the retirement epoch fencing, the `retractManagedRuntime` state machine, and the `writeStatusCondition` return-type migration. It's a clean, well-tested change: the tricky concurrency (a stale readiness tick republishing `DeploymentReady=True` after a retract) is genuinely closed, and the security-sensitive part of #606 (dropping `NetworkReady` must not create a network fail-open) holds up because the real gate consumers read is the `clerum.io/network-ready` annotation, not the condition. Nothing here blocks merge; what I found is all optional polish.

---

**🟡 Approve with nits** — nothing blocking, a few optional items.

---

## Optional (non-blocking)

The PR can merge without any of these — they're follow-ups or author's call.

1. **[R1-L1] `NetworkReady` can linger via `NetworkPolicyReconciler`** — the retired-type strip only lives in `McpServerReconciler.writeStatusCondition`. `networkPolicyReconciler.ts:3017` (`mergeStatusCondition`) preserves the full condition array when it publishes `ExternalEgressReady` (`:3132`), so an inherited `NetworkReady=True` can persist on a McpServer that only ever gets writes from `NetworkPolicyReconciler`. You already call this out in the body and `docs/architecture/overview.md:936`, and it's harmless today — no one reads the condition (WRC reads the annotation at `mcpDelegation.ts:1226`; grep for the condition across control-api/control-ui/mcp-host/workflow-recipes/desktop-app = 0 readers), and the disabled/fail-closed paths strip it via the retract write anyway. If you want to close the intent 100%, apply the same strip in the `NetworkPolicyReconciler` merge. Fine as a separate issue.

2. **[R1-L2] Retract-race test goes red on a setup timeout, not an assertion** — `reconciler.credentialsRotation.test.ts:878` (`in-flight ready tick cannot overwrite RuntimeNotDesired while retract write is in flight`). Reproduced: against the merge-base this test fails via a 5003ms timeout, not an assertion — its `await retractPatchGate` (`:917`) only releases when the code writes `RuntimeNotDesired`, which base never does, so it hangs. It satisfies T3 formally (red on the prior head) but the red points at a hung gate rather than the behavior. Worth an explicit guard (`expect.poll` / short timeout on `retractPatchGate`) so a failure points at the cause.

3. **[R1-L3] No property-based coverage on the condition-merge precedence** — `writeStatusCondition` (`reconciler.ts:2144`) concentrates precedence / retired-strip / `lastTransitionTime` rules. Example coverage is broad, but idempotency (`reconciler.test.ts:2881` is the only spot) and composability have no fuzz. Pulling the pure merge out and adding a couple of `fast-check` properties would harden it.

4. **[R1-L4] Hand-copied message literals in the test** — `reconciler.test.ts:146-147` copies `RUNTIME_NOT_DESIRED_DISABLED_MESSAGE` / `RUNTIME_NOT_DESIRED_FAIL_CLOSED_MESSAGE` (`reconciler.ts:144-145`, not exported) verbatim and asserts `condition.message === <literal>`. They match the producer today so nothing's shielded, but it's a producer-output fixture written by hand. Exporting the constants and importing them in the test removes the drift risk.

## What checked out clean

- **No network fail-open (reproduced).** WRC's `waitForNetworkReady` (`mcpDelegation.ts:1194`) reads the `clerum.io/network-ready` annotation, which `setNetworkReadyAnnotation` still writes untouched (`reconciler.ts:2558`). control-ui lists `NetworkReady` explicitly as an `IRRELEVANT_TYPE` (`resolveCredentialSurfaceInvariants.test.ts:200`). Dropping the condition takes signal from no one.
- **Delete ownership preserved.** Every delete in `retractManagedRuntime` goes through the HCC-owned guards; the `foreign` (WRC-owned) branch retires only the local readiness window and writes no `RuntimeNotDesired`.
- **TOCTOU on the retract closed.** `retireReadinessWindow` bumps the epoch before deleting the map entry and before the retract write; each tick checks `retiredNow()` before every write, and the PATCH carries `test` ops — uid + epoch + optimistic-lock, three layers. A stale tick can't republish `True` after retract.
- **`StatusWriteOutcome` migration is safe.** Only two callers use the value (`=== 'declined'`, `=== 'wrote'`); the other ten ignore it — no `if (wrote)` against an always-truthy string.
- **Suites/typecheck.** Full HCC suite green (110 files, 2403 tests), `tsc --noEmit` clean, and T3 verified — 25 of the new tests fail against the merge-base, 194/194 pre-existing stay green. The cluster e2e (`mcp-runtime-conditions.test.ts`) is read-only here and tracked in #636, not a #633 gate.

zach88 (approved):

PR #634, opened 2026-09-14
fix(control-ui): refine post-standardization layouts

josue-tech-94:
## Summary

- Keep shared table-header controls pinned to the top right without wrapping below the title.
- Preserve title/icon spacing while truncating long titles to one line.
- Clamp long presentational descriptions to two whole-word lines and expose the full value in a
  keyboard- and hover-accessible tooltip across the available content width.
- Leave interactive subtitle content fully visible and directly focusable.
- Remove the extra tab-to-table inset from Marketplace API Keys, Entries, and Images.
- Align the LLM Models section with the common section-card spacing and full-height layout.
- Consolidate Agent detail actions beside Back to agents, including the delete overflow menu.
- Guard unsaved Identity-file edits before a subtab can change.
- Stack this follow-up on #630 and synchronize its latest balanced grouped-table contract.

## Behavior

- Header identity content yields available width to a fixed, non-wrapping action group.
- Title icons remain fixed and vertically aligned while only the title text ellipsizes.
- Presentational descriptions wrap without splitting words, stop after two lines, and show the
  complete text only when truncation is present.
- Interactive subtitle content remains visible and focusable rather than being clipped.
- The full-description tooltip is available by hover or keyboard focus and stays within the main
  body width rather than extending beneath the sidebar.
- Marketplace Connectors keeps its existing spacing; API Keys, Entries, and Images now begin their
  tables at the same tab boundary.
- Catalog and Discovery Review retain four balanced provider-summary columns with a fixed Actions
  lane, plus independent expanded child-table schemas from #630.
- Identity sub-tabs now share their panel edge treatment and prompt operators to Save or Discard all
  before leaving unsaved edits.
- Models & creds, Connectors, Access, and Advanced expose their contextual CTAs in the shared title
  action area, while preserving existing modal and API behavior.

## Shared component contract

- `TablePanelHeader` owns icon/text separation and bounded presentational-description behavior for
  every Control UI consumer using this section-header pattern.
- Future sections receive the same nowrap action placement, title truncation, and safe description
  behavior without route-specific header CSS.
- The synchronized grouped-table primitive keeps summary and expanded child schemas independent,
  so future expandable tables cannot let child column counts move their summary headers.

## Testing

- `packages/frontend-components`: `npm test`; `npm run build`
- `control-ui`: `npm test` (187 files, 1998 tests)
- Round-two focused regression suite: 6 files, 38 tests passed
- R3-L3 focused ConfirmDialog / HostIdentityTab tests: 17 passed; Escape and backdrop resolve as Cancel
- R3-L1 title semantics: 2 focused tests passed; R3-L2 custom interactive subtitle + shared header: 10 passed
- R3-L4 description measurement lifecycle + shared header: 9 focused tests passed
- Agent detail focused tests: Identity, Access, Guardrails, and page integration
- `control-ui`: `npx tsc --noEmit`
- `control-ui`: `npm run build` (60 routes)
- Style rules: 0 errors; one pre-existing Desktop test warning
- Changed-file Prettier and `git diff --check`
- Sync semantic-conflict audit CLEAN
- Sync intent-preservation audit CLEAN
- Sync regression/security audit CLEAN / SAFE
- Exact-head GitHub Actions at `3d4c0dd329ba89c2834df4a7829cb2824155f3d6`: CI run 35248768653 and CodeQL run 35248768663 are queued.


zach88:
Took a deep pass through this one. It's a clean, well-scoped layout refactor — the section-actions seam and the shared header primitive both hold up under scrutiny. The one thing that stops me from a thumbs-up lives in the new unsaved-edits guard on the Identity tab: the standard "close the modal" gesture silently throws away the operator's edits.

## Verdict

🔴 **Not ready to approve**

- **Optional / non-blocking: 11**

1. **[R1-B1] Escape / backdrop dismiss on the Identity guard silently discards all edits**
   `components/HostIdentityTab/index.tsx:152-163` + `components/ConfirmDialog/index.tsx:19,44-46`.
   The new "Unsaved identity edits" dialog resolves `false` on three gestures: the "Discard all" button, **Escape**, and a **backdrop click**. In `setActiveField`, `false` runs `discardEdits()` and then switches tabs. So an operator who edits an identity file, switches sub-tab, sees the dialog, and hits Escape (the universal "cancel/close modal" reflex) **loses all their work and changes tabs anyway**, with no undo. Before this PR, edits survived navigating between sub-tabs.
   Extra sharp edge: in `ConfirmDialog`, initial focus lands on the "Discard all" button (the destructive one), so hitting Enter/Space by reflex also discards. The dialog inverts the usual convention where the default/dismiss action is the safe one.
   Repro: edit Identity → click the Soul tab → `keyDown Escape` → Identity's value reverts to the old text (edit lost).
   The root cause is that `confirm()` returns a boolean and can't tell "Discard all" (explicit) from "dismiss" (Escape/backdrop). A proper fix needs a tri-state outcome (Save / Discard / Cancel) so dismiss aborts the tab change and keeps the edits — that's decision logic, probably worth a quick design note before patching rather than a reflex fix.

## Optional (non-blocking)

The PR can merge without these — follow-up or author's call. Two of them (M2/M3) are the same class: the new behavior lacks real test coverage.

2. **[R1-M1] `saveActiveField` depends on the whole `state` object → full-page re-render per keystroke**
   `components/HostIdentityTab/index.tsx:142,168-205` + `app/hosts/[name]/page.tsx:196,1092`. `saveActiveField` carries `state` in its dep array, so it's recreated on every `setField` (every keystroke). The actions effect depends on it, so it fires `onActionsChange(null)` + `onActionsChange(node)` per key → two `setSectionActions` on `HostDetailsPage`, re-rendering the whole page as you type in the editor (header mode is the production path). Not a loop, just redundant work the sibling tabs avoid with narrow deps. Worth folding into the B1 fix since both touch `saveActiveField` — drop `state` from the callback (read via ref, or depend on the boolean slices).

3. **[R1-M2] The advanced update-loop invariant has no test, and the fix commit's Repro-test is invalid**
   `components/HostAdvancedTab/index.tsx` + `app/hosts/[name]/page.tsx`. Commit 3ed8ed5f cites `HostDetailsPage.identity.test.tsx` as its Repro-test, but that test renders with `tab: 'identity'` and never mounts `HostAdvancedTab`, so it'd pass against the previous head too. Nothing asserts the observable result (which buttons show in the header after switching tabs), and `HostAccessTab.test.tsx` renders without `onActionsChange`, i.e. it only exercises the fallback button, not the real header path. Suggest an integration test that switches tabs (identity → access → advanced/hooks) and asserts the header action set, plus one that mounts Advanced/hooks and checks it doesn't blow up with "Maximum update depth exceeded".

4. **[R1-M3] The most important new Identity behavior is untested**
   `components/__tests__/HostIdentityTab.test.tsx:128-151`. Only the Discard path is covered. Missing: Save-on-confirm (choosing "Save" → `updateHostPersonalization` with the edited value → tab switches), abort-on-save-failure (Save 409 → tab does NOT switch), and header mode (`renderTab()` never passes `onActionsChange`, so the whole action-registration path — the production one — is untested). The body invariant "prompt operators to Save or Discard all before leaving unsaved edits" only tests one branch.

5. **[R1-L1] `ClampedDescription` doesn't re-measure when `children` changes in place** — `components/TablePanelHeader/ClampedDescription.tsx:11`. `useLayoutEffect` deps are `[]`; edge case where a subtitle swaps truncated↔fitted but stays 2 lines tall won't retrigger the ResizeObserver. Current consumers pass static subtitles. Fix: add `children` to the deps.

6. **[R1-L2] The tooltip renders `children` a second time** — `ClampedDescription.tsx:36-40`. Harmless today (subtitles are text/`<code>`), but a subtitle carrying an `id` or interactive control would produce duplicate ids. Worth documenting the `subtitle` contract as presentational, or `aria-hidden` + strip ids on the tooltip copy.

7. **[R1-L3] a11y: `aria-describedby` points at a `display:none` node** — `ClampedDescription.tsx:29` + `globals.css`. Several screen readers won't announce a `display:none` node referenced by `aria-describedby`. Impact is bounded since the text still lives in the clamped value span.

8. **[R1-L4] The "first valid element = icon" title heuristic is fragile** — `components/TablePanelHeader/index.tsx:9-14,30-33`. A future consumer whose first title node is a text element would get it wrapped as an icon. All 34 current consumers are fine (verified). Robustness debt, not a live bug.

9. **[R1-L5] Header "Add hook" shows during `initialLoading`** — `components/HostAdvancedTab/index.tsx:37-54`. The action registers on `subTab === 'hooks'` without checking `initialLoading`, so the header offers "Add hook" while the body still says "Loading…". It only navigates, so harmless. Fix: `|| initialLoading` in the early-return.

10. **[R1-L6] Lost the "why" comment on the Add-hook route + duplicated it inline** — `components/HostGuardrailsSection/index.tsx` + `HostAdvancedTab/index.tsx:43-46`. The comment explaining why the list is filtered to guardrail hooks got dropped, and `orgEntriesFiltered({ type: GUARDRAIL_ENTRY_TYPE })` is now rebuilt inline with no rationale. Worth restoring the comment / sharing the constant.

11. **[R1-L7] Removed the "identity files are read-only to the agent" note** — `components/HostIdentityTab/index.tsx`. The `cu-identity-panel__intro` block with "Admin-managed identity files are readable by the agent but blocked from agent writes." is gone and not relocated anywhere. It was the only UI hint that these files are read-only to the agent — confirm that's intentional under "refine layouts".

12. **[R1-L8] Dead CSS** — `app/globals.css:8844,8866,8872`. `.cu-identity-panel__intro` and `.cu-identity-dirty-actions` are no longer referenced. Easy boy-scout cleanup.

## What checked out clean

- The single shared `sectionActions` slot doesn't clobber: React 18 runs all passive cleanups before setups, and `HostAdvancedTab` early-returns without a cleanup when `subTab !== 'hooks'`, so it never stomps `HostEnvTable`'s registration. The advanced update-loop fix (routerRef + dropping `router` from deps) holds.
- `onActionsChange?: (ReactNode | null) => void` is consistent across all four consumers and the parent.
- The `activeDirty` vs `anyDirty` path is NOT a bug — only the active field renders/edits, so `activeDirty === anyDirty` always, and Save never aborts a field switch wrongly (verified with a test).
- No security findings: no new `dangerouslySetInnerHTML` (React escapes the tooltip subtitle even if multi-tenant), env secret handling untouched, navigation via constant (no open redirect), `package.json` only bumps its own version (no supply-chain surface), no write gate weakened.
- Suites the review ran on the head: header 53/53, seam 109/109, identity 22/22 — all green.
- `tsc --noEmit`: the two errors are both pre-existing and outside this diff; the changed files compile clean.

zach88:
Took another deep pass through this one (round 2). Good news first: the round-1 blocker (Escape/backdrop silently discarding identity edits) is fully resolved. The tri-state dialog is solid, the dismiss path aborts the tab change and keeps edits, initial focus is on the non-destructive Cancel, and the new regression tests genuinely fail against the previous head. All twelve round-1 findings are closed or improved.

What's left is small: two low-severity regressions the follow-up commit introduced, plus a couple of should-fixes and some nits. The one substantive UX/a11y item lives in the shared header primitive.

🔴 **Not ready to approve**
- **Must resolve before approval: 2** (both are open regressions from the follow-up fix — hygiene, not severity: neither is user-visible today, but a fix that broke something shouldn't merge as-is)

1. **[R2-L1] `setActiveField` dep array references the wrong symbol** — `control-ui/components/HostIdentityTab/index.tsx:180` — *regression from `8a517b67`*. The hook destructures `{ choose, confirmDialog }` and the body uses `choose`, but the dep array still lists `confirm`, which no longer exists in scope and resolves to the global `window.confirm`. So the real dependency (`choose`) is missing and the array names a foreign symbol. Harmless at runtime today because `choose` is a stable `useCallback([])` (no stale closure) and `tsc` doesn't catch it since the global exists in the DOM lib — but it's incorrect, misleading, and would trip `react-hooks/exhaustive-deps`. In `3ed8ed5f` the hook returned `confirm` and the array was correct; the `confirm→choose` rename didn't update the deps. Fix: swap `confirm` for `choose`.

2. **[R2-L2] `ClampedDescription` dropped `aria-describedby` (a11y)** — `control-ui/components/TablePanelHeader/ClampedDescription.tsx:33-42` — *regression from `8a517b67`*. Fixing the double-render (R1-L2/L3) over-corrected: the tooltip is now `aria-hidden` with no `aria-describedby`, so the on-focus description association a truncated subtitle used to expose is gone. Mitigating factor: the full text is still in the accessibility tree via the visible `children` (line-clamp/overflow:hidden don't prune the a11y tree), so a screen reader reading in order still gets it — only the focus-time description announcement was lost. Low impact, but it's a behavior that existed and no longer does. Fix that closes it without reintroducing R1-L2: keep `aria-describedby` pointed at the flattened `tooltipText` string (not the children). If you'd rather call it a non-issue given the mitigation, that's a reasonable write-off too.

## Optional (non-blocking)

These don't block the merge; follow-up or your call. Worth calling out that #1 here is the most substantive real defect in the round even though it's not gating.

1. **[R2-M1] Clamped subtitle hides interactive content while keeping it focusable** — `control-ui/components/TablePanelHeader/index.tsx:50` + `control-ui/components/TablePanelHeader/ClampedDescription.tsx:36-44`. The shared primitive now applies `line-clamp:2 + overflow:hidden` and `tabIndex=0` to *any* subtitle. `control-ui/components/MarketplaceOrgImages/index.tsx:110-122` passes a subtitle containing a real `<a class="cu-link">API Keys</a>` (that link already existed at merge-base, so this PR aggravates a pre-existing affordance). Repro (by code reading + confirming the `<a>` at merge-base): when the subtitle truncates at the rendered width, the link drops to the clipped 3rd line — visually hidden but still focusable and clickable (WCAG 2.4.7, focus not visible), the focus tooltip shows the flattened text with no link, and the `tabIndex=0` wrapper adds a redundant tab stop before the `<a>`. No new module/decision logic needed: either don't clamp subtitles that contain interactive descendants, or make `subtitle` a text-only contract and move the link out of the subtitle in MarketplaceOrgImages. I'd lean toward doing this one before merge.

2. **[R2-M2] Access tab's production path (header-mode) has no test** — `control-ui/components/__tests__/HostAccessTab.test.tsx`. The PR changes HostAccessTab's contract (`index.tsx:329-342`, `:363-372`): a new effect registers "Add member"/"Add team" into the header slot and hides the inline button when `onActionsChange` is present. The existing suite only exercises the inline fallback (`!onActionsChange`), never the real path `page.tsx` uses. It's the sibling of the round-1 R1-M2 gap for the Access consumer. Verified correct by reading, so not a live bug, but the coverage gap is real. Suggestion: render HostAccessTab with an `onActionsChange` that captures the node (same `HeaderMode` pattern as the new HostAdvancedTab test) and assert the header exposes "Add member", flips to "Add team" on subtab switch, and the inline button doesn't appear.

3. **[R2-L3] HostEnvTable has no inline fallback if `onActionsChange` is absent** — `control-ui/components/HostEnvTable.tsx:47-53,111-132`. `onActionsChange` is optional but the inline Refresh/Add-variable buttons were removed, so mounting it without the prop leaves no way to add/refresh vars. The only caller (`HostAdvancedTab:92`) always passes it, so no live bug — but it's a latent trap and inconsistent with HostAccessTab, which keeps its fallback. Either make the prop required or keep an inline fallback.

4. **[R2-L4] Dead CSS class** — `control-ui/components/HostEnvTable.tsx:225` references `cu-host-env-tab__header`, which isn't defined in `globals.css` (only `.cu-host-env-tab` exists). It replaced inline flex styles and is cosmetically inert now (single child). Drop the class or define the rule.

5. **[R2-L5] Contradictory ARIA on the tooltip** — `control-ui/components/TablePanelHeader/ClampedDescription.tsx:41`. `role="tooltip"` together with `aria-hidden="true"` makes the role inert. Harmless; drop one of the two.

Nits (no need to track): the truncation test at `control-ui/components/__tests__/TablePanelHeader.test.tsx:95-104` hand-mocks `scrollHeight`/`clientHeight` (jsdom does no layout), so it certifies the mock rather than real truncation, and nothing exercises re-measurement when `children` changes (the R1-L1 fix). The ConfirmDialog backdrop-dismiss branch (`index.tsx:52`) has no dedicated test — only Escape is covered, though they share the `'dismiss'` resolution.

## What checked out clean

- **ConfirmDialog contract is intact.** `useConfirmDialog().confirm` still returns `Promise<boolean>` (the fix wraps `resolve(outcome === 'confirm')`); all four paths map identically to before. 55 `.confirm(` call-sites audited — all treat the return as boolean, none pass `discardLabel`. The tri-state lives only in the new `choose`, consumed solely by HostIdentityTab. No contract regression.
- **Security: clean.** No new `dangerouslySetInnerHTML`; the tooltip renders escaped text; no write gate weakened when moving CTAs to the header (compared line-by-line); HostEnvTable's `secret` handling unchanged; navigation is via an internal constant (no open-redirect); package.json only bumps its own version.
- **Section-actions seam:** single slot, no clobber across tab↔tab and hooks↔env↔approvals subtab transitions; the update-loop is closed (verified by crashing "Maximum update depth" against the pre-fix parent).
- **`saveActiveField` via `stateRef`:** no stale closure; narrowing deps to `[hostName, showToast]` is what fixes the per-keystroke re-registration.
- **`titleActions` was not renamed** — DetailPageShell/BodyLoadingSkeleton route it unchanged.
- **Suites (head `882804eb`):** Identity 27/27, HostDetailsPage 34/34, Advanced+Access 8/8, header+consumers 54/54. Green. T3 verified on the key regression tests (they fail against the prior head). Diff files typecheck clean.

zach88:
Took another deep pass through this one (round 3). Every open item from the last round is closed and verified against the current head, and the fix round didn't introduce any regressions this time. What's left is all shared-component robustness on `TablePanelHeader`/`ClampedDescription` plus a couple of nits, none of it blocking.

**🟡 Approve with nits**

## Optional (non-blocking)

This PR can merge as-is; these are follow-up / author's call.

1. **[R3-L1] Icon heuristic can blank the visible title text** — `components/TablePanelHeader/index.tsx:40-44,71-78`. `flattenTitleNodes` treats the first node as an icon whenever its `type` isn't a string (i.e. any function component), and sets the visible title to whatever's left. If a consumer passes a *single* component as `title` (e.g. `title={<HostTitle/>}`, which the prop type allows), that node lands in the `.cu-table-panel__title-icon` slot and `.cu-table-panel__title-text` renders empty → invisible title, silently. No live consumer hits this today (all ~34 use `<><IconX/> text</>` or a string), but this is a component half the repo consumes and the failure is quiet. Suggestion: only treat the first node as an icon when there's remaining text (`remainingTitleNodes.length > 0`), otherwise render the title verbatim.

2. **[R3-L2] Interactive-subtitle bypass only matches native tags** — `components/TablePanelHeader/index.tsx:18-27`. `hasInteractiveDescendant` (the clamp bypass that fixed the focusable-but-hidden subtitle) only detects native lowercase tags (`a/button/input/select/textarea`). A subtitle whose interactivity comes from a wrapper component (`<Button>`, `<Link>`, `<SelectInput>`) has a function `node.type` → returns false → it'd still be clamped with a trapped focus, which is the exact WCAG 2.4.7 issue the bypass targets. No broken sibling in production right now (interactive subtitles all use native tags), so this is a latent gap in the fix rather than a live bug.

3. **[R3-L3] `dismiss` vs `cancel` is indistinguishable to every consumer** — `components/ConfirmDialog/types.ts:2` + `index.tsx`. In `HostIdentityTab.setActiveField` both fall into `else { return }`, and the `confirm()` API maps both to `false`. The distinction isn't observable anywhere, so it's type complexity with no effect. Related nit: the `treats a backdrop click as dismissal` test uses the boolean `confirm()` API where backdrop already resolved `false` on the parent commit, so it goes green against the parent and doesn't exercise the new path — no actual coverage hole though, since the real "dismiss preserves edits" behavior is covered by the Escape test in `HostIdentityTab.test.tsx`.

4. **[R3-L4] Clamp re-measure resubscribes on every render** — `components/TablePanelHeader/ClampedDescription.tsx:33`. The `useLayoutEffect` depends on `children`, whose identity changes each header render, so it tears down and re-attaches the `ResizeObserver` and re-measures on every render (e.g. while typing in a search box in the same panel). Correct, just wasteful — depending on the derived `tooltipText` would be cheaper.

## What checked out clean

- **Unsaved-edits guard (Identity):** tri-state save/discard/cancel/dismiss mechanics are correct; the `anyDirty ⟺ activeDirty` invariant holds (the editor only touches the active field), so the prompt's "Save" is never a no-op trap. Covered by `HostIdentityTab.test.tsx`.
- **Shared action elevation via `onActionsChange`:** React's destroy-before-create ordering keeps stale actions from lingering when toggling hooks↔env subtabs or page tabs.
- **Interactive-subtitle bypass:** no hidden-but-focusable content for the native-tag consumers (`MarketplaceOrgImages`, `RecipeIntegrationsPanel`, etc.) — WCAG 2.4.7 OK.
- **a11y restore (R2-L2):** `aria-describedby` now points at a separate `sr-only` node (real clip-rect, not `display:none`), not the `aria-hidden` visual tooltip — so it reads to AT without reintroducing the hidden-node association.
- **Security:** clean. No `dangerouslySetInnerHTML`/`innerHTML`/`eval` in the diff; the flattened tooltip text renders as React-escaped text end to end; `GUARDRAIL_MARKETPLACE_ROUTE` is a static constant with no user input.
- **Suites:** 8 touched test files → 80/80 passing, adjacent regression suite 28/28, `tsc --noEmit` clean.

zach88 (approved):

PR #635, opened 2026-09-15
feat: agent creation polish + display-name propagation (TASK-229/230/231/234)

miguelo-999:
## What

The agent-rename feature was partially working: server-side propagation was already correct (`displayName = spec.host || name` via `buildAgentDirectoryEntry` → `/me/agents`), but frontend consumers kept rendering raw identifiers. This makes every surface re-read and render the display name.

## Scope (review R1-M2)

The change spans Control UI and Desktop because it delivers one display-name propagation audit, but the surfaces are independently deployable. This PR retains them together only for coordinated verification; future corrective work should split by application surface.

## Tickets

- TASK-229 — After creation, the wizard navigates to the new agent's Overview route and the header leads with the display name while its data loads. The identifier remains the route key; no route-URL eyebrow is rendered.
- TASK-230 — derive the agent identifier from a free-text name in the wizard (removes forced lowercase)
- TASK-231 — polish the connectors step (4th step) of the create-agent wizard
- TASK-234 — show agent display names in the member Agents tab (profile-admin user page)

## Extras found during the audit

- team Agents tab shows display names
- desktop fleet table shows display names
- desktop chat notifications show display names

## Review round 1

Independent review returned P0 none / 1 P1 / 2 P2 — all addressed in 36f50980 (toast body fallback uses displayName + regression test) and 396e4d65 (name-field error copy for names that kebab to empty + slug-note a11y via aria-live/aria-describedby/aria-invalid).

## Review round 2 (josue-tech-94)

- **R1-H1 (member + team Agents tabs)** — display names are not unique, so the immutable slug now stays visible as secondary identity whenever it differs from the display name, and the RowActionsMenu label carries it (`Actions for Name (slug)`). Duplicate-display-name regression tests added for both tabs; each revoke still targets the correct slug. (9f253a1f)
- **R1-M1 (qa-recorder spec)** — pre-creation assertion restored to the Agents list URL; the generated agent detail URL is asserted only after Create Agent completes. (3df5d8f7)
- **R1-M2** — scope rationale added above; TASK-229 description corrected (no route-URL eyebrow).
- Recorder run through creation → redirect → verification → cleanup: pending the authorized isolated environment (loopback control-ui/control-api, `E2E_AGENT_SECRET_NAME`, `QA_RECORDER_CONFIRM_MUTATIONS`) — not executed from this workspace.
- Desktop: no findings, no changes requested; untouched this round.

## Verification

- control-ui: HostWizard + ProfileAdminAgentCompatibility (incl. new duplicate-display-name regressions) + CreateMemberPage — green; `tsc --noEmit` clean
- desktop-app: FleetBoard.deeplink, AppHeader/notifications, useChatNotifications, useAgentChatController.*, useHostPrewarmController.chatIntegration — green; `verify:electron` OK (round 1; unchanged since)
- No backend changes


miguelo-999:
@claude please review this PR, check the previous reviews and provide your own independent review if there's feedback sort it by P0, P1, etc and if everything is good provide explicit green light.

miguelo-999:
@josue-tech-94 please take a look, both reviews using internal skills + claude output green light for merge. Waiting on your feedback.

josue-tech-94:
There are 3 items to address before the next review round.

### `control-ui/app/profile-admin/teams/[teamId]/page.tsx`

**R1-H1 — High**

The Agents access list now shows only the display name. Since display names are not unique, two different agents can both appear as something like `Support`, making navigation/revoke actions ambiguous.

**Required outcome:** keep the immutable agent slug visible somewhere in the row and make sure accessible/action labels also distinguish the agents.

**Suggested fix — not a strong implementation requirement:** keep the display name as the primary label and show the slug as secondary text underneath or beside it. An identifier column would also be valid. The important part is preserving both the human-facing name and immutable identity.

Please also cover the duplicate-name case in a regression test: two agents with the same display name but different slugs should remain distinguishable, and each action must still target the correct slug.

### `control-ui/app/profile-admin/users/[userId]/page.tsx`

**R1-H1 — High**

Same issue as the team page: replacing the unique slug with only `spec.host` makes duplicate display names indistinguishable in the member Agents access UI.

**Required outcome:** preserve the immutable slug as visible secondary identity and in any relevant accessible/action labels.

**Suggested fix — not a strong implementation requirement:** use the same display-name + secondary-slug treatment as the team list so these two access-management surfaces stay consistent.

The exact visual treatment is up to you; the requirement is that an admin can reliably tell which agent they are inspecting or revoking.

### `control-ui/e2e/qa-recorder-agent-create.spec.ts`

**R1-M1 — Medium**

Around line 94, the recorder expects the generated agent detail URL immediately after opening the Agents area, before the agent has actually been created.

**Suggested fix — strong:** keep the pre-creation assertion on the Agents/list route, then assert the generated agent detail URL only after Create Agent completes successfully.

Please run the recorder through creation → redirect → verification → cleanup in its authorized isolated environment.

### PR description

**R1-M2 — Medium**

The PR currently spans 28 files across Control UI and Desktop renderer, which exceeds the D1 review target. Please either split the independently deployable surfaces or add a concrete explanation for why keeping them together is the appropriate atomic scope.

The current implementation does not appear to establish a technical dependency requiring both surfaces to ship atomically, so splitting remains an option. If you keep them together, a rationale like this would satisfy the review concern:

> The change spans Control UI and Desktop because it delivers one display-name propagation audit, but the surfaces are independently deployable. This PR retains them together only for coordinated verification; future corrective work should split by application surface.

Also update the TASK-229 description. The current body says the detail header includes a route-URL eyebrow, but that UI was deliberately removed. Suggested replacement:

> After creation, the wizard navigates to the new agent’s Overview route and the header leads with the display name while its data loads. The identifier remains the route key; no route-URL eyebrow is rendered.

### Desktop validation

No Desktop code change is requested by this review. I did not find a Desktop correctness or security finding.

miguelo-999:
@josue-tech-94 
DONE
R1-H1 (High, both pages) — 9f253a1f
- Member + team Agents tab rows now lead with the display name and stack the immutable slug underneath whenever they differ (new cu-access-agent-copy/cu-access-agent-id, tokens only) — same treatment as FleetBoard.
- RowActionsMenu labels disambiguate: Actions for Duplicate Bot (agent-dup-a).
- Regression tests added for both tabs: two agents sharing a display name stay distinguishable, and revoking one row calls updateAdmin{User,Team}Agents with the correct slug.

R1-M1 (Medium, e2e) — 3df5d8f7
- Confirmed bug: the spec asserted the detail URL before creation (line 94–99, would always fail). Pre-creation assertion restored to the list URL; detail URL asserted only after Create Agent; comments fixed to match the shipped header (no eyebrow).

R1-M2 (PR description) — updated via gh pr edit: added the scope rationale, corrected the TASK-229 description, and documented round 2.

Verification: ProfileAdminAgentCompatibility 11/11 green (incl. 2 new), tsc --noEmit clean, version bumps 0.1.577/0.1.578 per convention. Desktop untouched (no findings).

josue-tech-94 (approved):

Issue #642, opened 2026-09-15
Raise mcp-host iteration limit to 1000 for long-running Desktop tasks


alfredolopez80:
## Reportante
Miguel (atribución solicitada por Alfredo). GitHub registra como autor técnico la cuenta autenticada que publica este issue.

## Problema
Un chat de Desktop durante una tarea larga termina con: “I've reached the maximum number of tool-use iterations (100). Please try a more specific request.” La captura muestra además 101 herramientas.

## Causa comprobada en el repositorio
`mcp-host/src/config.ts` lee `CLERUM_AGENT_MAX_TOOL_CALLS` con valor predeterminado 50. El ConfigMap de Minikube lo establece en 100. `TaskExecutor` lo pasa como `maxIterations` al bucle del agente, que genera el mensaje al agotarlo.
El límite es por ejecución del bucle, configurado por proceso de mcp-host; una iteración puede ejecutar varias herramientas. No es una cuota global ni un límite del proveedor. No se ha verificado el entorno del clúster que produjo la captura.

## Ajuste solicitado
Elevar a 1000 el valor predeterminado, el default de AgentConfig y la configuración de Minikube; actualizar documentación para aclarar que cuenta iteraciones. Mantener la posibilidad de sobrescribirlo mediante ENV.

## Criterios de aceptación
- Ambos defaults del agente y el ConfigMap de Minikube usan 1000.
- La documentación explica el alcance por Host y por iteración.
- Se conservan los límites independientes de tiempo y presupuesto.

## Alcance operativo
El cambio requiere despliegue para afectar a instancias existentes. Una ENV explícita sigue prevaleciendo sobre el default. Aumentar el máximo permite tareas más largas y también mayor consumo potencial; no añade continuación automática.

alfredolopez80:
## Timeout analysis for long-running chats

The current code parses CLERUM_AGENT_MAX_TASK_DURATION and CLERUM_AGENT_TASK_DELAY and passes them to AgentConfig, but TaskExecutor and SessionProcessor do not consume these fields. The configured 1800000 ms therefore does not enforce a 30-minute task deadline; 3 means milliseconds, not seconds. Raising either value would not extend the current execution path.

[...]

PR #643, opened 2026-09-15
fix(mcp-host): enforce durable task and tool execution limits

alfredolopez80:
## Problem and behavior

Long-running tasks need effective limits and must retain completed output when interrupted. The PR enforces a 24-hour active task budget, 1000 iterations, 25-minute shell/tool/MCP budgets and 3 ms between task dispatches. Approval waits are excluded; persisted consumption and original limits survive approval/resume/restart. Legacy approvals request new explicit approval on the same task, with atomic replacement and retained replay state.

Duration/iteration exhaustion remains a failed, incomplete task. Completed, provenance-checked artifacts are now retained through model/admission/batch interruption, approval resume and the single failure callback. Synchronous and asynchronous Host responses preserve the artifacts; Desktop retrieves and displays them on the error message, with its existing Download action. No second success callback is introduced.

## Adversarial validation of jozer-rami's review

Source: https://github.com/evenfire-ai/evenfire/pull/643#issuecomment-5698691807

1. **Partially confirmed and corrected:** a deadline check inside the persistence catch could clear the turn before the interruption message was saved. Both final checks now occur before that catch. Guarding and sanitization still consume active budget; pausing before validation was rejected because an unvalidated model answer is not completed work.
2. **Confirmed and corrected:** interruption dropped attachments and could report the configured rather than restored effective iteration maximum. Completed artifacts now survive with existing provenance/deduplication rules, the effective maximum is reported, and Desktop error rendering/download is covered.
3. **Consistency corrected:** omitted ENV values use defaults; explicitly empty, malformed and overflowing values are rejected by both execution-limit helpers.
4. **No functional defect established:** enforcement before forwarding an admitted-iteration event is retained. Forwarding an event before rejecting admission is not required by the telemetry contract.
5. **Privacy retained, diagnostics improved:** the unknown-pin warning includes cache size/pinned-count/capacity instead of restoring an identifying session key.

## Follow-up to Zach88

Reviewed [Zach88's findings](https://github.com/evenfire-ai/evenfire/pull/643#issuecomment-5703146294) and addressed the justified small changes in `f4234a8d07823b5ea27fca64a1127595fcba880d`:

- **R1-M2:** strict empty-ENV rejection retained intentionally. README already documents defaults only for absent variables; unrelated provider-default settings do not define execution-limit policy.
- **R1-M2b:** added 28 malformed/negative/fractional/overflow cases across both ENV readers.
- **R1-L2:** omit empty artifact arrays in both durable error recovery paths.
- **R1-L3:** iteration exhaustion now emits the declared `iteration_limit` reason explicitly.
- **R1-L4:** both LLM methods assert that the provider receives the combined signal, including propagated task abort and its reason.
- **R1-L6:** duplicate-suppression logs retain `priorTaskId` and `priorStatus`; private session/channel identifiers remain excluded. Pin warnings retain the previously added safe cache counters.
- Moved the TaskExecutor module header before its imports.
- **R1-L1 deferred:** recovering NULL accounting after unsupported rollback requires an explicit migration/recovery contract. Silently treating it as legacy could replenish task budgets; current fail-closed behavior remains.
- **R1-L5 unchanged:** retaining artifacts across arbitrary failures broadens the approved interruption contract without a demonstrated supported-flow defect.
- Dispatch timer `.unref()` not adopted: this can alter process lifetime while queued work remains, so it is not a cosmetic change.

## Synchronization and evidence

Certified head: `f4234a8d07823b5ea27fca64a1127595fcba880d`, merged with `origin/dev` at `15c4aa016311ec46513f216531107b3ed93dc041`.

- All 38 GitHub checks passed; PR is mergeable and awaits human review.
- Canonical local certification: **T0 PASS, T1 PASS, T2 PASS, NP08 HCC authorization PASS, Health PASS** on this exact head. The final orchestrator emitted `MINIKUBE_T2_PASS`.
- Node 24: **4068 Host tests passed** (one existing TODO), **2302 Desktop tests passed**; both production builds passed.
- Real PostgreSQL: 35 selected test files and 164 tests passed, zero skipped.
- Four focused regressions fail on previously reviewed head `659b947a` and pass after correction.
- Component/contract coverage verifies failed-result retrieval, retained attachments, visible Download action and downloaded filename/bytes.
- **Playwright NOT_RUN**: the existing fixture can delete the personal Desktop session; isolated browser certification remains a separate follow-up. Component tests and Health are not browser E2E evidence.
- Whole-renderer TypeScript checking still reports the same **38 baseline diagnostics**, with zero added or removed against dev `15c4aa01` at the prior reviewed head `77126d5f`. This supplementary check is not represented as passing.
- Repository formatting/style, public-boundary and release-manifest validation passed. Desktop retains dev's `v0.8.1` release metadata.
- [Claude's final re-review and audit](https://github.com/evenfire-ai/evenfire/pull/643#issuecomment-5703105089) on the prior head `77126d5f` found no introduced blocking defects and concluded it is ready for human approval. This was a static audit; the executed validation evidence above is refreshed for `f4234a8d`. [The final follow-up audit](https://github.com/evenfire-ai/evenfire/pull/643#issuecomment-5703462779) also found no introduced blocking defect and marked `f4234a8d` ready for human approval. No production deployment or merge is performed.

Closes #642


alfredolopez80:
## Timeout analysis for long-running chats

The current code parses CLERUM_AGENT_MAX_TASK_DURATION and CLERUM_AGENT_TASK_DELAY and passes them to AgentConfig, but TaskExecutor and SessionProcessor do not consume these fields. The configured 1800000 ms therefore does not enforce a 30-minute task deadline; 3 means milliseconds, not seconds. Raising either value would not extend the current execution path.

CLERUM_SHELL_TIMEOUT=600000 limits each shell command to 10 minutes. CLERUM_TOOL_TIMEOUT=660000 limits the outer wait for one tool to 11 minutes. The extra minute accommodates shell termination and result handling. Increasing the iteration count does not require increasing these per-operation limits.

CLERUM_MCP_TOOL_TIMEOUT_MS and CLERUM_MCP_TOOL_MAX_TOTAL_TIMEOUT_MS are both 3600000 (60 minutes) at the MCP boundary, subject to shorter caller budgets. Chat tools still encounter the outer 11-minute timeout first. The Promise.race in toolUseLoopSingleTool.ts does not cancel the underlying operation, so aligning cancellation and deadline propagation requires a runtime fix, not just larger ENV values.

An enforced 30-minute task deadline would allow 1000 iterations only at an average of 1.8 seconds each. Averages of 10 and 30 seconds would require approximately 2h47m and 8h20m. These are estimates, not measured execution times. Choose a task deadline from the intended runtime and observed latency, and keep individual operation timeouts separate.

Recommendation: retain current per-operation values for this iteration-limit change. Track and fix the unenforced task deadline and cancellation propagation before presenting MAX_TASK_DURATION as an effective safety limit. Existing configured per-task budgets remain independent; 1000 iterations is a maximum, not a guarantee.

Evidence: mcp-host/src/config.ts:717-719; mcp-host/src/main.ts:1635-1638; mcp-host/src/agent/stateMachine.ts createTaskExecutor; mcp-host/src/session/sessionProcessor.ts; mcp-host/src/agent/taskExecutor.ts; mcp-host/src/core/orchestration/toolUseLoopSingleTool.ts:109-124; mcp-host/src/mcp/requestOptions.ts.

Correction to the initial PR description: the duration setting is retained, but the current code does not enforce it. No additional numeric changes or runtime fixes have been applied by this analysis.

alfredolopez80:
@claude Please perform a fresh **full re-review and audit** of PR #643 at head `80661bd0774436bbb73822e96d3d00556c6f4394` against `dev`.

The change now enforces the settings rather than merely increasing defaults. Please independently trace:

- The 24-hour active task budget and 1000-iteration allowance across approval, resumption and restart, including exhaustion/error reporting and late provider results.
- The 25-minute shell/tool/MCP limits, caller-budget clamping, cancellation propagation, bounded shell process-group cleanup and timer disposal.
- Real dispatch spacing from `CLERUM_AGENT_TASK_DELAY`.
- Migration 014 and atomic legacy approval renewal: same task, new explicit approval, retained replay/source state, no execution before approval and no budget replenishment on ordinary restart.
- Persistence failures, cancellation races, budget-brake message/attachment compatibility, identity boundaries and the adequacy of the added regression tests.

The principal completed independent runtime/security/persistence review lanes and corrected the findings. Local Node 24.18.0 validation: **4008 passed, zero failed, one existing TODO**, with build/typecheck and repository hooks passing. All current GitHub checks are green on this exact head; the shell job required one retry after an `EADDRINUSE` runner failure.

Please distinguish introduced defects from pre-existing risks and provide exact locations and evidence for any actionable findings. No deployment or local Minikube T0/T1/T2 certification is claimed; CI and those runtime lanes are separate. Remote MCP cancellation remains cooperative. Please report your verdict and remaining coverage gaps; do not merge or deploy.

alfredolopez80:
@claude T2 is now certified on the latest head `659b947a01d99194bb38b3468af5111352fc96b3`. Please use this head for the requested full re-review/audit.

The canonical local orchestrator completed with `MINIKUBE_T2_PASS`:

- T0=PASS
- T1=PASS — 164/164 tests, 35/35 selected files, zero skipped/pending tests
- T2=PASS
- NP08_HCC_AUTHORIZATION=PASS
- Health=NOT_RUN
- Playwright=NOT_RUN

All 38 GitHub checks also passed on this exact head. The follow-up commit makes the unit-test Kubernetes configuration hermetic; it fixes six suites that depended on the machine's global context. Deployed numeric limits were checked inside the owned Host container and match the approved configuration.

Local runtime evidence remains in the ignored repository evidence directory. This is local Minikube certification, not a production deployment or a browser/E2E certification. Please retain those distinctions in your audit. No merge or deployment is requested.

alfredolopez80:
@claude Please perform another **complete re-review and full audit** of PR #643 at current head `659b947a01d99194bb38b3468af5111352fc96b3`, as explicitly requested by the author.

The previous review of this head has finished without blocking findings. All 38 current GitHub checks are SUCCESS, and the canonical local certification reported T0=PASS, T1=PASS (164 tests, 35 files, zero skips), T2=PASS and NP08_HCC_AUTHORIZATION=PASS. Health and Playwright remain NOT_RUN; no production deployment is claimed.

Review the complete PR against its base, covering correctness, security, concurrency, cancellation, effective limits, persistence/migration, compatibility and tests. Please explicitly resolve or retain the outstanding low-severity observation about attachment delivery on genuine iteration/duration exhaustion; it has not been silently accepted as fixed.

For test coverage, please inspect the existing direct equal-shell-deadline regression in `toolExecutionTimeout.test.ts` and the malformed/missing-accounting cases in `014-pending-approval-task-budget.test.ts` before carrying forward those earlier coverage gaps. The final approved tool belongs to its already-counted iteration; no additional model iteration is allowed after exhaustion. The same-task legacy renewal with a new explicit approval is an accepted product requirement.

Report actionable findings with exact locations, base/head attribution and supporting evidence, separating blockers from non-blocking observations. Confirm what you actually executed versus inspected. Please do not merge or deploy.

jozer-rami:
## Review of `659b947a` (vs `dev`)

Read the full diff at the certified head, and checked each finding below against the code; no blockers. Ordered by impact.

### 1. `taskExecutor.ts:1102` — a completed answer is discarded if the budget expires just before persisting (medium)

In `handleLoopResult` `'response'`, `executionBudget.assertTime()` runs before `pause()` and `completeTurn()`:

```ts
try {
  this.executionBudget.assertTime()
  this.executionBudget.pause()
  await this.deps.conversationManager.completeTurn(this.conversation!, content)
} catch (err) {
  await this.deps.conversationManager.failTurn(this.conversation!)
  throw err
}
```

If the duration budget runs out after the model has produced its final answer (e.g. during `guardProviderWorkflowTriggerClaim` or sanitization), the answer is thrown away. The interruption message is lost too: the inner `catch` calls `failTurn()` (state → Idle), so `run()`'s catch sees `state !== 'processing'` and takes the `failTurn()` branch instead of `completeTurn(error.message)`. The user gets `TASK_DURATION_LIMIT` with no persisted turn content for work that actually finished.

Pausing before the check (or not re-checking time once a final answer exists) would keep the result. Same pattern in `completeWithStaticResponse` (`:1208`); `:1081` is a related check after the workflow-trigger guard.

### 2. `taskExecutor.ts:1071` — iteration exhaustion no longer delivers collected attachments (medium)

On `dev`, `'exhaustion'` shared the `'response'` path, so the user received the exhaustion message plus `result.attachments`. Here the `TASK_ITERATION_LIMIT` throw happens before either is consumed; the failure path only sends `responseCallback({ error })` (`stateMachine.ts:1561`). Any files produced during a long task that hits the iteration cap are no longer delivered.

Minor, same site: the message uses `this.deps.config.maxToolCallsPerTask`, but `restore()` can tighten the effective limit to `min(config, snapshot.maxIterations)`. When it does, the reported count is wrong; reporting the budget's own `maxIterations` would stay accurate.

### 3. `config.ts:309` — empty-string handling differs from the sibling helper (low)

`getExecutionLimit` throws at import for `CLERUM_SHELL_TIMEOUT: ''` (and the other four limits), while `readPositiveSafeIntegerEnv` in this PR (`requestOptions.ts:23`) treats `''` as "use the default". Fail-fast is an improvement over `dev` (`parseInt('')` gave a `NaN` timer), but the two helpers should agree on one rule for empty ConfigMap values.

### 4. `taskExecutor.ts:1463` — budget check throws from inside the event emitter (low)

`consumeIteration()` (which calls `assertTime()`) runs as the first statement of `emit`, so when it throws, the `loop:iteration` event is never forwarded to `coreEvents`/`turnTiming` for that boundary, and `toolUseLoop.ts:106` unwinds through an emit call. The duration timer's abort already covers enforcement, so this mainly costs a telemetry event. Forwarding before checking, or checking at the top of the loop body, would keep emitters side-effect-only.

### 5. `sqliteConversationStore.ts:255` — log line lost its identifier (nit)

`logger.warn({}, 'Cannot pin unknown conversation')` drops the key the old `console.warn` included. If that was intentional (structured session keys embed user IDs), a non-identifying field would still make it actionable.

### Checked, no issues found

Legacy-approval renewal end to end: `renewing` guard, `replaceRequestId` binding on `session_id`/`task_id`/`'legacy'`, atomic delete + insert + `markAwaitingApproval`, rollback on persist failure. Every production path to `suspendForApproval` sets `task_budget` from `pause()`. Also `executeWithTimeout` (the abort signal reaches `callTool`, so timed-out MCP requests are cancelled rather than orphaned), `withAbort` rejection handling, `bindTaskSignal` coverage of `LlmPort`, the dispatch-delay recursion, and migration 014 idempotency.

alfredolopez80:
@claude Please re-review and audit current head `0faff70ee45406f1f0d4b614c6879d542df3699a` after adversarial validation of jozer-rami's review (comment 5698691807) and synchronization with `origin/dev` at `329241bc324393be6a81802e1b434ad20252ba66`.

All 38 current GitHub checks passed. Local reports show 4040 Host tests passed (one existing TODO) and 2297 Desktop tests passed. Four focused regressions fail on the reviewed head `659b947a` and pass with these changes.

Please focus on the surviving issues and their fixes:
- Final deadline checks are outside persistence catches, preserving an interruption record; guard/sanitization are still budgeted.
- Trusted completed artifacts survive interruption, partial tool-batch failure and resumed approvals; failure delivery remains single-shot and reports the effective restored iteration maximum.
- Sync/async Host failure responses retain artifacts. Desktop retrieves failed-task artifacts and displays/downloads them without changing the error state.
- Explicit empty execution-limit values now fail consistently; defaults apply only when absent. The pin warning exposes cache counters, not session identifiers.

The event-emitter observation was not treated as a demonstrated functional defect: budget rejection happens before an iteration is admitted. Please challenge that conclusion if there is contrary evidence.

Coverage limits are explicit: whole-renderer TypeScript checking has the same 38 pre-existing diagnostics as baseline, with no introduced diagnostics. The historical T2 certification applies to `659b947a`, not this new runtime head. Browser E2E and T2 have not been rerun for this revision. Do not interpret CI as runtime certification. Please report evidence-backed remaining findings; do not merge or deploy.

alfredolopez80:
@claude Please perform a complete final re-review and audit of PR #643 at `77126d5f01feca82eb0be622f9835d5e96b434d5`, synchronized with dev `15c4aa016311ec46513f216531107b3ed93dc041`, to leave it ready for human approval. Do not merge or deploy.

Your previous audit (5699694785) found no introduced blockers and was adversarially checked. Please inspect the final diff, merge integration, effective duration/iteration/tool limits, approval persistence and renewal on the same task, cancellation, and interrupted-task artifact delivery through Host and Desktop. Report only evidence-backed remaining defects and distinguish executed checks from static inspection.

Current evidence: all 38 GitHub checks SUCCESS; canonical local orchestrator emitted MINIKUBE_T2_PASS with T0/T1/T2, NP08 HCC authorization and Health all PASS on this exact head. Node 24 suites: 4040 Host tests passed (one TODO), 2302 Desktop tests passed; real PostgreSQL 164 tests across 35 files passed with zero skips. Both production builds passed.

Coverage limits: Playwright NOT_RUN because the existing fixture can alter a personal Desktop session. Supplementary whole-renderer typecheck has 38 pre-existing diagnostics, identical to current dev, zero introduced. Neither limitation is represented as green. The PR description contains the updated evidence.

zach88:
Took a deep pass through this one across the mcp-host core (budget/executor, orchestration, persistence/migration) and the desktop side, plus a dedicated security and end-to-end contract trace. Good news first: the durable-budget mechanics, the atomic legacy-approval renewal, migration 014, dispatch spacing and the interrupted-task artifact retention all hold up against the real code, and security came back clean. Nothing here blocks a merge — what's left is a couple of should-fixes and some nits.

**🟡 Approve with nits**
- Optional / non-blocking: **8** (2 Medium, 6 Low)

## Optional (non-blocking)

1. **[R1-M2] `getExecutionLimit` rejects empty-string env vars instead of falling back to the default** — `mcp-host/src/mcp/requestOptions.ts:24` (same class in `mcp-host/src/config.ts:309-321`). The diff dropped the empty-string tolerance, so `CLERUM_MCP_TOOL_TIMEOUT_MS=''` (and the execution-limit vars) now throw where the merge-base returned the default: `requestOptions` fails per MCP tool call, `config.getExecutionLimit` aborts boot. No current overlay sets these to `""`, but `deploy/base/control-plane/workflow-recipes.yaml` uses the `value: ""` / "leave empty to use provider defaults" idiom on neighboring MCP env vars, so an operator copying that idiom here breaks every MCP call (or startup). Reproduced with a throwaway test: base returns default, head throws. If the stricter validation is intentional, worth documenting and not advertising "leave empty" in the manifests; otherwise fold `''` back into the default path (`if (raw === undefined || raw === '') return defaultValue`).

2. **[R1-M2b] Stated env-validation invariant has no malformed/overflow test** — `mcp-host/src/config.executionLimits.test.ts`. The PR says `getExecutionLimit` rejects empty/malformed/overflow, but the suite only asserts the empty and absent cases. The malformed (non-numeric, negative) and overflow (> 2_147_483_647) branches have no assertion, so an accidental relaxation of the bound checks would stay green. Add an `it.each` for a malformed and an overflow value.

3. **[R1-L1] A NULL `task_budget` row hard-fails reconstruction instead of degrading** — `mcp-host/src/core/conversation/persistence/sqliteConversationStore.ts:1023` (producer) ↔ `reconstruct.ts:146` (reader). A row with SQL NULL `task_budget` isn't treated as legacy/undefined: `JSON.parse(null ?? 'null')` → `null` → `parseTaskExecutionBudget` throws, taking down `load_all_pending_approvals` on boot. Unreachable by the current producer (always writes JSON or `'legacy'`), but a rollback to old code (INSERT without the column → NULL) then roll-forward, or any future insert path that omits the column, drops the whole boot rather than degrading that one row. The fail-closed intent is fine; the blast radius is the concern.

4. **[R1-L2] Persisting `attachments: []` on the failed error message** — `desktop-app/ui/src/hooks/domain/useAgentChatController.ts:1705` and `:2002`. Both durable-error-recovery paths assign `attachments: buildResponseFileAttachments(taskResult)` unconditionally, writing an empty array when there are no artifacts, while the sibling paths (`:1728`, `:2185`, `:2199`) use the conditional spread. Harmless for rendering (`displayAttachments` gates on `.length`), just inconsistent. Same conditional spread fixes it.

5. **[R1-L3] Dead union member `exhaustion.reason: 'iteration_limit'`** — `mcp-host/src/core/types.ts:550` ↔ `toolUseLoopRuntime.ts:206-223`. The union declares `'iteration_limit'` but no producer emits it (`exhaustionResult` omits `reason`; only `taskBrakeResult` sets `'task_budget'`). The consumer at `taskExecutor.ts:1094` treats `undefined` as the iteration case, so it's functionally correct — just a dead member.

6. **[R1-L4] abortableLlmPort test doesn't assert the combined signal is forwarded** — `mcp-host/src/core/adapters/__tests__/abortableLlmPort.test.ts:13`. The mocked `call` ignores its `request`, so the test proves "late result ignored" but not "the underlying op receives the cancellation" — which is the exact point of the runtime fix. The code does forward it; this is a coverage gap. Assert that the `request` handed to `inner` carries an `AbortSignal` that aborts with the task signal.

7. **[R1-L5] Completed attachments dropped on a generic mid-task failure** — `mcp-host/src/agent/taskExecutor.ts:534-538` and `:863-867`. Completed artifacts are only forwarded to `onFail` when `error instanceof TaskLimitError`. A generic failure after tools already produced artifacts (e.g. an `LlmError`) drops `completedAttachments`. This matches the stated scope ("across interruption"), so flagging only so the narrowing is deliberate.

8. **[R1-L6] Log lines lost actionable context in the structured-logger migration** — `mcp-host/src/core/conversation/persistence/sqliteConversationStore.ts:253` (the "pin unknown conversation" warn dropped the `key`, so an incident can't be traced back to a conversation) and `mcp-host/src/messageHandler.ts:320-324` (the dup-suppression log now emits the new suppressed `taskId` and dropped `priorTaskId`/`priorStatus`/the `channelType:channelId:messageId` coordinates that made it actionable). Observability only — a non-sensitive id would restore traceability without reintroducing the session key.

Two trivial nits alongside: `import { randomUUID }` sits above the module JSDoc in `taskExecutor.ts:1`, and `dispatchTimer` in `sessionProcessor.ts:106` has no `.unref()` (negligible at a 3ms delay, just asymmetric with the budget timer).

## What checked out clean

Verified, not assumed:
- **Security**: no introduced vulns. Limit clamps reject 0/negative/overflow (except `agentTaskDelay`, intentionally `allowZero`); legacy-renewal ownership binding (session_id + task_id, prior must be `'legacy'`); no tool runs before approval; `restore()` uses `min(env, snapshot)` so a resumed task can't gain budget; shell process-group cleanup; late-result fencing; no secret leaked in logs (the session key was removed, not reintroduced).
- **Migration 014**: additive, nullable, idempotent, fail-closed; renewal is genuinely atomic in one `immediate()` transaction with correct rollback and double-renewal blocked.
- **Dispatch spacing**: gates on `nextDispatchAt`, single catch-up timer, stamps only on applied transitions, no orphaned timers (`getTimerCount() === 0` in tests).
- **Iteration accounting**: loop runs exactly `remaining`, no replenishment on resume/restart; effective (restored) max is reported, not the configured one.
- **Timeouts**: `executeWithTimeout` single-settle, timers/listeners cleared in `finally`; shell SIGTERM→SIGKILL grace sits inside the 6s cleanup window, partial output preserved.
- **Contracts**: Host↔Desktop attachments, DB worker protocol, migration↔readers and env↔consumers all line up end-to-end and are additive/back-compatible both directions.
- **Suites**: all touched-area suites green on the PR head; the three new desktop tests fail on the merge-base (real regression coverage). Note the mcp-host `tsc --noEmit` reports 4 errors, all pre-existing in files byte-identical to the merge-base.

zach88 (approved):

alfredolopez80:
@claude Please complete a final re-review/audit at `f4234a8d07823b5ea27fca64a1127595fcba880d` after the small follow-up to Zach88 (5703146294). The PR description records the disposition of every finding: intentional empty-ENV rejection retained, 28 malformed/overflow tests added, combined cancellation signal asserted, empty error attachments omitted, explicit iteration-limit reason, safe prior-task log context, module-header cleanup. NULL-accounting recovery and generic-error artifact retention remain outside scope with reasons; do not infer that those behaviors were changed.

All 38 GitHub checks passed. Exact-head canonical certification emitted MINIKUBE_T2_PASS: T0/T1/T2, NP08 and Health PASS. Node 24: 4068 Host tests (+1 existing TODO), 2302 Desktop tests; real PostgreSQL lane PASS. Playwright remains NOT_RUN and the prior supplementary renderer check has known baseline errors, not a green claim. Please verify the small delta, report evidence-backed remaining defects, and distinguish inspection from tests you execute. Do not merge or deploy.

PR #658, opened 2026-09-16
feat(evendrive): breadcrumb folder menu, EvenDrive rename, scroll fix, inherited file permissions

miguelo-999:
## What this PR does

Four EvenDrive (formerly GFS / Global File System) improvements, one lane per EvenTasks ticket. All UI work; no backend changes.

### TASK-235 — GFS → EvenDrive rename
- Renames user-facing GFS / Global File System strings to EvenDrive across desktop-app and control-ui: nav sidebar entries, dialogs, toasts, command palette, attach menus, e2e specs.
- Untouched by design: `gfs://` URIs, code identifiers, API paths, route hrefs, data-testids.

### TASK-242 — Active-folder breadcrumb context menu
- Exactly one ⋮ menu, on the active (currently open) folder segment, positioned right of its name — no per-crumb dots.
- Menu is built through the same construction path as the parent-view folder-row menu, so option parity is guaranteed by construction; includes the Open EvenDrive link action (resolve + navigate).
- Mirrored in both apps; no menu on file segments; breadcrumb navigation regressions covered by tests.

### BUG-133 — File-list scrolling in Control UI
- The EvenDrive file list inside its card did not scroll when a folder held many files. The list frame is now the scroll container (`overflow: auto`) with a sticky list header, matching the `.cu-table-wrap` contract.

### TASK-243 — Inherited file permissions (Google Drive model)
- File Share dialogs show exactly ONE row per member: direct grants merged with inherited ancestor access, strongest role wins. Inherited rows render as normal, toggleable rows — no duplicates, no lock/muting.
- ANY role change on an inherited row (upgrade or downgrade, no exceptions) or removal opens a confirmation modal: "Update role on parent folder?" with the Google-Drive-style two-column before/after (parent folder vs file) and a Learn more help panel.
- Confirm applies the change to the PARENT folder's grant so it cascades to everything inside; any stale direct grant on the file is aligned so it cannot mask the new role. Cancel reverts and sends nothing.
- Folder Share dialogs are unchanged (direct management as before).
- Inheritance is derived on the client: the grants/shares GETs only return direct rows, so ancestor grants (`inherit=true`) and shares (`includeDescendants=true`) are walked per the server's own `holds()` rule. Derivation runs for file dialogs only.

## Testing
- desktop-app targeted suites (GfsGrantList, inheritedAccess, GfsInheritedAccessDialog, FilesPage, useGfsBrowserController): 115 passed.
- control-ui targeted suites (GfsGrantPanel.inherited, GfsBrowser): 64 passed.
- Prettier + commitlint enforced via pre-commit hook.

## Reviewer notes
- The confirm flow also aligns the file's own direct grant when one exists — intentional, to prevent a stronger stale direct grant from masking the parent-level change.
- Deep ancestor trees add a few reads per file dialog (frontend derivation); a future backend effective-grants endpoint would remove that.
- EvenTasks tickets: TASK-235, TASK-242, TASK-243, BUG-133.


miguelo-999:
# PR #658 review — round 1 (feat/evendrive-improvements @ 6d28ae41)

**Decision: Blocked** — 2 must-fix findings, both in the TASK-243 inherited-permissions flow. The rename (TASK-235), breadcrumb menu (TASK-242), and scroll fix (BUG-133) lanes verified clean.

Reviewed `origin/dev...6d28ae41` (58 files, +3627/−305, control-ui + desktop-app). Findings carry stable IDs — please reference them in fix commits.

---

## P0 — must fix before merge

### R1-H1 — Confirmed remove/downgrade only mutates ONE ancestor; other contributing ancestors silently keep the access
- **Severity:** High · **Confidence:** code-verified
- **Evidence:** `desktop-app/ui/src/gfs/inheritedAccess.ts:84-92`, `control-ui/components/gfsInheritedAccess.ts:134-158` (source selection); `desktop-app/ui/src/pages/FilesPage.tsx:629-632`, `control-ui/components/GfsGrantPanel.tsx:522-594` (`confirmParentUpdate`)
- **What is wrong:** the per-subject merge unions permissions across all ancestors, but the editable `source` keeps only one folder (nearest editor-class). The confirmation dialog's file column promises the outcome of the whole change ("No access" on remove, the new role on downgrade), yet confirm mutates only the source folder's rows. `inheritedFrom[]` (multiple contributing folders) is computed but never rendered anywhere.
- **Scenario:** subject has an `inherit` grant `[read]` on `/` and `[read,write,delete]` on `/docs`; file `/docs/report.md`. Remove → deletes the `/docs` grant only; subject retains read via the root. Dialog said file → "No access". Same for downgrade when two ancestors contribute editor-class rows (source = nearest editor; the farther editor row survives).
- **Impact:** an operator is told access was removed when it was not — on the permissions-management surface this is a trust-relevant misstatement, reachable in ordinary configurations (root-level grants are the drive-wide-access mechanism).
- **Required fix:** compute the file column's remaining role across **all** contributing ancestors (not just direct rows + source), and either block/warn on multi-contributor downgrades/removals or apply the change to every contributing ancestor. Surface `inheritedFrom` in the dialog.
- **Required test:** multi-ancestor remove and downgrade cases asserting the disclosed outcome matches the effective post-state (both apps).

### R1-H2 — Desktop: confirmed role change leaves the file's direct *shares* in place, masking the new role
- **Severity:** High · **Confidence:** code-verified
- **Evidence:** `desktop-app/ui/src/pages/FilesPage.tsx:621-624` — only `request.row.grant` is aligned; `request.row.shares` (the merged row's direct file shares, `desktop-app/ui/src/gfs/GfsGrantList/index.tsx:102-125`) are never revoked.
- **What is wrong:** the PR body's stated invariant is "any stale direct grant on the file is aligned so it cannot mask the new role". control-ui honors it for grants *and* shares (deletes `row.direct.shareIds`); desktop revokes the parent's shares but not the file's own direct shares for that member.
- **Scenario:** member has a direct URI share `[read,write]` on the file plus inherited read from the parent → merged row shows Editor → operator downgrades to Read and confirms → parent grant becomes read, the direct editor share survives → member is still Editor; toast says "is now Read-only".
- **Required fix:** in the change-role branch, revoke (or role-align) `request.row.shares` exactly like control-ui does.
- **Required test:** merged row with a direct share; downgrade asserts the share is revoked (current test only covers the direct-grant case, `FilesPage.test.tsx:1766-1831`).

## P1 — should fix

### R1-M1 — Parent-folder grant bits are shaped by the *file's* affordances (desktop)
`FilesPage.tsx:616-619`: `bits = requested.filter(bit => affordances?.grantableBits.includes(bit))` uses the open file's grantable bits to build a grant on the **parent folder**. If `affordances` is null/failed, `bits` is `[]` → server rejects with `400 permissions_invalid` (fail-closed, but a confusing error for a confirmed action). When the file's grantable bits are narrower than the role, the confirmed "Editor" is silently weaker on the parent than the dialog promised. Server-side `assertMayGrant` keeps this conservative, so not a security hole. Fix: derive/validate bits against the parent folder, or document the deliberate reuse.

### R1-M2 — Merge/precedence decision logic has no property-based coverage (T2)
The subject-merge + strongest-source selection is precedence logic implemented twice (`gfsInheritedAccess.ts` ×2 apps) with hand-written cases only; the pure core isn't extracted, which blocks property tests (idempotency `merge(merge(a,b),b)`, total order, no-drop). Extract the pure merge and add `fast-check` properties. A decision table written up front would also have caught R1-H1.

### R1-M3 — Derivation failures are completely silent (both apps)
`GfsGrantPanel.tsx:202-231` (catch → `setInheritedAccess([])`) and `useGfsBrowserController.ts:351-367` (query failure → `[]`): a permissions dialog that silently omits inherited rows invites wrong operator decisions. Per-ancestor best-effort skipping is by design, but total derivation failure deserves at least a quiet inline notice.

## P2 — nits / follow-ups

- **R1-L1** `control-ui/components/GfsBrowser.tsx:746,857` — user-visible toasts still say "GFS Upload v2" after the EvenDrive rename (TASK-235 completeness; owner call if it's a protocol name).
- **R1-L2** Desktop `GfsInheritedAccessDialog/index.tsx:61-66` — `alertdialog` lacks `aria-describedby` and focus restore; the control-ui counterpart has both.
- **R1-L3** Walk semantics diverge: control-ui skips-and-continues past unresolvable ancestors (path walk, unbounded); desktop breaks at the first (rid-walk, capped at `GFS_BREADCRUMB_MAX_DEPTH`=64). Display-only; document or align.
- **D1 note:** 58 files across two apps exceeds the ≤25-files/1-service soft budget; the four-ticket justification in the body is accepted, but all findings concentrate in TASK-243 — consider extracting it if more rounds are needed.

## What looks good (verified)

- **Menu parity by construction:** `folderMenuProps` (control-ui `GfsBrowser.tsx:1021`) / `folderMenuPropsFor` (desktop `FilesPage.tsx:1122`) — breadcrumb and row menus share one construction path; no menu on file crumbs or the synthetic root; "Open EvenDrive link" resolve+navigate flow resets stale dialog state.
- **BUG-133:** `overflow:auto` on `.cu-gfs-list-frame` with a fully bounded `min-height:0` chain and an opaque sticky header; row menus portal to `document.body` so they aren't clipped.
- **Rename discipline:** `gfs://` URIs, code identifiers, API paths, `/global-file-system` route, data-testids all untouched (grepped); e2e/unit assertions updated consistently.
- **Robustness details:** request-id + AbortController race guards in `loadInheritedAccess`; `isAlreadyMissing` 404 tolerance; desktop Escape-layering (parent-update dialog cancels alone); stable list keys (`subjectKey`); folder dialogs verifiably unchanged (`mergeInherited`/`resourceIsFile` gates).
- **Security:** SAFE. No new endpoints, secrets, logs, K8s/NP, or dependencies (package.json = version bump). All mutations go through existing server-authorized surfaces; the server enforces grantability and rejects empty permission arrays. Client derivation is display-only.

## Validation run (Node 24, read-only)

| Check | Result |
|---|---|
| desktop vitest: inheritedAccess, GfsInheritedAccessDialog, GfsGrantList, moveDialog, useGfsBrowserController | 55 passed |
| desktop vitest: FilesPage.test.tsx | 66 passed |
| desktop vitest: SidebarNav.test.tsx | 15 passed |
| control-ui vitest: GfsGrantPanel.inherited, GfsBrowser, Sidebar, ControlAdminsPanel.gfsOperatorLink | 92 passed |

Gaps: Playwright e2e not run (separate evidence lane); no runtime reproduction of the two High findings (code-traced only).

## PR description updates

Accurate except: "any stale direct grant on the file is aligned so it cannot mask the new role" is false on desktop for shares (R1-H2), and the implied "file becomes the chosen role / No access" outcome is false in multi-ancestor cases (R1-H1). Update once fixed, or now to state the limitation.

---

*Review round 1. Out-of-scope/pre-existing items (2 Low) withheld from this comment per review policy — available on request. Fix commits: one finding per commit, referencing the ID, with a `Class-check:` line and a regression test that fails at `6d28ae41`.*

miguelo-999:
## Round-1 review findings — implemented

All findings from round 1 (#issuecomment-5706493922) are addressed, plus the previously deferred items. One commit per finding, each carrying a regression test that fails at the prior head.

| Finding | Commit | What changed |
|---|---|---|
| R1-H1 (P0) | `78e788480` | Drive-aligned multi-ancestor semantics (below) |
| R1-H2 (P0) | `24bc7f405` | Confirmed role changes revoke/align the file's direct shares too (desktop), matching control-ui |
| R1-M1 (P1) | `fa08117f7` | Parent-role bits pre-flighted against the affected folder's own affordances; same-role no-ops don't open the modal |
| R1-M2 (P1) | `f72911ef2` | Pure merge core extracted per app + fast-check property suites (idempotency, no-drop, strongest-wins, total role order) |
| R1-M3 (P1) | `e96a7feb7` | Total derivation failures surface a quiet inline notice in both share dialogs |
| R1-L1 (P2) | `09ad68937` | "GFS Upload v2" toasts renamed to EvenDrive Upload v2 |
| R1-L2 (P2) | `1d974d5da` | Desktop dialog gains aria-describedby + focus restore |
| R1-L3 (P2) | `7a8f9feb4` | Walk semantics unified: bounded depth, skip-and-continue, documented at both sites |
| — | `6263036df` | EvenDrive nav/header icon (hard-drive) in control-ui + desktop navbar |

### R1-H1 semantics (aligned with real Google Drive behavior — nested-folder test: Marketing → Josue Viewer, Marketing/Campaigns → Josue Editor, report.svg inside Campaigns)
- Derivation keeps EVERY contributing ancestor per subject; effective role = strongest across sources.
- **Remove** from a file revokes grants + shares on every contributing ancestor and aligns the file's direct rows. Dialog uses Drive's wording ("Remove from parent folder?" / "Removing … will also remove them from a parent folder…") and lists each affected folder → Remove.
- **Downgrade** lowers every contributing ancestor above the target role; ancestors at or below the target stay untouched.
- **Upgrade** raises the strongest contributing ancestor once.
- **Partial-failure honesty**: sequential per-folder calls; first failure stops, state refreshes, and the toast states reality ("Removed from 1 of 2 folders — Marketing still grants access").

Verification: desktop-app targeted suites 132/132, control-ui 71/71.

miguelo-999:
@claude please review this PR, check the previous reviews and provide your own independent review if there's feedback sort it by P0, P1, etc and if everything is good provide explicit green light.

miguelo-999:
@claude round 2 done bro, N1 N2 N3 all fixed with regression tests and the dev conflicts are resolved (merge commit b85fb98ba, kept both the icon cleanup from dev and our EvenDrive changes). everything green on my side, take another look pls, need that green light

miguelo-999:
@josue-tech-94 full review cycle completed please take a look

zach88:
Took a deep pass through this one. The rename, breadcrumb menu, and scroll fix all came out clean; basically everything worth flagging is concentrated in the inherited-file-permissions work (TASK-243), which is implemented twice (desktop-app + control-ui) and the two sides have drifted apart. There's also one real regression that a later fix commit reintroduced.

### 🔴 Not ready to approve
- **Must resolve before approval: 2**
- **Optional / non-blocking: 12**

1. **[R1-B1] Downgrade leaves a stale direct Editor grant while the toast claims success** (regression from `61818072c`) — `desktop-app/ui/src/pages/FilesPage.tsx:738-760`.
   The N2 fix ("skip empty file-alignment grant") reintroduced, for the direct *grant*, the masking that R1-H2 (`24bc7f405`) had closed for shares. When you downgrade a merged row to Read and the file's `grantableBits` come back empty (`affordances` null / refetch failed), N2 skips the alignment grant, so the member's direct Editor grant is never lowered or revoked, and the flow still reports success: *"<label> is now Read-only on <folders> and everything inside it."* The member stays Editor via the direct grant while the UI says access was reduced. On a permissions surface, a downgrade that silently doesn't happen but claims it did is the dangerous case.
   Repro: merged row for a member with a direct grant `['read','write']` on the file + inherited Editor, `affordances.grantableBits = []`, downgrade to Read → success toast fires, the grant is never called, the Editor grant survives. Verified with a throwaway test.
   Fix: for the empty-bits case, *revoke* the direct grant (like the `remove` path does) instead of skipping it. Also fix its test (see R1-L5): the current fixture uses a read-only direct grant so it never exercises the dangerous case, and it asserts an intermediate effect ("grant not called") instead of the observable effective role.

2. **[R1-H1] control-ui has no per-folder affordances pre-flight; desktop does** — `control-ui/components/GfsGrantPanel.tsx:585-640` vs `desktop-app/ui/src/pages/FilesPage.tsx:603-648`.
   Desktop walks every affected folder before opening the modal, calls `gfs.affordances(resourceId)`, validates `canDelegate` + `grantableBits`, and fails loudly without opening the modal if a folder can't express the role. control-ui has zero affordances/grantable/canDelegate references: it opens the modal optimistically and applies `rolePermissions(nextRole)` uniformly, only finding out at the server. Same business rule, two implementations, different observable outcome: on control-ui a partial state is possible ("Updated 1 of 2 folders") where a member stays effectively Editor via the folder that failed.
   Worth calling out: this is **not** a privilege escalation. Enforcement is server-side and deny-by-default, so the client walk is display-only. Since control-ui is the admin plane (full authority via control-api), if an admin never lacks `manage_acl` over a visible ancestor, the divergence is defensible and can be closed in writing with that justification instead of aligned. Your call: extract the pre-flight into the shared module and run it on both sides, or close-by-written. Flagging it as gating because it's the biggest design gap in the PR, but it may legitimately resolve to "documented + tested".

## Optional (non-blocking)

The PR can merge without these; follow-up or author's discretion.

3. **[R1-M1] Inherited-walk divergence: `break` vs `continue` on an unresolvable mid-ancestor** — `desktop-app/ui/src/gfs/inheritedAccess.ts:57-64` vs `control-ui/components/gfsInheritedAccess.ts:113-128`. For `/a/b/c.md` where `/a` grants `inherit=true` but `/b` can't be resolved, desktop breaks at `/b` and the `/a` grant disappears; control-ui skips `/b` and keeps it. Same file, different inherited access per app. It's documented in both comments as "one divergence" but it's still observable authz drift.

4. **[R1-M2] Total-failure semantics applied to only one side** — `control-ui/components/gfsInheritedAccess.ts:144-146` vs `desktop-app/ui/src/gfs/inheritedAccess.ts:48`. When no ancestor resolves, control-ui throws `inherited_access_derivation_failed` and surfaces a notice; desktop returns `[]` silently and its `inheritedAccessError` never fires for that case, hiding that the member might be inherited. The R1-M3 fix only landed on one mirror.

5. **[R1-M3] Direct-grant alignment sends full permissions unconditionally on control-ui** — `control-ui/components/GfsGrantPanel.tsx:659-676`. The other side of R1-B1: control-ui `putGfsGrant` with the full role, no affordances filter, no guard; if the file's affordances don't allow those bits the server rejects it and you get "direct access still shows the old role" even though the parent changed. It fails loud (better than B1's silence) but it's still an unhandled path.

6. **[R1-M4] Merge property tests miss composability and a real no-duplicates assertion** — both `desktop-app/ui/src/gfs/__tests__/inheritedAccessMerge.prop.test.ts` and `control-ui/components/__tests__/gfsInheritedAccessMerge.prop.test.ts`. They cover total order, no-drop, idempotence, strongest-role, but not composability `mergeOuter(mergeInner(a,b),a) === mergeInner(a,b)`, and "no duplicate ids" compares `new Set(...)` on both sides so it'd never catch a repeated source. Repro: I ran composability as a throwaway property test (500 runs) and it passes, so this is a coverage gap, not a hidden bug. Add both properties to both mirrors.

7. **[R1-M5] "Open EvenDrive link" folders lose their version, so breadcrumb Move/Rename/Delete send `ifMatch: 0`** — `control-ui/components/GfsBrowser.tsx:1041-1068` + `:1008-1019`. The resolve view returns `updatedAt` but no version, `crumbToChild` defaults `version` to `0`, and Rename/Move/Delete all send `ifMatch: 0`. A folder at real version > 0 gets a 409; a folder at version 0 mutates with no concurrency guard. The new breadcrumb actions are effectively unusable on link-navigated folders. Repro: verified `ifMatch` is `0` on a link-navigated folder vs the correct value via `openDirectory`, with a throwaway test. Fix: expose the version from the resolve view and populate the crumb, or disable the mutating actions on link-navigated crumbs until a real version is known. Also lock it with a non-zero-version fixture (R1-L7 — the current `version: 0` fixture hides this).

Nits (Low):
- **[R1-L1]** `desktop-app/.../GfsInheritedAccessDialog/index.tsx:103` uses `key={folder.name}`; two ancestors with the same visible name (different resourceId) produce duplicate keys. React still renders both rows, just warns. Key by resourceId.
- **[R1-L2]** `desktop-app/.../GfsGrantList/index.tsx:220` merged-row `data-testid` is `gfs-access-row-inherited-${subject.type}` (type only); two members of the same type collide and `getByTestId` would throw. React's `key` already uses `subjectKey`; do the same for the testid.
- **[R1-L3]** `desktop-app/ui/src/pages/FilesPage.tsx:708` `sources[0].name` trips TS2532 under `ui/tsconfig.json`. Runtime-safe (guarded by `sources.length === 1`) and not CI-gated, but easy to fix: `const [firstSource] = sources`.
- **[R1-L4]** `control-ui/components/GfsGrantPanel.tsx` opens the "Update role on parent folder?" modal with an empty folder list when the change only realigns the direct grant (direct=Editor, inherited=Read floor, drop to Read). Frames a purely-direct change as a parent-folder update.
- **[R1-L6]** `control-ui/components/gfsInheritedAccess.ts` skip-and-continue walk (the behavior behind R1-M1) has no explicit test; desktop does (`inheritedAccess.test.ts`).

## What checked out clean

- **Security: safe.** The client inheritance walk is display-only; enforcement stays server-side and deny-by-default, and every mutation (`putGfsGrant`/`deleteGfsGrant`/`deleteGfsShare`) is server-authorized. The replicated `holds()` matches the server's `allow()`, the cascade blast radius is disclosed in the modal, stale-grant alignment is scoped per subject, no secrets logged, no injection in `getGfsResourceByPath` (URLSearchParams), no new deps.
- **Merge core + precedence rule** are byte-identical across both apps (`roleRank` Read<Editor, `strongestInheritedSource`, `planInheritedRoleChange`, `mergeInheritedAccessItem`), ties resolved the same, depth cap 64 on both. The extraction (R1-M2 commit) is output-equivalent to the old inline logic.
- **Breadcrumb (TASK-242):** parity by construction verified (single `folderMenuProps` helper), exactly one menu on the active folder, none on crumbs/files/root, navigation intact, "Open EvenDrive link" resolves through the real endpoint.
- **BUG-133 scroll:** `.cu-gfs-list-frame overflow:auto` + sticky head match the `.cu-table-wrap--sticky-header` contract; no collapse with few files, no double scroll.
- **Rename (TASK-235):** user-facing strings only; `data-testid`, `gfs://` URIs, `/api/` paths, hrefs, `.da-gfs-*` classNames, and `window.clerum.gfs.*` all left intact; e2e specs updated in lockstep.
- **Suites the review ran:** desktop core 33/33, desktop breadcrumb area 122, control-ui core 26/26, control-ui browser area 80; gated desktop typecheck (`src/**`) green.

One process note, take it or leave it: this is 64 files across two services and four independent tasks. Since rename/breadcrumb/scroll all came out clean, extracting those three into their own PR and merging them now would leave a much smaller TASK-243-only PR for round 2, so we're not re-reviewing ~40 healthy files to find bugs concentrated in ~10.

josue-tech-94:
@miguelo-999  Following up on Zach’s review with an independent pass over the current head.

Zach’s existing comments remain the primary findings; this review was intentionally deduplicated against them. We **confirmed the two blocking items are still open** and found **three expansions to Zach’s existing findings**, but no separate new root-cause issue.

### Must address before approval

* **R1-B1 — Desktop downgrade can leave a direct Editor grant while reporting success.**
  Confirming Zach’s finding. Please handle the empty-`grantableBits` case so a stale direct Editor grant cannot continue masking the requested Read-only role. The regression test should use the dangerous Editor case and assert the **observable effective role**, not just whether a grant call was skipped.

* **R1-H1 — Control UI has no per-folder affordance preflight.**
  Confirming Zach’s finding. Either align Control UI with Desktop by validating each affected folder before applying mutations, or, if the admin-plane contract genuinely guarantees the required authority, close this explicitly in writing with that contract documented and tested. The current partial-update behavior needs a deliberate disposition.

### Expansions to Zach’s findings

These are **not new IDs**; they expand issues Zach already identified:

* **R1-H2 — Desktop direct-share state is stale after inherited updates.**
  The server-side share revoke is present, but the UI does not refresh the direct-share query after successful or partially successful inherited updates. Please refresh shares wherever those mutations can occur and add a test asserting the rendered effective role after refresh.

* **R1-M5 — Breadcrumb version freshness also breaks after a normal rename.**
  Zach identified the `ifMatch: 0` problem for link-opened folders. The same version contract also becomes stale after a normal breadcrumb rename because the returned new resource version is not stored. Please preserve the real version for link-opened crumbs **and** update the crumb version after mutations; cover consecutive mutations with a non-zero version fixture.

* **R1-L2 — Subject-type-only test IDs also collide in Control UI.**
  Zach found this in Desktop; Control UI has the same pattern. Use the stable subject key/identity rather than subject type alone in both surfaces and cover two subjects of the same type.

Zach’s other non-blocking findings remain covered by his original comment; no need to duplicate them here.

Once the above is addressed, please update the PR description so the inherited-permissions work does not overstate completion, and include validation for:

* direct Editor → Read-only downgrade with empty file affordances;
* Control UI per-folder preflight / documented authority contract;
* direct-share refresh after inherited mutations;
* breadcrumb version continuity across consecutive mutations.

After the fixes, the next pass should verify Zach’s findings against the new head and run the bounded collateral review from the currently reviewed head.

miguelo-999:
@zach88  @josue-tech-94 Thanks for the review. I addressed the blocking findings, follow-up items, and applicable nits.

Blocking items:

- R1-B1: Empty `grantableBits` during a downgrade now revokes the stale direct grant instead of silently skipping alignment. The regression test now covers a direct Editor grant plus inherited access and verifies the effective role after refresh.
- R1-H1: Control UI now performs per-ancestor affordance preflight, validating `canDelegate` and `grantableBits` before opening the parent-folder confirmation or applying updates. This now matches Desktop behavior and fails closed when the requested role cannot be expressed.

Follow-up items:

- R1-H2: Desktop refreshes direct shares after inherited-access mutations, including partial-success paths. Tests verify the rendered effective role after refresh.
- R1-M5: Breadcrumbs retain the real resource version for link-opened folders, and mutation responses update the crumb version. Added coverage for consecutive mutations with non-zero versions.
- R1-L2: Desktop and Control UI test IDs now use stable subject identity (`subject.type:id`) instead of subject type alone. Added coverage for multiple subjects of the same type.

Additional nits:

- R1-L1: Inherited-access dialog rows are keyed by `resourceId`, avoiding collisions for folders with the same name.
- R1-L3: Replaced unsafe `sources[0]` access with a destructured, guarded first source.
- R1-L4: Control UI now handles direct-only realignment without opening an empty parent-folder modal. Added regression coverage.
 EvenFire Land… (1)＋- R1-L6: Added an explicit Control UI test proving that an inaccessible ancestor is skipped while higher ancestors continue to be processed.

josue-tech-94:
Round-four follow-up on PR #658.

The important point from this pass is that Zach’s existing finding IDs remain the source of truth. Most of the blocking/follow-up work from the previous round is now fixed, but a few items remain, and **the round-four design stop now applies** because the repeated issues continue to concentrate in TASK-243 across Desktop, Control UI, and now a supporting Control API contract.

### 1. First: write the TASK-243 mini-spec before more isolated patches

Please pause further TASK-243 feature/fix work long enough to define the shared contract for:

* effective permission role + inherited source model;
* Desktop delegated-user authority vs Control UI operator authority;
* per-ancestor feasibility/preflight;
* direct grant/share alignment;
* sequential partial-success behavior;
* who owns query refresh after mutations;
* who owns resource-version freshness;
* producer-backed cross-app contract tests.

The recurring issue is that the same permission behavior exists in multiple places and has drifted between them. The next fixes should follow one agreed model rather than another local patch.

### 2. R1-M5 — still partially open: breadcrumb version becomes stale after Move

**Area:** `control-ui/components/GfsBrowser.tsx`

The link-open version issue and rename version update are fixed, but the **Move** path still sends the current version, ignores the version returned by the successful mutation, and leaves the active breadcrumb crumb holding the old version.

That means:

1. active folder is version N;
2. Move succeeds and server advances it to N+1;
3. breadcrumb still stores N;
4. next Rename/Move from that breadcrumb uses stale `ifMatch: N`;
5. mutation can fail with a precondition conflict.

Please consume the successful Move response and update the matching breadcrumb’s stored version, just as the rename path now does.

**Test gap:** `control-ui/components/__tests__/GfsBrowser.test.tsx`

The new non-zero consecutive-version test currently exercises listing-row renames, not repeated breadcrumb mutations.

Please add a test that:

* opens/navigates a breadcrumb folder with a real non-zero version;
* performs a breadcrumb mutation including Move;
* receives the next version from the server;
* performs another breadcrumb mutation;
* proves the second request uses the version returned by the first.

Keep this under **R1-M5 / R1-L7**, not a new finding.

### 3. R1-M1 — Desktop and Control UI still disagree when an ancestor cannot be resolved

**Areas:**

* `desktop-app/ui/src/gfs/inheritedAccess.ts`
* `control-ui/components/gfsInheritedAccess.ts`

Desktop still stops its inherited-access walk when an ancestor cannot be resolved, while Control UI skips that ancestor and continues higher.

That means the same resource can display different inherited access depending on which app is used.

Please use the mini-spec to choose one canonical behavior and either align both implementations or explicitly disposition the difference with a documented reason and contract test.

### 4. R1-M2 — Desktop can still silently produce no inherited access after ACL-read failures

**Area:** `desktop-app/ui/src/gfs/inheritedAccess.ts`

The total resolve-failure notice improved, but there is still a separate case where ancestor ACL-list reads fail and are swallowed. If all relevant reads fail, Desktop can return an empty inherited result that is indistinguishable from “there is no inherited access.”

Please make the failure semantics explicit according to the mini-spec so an ACL-read failure cannot silently look like a real empty permission set.

### 5. R1-M4 — property-test coverage is still incomplete

**Areas:**

* `desktop-app/ui/src/gfs/__tests__/inheritedAccessMerge.prop.test.ts`
* `control-ui/components/__tests__/gfsInheritedAccessMerge.prop.test.ts`

Two gaps remain:

* the duplicate-source property is effectively vacuous because both sides are converted to `Set`s, so duplicate IDs cannot be detected;
* the previously requested composability property is still missing.

Please add a real uniqueness assertion and the composability invariant to both implementations.

### 6. R1-H2 — implementation fixed, but strengthen the regression proof

**Area:** Desktop inherited-permission flow / `FilesPage` tests.

The implementation now refreshes shares on success and partial failure, so the code-side stale-share problem appears fixed.

The remaining issue is the regression test: the `revokeShare` mock itself removes the share from component state while `refreshShares` is effectively a no-op. Therefore the rendered Read result can pass without proving that **refreshing the query** is what removed the stale Editor share.

Please adjust the test so:

* revoking the share does **not** directly update the rendered/query state;
* the stale share remains until `refreshShares` returns refreshed data;
* the final rendered role proves the refresh contract;
* include the partial-failure path as well.

Keep this under **R1-H2**; no new ID is needed.

### 7. Update the PR description

The current description still says:

> “All UI work; no backend changes.”

That is no longer true.

The PR now changes the Control API, including:

* `control-api/src/gfs/resolve.ts`
* `control-api/src/routes/gfs/resolve.ts`
* `control-api/src/routes/gfs/resources.ts`
* related Control API tests/dependencies.

Please update the description to mention the resource-version/affordance API support and route protections.

A suitable replacement for that sentence is:

> The UI changes use Control API support for resource versions and authenticated GFS affordances; GFS resource routes are rate-limited before authentication and again through the distributed per-operator buckets.

Also avoid describing TASK-243 version continuity as complete until the breadcrumb Move path above is fixed.

### CodeQL

No additional CodeQL fix is requested from this review. The rate-limit reports were consolidated and checked against the current code:

* the existing PATCH omission was pre-existing at the merge base;
* the new affordance GET is now protected;
* the current implementation rate-limits both routes;
* current CodeQL passes.

The older inline CodeQL threads being outdated/unresolved in GitHub does not represent a separate open code finding.

### Already confirmed fixed

No additional work is requested for Zach’s:

* **R1-B1**
* **R1-H1**
* **R1-L1**
* **R1-L2**
* **R1-L3**
* **R1-L4**
* **R1-L5**
* **R1-L6**

**R1-M3** is also no longer applicable under the clarified authenticated global-operator contract.

miguelo-999:
Scope clarification for reviewers

Commit 5a9dece00 reverts the control-api/backend changes from this PR. The branch no longer adds the GFS affordances endpoint, resolver version fields, or route rate-limiting middleware; control-api now matches origin/dev for those paths.

The backend follow-up is tracked in issue #774: https://github.com/evenfire-ai/evenfire/issues/774. Backend team ownership is required for the route rate limits and for the supported affordances/resource-version contracts.

For R1-H1, Control UI remains the admin plane and server authorization remains authoritative. This PR keeps the inherited-access mutation flow and honest partial-failure refreshes, but does not call a new backend affordances endpoint or claim client-side preflight parity.

For R1-M5, normal breadcrumb mutations still use the server-returned version. A link-opened breadcrumb does not expose mutation actions until the backend resolve contract supplies a real version, so the UI never sends a fabricated ifMatch value. The remaining desktop and Control UI review fixes are still included.

zach88:
Took another deep pass through this one for round 2. Good news up front: the round-1 Blocker (stale Editor grant on downgrade) and both Highs (partial-authz state, direct-share refresh) are genuinely closed and I verified the B1 regression test actually fails against the pre-fix head. The fixes didn't reintroduce any runtime regression either. What's left is concentrated in TASK-243 and the new control-api surface: mostly test-quality gaps, one type-error the fix introduced, and a stale PR body.

**🔴 Not ready to approve**
- **Must resolve before approval: 1** (an open regression from a round-2 fix)
- **Optional / non-blocking: 12** (Medium + Low)

Only blocking item is a regression, and it's a cheap fix. Flagging it because a fix that broke something shouldn't ride in as-is, not because it's scary. This is my read from the review, not a formal GH approval.

1. **[R2-M1] The R1-L1 key fix broke the type-check and doesn't do what it intended** `regression from 26810b18f`
   The nit fix swapped `key={folder.name}` for `key={folder.resourceId}` in `desktop-app/ui/src/gfs/GfsInheritedAccessDialog/index.tsx:103`, but `GfsInheritedAccessDialogFolder` (`types.ts:3-7`) has no `resourceId`, and `FilesPage.tsx:2104-2111` builds those folders as `{ name, currentRole }`. **Repro:** `tsc -p ui/tsconfig.json --noEmit` emits `TS2339: Property 'resourceId' does not exist`. At runtime `key` is always `undefined`, so React falls back to index and the original goal (stable keys across homonymous ancestors) still isn't met. CI doesn't catch it because `ui/` isn't type-checked. Fix: add `resourceId: string` to the type and pass `source.resourceId` / `update.source.resourceId` in the `.map`.

## Optional (non-blocking)

The PR can merge without these; they're follow-ups or your call.

2. **[R2-M2] The breadcrumb-version tests don't cover the bug** (T3/T4)
   The M5 code fix is correct, but both tests in `control-ui/components/__tests__/GfsBrowser.test.tsx:1097,1480` pass green against the pre-fix head. **Repro:** copied them onto the round-1 head and both stay green. The renames test operates on file rows (never touches `crumbToChild`); the ⋯-menu test only checks the menu renders, never asserts the `ifMatch` of a breadcrumb mutation (fixture uses `version:7` but nothing asserts it). Add a test that opens a folder via "Open EvenDrive link" and fires Rename/Move/Delete from the crumb's ⋯, asserting the PATCH carries `ifMatch:7` (and a second rename carries the returned version, not 0). It should fail against the pre-fix head.

3. **[R2-M3] No tests for the new affordances endpoint or the rate-limiting** (T5)
   `control-api/src/routes/gfs/resources.ts:406-440` adds an authenticated route that returns authority info plus a double rate-limiter, with zero coverage: no 200 shape/values, no 400 on bad UUID, no 429/`Retry-After`. The existing `affordances` tests cover the external endpoint, not this one.

4. **[R2-M4] The affordances endpoint hardcodes its response instead of reusing `heldPermissions`** (D4)
   `resources.ts:425-431` returns `held/grantableBits = [...PERMISSIONS]`, `canDelegate/canCreateShare: true` for any valid UUID without touching the authority engine. `heldPermissions(db, caller, drive, id)` already exists (`grants.ts:512`, "the all-bits answer the affordances panel needs") and the twin external endpoint (`routes/external/gfs.ts:700`) uses it, with a different response shape (`{held, isOperator}` vs the five-field one here). Two implementations of the same concept in the seam. `grantableBits: all` is also inaccurate for a `1st:*` host (`assertTargetPolicy` blocks `delete/manage_acl/share` even for operators), and it makes the control-ui preflight a no-op, so the "cannot make X an Editor" branch (`GfsGrantPanel.tsx:590-600`) is dead in production. Not a leak today (operator-only, equivalent to `heldPermissions(operator)`). Either derive from `heldPermissions`, or document that the preflight is decorative and H1's close rides on the honest partial-failure message.

5. **[R2-M5] PR body says "All UI work; no backend changes", which isn't true anymore** (R4)
   Round 2 added `control-api/src/routes/gfs/resources.ts` (endpoint + rate-limit) and `control-api/src/gfs/resolve.ts` (the `version` field). The Testing section also still shows the round-1 numbers (115/64). Worth syncing the body, since it's part of what gets reviewed.

6. **[R1-M2] Desktop silently returns `[]` when the ancestor walk fully fails** (re-triage: 2 rounds open, needs to escalate or be closed in writing)
   `desktop-app/ui/src/gfs/inheritedAccess.ts:85-90` returns `[]` with no zero-resolved guard, while control-ui throws `inherited_access_derivation_failed` and surfaces a notice. **Repro:** file with a readable direct grant + first parent rate-limited, desktop renders "no inherited access" and hides that the member may be inherited. Safe direction (desktop under-reports, server still enforces), so closing it in writing is defensible, but it shouldn't drift into a third round undecided.

7. **[R1-M4] Merge prop-tests still miss composability and a direct no-dup assertion** (T2; re-triage: 2 rounds open)
   Both mirrors (`desktop-app/ui/src/gfs/__tests__/inheritedAccessMerge.prop.test.ts`, `control-ui/components/__tests__/gfsInheritedAccessMerge.prop.test.ts`) cover idempotence/no-drop/strongest/total-order but not composability, and "never drops a source" compares `new Set(...)` on both sides, which masks a duplicate resourceId. The fuzz also never feeds the same folder twice. It's a coverage gap, not a masked bug. Adding the two properties is cheap.

8. **[R2-L1] `refreshCurrent`/`load` don't repopulate the active crumb's version** `control-ui/components/GfsBrowser.tsx:562-565`. Only rename does. Safe (a stale version derives a legit 409, not `ifMatch:0`), low impact.

9. **[R2-L2] Affordances doesn't validate resource existence or honor `?drive=`** `control-api/src/routes/gfs/resources.ts:412`. Not an enumeration oracle (constant response), resolves itself if you fix R2-M4.

10. **[R2-L3] Pre-auth edge rate-limiter keys by IP** `control-api/src/routes/gfs/resources.ts:144-151`. With a second proxy hop under `trust proxy: 1`, operators behind one NAT share a bucket. Mitigated: the post-auth PG limiter (per hashed subject) is the authoritative one; the edge is an explicit backstop. Worth watching if a hop gets added.

11. **[R2-L4] A new test certifies pre-existing behavior** `control-ui/components/__tests__/gfsInheritedAccess.test.ts`. Its target is byte-identical across the range, so it passes against the prior head (T3): added coverage, not a repro test.

12. **[R2-L5] ~15 type errors in the PR's own test files** (`FilesPage.test.tsx`, both prop-tests, `GfsGrantList.test.tsx`). They run green under vitest/esbuild but are type debt CI can't see.

13. **[R2-L6] The fix commits skip the ledger discipline** (F4). `26810b18f` bundles every round-1 finding into one commit with no ledger ID, `Class-check:`, or `Repro-test:` in the message. Makes it impossible to bisect or selectively revert a future regression back to its cause.

## What checked out clean

- **B1, H1, H2 closed and verified.** Empty-bits now revokes the stale direct grant (`FilesPage.tsx:761-766`), the B1 test uses a real Editor grant and asserts the observable role, and I confirmed it fails against the pre-fix head. `refreshShares()` is in all four terminal paths.
- **No runtime regressions** from the round-2 fixes. Suites green at HEAD: control-ui 92, desktop 142, control-api 226. The testid generalization (subjectKey) landed on all three producers with no old-form consumer left.
- **Security: clean** on the new backend surface. The affordances endpoint is operator-only and non-authoritative (returning full bits grants nothing; the operator is already root-of-trust and every mutation stays server-side gated). Rate-limit keys hash the subject, and `trust proxy: 1` keeps `req.ip` unforgeable for the per-subject bucket. The empty-bits `revokeGrant` can only remove access, never elevate.
- **Merge core** stays byte-identical between the two mirrors. Contract shapes (affordances, `version`) line up across all three layers.
- **No CRD changes.**

One process note: this is still 73 files across 3 services (control-api joined in round 2) and 4 independent tasks. Rename/breadcrumb/scroll came out clean both rounds and everything live is in TASK-243 + the new backend. Extracting the three clean tasks into their own PR and merging them now would shrink what each round has to re-review. Also, since the affordances endpoint is new: deploy control-api before control-ui (a new UI against an old API gets a 404 and fails closed on the inherited role-change, which is safe but blocks the flow).

miguelo-999:
@josue-tech-94 @zach88 all requested changes applied and done

josue-tech-94:
Thanks for the latest fixes. I re-reviewed the current head `e724f112f008b88ba26fa578ed33e0f3a6aa5156`.

**This is not ready to approve yet.** Most prior findings are now closed, but there is one new correctness issue plus one remaining producer/test-contract gap.

### Must address

**R5-M1 — Moving the active breadcrumb folder leaves the breadcrumb path stale**

**Area:** Control UI EvenDrive breadcrumb / move flow.

Scenario:

* current folder is `/org`;
* move `/org` under `/archive`;
* the move succeeds and the folder’s name/version state is updated;
* the active breadcrumb still represents `/org` instead of `/archive/org`.

The move changes the folder’s ancestry, so updating only the active crumb is insufficient.

Please rebuild the breadcrumb trail from the folder’s new location after a successful move, rather than only updating the existing crumb metadata.

Add regression coverage for moving the **currently active breadcrumb folder** to another parent and verify the resulting breadcrumb path reflects the new ancestor chain.

### R1-M5 / R2-M2 — link-opened breadcrumb version coverage is still incomplete

The new `7 → 8 → 9` test assumes link resolution returns a resource `version`.

The current real resolver does **not** provide that version.

The UI currently behaves safely when the version is absent by hiding mutation actions, but the test is using a shape the real producer does not emit. Therefore it does not prove that a folder opened through an EvenDrive link can actually use breadcrumb mutations with a valid concurrency version.

Please do one of these:

1. Establish a real producer contract that supplies the version and make the regression test derive its fixture from that real contract; or
2. Explicitly defer link-opened breadcrumb mutations and keep them unavailable when no real version exists.

Do not keep a synthetic test fixture that assumes a producer shape that does not exist.

miguelo-999:
@josue-tech-94 Thanks for the thorough re-review. Both findings are addressed in cb02133f9 (Control UI only — no backend changes), rebased on top of the latest dev merge (7dc0bbbdb).

R5-M1 — stale breadcrumb after moving the active folder

Agreed: the move only patched the moved crumb's name/version, so the ancestors above it still described the old location. moveResource now rebuilds the trail from the folder's new location whenever the moved resource is part of the open trail:

- After a successful PATCH, the moved folder is resolved (/api/v1/gfs/resolve), and its new path drives a by-path ancestor walk (getGfsResourceByPath → /api/v1/gfs/by-path) — so /org moved under /archive now yields main / archive / org.
- Crumbs below the moved folder keep their relative order (they moved with it); the moved crumb carries the receipt version, and if the receipt supplies none, the stale pre-move version is dropped (the crumb honestly loses its mutating menu instead of issuing a guaranteed-conflict ifMatch).
- The old parent's cached children listing and the destination's are both dropped so later navigation revalidates.

Regression test added (rebuilds the breadcrumb trail after moving the open folder to another parent): opens /org by navigation (listed version 7), moves it under /archive via the breadcrumb ⋯ menu, asserts the PATCH ran with ifMatch: 7, the resolve + by-path walk fired, the breadcrumb renders main / archive / org in order, and a follow-up breadcrumb rename uses the post-move version (ifMatch: 8).

R1-M5 / R2-M2 — link-opened breadcrumb version coverage

We took the second option: explicit deferral, no backend changes. Confirmed the real producers first — control-api toResolveView returns no version (gfsc /v1/resolve does, which is why the Desktop fixtures are legitimate), so the synthetic version: 7 resolve fixture was a producer shape that does not exist. It is removed.

- The replacement test derives its fixture field-for-field from the real resolve contract (resourceId, rid, gfsUri, drive, name, kind, path, updatedAt — no version) and pins the deferral: the link-opened folder still lists children, its active crumb renders no mutating ⋯ menu, and zero mutation requests are issued against an invented ifMatch.
- The deferral is now documented explicitly at openEvenDriveLink in GfsBrowser.tsx: until the resolve contract supplies a version (backend issue #774 owns that producer change), crumbToChild returns null for versionless crumbs and breadcrumb mutations on link-opened folders stay unavailable — not run against a fabricated version.
- The earlier link-open test's resolve fixture was also aligned to the complete real contract shape so no test implies a versioned resolver.

josue-tech-94:
Thanks for the latest update. I re-reviewed the current head `cb02133f911629f656f730fd46882e045c4c34e8`.

**Round 6 is still blocked**, but the remaining code issue is now narrow.

### R5-M1 — still partially open

The successful active-folder Move path is improved and now rebuilds the breadcrumb correctly.

For the happy path:

- `/org` moved under `/archive`
- breadcrumb becomes `main / archive / org`
- the moved crumb carries the new version
- the follow-up mutation uses the updated version

That part checks out.

The remaining problem is the **post-move reconstruction failure path**.

After the Move PATCH has already succeeded, if either:

- `/api/v1/gfs/resolve` fails, or
- one of the ancestor `/api/v1/gfs/by-path` lookups fails,

the code currently drops the rebuilt ancestors and still presents a shortened breadcrumb such as:

`main / org`

while also reporting the Move as successful.

That breadcrumb is no longer the resource’s real location, so the UI is presenting incomplete path state as if reconstruction succeeded.

Please keep this under **R5-M1** rather than creating another ID.

#### What to change

After a successful Move, if canonical path reconstruction fails:

- do **not** silently replace the breadcrumb with a misleading partial ancestry;
- preserve an explicitly stale/incomplete state or otherwise make it clear that the move succeeded but the path could not be refreshed;
- surface a retry/error state so the user can recover;
- do not present the shortened path as authoritative.

Please add regression coverage for both failure cases:

1. Move succeeds, then `resolve` fails.
2. Move succeeds, `resolve` succeeds, but an ancestor `by-path` lookup fails.

The test should verify both the visible breadcrumb state and the user-visible error/retry behavior.

### R1-M5 / R2-M2 — closed in writing under the explicit deferral

The link-open version behavior is now acceptable under the chosen contract.

The real Control API resolve producer does not return `version`, and the Control UI now reflects that honestly:

- link-opened folders still open and list normally;
- versionless crumbs do not expose mutating breadcrumb actions;
- no fabricated `ifMatch` is sent;
- the tests now use the real producer shape;
- backend issue **#774** owns the future resolver-version contract.

So **R1-M5 and R2-M2 can remain closed in writing under this deferral**, and **R1-L7 is obsolete**.

No further work is requested there in this PR unless the product decision changes.


### PR description

The scope statement:

> “All UI work; no backend changes”

is accurate for the current PR.

The exact Testing counts are still stale:

- Desktop: `115`
- Control UI: `64`

Please update them to the current numbers or remove the exact counts if they are no longer intended to be maintained.

### Existing follow-ups

The historical low-priority items:

- R2-L1
- R2-L4
- R2-L5
- R2-L6

remain documented follow-ups. They are **not new blockers from this round**.

No new security issue was found, and the old CodeQL rate-limit threads are outdated/duplicate context outside the current merge-base diff.

miguelo-999:
Pushed: 7cdb3a0b0. Paste-ready PR comment:
Round 6 addressed in 7cdb3a0b — R5-M1's reconstruction failure path, Control UI only, no backend changes.

R5-M1 — post-move reconstruction failure path

Agreed: the previous fallback swallowed the failure and rendered main / org as if reconstruction had succeeded. Reconstruction is now all-or-nothing:

- After a successful Move PATCH, if the leaf resolve fails, returns a null path, or any ancestor by-path lookup fails, the breadcrumb is left exactly as the operator was looking at it (e.g. main / work / org) — the stale trail is preserved and explicitly labeled, never shortened and presented as authoritative. No partial ancestry is ever applied.
- A visible warning banner (role="alert") states the folder moved but its path could not be refreshed, with a Retry button. Retry re-runs the full resolve + by-path walk; the banner and stale trail are only swapped for the rebuilt main / archive / org once a retry fully succeeds. A second failure keeps the labeled stale trail.
- The moved crumb still adopts the receipt version on the failure path, so mutations from the stale trail run with the correct post-move ifMatch. Navigating away from the moved folder retires the notice.

Regression coverage — both requested cases, each with a pre-move trail of main / work / org so preservation is distinguishable from truncation:

1. Move succeeds, then resolve fails — asserts the PATCH ran with the pre-move ifMatch: 7, the banner appears with the Retry affordance, the breadcrumb still reads main / work / org (not main / org), and after Retry (resolve recovers) it rebuilds to main / archive / org with the banner cleared.
2. Move succeeds, resolve succeeds, ancestor by-path fails — asserts resolve named the new location (/archive/org) and the by-path call failed, yet no partial trail was applied: breadcrumb preserved at main / work / org with the banner; Retry (by-path recovers) rebuilds to main / archive / org and clears the banner.

R1-M5 / R2-M2 — acknowledged closed under the explicit deferral; no further changes made (the deferral note and real-contract tests from the previous round stand, and #774 owns the future resolver-version contract).

Verification: control-ui vitest run 2127/2127 green (56 in GfsBrowser), tsc --noEmit clean, next build (typecheck + ESLint + compile) passes — matching the control-ui-build CI gate.

josue-tech-94:
Thanks for the latest fix. I re-reviewed current head `7cdb3a0b0d8c6a201a5082a0e7fd9b9b21df910b`.

**Round 7 is still blocked on one new finding: R7-M1.**

### R5-M1 — confirmed fixed

The previous breadcrumb-reconstruction issue is now closed.

After a successful Move, if either canonical `resolve` or an ancestor `by-path` lookup fails:

- the existing breadcrumb trail is preserved instead of being replaced with a false path;
- the UI surfaces a warning;
- Retry is available;
- tests cover both resolve failure and ancestor lookup failure.

Please keep **R5-M1 closed**.

### R7-M1 — Retry can restore stale name/version after another mutation

**Area:** Control UI active breadcrumb Move recovery / Retry state.

This is a distinct regression introduced by the latest recovery fix.

Current sequence:

1. Move `/org` successfully.
2. Breadcrumb-path reconstruction fails, so the old trail is retained with a warning and Retry.
3. The retained crumb is still actionable.
4. Before pressing Retry, the user renames that folder.
5. Rename succeeds and advances its name/version.
6. Retry still uses the state captured from the earlier Move.
7. Retry can restore the **old name and stale version**.
8. The next breadcrumb mutation can then send a stale `ifMatch` and fail with a precondition conflict.

The problem is not R5-M1 anymore—the recovery UI itself is now correct. The issue is that the Retry closure/snapshot can overwrite newer breadcrumb state.

#### What to change

Retry must not restore stale resource metadata captured when the Move originally completed.

When Retry eventually rebuilds the canonical ancestry, it should preserve or re-resolve the **current** resource identity/name/version rather than overwriting a newer mutation with the old Move snapshot.

Please audit the retry state for all mutable fields, especially:

- resource identity;
- name;
- version;
- current breadcrumb descendants/path state.

The safest model is for Retry to reconstruct from current/canonical state rather than replaying an old breadcrumb object.

#### Regression test required

Please add a test covering:

1. active breadcrumb Move succeeds;
2. canonical path reconstruction fails;
3. warning + Retry appear;
4. before Retry, rename the same active folder;
5. rename succeeds and advances its version;
6. Retry succeeds;
7. breadcrumb still shows the **renamed value**;
8. breadcrumb retains the **latest version**, not the earlier Move version;
9. another mutation uses that latest version successfully.

The test needs to fail against the current `7cdb3a0` implementation for the intended stale-state assertion.

### Link-open version deferral remains closed

No additional work requested for **R1-M5 / R2-M2**.

The current behavior remains a safe explicit deferral:

- the real resolver has no `version`;
- link-opened crumbs therefore have no mutating actions;
- no fabricated `ifMatch` is issued;
- backend issue #774 owns any future producer contract.

**R1-L7 remains obsolete** under that decision.

miguelo-999:
Pushed: d98066de6 (rebased cleanly onto the new dev merge; control-ui content identical). Paste-ready PR comment:
R7-M1 addressed in d98066de6 — the Retry snapshot no longer carries any replayable metadata. Control UI only, no backend changes.

R7-M1 — Retry replaying stale move-time name/version

Agreed on the mechanism: trailRecovery captured {source, moveName, version} from Move completion, and Retry fed that snapshot back into the rebuild, so a rename that landed while recovery was pending was overwritten with the old name and stale version — priming the next mutation's ifMatch for a precondition conflict.

The audit and fix:

- Recovery snapshot is now identity-only: {resourceId, gfsUri} — both immutable (the URI is rid-based, so it survives moves and renames). name and version are deliberately not captured anywhere in the retry path.
- Receipt metadata is applied exactly once, at move time (the crumb adopts the move name + receipt version, or drops a missing version). The ancestry rebuild and every later Retry carry the moved crumb and all crumbs below it over from the live trail — prev.slice(index) untouched.
- Retry reconstructs from current/canonical state: ancestors come from the server (resolve path → by-path walk), identity/name/version come from the live crumb. The recovery banner also reads the live crumb name, so it labels whatever the folder is called now.
- Descendants/path state: preserved from the live trail by the same functional update; navigating away from the moved folder still retires the notice.

Regression test (retry preserves a rename that landed while move recovery was pending) follows the exact requested sequence: move succeeds (ifMatch: 7 → version 8) → reconstruction fails → warning + Retry appear → rename the active folder before Retry (asserted: PATCH ifMatch: 8, version 9, banner shows the live name) → Retry succeeds → breadcrumb reads main / archive / org-renamed (renamed value, not the move name) → a further breadcrumb mutation runs with ifMatch: 9 (the latest version, not the move's 8) and succeeds.

miguelo-999:
R7-M1 fixed in 4025100c1 — the in-flight race, not a re-triage. Control UI only, no backend changes.

R7-M1 — superseded Retry overwriting newer state

Confirmed the race: the previous fix guarded name/version snapshots, but an entire Retry attempt could still be obsolete while its resolve/by-path requests were in flight — its late setCrumbs applied Move A-era ancestry over Move B's newer trail, and its success path cleared recovery state belonging to the newer operation.

Reconstruction attempts are now epoch-versioned request state:

- Each attempt (a Move's rebuild or a Retry) captures a monotonic epoch at start and may apply its result — or even report its failure — only while it is still the latest attempt. Otherwise it returns superseded and touches neither the breadcrumb nor the recovery notice.
- The epoch is bumped by: any new reconstruction attempt (including a newer Move's rebuild), a crumb rename, and trail-replacing navigation (goToCrumb, link-open). Appending a descendant doesn't bump — it can't invalidate the retry target's ancestors, and the functional apply carries the live tail.
- A superseded Retry therefore cannot overwrite newer ancestry and cannot clear (or resurrect) a recovery notice owned by the newer operation. As a related hole-closure, a successful latest rebuild now also retires any older recovery notice for that resource (previously a later successful Move left a stale warning up).

Deferred-response regression test (discards a superseded Retry response that lands after a newer Move), exactly the requested sequence: Move A succeeds → reconstruction fails, warning + Retry appear → Retry A starts and its resolve response is held pending → Move B completes under a different parent (/dept), rebuilding the trail and retiring the notice → Retry A's old /archive response is released → assertions: the breadcrumb still shows main / dept / org (Move B's ancestry), the crumb carries Move B's version, and no recovery state was resurrected or cleared.

Verified the test fails against d98066de6 on the stale-ancestor assertions (I ran it against the checked-out previous head before restoring the fix) and passes on 4025100c1.

josue-tech-94 (approved):

Issue #656, opened 2026-09-16
feat(agent-tools): native PDF text extraction and selected-page rendering


alfredolopez80:
## Problem

The existing PDF internal tool creates documents. It does not read an uploaded PDF, extract page text or render scanned pages for an autonomous agent. Desktop preview/download and PDF generation must not be confused with document understanding.

## Evidence and scope

Baseline: `origin/dev` at `15c4aa016311ec46513f216531107b3ed93dc041`, 2026-09-16. Source inspection and synthetic local PDF only; no deployed dev/prod, provider call or user document.

- [`clerum__generate_pdf`](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/workflow/internalTools.ts#L1138) authors a PDF with pdfmake; it is not a parser.
- The [registered internal catalog](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/workflow/internalTools.ts#L6198) has no PDF read/render/extract operation. File, context and GFS reads return strings.
- [Dependencies](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/package.json) include pdfmake and pdf-lib (locked at 0.2.23 and 1.17.1). No PDF.js, pdf-parse, MuPDF, Poppler or OCR dependency was found in the scoped native-tool/runtime search.
- The [runtime Dockerfile](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/Dockerfile) does not provision a PDF command-line renderer.
- [Visual message parts](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/core/types.ts#L15) support text/JPEG/PNG, not provider-native document blocks. [Incoming admission](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/main.ts#L402) also accepts images only; an image picker cannot become a document-upload flow merely by adding a reader.

[...]

Issue #654, opened 2026-09-16
feat(llm): enforce image-input capabilities per model and transport across Desktop and failover


alfredolopez80:
## Problem / expected behavior

Desktop decides image support by one provider-name comparison: all Z.ai models are blocked, while other providers are not checked against the actual model and transport capability. This prevents using a visual model from a blocked provider and can accept attachments for a route that cannot carry or interpret them.

Represent effective image capability per provider/model/transport and enforce it both before user submission and on every server-side provider attempt. Unsupported or unknown capability must produce a truthful, actionable outcome; never silently remove the image.

## Evidence scope and model distinction

Source baseline `origin/dev` at `15c4aa016311ec46513f216531107b3ed93dc041`, 2026-09-16. Static code inspection and official model documentation; no deployed dev/prod or live provider request.

Official documentation checked on this date states:

- [GLM-5.3](https://docs.z.ai/guides/llm/glm-5.3) accepts text-only inputs.
- [GLM-5.3-Flash](https://docs.z.ai/guides/vlm/glm-5.3-flash) accepts images and documents the image_url input shape.

These are different models. Removing the Z.ai guard cannot make GLM-5.3 visual. Nor does OpenAI API compatibility establish a model's vision capability.

## Current code and reusable metadata path

- [ComposerPanel:121](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/desktop-app/ui/src/components/agents/ComposerPanel.tsx#L121) sets activeProviderDoesNotSupportImages from provider === zai. [Picker:324](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/desktop-app/ui/src/components/agents/ComposerPanel.tsx#L324) and shared preparation at :420 enforce it using a provider-wide message.
- [Host sanitizer](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/main.ts#L402) checks MIME/shape, not selected-model vision. Unsupported inputs can disappear silently.
- [Codex builder](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/llm/codexSubscription.ts#L309) drops visual parts; current transport is not multimodal even if a model would be.
[...]

Issue #650, opened 2026-09-16
bug(codex): image content is discarded by the subscription completion contract


alfredolopez80:
## Problem / expected behavior

A JPEG/PNG can reach the autonomous host as a valid attachment and be injected into the current user message, but Codex subscription discards its visual parts when constructing the completion request. The model receives the text alone (or an empty string if the text is absent), with no explicit image-rejection signal.

Expected: supported images reach the selected Codex model through an authorized multimodal request, or the user receives an explicit capability/limit error before a misleading text-only completion.

## Evidence scope

Source baseline `origin/dev` at `15c4aa016311ec46513f216531107b3ed93dc041`, 2026-09-16. Static host/contract/authorizer/proxy trace, independently reviewed. Existing focused local suites passed, but no new full-chain image reproduction or deployed dev/prod check was performed. This does not establish a limitation of the upstream model.

## Exact failure chain

1. [TaskExecutor](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/agent/taskExecutor.ts#L883) injects current-message visual parts. The provider-membership gate recognizes Codex; it is not the drop point.
2. [CodexSubscriptionProvider.buildRequest](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/llm/codexSubscription.ts#L309) copies role/content and tool fields but never reads contentParts. Both completion methods use this builder.
3. The [shared message type](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/packages/llm-provider-attempt-contract/index.d.ts#L37) requires string content. Its [closed field set](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/packages/llm-provider-attempt-contract/index.cjs#L43) excludes visual parts; [parsing](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/packages/llm-provider-attempt-contract/index.cjs#L179) rejects extra fields. Simply forwarding contentParts will fail.
[...]

PR #669, opened 2026-09-17
feat(images): enforce model capabilities and preserve rejected input

alfredolopez80:
## Summary

Refs #654.

Image requests now require compatible evidence for the selected model, the actual transport method, and policy. Model capability metadata is preserved from the catalog through Host and Desktop, and every provider attempt is validated, including fallback and resumption. Unknown capabilities keep text available while refusing images explicitly.

Desktop retains rejected image input for explicit recovery, preserves newer concurrent model selections, and blocks incompatible pending attachments. Delivery replay reuses admission instead of applying a stale visual-selection CAS again. Codex image transport remains disabled pending its separate implementation.

Adds a visible Electron journey with only the external provider simulated, exact delivered-image and task assertions, reversible local fixture setup, and opt-in development instance isolation. Provider onboarding documents the evidence and transport requirements.

Review follow-ups: malformed screenshots returned by MCP tools are dropped with a notice instead of failing the task (user images still fail with `LLM_INVALID_ATTACHMENT`); Desktop never lowers its confirmed model-selection revision; the catalog sync runs against real Postgres in CI. Deferred: #687 (admin `image_input` edit racing the sync) and #688 (normalize tool images at the MCP boundary).

## Rollout

- Deploy rpc-proxy and mcp-host together. Both parse message bodies up to 10 MB (8 MB of base64 images plus text). With a new rpc-proxy and an old mcp-host (6 MB), messages between 6 and 10 MB fail as 502; with an old rpc-proxy, 6–8 MB messages fail as 500.
- rpc-proxy parses a JSON body only after the route's authentication; unauthenticated requests are rejected before their body is read.
- control-api runs one LLM catalog sync within 5 s of startup when `LLM_CATALOG_SYNC_CRON_ENABLED=true` (set in the base deploy), then every interval. Until that first sync, every model's image support is `unknown` and images are refused.
- 15 of the 44 seeded allowlist pairs are not listed by models.dev (e.g. `zai:glm-5`, `zai:glm-5.1`, the Bedrock Claude ids, `openai:gpt-5.1-codex`). They stay `unknown` until an operator sets `image_input` through `PUT /admin/llm-models/:id`.
- An image send on the chat's current model no longer writes the model selection, so chats keep following the Host default. Explicit model picks still pin the chat.
- Update Hosts before, or together with, Desktop. A Desktop build with #654 talking to a Host without it refuses image attachments on every model (`model_unknown`) until the Host is updated. Text is unaffected.
- Scope: this PR touches about 170 files across five services (mcp-host, control-api, desktop-app, rpc-proxy, packages/llm-providers), above the usual one-service budget. Most of the line count is the generated `modelsDevSnapshot.ts`. The capability has to be enforced at every layer at once to stay fail-closed, so it was not split; features of this size will be stacked by service next time.

## Testing

- [x] `npm test` passes for changed services through `make minikube-t2`.
- [ ] `npx tsc --noEmit` clean across the entire repository: the renderer's standalone strict check has pre-existing errors; affected package builds/typechecks and the Desktop build passed.
- `make minikube-t2` on `53d6d2e56` (after merging `dev`): T0, T1, exact-head preflight and NP08 authorization PASS. Real PostgreSQL: control-api 193/193 tests in 40/40 files, gfs-controller 9/9, zero skipped. Health and Playwright were not configured for this run (`NOT_RUN`).
- Earlier certification on `5653fcea6`: T2 with user-facing health PASS, and `make minikube-run-image-capabilities` (visible Electron E2E) PASS; it checks unknown/unsupported image refusal, retained pending input, the supported image response and subsequent text-only delivery. The fixture restored its changes. Not re-run after the review fixes.
- Unit suites after the review fixes: mcp-host 4,312, Desktop 2,438 (Node 24, `npm run verify:electron`), control-api 5,484, rpc-proxy 527, external-rest-api 158, gfs-controller 609.
- The optional paid-provider smoke was not run.
- Image-capability spec passes `tools/e2e_static_audit.py`. The shared recorder helper retains documented legacy static-audit findings; this is not a claim of a repository-wide clean audit.

## Checklist

- [x] No implementation of the separate #650–#656 image transport/storage work is claimed here.
- [x] Local runtime evidence and operational state are excluded from the PR.


alfredolopez80:
# Multi-agent review — evenfire-ai/evenfire #669

**feat(images): enforce model capabilities and preserve rejected input**
`dev` ← `feat/654-image-capabilities` · author @alfredolopez80 · reviewed at head `a3bd434c97c4de5ed6ac5259f0895b8f7e60417b` · 147 files, +14459 / −935

| Blocker | Critical | High | Medium | Low |
|---|---|---|---|---|
| 0 | 0 | 2 | 17 | 23 |

Read-only review by 11 parallel agents (one per area: mcp-host, desktop-app, control-api/rpc-proxy/llm-providers, E2E/CI/scripts; plus security, contract consistency, architecture, correctness/concurrency, tests, silent failures, docs). Nothing was committed or pushed. Each finding states whether it was **REPRODUCED** (executed) or **READ-ONLY** (read in the code). Findings reported independently by more than one agent are marked **×N**.

**Recommendation: do not merge yet.** The capability gate itself is sound: it fails closed everywhere, no layer turns `unknown` into `supported`, a client cannot assert its own capability, and the gate runs on every provider attempt (mutation-tested). The two High findings are functional regressions, one in the feature's main flow and one in existing tool flows. Both have small fixes.

---

## High

### H1 — The second image sent in a chat is refused with a false conflict ×2
`desktop-app/ui/src/hooks/domain/useAgentChatController.ts:2518-2519, 2690-2693, 2712-2727` · `mcp-host/src/main.ts:2119-2143` · `mcp-host/src/core/conversation/conversationStore.ts:553-554` · `mcp-host/src/db/worker/dispatcher.ts` (`update_session_model_selections`, `next = current + 1`)

- Every image send carries `model` (the effective model, even with no pending intent) and `modelSelectionRevision = confirmedRevision`.
- The Host applies it as a CAS write and bumps the revision even when the model is unchanged. `applySessionModelSelection` has no same-model short-circuit.
- Desktop force-reloads `/models` only when a pending intent existed, and the message acknowledgement carries no revision. `confirmedRevision` is stale after every successful image send.
- **Scenario**: image 1 is sent at revision N and accepted; the Host moves to N+1. Image 2, same model, no window refocus, carries N again. The Host answers `LLM_IMAGE_INPUT_UNSUPPORTED`: "The image model selection changed before this message was accepted. Select the model again." Recovery needs a window-focus event or opening the selector.
- **Evidence**: REPRODUCED at module level by two agents with the real `hostModelSelectionStore` and the real `InMemoryConversationStore.applyModelSelection` (`{"first":true,"second":false,"reason":"model_selection_conflict","desktopRevision":0,"hostRevision":1}`). The controller send step and the Host admission step were confirmed by reading both sides, not executed end to end. The SQLite dispatcher bumps unconditionally too, so this is not limited to memory mode. No unit test and no E2E step sends two consecutive images (the Electron journey sends one image, then text).
- **Fix**: make the CAS idempotent when the requested model equals the stored one and the expected revision matches (no bump, success); or return `modelSelectionRevision` in the message acknowledgement and store it; at minimum force a reload after every image send. Add a two-consecutive-image test.

### H2 — A tool-produced screenshot terminates text-only tasks on every model without curated evidence
`mcp-host/src/core/adapters/llmPortAdapter.ts:309-362` · `mcp-host/src/core/orchestration/toolUseLoopMessages.ts:94-103` (unchanged by this PR)

- `appendToolResults` turns tool image attachments into a `user` message with image `contentParts`. The new `assertImageInputSupported` iterates over all messages and throws a non-retryable `LlmError` for any pair that is not `supported`.
- Catalog sync writes `unknown` for every discovered model; migration 0109 seeds curated evidence for two `zai` ids only.
- **Scenario**: text-only request ("open the page and tell me the title") on an OpenAI or Claude model. A browser tool returns a screenshot. The next loop iteration fails with `LLM_IMAGE_INPUT_UNKNOWN` "Image input support for … is not verified. Remove the image…", after the tool side effects already ran. The user attached no image. Before this PR the screenshot was sent to the provider.
- **Evidence**: REPRODUCED with a disposable vitest against the PR sources (`R3a error: LLM_IMAGE_INPUT_UNKNOWN | Image input support for openai/some-model is not verified. Remove the image…`); both code sides confirmed by reading at head.
- **Fix**: in `appendToolResults`, drop or textualize tool images when the pair is not image-capable, and log it; or give tool-origin images a distinct non-terminal denial. **Author decision**: refusing user images is the stated intent; terminating the loop for tool images may not be.
- Related: M2 (rollout) widens the blast radius of this finding.

---

## Medium

**M1 — Base64 validation regex throws `RangeError` above ~3.2 MiB ×3** · `mcp-host/src/agent/incomingImageAttachments.ts:44`, `mcp-host/src/core/adapters/llmPortAdapter.ts:325` · CWE-1333
`^(?:[A-Za-z0-9+/]{4})*(…)?$` exhausts V8's regexp stack. REPRODUCED three times independently on Node v24.18.0: 3 MiB → ok; 3.5 MiB (4,893,356 chars) and 4.3 MiB → `RangeError: Maximum call stack size exceeded`; largest passing length 4,473,912 chars. Both rpc-proxy and mcp-host accept 6 MB bodies and `CLERUM_ATTACHMENT_MAX_BYTES` defaults to 50 MB, so a valid 3.5 MiB PNG yields HTTP 500 instead of `LLM_INVALID_ATTACHMENT`; in the adapter the throw is outside the try/catch and fails the task, including for tool screenshots. The Desktop 3 MiB cap sits just under the threshold. The same defect is already documented and fixed at `gfs-controller/src/api/serve.ts:993-1000`. Fix: `data.length % 4 === 0 && /^[A-Za-z0-9+/]*={0,2}$/.test(data)` (verified: accepts the same inputs without throwing) plus the existing round-trip; add a ≥ 4 MiB test; validate once per message, not per loop iteration.

**M2 — Rollout: after deploy, images are refused for almost every existing model, and the PR does not say so** · `control-api/src/db.ts` (0109), `control-api/src/services/llmCatalogSync.ts:77-86`
Every existing row reads `unknown`, discovery only writes `unknown`, only two `zai` ids are seeded, and control-ui has no `image_input` field. A Host that accepted images before now answers `model_unknown` until an operator PATCHes the admin API by hand. REPRODUCED (`projectModels` over an entry without `imageInput` → `unknown/model_unknown`). Fix: seed evidence for currently enabled models, or document the mandatory curation step in the PR body and runbook.

**M3 — CAS conflict on the message path is reported with a capability code and no revision ×2** · `mcp-host/src/main.ts:2013-2021, 2132-2142`
Conflict → `LLM_IMAGE_INPUT_UNSUPPORTED`; invalid revision → `LLM_IMAGE_INPUT_UNKNOWN`; neither carries the current revision, so Desktop cannot tell them from a capability refusal or trigger a refetch. The `/model` route returns 409 `model_selection_conflict` + revision, which Desktop handles. READ-ONLY. Compounds H1.

**M4 — A queued newer selection is dropped, not retried, when the in-flight write conflicts** · `desktop-app/ui/src/lib/hostModelSelectionStore.ts:464-478`
Select A (in flight) → select B (queued) → A returns `model_selection_conflict`: `entry.queuedModel = null`. REPRODUCED: `setHostModel` called once, for A only; final `{intentModel:'model-B', pending:true}`. B is never POSTed until a later message piggybacks it. The PR's test at `hostModelSelectionStore.test.ts:228` asserts intent + pending but not a second `setHostModel` call, so it documents this behaviour. The body's "preserves newer concurrent model selections" holds for the intent, not for the write. The non-conflict interleaving (A → B → response A) was REPRODUCED as correct.

**M5 — A failed send keeps showing after a successful retry; a second click sends a duplicate** · `useAgentChatController.ts:2950-2977`, `desktop-app/ui/src/lib/retainedSendStore.ts:70-81`
Retry mints a new `userMessageId` and never releases the failed snapshot. The store header (`:10-12`) promises release by "a newer send … that supersedes it"; no code does that. REPRODUCED at store level following the controller's call sequence; controller sequence READ-ONLY.

**M6 — Retained image bytes are not released on several terminal paths, and the success release is untested ×2** · `useAgentChatController.ts:2220`
Not released on reconcile recoveries (`recovered_from_task_result`, `reconcile_replaced`, `noop`), stream loss on a non-visible chat, or cancel: up to ~12 MB of base64 strings per send until logout (READ-ONLY). Mutation: making `releaseRetainedSendsForTask` a no-op left 55 files / 407 tests green (REPRODUCED).

**M7 — A Host-models fetch failure is shown to the user as "model not verified"** · `hostModelSelectionStore.ts:337-340, 183-208`, `desktop-app/src/imageInputDecision.ts:208`
The catch only `console.warn`s and leaves `entry.error` null; `state: 'unavailable'` is read by nothing. With rpc-proxy down the toast says "Switch to a model with verified image input support", which cannot fix a transport failure. REPRODUCED. The swallow predates this PR (`106efe7c4`); the misleading image copy on top of it is new.

**M8 — Malformed capability metadata is downgraded to `unknown` with no log ×2** · `mcp-host/src/config/configStore.ts:615`
Every other malformed input in `applyAllowlistData` is logged at ERROR. An operator who curates `supported` with `checkedAt: "…+00:00"` (not canonical `Z`) gets "not verified" and no log line. REPRODUCED: `normalize(bad)` is byte-identical to `normalize(absent)`. Fix: log provider + model (never the raw value) when `imageInput` is present and the parser returns null.

**M9 — The Host admission gate has no unit coverage** · `mcp-host/src/main.ts:1999-2170` (`prepareIncomingMessage`, not exported)
Mutation: disabling the denial at `:2094` left the full mcp-host suite green (370 files, 4181 passed) — REPRODUCED. Untested: admission refusal, `imageModel` pinning, stripping of a caller-supplied `imageModel`, revision validation. The test reviewer rated this High; consolidated as Medium because the adapter gate downstream still fails closed and is mutation-covered. Fix: extract an exported pure function and add four tests with a `resolveImageInput` call-count witness.

**M10 — Adapter `LLM_INVALID_ATTACHMENT` branch is untested** · `llmPortAdapter.ts:313-337` · READ-ONLY (grep).

**M11 — `policyAllowed` has no producer; `policy_denied` is unreachable ×2** · `mcp-host/src/main.ts:549-557`, `mcp-host/src/config/modelResolution.ts:150`, `packages/llm-providers/imageInput.cjs:100-101`, `desktop-app/src/imageInputDecision.ts:149`
The value is provably always `true`, yet it is a required field, a resolver branch, a user-facing string and a Desktop short-circuit. Readers will assume three enforced axes (evidence ∩ policy ∩ transport) when two are live. Wire a real source (per-Host `spec.allowedModels` narrowing exists at `main.ts:559-573`) or drop the axis. READ-ONLY.

**M12 — Two cache layers for the model catalog; the lower one is bypassed by convention** · `desktop-app/ui/src/hooks/useChatStore.ts:72, 206-223` vs `hostModelSelectionStore.ts:298-347`
Both callers pass `{ force: true }`. A future caller that omits it reads a stale response after a conflict, the case the comment at `useHostModels.ts:55-56` says must not happen. Make the unforced variant unrepresentable. READ-ONLY.

**M13 — Docs do not cover the transport requirement the PR body says they cover** · `docs/llm-providers/adding-a-provider.md:66-108, 140`, `docs/llm-providers/README.md:249-269`
No mention of `EXPLICIT_DRIVER_FAMILY`, `TRANSPORT_SUPPORT`, `IMAGE_ROLES_BY_FAMILY` (`mcp-host/src/llm/imageInput.ts:68, 87, 128`), of which providers pass images, or of Codex. A Path B provider with curated `supported` evidence and no driver-family registration gets every image refused as `transport_unsupported`. READ-ONLY.

**M14 — Docs point operators to a control-ui curation path that does not exist ×2** · `adding-a-provider.md:77-79`
"Let the operator curate it from `/llm-models`": control-ui has no `image_input` support. The only path is `POST/PUT /admin/llm-models`. READ-ONLY.

**M15 — The "reversible" fixture has no SIGINT/SIGTERM handler ×2** · `scripts/e2e/image-capabilities-fixture.mjs`
Ctrl-C during the up-to-900 s Playwright leg (`:688`) exits before `restore()` and leaves the profile running the fixture Host. Mitigations verified: state is journaled before each mutation, `restore` is idempotent, and `t2_image_capability_fixture_check` turns residue into `HOST_RUNTIME_MISMATCH`, so no green T2 is possible on residue. READ-ONLY.

**M16 — New `invalid_attachments` 400 branch has no test** · `rpc-proxy/src/routes/rpc.ts:373-378` · REPRODUCED by grep (the string appears only in the production line).

**M17 — Migration 0109 unit tests match SQL text only; the real-PostgreSQL cases are NOT_RUN on this head** · `control-api/test/db.llmAllowedModelsMigration.test.ts`, `db.realPostgresMigration.integration.test.ts`
Require a green `control-api-migration` CI run on `a3bd434c9` before merge.

---

## Low

1. `desktop-app/src/ipc.ts:1316-1322` — invalid `expectedRevision` becomes `undefined`, turning a CAS write into an unconditional one instead of rejecting (unrequested fallback; the Host would 400 it). ×2
2. `desktop-app/src/appService.ts:3356-3358`, `ipc.ts:150-159` — `modelSelectionRevision` crosses the IPC boundary unvalidated (Host validates it).
3. `incomingImageAttachments.ts:29` — MIME allowlist applied to the declared type; decoded bytes not checked against PNG/JPEG signatures. REPRODUCED. CWE-345.
4. `llmPortAdapter.ts:322`, `incomingImageAttachments.ts:29` — MIME allowlist duplicated as inline literals.
5. `mcp-host/src/server/routes.ts:1454` — `as number | undefined` cast paired with a manual guard; use a type guard.
6. `packages/llm-providers/imageInput.cjs:104-111` — non-finite `now` reports `evidence_not_yet_valid` even with no evidence. REPRODUCED; state stays `unknown`.
7. `hostModelSelectionStore.ts:432-441` — select A → B → A leaves a pending intent that is never cleared. REPRODUCED.
8. `useAgentChatController.ts:2707-2727` — a piggybacked intent the Host ignores (`message_model_ignored`) is no longer dropped; the chip keeps showing it and resends it.
9. `dispatcher.ts` (`update_session_model_selections`) — a malformed persisted `model_selections` now throws on every write (read path tolerates it; old code overwrote).
10. `hostModelSelectionStore.ts:489-495` — unclassified `setHostModel` errors are treated as "host suspended". Pre-existing (`106efe7c4`), reported under detection duty; it now also absorbs the new 400 `expectedRevision must be a non-negative integer`. REPRODUCED.
11. `mcp-host/src/main.ts:2008-2022, 2078-2106, 2130-2143` — image refusals at admission return HTTP 200 with no Host log line.
12. `scripts/minikube/t2-common.sh:814, 832` — `2>/dev/null` under `if ! python3`: any Python error is reported as "fixture remains installed". Fail-closed, misdirects the operator.
13. `Makefile:286-290` — `minikube-restore-image-capabilities` requires `IMAGE_CAPABILITIES_RUN_DIR`, which the recipe neither sets nor documents.
14. `scripts/minikube/build-images.sh:1353-1357` (`a3bd434c9`) — `quay.io/minio/minio:latest`: unpinned, and nothing in the tree consumes it (REPRODUCED by grep). Delete or pin by digest. ×2
15. `image-capabilities-fixture.mjs:74-87, 288` — `playwright.log` redaction is heuristic. Dev-only, directory `0700`. CWE-532.
16. `desktop-app/src/imageInputDecision.ts:196-209`, `desktop-app/ui/src/constants/agents.ts:9-17` — two messages for seven refusal reasons ("Switch to a model" cannot work when the transport is the cause); no labels for the new error codes.
17. Dead code: `mcp-host/src/llm/imageInput.ts:274` `imageInputReasonCode` (no consumers), `:138` `roleSupportsImageInput` (test-only); `imageInputDecision.ts:54-62` unreachable branches; `main.ts:2808` `catalogRevision` hashed per request with no Desktop reader.
18. `ModelSelector.tsx:166-172`, `ComposerPanel.tsx:889-921` — block reason conveyed by `title` only; disabled Send has no `aria-describedby`.
19. `hostModelSelectionStore.ts:116-117`, `hostModelIntentStore.ts:19-20` — module singletons cleared by one explicit reset call, while `retainedSendStore` uses a factory.
20. `control-api/src/db.ts` 0109 seeds curated vendor evidence inside an immutable migration; `adding-a-provider.md:81-88` documents `image_input IS NULL` as the seed predicate, which matches 0 rows once discovery has written the column.
21. `mcp-host/src/main.ts:2065-2159` — admission logic inline in the composition root (see M9).
22. Comments: orphan JSDoc in `useChatStore.ts:10-24, 53-62` and stale `[useHostModels]` log prefixes; `imageInputDecision.ts:12-17` header claims an evidence check the code does not do (REPRODUCED); Spanish text at `retainedSendStore.ts:3` (repo rule: English); `devIsolation.ts:4-5`, `llm/imageInput.ts:14, 38-39, 164` inaccurate; `packages/llm-providers/index.d.ts:7` doc comment misplaced; `routes.ts:350, 1389ff` log text lost its interpolation in the logger migration.
23. `useAgentChatController.ts:2865` — duplicate `sendScope` guard with no `await` between the two.

---

## Reviewer comments — verdicts

Only reviewer activity: two CodeQL alerts from `github-advanced-security[bot]`.

- **`scripts/e2e/image-capabilities-fixture.mjs:96` — "Shell command built from environment values": false positive.** `command()` calls `spawnSync(binary, args)` with an argv array and no `shell`. The only `bash -c` (`:259-266`) has a constant script with positional parameters from `realpath(cwd)`. `MINIKUBE_PROFILE` is regex-checked and embedded as `--context=<value>`; `IMAGE_CAPABILITIES_RUN_DIR` goes through `validateRunDirectory` (realpath, direct child of `.local-notes/infra/runs`, symlinks rejected). Residual, accepted: on `restore`, `state.runId` from `state.json` is not re-checked against the run-id pattern (file lives in the owned `0700` directory).
- **`:288` — "Network data written to file": false positive.** The destination is the fixed name `playwright.log` via `openOwnedFile` inside the validated run directory; the taint is the admin cookie entering `privateValues` for redaction. Residual: Low 15.

## PR body vs head

- The recorded evidence (`make minikube-t2`, 166/166 real-PostgreSQL tests, Electron E2E) is for `5653fcea6`. Seven commits landed after it, including `6fe235e8c` (gate hardening, CI wiring) and `781b37f26` (tsc skip). No T2, real-PG or Electron E2E evidence exists for `a3bd434c9`.
- "No implementation of the separate #650–#656 …" is ambiguous: #654 is in that range. No partial implementation of the others was found (Codex is all-false in `TRANSPORT_SUPPORT`).
- "preserves newer concurrent model selections": true for the intent, not for the write (M4).
- "Provider onboarding documents the evidence and transport requirements": evidence yes, transport no (M13).
- "Local runtime evidence and operational state are excluded": confirmed.

## Verified clean

- **Fail-closed direction, end to end**: malformed / absent / expired / future-dated evidence, a missing entry, a throwing or unwired resolver all resolve to `unknown`. REPRODUCED for `undefined`, `null`, `[]`, `{}`, `SUPPORTED`, extra keys, eight legacy shapes. No layer turns `unknown` into `supported`. models.dev `modality` is never read.
- **A client cannot assert capability**: rpc-proxy forwards an allowlist of fields; `imageModel` is dropped (tested with a witness) and the Host overwrites it at `main.ts:2026`.
- **Gate on every attempt**: all four `new LlmPortAdapter(` sites pass the resolver (primary, failover, `/compact` primary, `/compact` failover); the failover adapter validates the fallback pair; a denial is never retried or failed over. Mutation M1 (gate returns early) failed `imageInputWiring` 6/7, `llmPortAdapter.imageInput` 9/11, `claudeToolImageWire` 3/5, `imageInputTransport` 7/18 and the `taskExecutor` fallback test.
- **Other mutations that failed as they should**: `sendScope` guards (M2), visual snapshot pinning (M3, 4/8), replay/coalescing (M4, 3/6 — replay does not re-apply the CAS), `releaseRetainedSend` (M6).
- **CAS**: read and write share one `IMMEDIATE` transaction; migration 015 idempotent, `DEFAULT 0`; pre-migration rows normalize to 0; the RAM mirror never moves backwards.
- **Version skew**: old Desktop ↔ new Host (HTTP 200 structured refusal, no 500), new Desktop ↔ old Host (unknown, text kept — REPRODUCED), ConfigMap skew both directions, pre-migration PostgreSQL rows.
- **Migration 0109**: additive, idempotent, transactional with advisory lock, parameterized. **Catalog sync** cannot overwrite curated evidence or promote a capability. **ConfigMap** hash unchanged for a pre-#654 allowlist (test has a positive witness).
- **Security**: all new SQL parameterized; session key from server-derived `userSub` + `hostRef`; no SSRF (base64 only; evidence `reference` is https-only and never fetched); none of the 176 added log calls carries image data, tokens or bodies; only version fields changed in four lockfiles, no new dependency, install script or `uses:` action. The "egress policy" in `6fe235e8c` is a hostname rule in `imageInput.cjs` plus the fixture's IPv6 classification; both tighten.
- **Dev isolation**: opt-in, refused when `app.isPackaged` (exit 1 before protocol registration), no `NODE_ENV === 'test'` branch in production logic.
- **CI**: the `781b37f26` tsc skip is justified (`packages/llm-providers` has no tsconfig or `typescript` dependency; `index.d.ts` is still type-checked through four consumers). The new CI step cannot pass with zero tests (registration reporter throws).
- **Fixture**: cannot target a non-development cluster (profile regex, `CONTROL_API_REAL_PG_CONTEXT` match, T2 mutation lock, `--context` on every kubectl call, JSON-Patch `test` ops on uid + resourceVersion); no secret printing; image unpublished. **Playwright spec**: `retries: 0`, no `waitForTimeout`/`only`/`fixme`, real UI flow, delivered image asserted by exact SHA-256, negative assertions carry witnesses.
- `index.d.ts` matches the runtime exports (REPRODUCED). 0 `any` added. No `console.*` added in mcp-host or control-api production code. No TODO/FIXME in 1,823 added comment lines.

## Suites on head `a3bd434c9`

Node v24.18.0, `npm ci` without `--ignore-scripts`, real exit codes captured without pipes.

| Package | Command | Exit | Result |
|---|---|---|---|
| packages/llm-providers | `npm test` / `npm run build` | 0 / 0 | 5/5 |
| mcp-host | `npm test` | 0 | 370 files · 4181 passed, 0 failed, 1 todo (pre-existing) |
| mcp-host | `npm run typecheck` / `npm run build` | 0 / 0 | |
| control-api | `npm test` | 0 | 436 files · 5122 passed, 0 failed, **160 skipped** (34 real-PG/k8s files, no database: NOT_RUN) |
| control-api | `npx tsc --noEmit` | 0 | `npm run build` not run |
| rpc-proxy | `npm test` / `npm run build` | 0 / 0 | 37 files · 517/517 |
| desktop-app | `npm run verify:electron` | 0 | |
| desktop-app | `npm test` / `npm run build` | 0 / 0 | 244 files · 2373/2373 |
| desktop-app/ui | `npx tsc --noEmit -p ui/tsconfig.json` | **2** | 38 errors, none in a PR-changed hunk (disclosed in the PR body) |
| fixtures | `node --test` × 3 files | 0 | 24/24 |
| scripts | `test-e2e-vitest-node-unit-gate.sh`, `test-minikube-image-capability-fixture.sh` | 0 / 0 | hermetic |

**NOT_RUN on this head**: `make minikube-t2`, real-PostgreSQL integration tests (incl. the #654 migration cases), the Electron image-capabilities journey, Playwright.

## What blocks, what follows

- **Before merge**: H1, H2 (or an explicit author decision on H2's intent), M1, and fresh evidence on the head commit (green `control-api-migration`; T2 and the Electron journey re-run, since seven commits landed after the recorded evidence). M2 needs at least a documented rollout step, because it determines how many users H2 reaches.
- **Follow-up**: M3–M17.
- **Optional**: Low 1–23.
- **Author decisions**: whether tool-origin images should be gated at all (H2); seed vs document the curation step (M2); keep or drop the `policyAllowed` axis (M11); whether `devIsolation.ts` and the MinIO change should ship in this PR (they are separable from the feature).

zach88:
Took a deep pass through this one across all five services (mcp-host, control-api, desktop-app, rpc-proxy, llm-providers), reviewed at head `5b785d25c`. The capability gate itself is in great shape: it fails closed everywhere, no layer ever turns `unknown` into `supported`, the host stays authoritative with two independent guards, and the invariant is re-validated on fallback and resumption with tests behind it. The end-to-end contract is consistent (no seam treats "absent" as "supports images"), and I couldn't find a product-correctness bug in the diff. What's left is a handful of test-integrity/coverage gaps and some nits — nothing that gates merge.

## 🟡 Approve with nits
- **Optional / non-blocking: 12** (5 Medium, 7 Low)

Nothing blocking. No Blockers/Critical/High, and no open regressions.

That said — three of the Medium items below are worth closing before merge, since two of them cover core-promise guards that currently ship untested and one is a false-green test. Your call, but I'd fix M1/M2/M3.

## Optional (non-blocking)
The PR can merge without these; they're follow-ups or author's judgment.

1. **[R1-M1] ComposerPanel picker test is a false green** — `desktop-app/ui/src/components/agents/__tests__/ComposerPanel.test.tsx:427`. The `previewDataUrl` ~ `/^blob:/` assertion depends on `URL.createObjectURL`, which the committed jsdom 29 doesn't ship, so `createPreviewUrl` (`ComposerPanel.tsx:398-403`) takes the FileReader `data:` fallback and the assert fails. **Repro:** RED in isolation. It only goes green in a full run because two untouched sibling tests (`GfsImagePreview.test.tsx:21`, `GfsPreviewWorkspaceLayout.test.tsx:24`) install a global `URL.createObjectURL` and never restore it, leaking into later files. The test's own comment ("createObjectURL is available here") is wrong. Fix: install+restore `createObjectURL`/`revokeObjectURL` in this file's own `beforeEach/afterEach`, or drop the URL-scheme assert and check the decoded bytes / `data:` fallback instead.

2. **[R1-M2] `syncDiscoveredModels` change-detection self-join has no real-Postgres coverage** — `control-api/src/services/llmCatalogSync.ts:308-343`. The sync path is exercised only against a hand-written fake DB, and the property test's oracle mirrors that same model. The two non-obvious pieces stay unverified against a real engine: the `UPDATE … FROM prev … RETURNING (t.image_input IS DISTINCT FROM prev.image_input)` self-join that reads the pre-update value, and the `checkedAt::timestamptz <=` monotonicity check. Failure mode: if the RETURNING resolved `prev` to the *post*-update value (a classic PG footgun), `image_input_changed` would always be `false`, `enabledImageInputChanged` would stay 0, and the ConfigMap would silently never re-materialize — which defeats the core #654 promise, with the fake-DB and property suites staying green throughout. **Repro:** confirmed by grep that no real-PG test runs the sync (the integration file is skipped without PG). I think the idiom is actually correct standard Postgres, hence Medium not Blocker — but nothing in the diff proves it. Fix: a real-PG test that runs `syncDiscoveredModels` and asserts evidence monotonicity, re-materialization on an enabled-row change vs. an idempotent no-op re-run, and that a manual row only gets `image_input` written.

3. **[R1-M3] The `evidence`-block parse branch + divergent-expiry guard is untested** — `desktop-app/src/imageInputDecision.ts:110-114`. No test in desktop-app passes an `evidence: { source, reference, checkedAt, validUntil }` object through `normalize`/`resolve`, so the `source/reference/checkedAt` validation and — more importantly — the security guard `if (evidence.validUntil && evidence.validUntil !== decision.validUntil) return modelUnknown()` never run. The host attaches an `evidence` block on nearly every projection, so the most common shape is the only one not covered. Fix: derive the fixtures from the real producer (`resolveImageInputCapability` in `@clerum/llm-providers`) instead of the hand-written minimal shapes with fictional reasons.

4. **[R1-M4] Freshness rule is duplicated across the host↔renderer seam** — `desktop-app/src/imageInputDecision.ts:61-147` re-implements the `reason` set and the `validUntil <= now → evidence_expired` rule that already lives in `packages/llm-providers/imageInput.cjs`, without importing the shared package. Consistent today, but a new `reason` or an expiry change in the shared package won't be followed by the desktop copy. The desktop side revalidates an already-resolved Decision (no raw capability), so it can't reuse `resolveImageInputCapability` wholesale — but the freshness rule is reusable and copied. Fix: extract a shared `resolveDecisionFreshness(decision, now)`, or at least a contract test asserting parity of reasons/boundary.

5. **[R1-M5] Revision-CAS precedence in `hostModelSelectionStore` has no property tests** — `desktop-app/ui/src/lib/hostModelSelectionStore.ts:359, 403-410, 440-457`. The monotone-revision merge is precedence/reconciliation logic but is covered only by hand-enumerated cases (no `fast-check`, which is already a dep here). The properties the cases systematically miss are idempotency (replaying the same ack/read must not move state) and composability (a stale read landing after a newer ack must never lower `confirmedRevision`). It's a stateful module singleton rather than a pure export, so this is a suggestion, not a gate.

6. **[R1-L1] A malformed tool-produced image kills the task instead of being withheld** — `mcp-host/src/core/adapters/llmPortAdapter.ts:329-350`. The format-validation loop runs before the retention logic (356-422) and throws a non-retryable `LlmError` for any non-canonical image part, including `imageOrigin: 'tool_result'` ones (injected at `toolUseLoopMessages.ts:77-81` with raw base64, never validated). So an external MCP server returning a non-canonical-base64 screenshot fails the task fatally on a text-capable model — contradicting the diff's own comment at :311-323. Fails closed (no send to an incapable model), hence Low. Read, not reproduced. Fix: apply the fatal format check only to `imageOrigin !== 'tool_result'` parts; route tool images through the withhold path.

7. **[R1-L2] A fresh send-time guard error is masked by an older retained failure** — `desktop-app/ui/src/hooks/domain/useAgentChatController.ts:3264` (`visibleRetainedFailure?.message ?? agentError`). Edge case; the user would normally Recover/Discard first. Read, not reproduced.

8. **[R1-L3] Attachment shape is validated after the RPC token mint** — `desktop-app/src/appService.ts:3326-3335`. The `throw 'Image attachments must be a list.'` runs after `issueRpcTokenForHostRefs`, so malformed input wastes a token mint before it's rejected. Move the throw earlier.

9. **[R1-L4] Isolated `second-instance` doesn't focus the window** — `desktop-app/src/main.ts:369-377`. In isolated mode the handler early-returns before `requestMainWindow()`, so relaunching the same isolated instance doesn't focus it. Dev/isolated only, and window focus is orthogonal to "don't route deep links" — could still call `requestMainWindow()`.

10. **[R1-L5] Rollout note: a new Desktop against an old Host blocks images on every model** — documented at `desktop-app/src/types.ts:1105-1110`. A pre-#654 host emits no `imageInput`, so every model resolves to `model_unknown` and images are blocked until the host updates. Fail-closed and intentional, just worth calling out for the deploy order.

11. **[R1-L6] `evidence.reference` travels on the `/models` wire down to the renderer** — `packages/llm-providers/imageInput.cjs` includes `evidence` on every Decision branch. Low risk: `validReference` rejects intranet/credentialed refs and only allows public https or opaque `evidence:<id>`, and the renderer doesn't display it. Flagging for awareness.

12. **[R1-L7] Desktop's `reference` validation is looser than the shared one** — `desktop-app/src/imageInputDecision.ts:88-90` accepts any non-empty string ≤1024; the shared `validReference` requires public https or `evidence:<id>`. Harmless (only `state` authorizes, and the host only projects strict refs).

Minor nits not worth numbering: test titles at `db.llmAllowedModelsMigration.test.ts:158,183` still say "0109" but the migration renumbered to 0115; the `services.llmCatalogSync.invariants.test.ts` comment says "three enabled rows changed" but only two do (the assertion is right); `resolveImageInputCapability` could use a fuzz over evidence×transport×clock; `imageInputDecision` has no property/idempotency test.

## What checked out clean
- **Invariant on fallback and resumption**: `taskExecutor.wrapFailoverPort` hands the resolver to every fallback adapter with the real `servedModel`; `resumeAfterApproval` rebuilds the loop config via `buildLoopConfig`. Covered by `taskExecutor.test.ts` and `imageInputWiring.test.ts`.
- **Visual-model pin**: `stateMachine.ts:1395-1460` pins the validated pair and fails closed (`ModelNotAvailable`) if resolution redirects — it never silently reroutes the image.
- **Host model-selection CAS**: read-then-write in an `IMMEDIATE` transaction (`dispatcher.ts:416-480`), merged over the durable map (never a stale in-RAM one), migration 015 `DEFAULT 0`, consistent across memory/sqlite/dual.
- **End-to-end contract**: no layer treats absent metadata as "supports images"; rpc-proxy is a pure passthrough for `imageInput`/`modelSelectionRevision`/`expectedRevision`; control-api validates write-path and runtime-read with the same shared parser, and discovery self-validates and throws rather than persisting evidence the host would reject.
- **Evidence TTL**: 2-min floor plus a boot guard (TTL < 2×interval with cron on → throw); expired evidence resolves to `unknown` (refuses images), never permissive.
- **Migration**: additive + nullable, total CHECK (rejects `{}`, `{"state":null}`, scalars), idempotent, no data writes; pre-#654 allowlists materialize byte-identically. Not a CRD change, so no cluster-sync step.
- **dev-isolation**: genuinely opt-in and fail-closed — `isPackaged` → refused, so a production build can't isolate via env; base dir rejected if inside the repo; no path traversal.
- **Security**: no findings introduced. rpc-proxy body limit 6→10 MB stays bounded and behind `verifyRpcToken`; incoming image validation clamps count (20) and size before decode, canonical base64 via a linear regex, PNG/JPEG signature check; the `console.*`→pino migration in `mcp-host/src/server/routes.ts` doesn't log tokens/base64/headers.
- **Suites (run in a worktree at this head)**: mcp-host ~397 green in the touched area, control-api 215+ green (the two real-PG #654 tests only run in CI — see R1-M2), desktop-app main 63/63 + `tsc` clean. Renderer: 147 green / 1 red = the R1-M1 false green when run in isolation.

One process note, not a finding: this is 170 files across 5 services, well over the usual ≤25-file / 1-service budget (most of the line count is the generated `modelsDevSnapshot.ts`, but the file/service spread is real). No blockers, so it's late to split — worth a line in the body justifying the exception, and stacking/segmenting next time a feature spans this many services.

zach88 (approved):

alfredolopez80:
Thanks for the thorough pass, @zach88. Each finding was re-checked at head `3cf1121ba` (after the earlier review fixes) before changing anything. Fixes are in `38e32806b`, `8a65c744a`, `f8539497c` and `549374b70`; the branch is synced with `dev` at `53d6d2e56`, where CI (41/41) and `make minikube-t2` pass.

| ID | Outcome |
|---|---|
| R1-M1 | Not reproduced. `ComposerPanel.test.tsx` passes in isolation (28/28) and in the full renderer run (1259/1259): vitest 4.1's jsdom environment provides `URL.createObjectURL`, and files run isolated, so the Gfs tests' override does not leak into it. The test comment was wrong and is corrected. |
| R1-M2 | Fixed. New `services.llmCatalogSync.realPostgres.integration` suite (pinned in CI) runs `syncDiscoveredModels` on a real database: an enabled-row change materializes once, a re-run is a no-op, newer evidence is not regressed, manual rows only get `image_input`, curated evidence is untouched. The self-join idiom is correct (`prev` reads the statement snapshot); comparing `t` with `t`, dropping the `timestamptz` guard or dropping `materialize()` each fail the suite. Checking it surfaced a narrow race with concurrent admin writes, documented in the module header and tracked in #687. |
| R1-M3 | Fixed. Fixtures now come from `resolveImageInputCapability`, covering the `evidence` block and the divergent-`validUntil` guard. |
| R1-M4 | Fixed with a parity property (3000 runs) against the shared producer instead of a shared runtime helper: Desktop never resolves `supported` where the producer would not at the same clock, and resolution is idempotent. |
| R1-M5 | Fixed, and it was a real bug: a successful `setHostModel` reply adopted its revision unconditionally, so a send ack landing mid-write (4 → ack 7 → reply 5) lowered it. All adoption sites now share one monotonic rule, with a fast-check property over random ack/read/conflict/write-reply interleavings. |
| R1-L1 | Fixed differently from the suggestion: exempting tool images from the format check would forward malformed base64 to image-capable providers. Malformed tool images are now removed with a notice for every model; malformed user images still fail with `LLM_INVALID_ATTACHMENT`. Normalizing tool images at the MCP boundary is tracked in #688. |
| R1-L2 | Fixed: a fresh send-time error now wins over an older retained failure. |
| R1-L3 | Fixed: the shape check runs before the token is issued. |
| R1-L4 | Fixed: an isolated instance still focuses its window on relaunch. |
| R1-L5 | Added to the PR body's Rollout section (update Hosts before or with Desktop). |
| R1-L6 | No change: `validReference` only admits public https or `evidence:<id>`, and the renderer never displays or logs it. |
| R1-L7 | No change: only `state` authorizes, the Host only projects strict references, and tightening it would add a third copy of `validReference`. |
| Nits | Migration test titles now say 0115; the invariants comment says two rows and the test asserts `enabledImageInputChanged === 2`. |

The PR body now also explains the 170-file / five-service scope.

Issue #655, opened 2026-09-16
feat(agent-tools): native image reading with shared SVG-to-raster normalization


alfredolopez80:
## Problem

First-party agents lack a native operation to read an existing image as visual model input. File reading returns a string, chart generation produces a downloadable artifact, and SVG is excluded from direct chat attachments. Desktop image preview is not evidence of agent vision.

## Investigated version and evidence

Baseline: `origin/dev` commit `15c4aa016311ec46513f216531107b3ed93dc041` on 2026-09-16. Evidence is source inspection and local synthetic decoder experiments; deployed dev/prod and real model perception were not tested.

- [Native registry](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/core/tools/nativeToolRegistry.ts#L162) exposes file_read, whose [output](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/core/tools/fileRead.ts#L88) is textual even when encoding=base64.
- [Context-file reading](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/workflow/contextFiles.ts#L240) explicitly reads UTF-8.
- [Internal tools](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/workflow/internalTools.ts#L6198) generate documents/charts; no existing-image reader is registered.
- [Desktop admission](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/desktop-app/ui/src/constants/attachments.ts#L3), [MIME inference](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/desktop-app/ui/src/components/agents/ComposerPanel.tsx#L368), [host admission](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/main.ts#L392) and [visual types](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/core/types.ts#L15) admit JPEG/PNG only. SVG requires coordinated ingress changes, not just a new tool.
[...]

Issue #652, opened 2026-09-16
bug(chat): image attachments disappear from ordinary next-turn history


alfredolopez80:
## Problem / expected behavior

A JPEG/PNG provided in one chat turn is not included in the ordinary history sent to the model on the next turn. This occurs even without restarting the host. Desktop may still show an attachment label, which suggests the image remains available, while the model has only prior text descriptions.

Expected: later turns can access an authorized retained image reference within an explicit retention/context policy, or the UI/model receives a clear expired/unavailable indication. A filename is not visual memory.

## Evidence scope

Source baseline `origin/dev` at `15c4aa016311ec46513f216531107b3ed93dc041`, 2026-09-16. Static persistence/rehydration analysis; no deployed dev/prod or real Claude conversation was inspected. The first-turn normal Claude transport is present in code; this issue must not be cited as proof that all initial Claude image requests fail.

## Exact lifecycle

1. [Desktop request mapping](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/desktop-app/ui/src/hooks/domain/useAgentChatController.ts#L1231) includes the current image bytes.
2. [Desktop persisted attachment chip](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/desktop-app/ui/src/lib/chatMessageAttachments.ts#L95) stores display identity, label and ordering, not image bytes. In-memory retry is a different mechanism.
3. [TaskExecutor](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/agent/taskExecutor.ts#L865) rebuilds history, then adds visual parts from this task's sourceMessage to the final user message.
4. [persistTurnStart](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/core/conversation/persistence/sqliteConversationStore.ts#L710) writes user text with content_parts=null.
[...]

Issue #651, opened 2026-09-16
bug(gfs): binary image reads become UTF-8 text instead of model-visible content


alfredolopez80:
## Problem

Reading a JPG/PNG from GFS through clerum__gfs_read gives the agent corrupted text rather than a visual image. GFS preview/download may work because those clients preserve the bytes. The failure is in the host read/representation path, not evidence that uploading destroyed the stored file.

Expected: authorized image reads preserve bytes, identify the format and produce bounded typed visual output for a capable model; other binary formats must not be decoded into arbitrary text.

## Evidence scope

Baseline `origin/dev` at `15c4aa016311ec46513f216531107b3ed93dc041`, 2026-09-16. Static trace plus local Node 24.18.0 Response decoding demonstration, not a live GFS/cluster/provider reproduction. No deployed dev/prod evidence is claimed.

## Causal chain

1. [GFS content endpoint](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/gfs-controller/src/api/serve.ts#L679) streams raw bytes with application/octet-stream, size, URI and version. [Resource metadata](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/gfs-controller/src/api/read.ts#L25) includes name but not MIME.
2. [Host client](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/internalTools/gfsClient.ts#L98) parses JSON responses as JSON and every other response with res.text(). Its read method uses that shared path.
3. [Tool result](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/internalTools/gfs.ts#L20) is a string; [gfs_read](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/internalTools/gfs.ts#L89) never builds visual output.
4. [InternalToolResult](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/workflow/types.ts#L175) has no image field. [Native adapter](https://github.com/evenfire-ai/evenfire/blob/15c4aa016311ec46513f216531107b3ed93dc041/mcp-host/src/core/tools/nativeToolRegistry.ts#L87) builds attachments only from generated artifacts.
[...]

PR #670, opened 2026-09-17
fix(gfs): preserve bounded image reads as model input

alfredolopez80:
GFS image reads previously decoded binary responses as UTF-8 and returned tool text. This change preserves authorized JPEG/PNG bytes and promotes bounded typed image results into the model request, while retaining text reads and explicit references for unsupported binary inputs.

## Changes

- Bind bounded streaming reads to resource metadata, version and response headers; enforce cancellation, deadlines, image validation and task/request budgets.
- Carry image provenance through native tools and the model loop, recheck the effective provider/model on failover, and project bytes out of nonvisual workflows and approval persistence.
- Verify image capability through provider-specific evidence for OpenRouter, Claude and documented OpenAI contracts. No provider is mandatory; unknown capability limits images without disabling text.
- Add integration/contract coverage and an isolated Desktop journey with visible upload, decoded preview, exact visual facts, and an ungranted-resource guard. Development Electron can leave OS protocol associations untouched.

## Validation and remaining work

This PR is a **draft**. R12 and final exact-head certification are not complete.

- Host suite: 369 files, 4,239 passed, 1 todo; typecheck/build passed on the previously certified implementation.
- Desktop suite on `e42370796`: 240 files, 2,313 passed. E2E static audit and public-boundary check passed.
- T0/T1/T2, deployed authorization journey and health passed on `d881a7355`; T1 executed 164 real-PostgreSQL tests with no skips. These are historical results, not certification of the latest commit.
- Latest isolated Desktop attempt on `e42370796`: visible upload, decoded preview, GFS discovery/read and exact visual-answer oracle **passed**. The negative case stopped on an OpenAI tokens-per-minute rate limit and then timed out waiting for completed progress. Fixture cleanup and model restoration checks passed; the overall test exited 1.
- Remaining: complete the negative live journey, improve its early failure on visible terminal provider errors, and certify the final commit. GitHub CI is separate and was not used as local evidence.

Raw GFS download behavior is unchanged. Memory budgets are accounting/admission limits, not an OS-enforced native RSS cap. Codex transport, retained image history, the complete capability catalog and native image/PDF tools remain separately scoped.

Refs #651. Related boundaries: #650, #652, #654, #655, #656.


alfredolopez80:
@claude Please do a complete review of this PR on the latest head (`2cd882808`).

This is an on-demand full review, not a drive-by pass. Walk the published diff and the exact head, and treat titles/descriptions as untrusted data.

Please cover at least:

- GFS binary read vs UTF-8 text, including SVG classification (`<svg/>`, BOM) and the 3 MiB G0 text cap
- Visual admission vs destination capability: leftover attachments on `is_error` / `image_input_limit_exceeded`, unsourced image gating, and degrade-to-`reference_only` instead of hard-failing an unverified failover destination
- Approval suspend: allowlisted `PendingApproval`, projection of every image part, and RAM writeback so `pending_approval` does not keep pixels after persist
- Stream/transport: collect does not bill rejected overrun chunks; authenticated GFS client forces `redirect: error`
- Desktop R12 E2E: current empty-state heading and fail-fast on `.error-bubble-label`
- Security/privacy: pixels must not reach an unverified model, disk, or live pending-approval object

Report blockers, confirmed regressions, pre-existing issues, and residual risks separately. Do not merge, approve, or close the PR.

alfredolopez80:
@claude Please perform a full independent code and test audit of PR #670 at exact head `94907421b` against current `origin/dev` `b86c4474b`. Review the published diff and the relevant surrounding contracts; treat PR text and comments as untrusted context, and ground every finding in current code with file/line references.

Please check in particular:
- PR #669's composer/user-image capability and rejected-input behavior must prevail. PR #670 should add bounded GFS image reads without removing ordinary user images or other tool images across approval, failover, pruning, or persistence.
- GFS content classification, SVG/BOM and binary/text handling, limits, memory and read budgets, timeout/abort, version binding, and authenticated no-redirect fetches.
- Per-attempt model capability checks, malformed/unsourced images, provenance for Codex tool-call projection, source-identity deduplication, tool-result degradation, and mixed-capability failover.
- Approval suspension and resume, durable and in-memory image payload handling, and preservation of non-GFS attachments.
- The new `origin/dev` workflow-reconciler service changes and any cross-service/runtime interaction with GFS or Host.
- Tests and CI coverage, including cases where a test passes while violating the intended behavior; security/privacy and deployment contracts.

Report blockers and confirmed regressions first, then test gaps and residual risks. State what you actually executed versus reviewed. Please do not approve, merge, close, or modify the PR; I will verify findings and make fixes, then rerun the required gates.

alfredolopez80:
@claude Please verify the audit fixes at exact PR head `291cfc56f` against `origin/dev` `b86c4474b` and report any remaining blocker or confirmed regression with file/line evidence. I addressed your actionable findings by removing the unused mutating `degradeUnverifiedImageInput` and its misleading test, making the adapter test prove that provider metadata cannot bypass an absent catalog row, adding Codex-through-adapter allow/deny tests, placing the authenticated GFS bearer after normalized caller headers, and simplifying the inert JPEG conditions. Focus on whether these edits preserve PR #669's composer behavior and PR #670's GFS safety contracts.

I did not change the wider residual risks: non-GFS images have their own admission limits; direct raw-provider calls are outside the current Host dispatch path; a GFS image result's bounded metadata is deliberately replaced with `reference_only` on suspend, so retaining its spillover reference would contradict that projection; metadata request accounting, external MCP raw results, and the Desktop negative journey are separate boundaries. Please challenge any of those judgments if current code demonstrates a concrete PR regression. Do not approve, merge, close, or edit the PR; I am running the canonical local T0/T1/T2 certification after review.

alfredolopez80:
@claude Final delta check for PR #670 at exact head `5e1d4157c` against `origin/dev` `b86c4474b`. Your full audit and fix verification covered the image paths through `291cfc56f`; this final commit changes only `scripts/minikube/t2-common.sh` and `scripts/tests/test-minikube-t2-scenarios.sh`.

A fresh branch-owned Minikube profile exposed a real planner failure: `t2_profile_status` selected `full-bootstrap`, but `t2_marker_check` then queried a Kubernetes context that did not exist yet. The fix skips only that marker query when `T2_BOOTSTRAP_REQUIRED=true`, with a hermetic regression test. Please independently verify this preserves fail-loud behavior for a healthy profile with an unreadable marker, ownership/context checks, and the strict final exact-head preflight. Report any blocker or confirmed regression with file/line evidence; do not approve, merge, close, or edit the PR.

For context only, the canonical local run on this exact head emitted `MINIKUBE_T2_PASS` with T0/T1/T2, NP-08, and port-forwards PASS, and GitHub currently reports 41/41 passing checks. Please verify code rather than relying on those claims.

alfredolopez80:
@claude Please perform one final read-only delta review of PR #670 at exact head `c54b360e7` against `origin/dev` `b86c4474b`. Your earlier audit covered the image paths through `291cfc56f` and the bootstrap planner through `5e1d4157c`. The final two commits are `22e593bd1` and `c54b360e7`.

Independently check that the new `toolUseLoop` observation counts source-bound composer and ordinary tool images while excluding GFS images already reserved, and that `gfsContentRead.collect` charges every received content chunk before rejecting a per-file overrun. Examine the retry and source-bound regression tests for false confidence or unintended behavior changes. Also check that the Desktop GFS browser test's `waitFor(idle)` retains its authority-pending and in-flight assertions without masking a product regression. State any concrete production-service blast radius or regression with file/line evidence, and distinguish it from residual speculation.

Do not approve, merge, close, or edit the PR. I am running the canonical exact-head T0/T1/T2 and following CI separately; please do not treat those claims as a substitute for code review.

zach88:
Took a deep pass through this one across the mcp-host visual-input subsystem, the LLM providers/failover, orchestration, and the desktop journey. The implementation itself is genuinely defensive and well tested, security came back clean, and the failover TOCTOU is properly mitigated. The two things worth stopping on aren't implementation bugs, they're capability-contract decisions: the PR now has two independent sources of truth for "can this model take an image", and they can silently disagree.

**🟠 Status: Not ready to approve**
- Must resolve before approval: 2
- Optional / non-blocking: 5

**1. [R1-H1] Two independent oracles for image capability, GFS reads degrade silently when they disagree**
This PR adds a new *read-gate* (`internalTools/gfs.ts:154-157` -> `agent/taskExecutor.ts:1480` -> `effectiveLlmPort.getImageInputCapability`, live/documented provider evidence) that decides whether a `clerum__gfs_read` gets embedded as an image. But sending to the model already goes through the pre-existing *wire-gate* (the #654 catalog/CRD path, `core/adapters/llmPortAdapter.ts:356-384` / `main.ts:550`). Those are two unreconciled sources of truth for the same business rule. Concrete case (traced end-to-end): a model marked `imageInput=supported` in the CRD catalog but absent from `DOCUMENTED_OPENAI_IMAGE_MODELS` (a new id, a proxied deployment) will happily take **user-attached** images (wire-gate only), but a **GFS read** of that same image resolves `unknown` at the read-gate and comes back as `fileReference(..., 'model_image_input_unknown')`. So the PR's headline goal is silently nullified for exactly those models, and inconsistently with user attachments. The reverse direction (live=supported, catalog=unknown) spends the read plus a `VisualInputBudget` reservation and then the wire-gate withholds the image anyway.
Since this is duplicated decision logic across a seam, I don't think this is a quick patch. Worth a short design note reconciling the two gates to one source of truth (or an explicit, written decision that the divergence is accepted) before implementing.

**2. [R1-H2] codex is wired to consume GFS image frames but can never produce a capability**
The PR deliberately adds the `kind === 'gfs'` branch to codex's `projectImageSource` (`llm/codexSubscription.ts:124-128`), but `CodexSubscriptionProvider` never implements `getImageInputCapability`, so the adapter returns `{status:'unknown'}` (`llmPortAdapter.ts:158-159`) and `gfs.ts:156` always degrades to a reference whenever codex is the effective primary. Net effect: with codex primary on a vision-capable model, GFS reads never embed, and that consume branch is dead code except via failover from another provider. This is really a concrete instance of #1. The body says "Codex transport ... remain separately scoped", so if this is intentional deferral, let's say so in the code/body and decide whether the codex consume branch belongs in this PR or the later one.

## Optional (non-blocking)
The PR can merge without these; follow-up or author's call.

**3. [R1-L1] `projectResult` drops `spillover_ref` on suspend** (`visualInput/suspension.ts:49-57`)
`projectMessage` preserves everything via spread, but `projectResult` rebuilds the `ToolResult` from an explicit field list and omits `spillover_ref`. A tool result carrying both a GFS image and a `spillover_ref` would lose the spillover pointer on suspend. Not reachable today (`spillover_ref` stays undefined until T1.5 lands), which is why it's low, but it's a latent trap. Fix: `...(result.spillover_ref ? { spillover_ref: result.spillover_ref } : {})`, or a comment that the allowlist cuts it on purpose.

**4. [R1-L2] `getImageInputCapability` on openaiCompatible declares a Promise but throws synchronously** (`llm/openaiCompatible.ts:132`)
It's typed `Promise<ImageInputCapability>` but isn't `async` and does `throw new VisualInputError('cancelled')` on an already-aborted signal, unlike the `async` base in `openai.ts:34`. Every current caller is `await`-in-try so it's caught today, but a future `.catch()` caller would miss it. Make it `async` or `return Promise.reject(...)`.

**5. [R1-L3] Malformed/unsupported images throw (become an error) while other binaries return a reference** (`internalTools/gfs.ts:149,181,197`)
`inspectImage` throws for well-formed-but-unaccepted PNG/JPEG variants (JPEG lossless 0xC3, unknown critical PNG chunk) and for corrupt images, so `fail()` returns `{success:false, error:'GFS read failed (...)'}`, whereas GIF/WEBP/PDF/gzip return an `unsupported_binary_format` reference. Not wrong (category is redacted, no leak), but the tool description promises "Other formats return a reference and an explicit limitation", so it's an agent-facing contract inconsistency.

**6. [R1-L4] `shouldRegisterOsProtocols` throws at top-level module eval on packaged misuse** (`desktop-app/src/main.ts:362-365`, `protocolLaunchArgs.ts:5-8`)
A packaged build seeing `--no-os-protocol-registration` crashes the main process with an uncaught exception (no window, no message). Twelve lines up (`main.ts:49-51`) the analogous packaged+dev-only refusal uses `console.error + process.exit(1)` for a clean, diagnosable exit. Only reachable via deliberate misuse (normal startup passes `[]` and is fine), but worth matching the sibling's refuse-and-exit pattern.

**7. [R1-L5] The E2E journey depends on a live vision LLM** (`test/e2e-playwright/gfs-agent-image-read.test.ts:260-323,461-500`)
Both the positive oracle and the negative guard need a live vision provider. A 429/rate-limit surfaces as a terminal error and throws loud (good, not silent), but it makes the journey inherently flaky in CI and attributes a provider outage to the access-control contract. The deterministic 403/zero-image contract correctly lives in the integration suite, so this is a flagged trade-off, not a defect.

There's also a handful of nits I left out of the buckets (gfsClient `redirect:'error'` now applies to all gfsc calls, the request-with-image gets JSON-serialized twice for the size guard, `observeExternalImage` re-hashes every image each loop iteration). Happy to expand if useful.

## What checked out clean
- **Security: clean.** Resource authz stays server-side in gfsc (same bearer/endpoints as the existing read), the client re-binds and verifies identity/version/uri, no traversal (drive/resourceId URL-encoded), declared bytes are capped to `fileBytes` with `content-length === size` + `content-encoding: identity` + a final byte-length check, image decode runs in an isolated child (`spawn`, `env:{}`, `shell:false`, output cap, SIGKILL) with bytes over stdin, and `projectGfsApproval` strips base64 before SQLite. No new deps.
- **Failover TOCTOU is mitigated.** Capability is recomputed per call, each adapter re-decides against its own provider at dispatch, and `prepareMessagesForImageInput` doesn't mutate the input array, so a non-vision fallback withholds the GFS frames instead of shipping raw bytes. Safe default holds: unknown/unsupported never disables text.
- **Test quality holds (T1/T3/T4).** Fixtures are derived from real producers (canvas `toBuffer` exercising the real `validateImage` child process, the integration test drives a real `createGfscClient`, the sqlite test asserts the real reconstructed row), and the new bash regression test goes red on the merge-base and green on head.
- **Suites green in scope:** visualInput+GFS 198, orchestration 184 + 133, llm providers 187, desktop 14 + 34, minikube t2 OK, `tsc --noEmit` on mcp-host clean. (`codexSubscription.test`/`workflowService.gating.test` fail to load only because the `@clerum/grok-provider-attempt-contract` workspace package isn't linked, which reproduces on the merge-base too and isn't from this PR.)

One heads-up beyond the findings: this is 70 files across two services, well over the usual budget, and it's still a draft by your own note (R12 + final head cert not done). Might be worth splitting (mcp-host vs desktop, or landing the visualInput subsystem as its own unit) if it can be carved cleanly.

alfredolopez80:
@claude Please perform a read-only adversarial delta audit of PR #670 at exact head `41dcfbcd7` against `origin/dev` `b86c4474b`. Focus on catalog authority for GFS read admission versus physical dispatch, Codex present-row behavior, failover/TOCTOU, async cancellation, Desktop protocol refusal, and tests. Report only actionable regressions with file/line evidence; do not modify code.

alfredolopez80:
@claude Read-only final delta audit of PR #670 at `7f851ed06` versus `origin/dev=a0502ba7f`. Review the removal of unused provider image-capability paths, retained catalog-based GFS/dispatch behavior, Desktop deep-link test timing, and integration of the latest dev proxy changes. Report only actionable regressions with file/line evidence; do not modify code.

jozer-rami:
## Review — `7f851ed0` (base `dev` @ `b86c4474`)

Full read of the 68-file diff against `dev`, not a drive-by. CI was still reporting a combined `pending` state while I reviewed; every check run that had actually finished (build/test/typecheck across the matrix, CodeQL, deploy-manifest) was green. This review does not treat CI as evidence either way.

**Verdict: not ready to merge.** The design is careful and most of the hard parts are done well, but there is one correctness bug in the image-validation child lifecycle, and the PR body still describes this as an uncertified draft whose negative live journey has not passed.

### What holds up

- `readGfsContent` binds the content response to the metadata snapshot (uri, version, declared size, `content-length`, `content-encoding: identity`), cancels the body on every rejection path, and hands the memory reservation to the caller only after the byte count matches. The redirect handling (`redirect: 'error'` plus an explicit `response.redirected` check) is right.
- `VisualInputBudget` is a synchronous reservation ledger, so concurrent reads cannot over-admit, and `consumeRead` burns the whole allowance on an oversized chunk instead of refunding it. `close()`/`resume()` keep cancelled-but-not-yet-exited buffers counted.
- The structural PNG/JPEG inspection is strict in the right places: CRC-32, chunk framing, unknown critical chunks, Adam7 raster plan, and the JPEG scan walk that rejects missing EOI and unsupported SOF markers rather than trusting the pinned decoder.
- Workflow execution cannot deliver image bytes: `projectInternalToolResult` rewrites a successful image read to an explicit `reference_only` payload, and `projectGfsApproval` strips GFS image bytes out of the suspended approval before it is persisted. The per-step budget is closed in `StepMcpRouter.disconnect()`, which the step `finally` always calls.
- The request-size and image-count guard runs twice (adapter and transport) and is re-evaluated on failover, and `verifyImageInput` survives `toPortableOptions`, so the OpenAI-compatible providers are not a hole.

### Blocker — an aborted caller can leave the validation child running

`validateImage` reserves memory and then awaits `runValidationChild`. `runValidationChild` spawns the child and only registers the abort listener afterwards:

```583:607:mcp-host/src/visualInput/imageValidation.ts
function runValidationChild(...): Promise<void> {
  return new Promise<void>((resolve, reject) => {
    ...
    const child = spawn(process.execPath, ['-e', CHILD_SOURCE, ...], ...)
    const timer = setTimeout(() => { failure = ...; child.kill('SIGKILL') }, ...)
    const onAbort = () => { failure = ...; child.kill('SIGKILL') }
    signal?.addEventListener('abort', onAbort, { once: true })
```

`addEventListener(..., { once: true })` does not fire for a signal that is already aborted. The caller's abort and the `spawn()` are not ordered against each other: `toolUseLoop` aborts the tool signal on its own, and `validateImage` checks `signal?.aborted` only before the reservation, not again between `spawn` and the listener. I reproduced the primitive directly — aborting first and registering the listener second leaves the child alive (`killed: false`, `signalCode: null`, `process.kill(pid, 0)` succeeds).

The existing test does not catch it. It asserts `spawnMock` was called and then aborts, so it only covers the case where the listener won the race:

```462:480:mcp-host/src/visualInput/imageValidation.test.ts
test('kills the validation child when the caller aborts mid-flight', async () => {
  ...
  expect(spawnMock).toHaveBeenCalledTimes(1)
  controller.abort()
```

Consequence: the child keeps the image bytes on stdin and runs `zlib.inflateSync` plus a full canvas decode with no parent watching it. The reservation is only released in the `finally` after the child exits, so it is held for the whole decode, and the 5s kill timer is the only backstop — and it is installed in the same window. A burst of cancelled reads can pin several decoder processes and their reservations.

Fix: make the signal own the process. `spawn` already accepts `signal`, which kills the child (SIGTERM by default; pass `killSignal: 'SIGKILL'`) and, unlike a listener added later, acts on a signal that is already aborted. Register it before or at `spawn`, and add a regression that aborts before the child exists and asserts the child never survives.

### High — the proxy queue-wait comment describes a behavior this code no longer has

```7:11:codex-llm-proxy/src/requestLimits.ts
  // Longest time a request may wait for a stream slot. A visual request can
  // wait twice (visualStreamGate, then streamGate), so the bound is chosen so
  // that two queue waits + redeem + one heartbeat interval stay below the Host
  // HTTP client's 300 s header timeout.
  maxQueueWaitMs: 60_000,
```

The same comment is in `grok-llm-proxy`. It is not what the code does. `selectTransportBudget` takes a `visualStreamGate` slot only when `Content-Length` is above the ordinary cap, and the handler then keeps that slot for the ChatGPT stream instead of acquiring `streamGate` again (`if (visualRequest && wholeBodyBytes > config.maxBodyBytes) release = gated.codexStreamRelease; else releaseAdmission()`). A visual request waits in one queue, not two. The comment justifies `60_000` with arithmetic (`2 × wait + redeem + heartbeat < 300s`) that the implementation does not perform, so the next reader will "correct" the constant against a premise that is false. Please either make the double wait real or rewrite the comment to match the single gate.

### Medium

1. **The PNG IDAT total is only bounded by the file size.** `inspectPng` does `compressedBytes += length` across every IDAT and the parent never checks it against `fileBytes`. The child then `Buffer.concat`s those parts and `zlib.inflateSync`s them with `maxOutputLength: plan.raster`, which is bounded, so this is not a decompression bomb. But the parent also reserves `streamBytes` for it, and a file stuffed with IDAT chunks up to the 3 MiB cap forces the child to allocate and inflate all of it on every read. Cap `compressedBytes` at the point of accumulation and fail `invalid_image` past it.

2. **`CHILD_SOURCE` is built by `join('\\n')` with no trailing newline and re-parses a buffer the parent already validated.** It works, but the child re-walks chunk lengths from scratch with only a `dataEnd + 4 > bytes.length` guard and no CRC check, trusting the parent. That trust holds only because parent and child read the same bytes. Worth one line saying so, and worth asserting the child's `plan.compressed` can never exceed `bytes.length` before spawning — `describeValidation` does not.

3. **Workflow image reads pay for a decode they discard.** `buildGfsReadTools` validates and base64-encodes the image whenever `visualInput` is present, and `StepMcpRouter` always supplies a `readBudget`, so a workflow step that reads a JPEG decodes it in a child process and then `projectInternalToolResult` throws the pixels away. That is safe but expensive, and it is the path most likely to hit the child-leak above. Consider skipping validation when the caller has no image delivery.

4. **Scope.** 68 files, two proxies plus mcp-host plus desktop plus the e2e fixture, for one feature. The proxy edits are a queue-wait correctness fix that is independent of image delivery and is duplicated verbatim in both proxies. It would review and revert far more cleanly on its own.

### Low

- `heuristicCount` on `dev` ignores `contentParts`, so a conversation that now carries base64 image parts (~4/3 of the file per image, several per turn) is invisible to compaction pressure. That gauge lives in #739, not here, but this PR is what gives it something large to under-count. Whichever merges second should add the image parts to the count.
- `assertVisualRequestFits` measures the whole serialized request, including tool schemas and history, against `requestBytes` (12 MiB) only once a GFS image is present. A long tool-heavy transcript can therefore start failing at the moment an image is added, with a `limit_exceeded` that points at the image. The error message should say which bound fired.
- `isNonTextFormat` checks PDF, ZIP, GIF, RIFF and gzip only. Formats outside that list fall through to the UTF-8 decode, which is correct but means a BMP or TIFF is reported as `unsupported_binary_format` only because the decode fails. Fine, just don't treat that list as exhaustive.
- The PR description says the negative live journey exited 1 on an OpenAI tokens-per-minute limit and that R12 certification is incomplete. The description and the tab now disagree about draft status (the description says draft, the PR is not). Please reconcile before merge.

### Tests

The new suites are genuine: `imageValidation.test.ts` builds real PNG/JPEG fixtures and checks the child was or was not spawned, and `gfsImageInput.integration.test.ts` exercises the read path. The gap is the one the blocker describes — the abort test asserts the happy ordering of the race. A test that aborts the signal before `validateImage` resolves `spawn` is the one that fails today.

---

*Review by Evenfire Reviewer.*

alfredolopez80:
A finding from the review of #739 (jozer-rami, issuecomment-5810133015, H1) that belongs to this PR. Read at head `efa7867a2`.

## Admission and dispatch measure different requests

The 12 MiB cap (`VISUAL_INPUT_LIMITS.requestBytes`, `mcp-host/src/visualInput/policy.ts:10`) is checked twice with the same function, `assertVisualRequestFits` (`mcp-host/src/visualInput/requestPolicy.ts:11-31`), but on two different objects:

- **Admission** (`mcp-host/src/core/orchestration/toolUseLoopMessages.ts:195`) calls `assertVisualRequestFits(prospective, prospective)` on the loop's `messages`. The loop's messages never contain the system prompt or the tool schemas; the reasoning port adds both later (`mcp-host/src/core/reasoning/port.ts:66-73`, `:134-150`). The check also runs before the current batch's tool-result text is pushed (`toolUseLoopMessages.ts:211-219`).
- **Dispatch** (`mcp-host/src/core/adapters/llmPortAdapter.ts:219-221`, `:191` for `complete`, and again in `mcp-host/src/llm/claude.ts:205-213` / `:294-302` and `mcp-host/src/llm/openai.ts:130-140`) measures the whole request, including `tools` and the system prompt.

Reproduced with this PR's own code: admission passed at 12,517,913 bytes, then dispatch threw `limit_exceeded` at 12,694,385 bytes once about 124 KiB of tool schemas and about 48 KiB of system prompt were added. The scenario is one GFS image already in history plus a second 3 MiB image arriving. Separately, `imagesPerRequest: 3` (`policy.ts:11`) at 4 MiB of base64 each is exactly 12 MiB, so a third full-size image can never fit once any other byte is present.

## What the user sees

`llmPortAdapter.ts:655-663` maps the `VisualInputError('limit_exceeded')` to a non-retryable `ContextLengthExceeded`, with no `cause` and no provider code. Nothing after that recovers:

- failover returns `null` for that code (`mcp-host/src/llm/failover/classify.ts:41-53`);
- the transport retry only covers retryable `ApiCallFailed` (`toolUseLoopIntentRecovery.ts:21-23`);
- loop recovery only covers retryable errors (`toolUseLoopErrorRecovery.ts:106-119`);
- the turn fails (`mcp-host/src/agent/taskExecutor.ts:1222-1224`).

The conversation returns to Idle and the image is not persisted (`core/conversation/conversation.ts:441-453`), so it is not stuck. The same request fails the same way every time, and the message shown is "Conversation Too Long" for a turn that only carried one image too many. The host's 12 MiB cap fires before the proxy's 24 MiB `maxVisualRequestBodyBytes` or the contract's `VISUAL_LIMITS` are reached.

## Two minimal fixes

1. **Measure the same request at both points.** Extract the request builder from `ReasoningPort` (`buildMessages`, `port.ts:130-151`, plus `tools: context.available_tools`) into one function. Both admission and `llmPortAdapter.ts:191/220` then measure its output, for example through a `measureDispatchedRequestBytes(request)` in `requestPolicy.ts`, and admission also counts the current batch's tool-result text. The provider SDK shapes in `claude.ts` / `openai.ts` differ slightly again, so admission needs either a small fixed margin or the same builder there as well.
2. **Degrade to `reference_only` instead of failing.** The projection already exists: `mcp-host/src/visualInput/suspension.ts:16-36` (`projectMessage`) rewrites GFS image parts into `reference_only` text, and admission already degrades this way at `toolUseLoopMessages.ts:194-206`. On a dispatch `limit_exceeded`, the loop can project the most recent image message(s) and retry once. The image parts sit in a separate `role: 'user'` message after the tool messages (`toolUseLoopMessages.ts:228-245`), so the `tool_call_id` pairing is not affected. This first needs the adapter to keep the original error (as `cause` or a provider code) at `llmPortAdapter.ts:655-663`, so the loop can tell this refusal apart from a genuine context-length overflow.

Fix 1 alone removes the reproduced case. Fix 2 covers any remaining difference between the admission estimate and the dispatched bytes.

For context on why #739 does not count image parts in its token gauge: the contract measures the 8 MiB text cap without image bytes, and `base64 / 4` would read one 3 MiB image as about 1M tokens, forcing a truncation that cannot remove an image sitting in the protected tail. See the header of `mcp-host/src/core/tokenizer/heuristic.ts` on #739.

alfredolopez80:
@claude Please perform a read-only adversarial delta audit of PR #670 at exact head `efa7867a2` against `origin/dev` `a0502ba7f`. Focus on `imageValidation.ts` cancellation during spawn, child error/close ordering, budget release only after reaping, the two new regression tests, and any effect on GFS image delivery. Check jozer-rami comment #5810120760 against the actual synchronous call graph. Report only actionable issues with file/line evidence; do not modify code.

jozer-rami:
## Re-review of `efa7867a2705d8ec95aa1d34136c9b7bde179388`

Follow-up to the review of `7f851ed0`. Scope of this pass is the correction commit `efa7867a` (`fix(gfs): reap validation child before releasing image budget`, +67/−8 across `imageValidation.ts`, `imageValidation.test.ts` and the mcp-host version bump), audited for what the fix itself changed rather than a second full-head pass.

### Verdict

The finding from the previous review is fixed, and the fix does not open a new one. The change is small, the two new tests exercise the two behaviours that were actually wrong, and I reproduced the remaining edge on Node 24 and it settles correctly. Nothing here blocks the PR on review grounds.

### What was reported, and what changed

`runValidationChild` used to register the abort listener before `spawn` returned, and `EventTarget.addEventListener` does not fire for a signal that is already aborted, so a signal that flipped during the spawn left the child alive with the bytes on stdin and the memory reservation held until the 5s timer. The commit moves the registration to after every handler is attached and adds the missing check:

```676:684:mcp-host/src/visualInput/imageValidation.ts
    signal?.addEventListener('abort', onAbort, { once: true })
    if (signal?.aborted) {
      onAbort()
      return
    }

    // The payload is bounded before this point, so this write is capped too.
    child.stdin.end(bytes)
```

The early `return` is the right call, not just a kill: it skips `child.stdin.end(bytes)`, so an aborted request no longer pushes the image into a process that is about to be killed. `onAbort` sets `failure` and sends `SIGKILL`, and `settle` still prefers `failure` over any reply, so a decode result that lands after the abort cannot be reported as validated.

### Collateral of the fix

Two neighbouring behaviours moved with it, and both look deliberate rather than accidental.

- **`error` no longer settles on its own.** Previously `child.on('error')` rejected immediately, which released the reservation in `validateImage`'s `finally` while the child could still be running. Now it only records `failure` and the reservation is held until `close`. `close` is emitted after `error` and after any `exit`, including when the underlying spawn fails, so the promise still settles. The new test asserts the reservation is still held one microtask after the synthetic `error` and is back to zero once the rejection lands.
- **The timeout no longer overwrites an earlier cause.** The timer used to unconditionally assign `failure = timeout`, so an abort or a reply-limit failure that happened first was reported as a timeout if the kill was slow. It now assigns only when `failure` is unset. Note the asymmetry is intentional and correct: `onAbort` still overwrites, because cancellation has to outrank a timeout that already fired.

One path the new tests do not pin: the signal aborting in the window where `spawn` fails outright (binary missing, or an `EMFILE`/`ENFILE` from the descriptor ceiling). I reproduced that on Node 24 with the same control flow — `error` then `close` with code `-2`, no `exit` in between. With the signal already aborted, the promise rejects `cancelled` and the `close` handler settles it, so the reservation is released. Without the abort it rejects `invalid_response` the same way. `child.kill('SIGKILL')` returns `false` there and does not throw, so the `onAbort` call is safe. Not worth another test unless you want the spawn-failure path pinned, but it is not a leak.

### Tests

`reaps a child when the signal aborts during spawn` aborts from inside the `spawn` mock, after the real child exists and before the implementation can register anything, and asserts `SIGKILL`, `code: 'cancelled'` and `residentBytes === 0`. That is the exact interleaving from the previous review, and it fails on `7f851ed0`. `keeps the reservation until close when the child emits an error` covers the second behaviour change. The pre-existing abort-mid-flight and pre-aborted-signal tests still bracket both sides of the new one.

### CI

At `efa7867a` all 41 checks are green, including `Build / test / typecheck (mcp-host)` and CodeQL. The earlier combined `pending` state was this run still in progress.

### Still open, unchanged by this commit

The `maxQueueWaitMs` comment in `openaiProxy.ts` and `anthropicProxy.ts` still describes the old double queue wait. Pre-existing relative to this fix, and still inside the PR's diff, so it remains worth correcting but it is a comment, not a behaviour bug.

---
*Review by Evenfire Reviewer*

jozer-rami (approved):

Issue #629, opened 2026-09-14
HCC sends ~230 non-converging updates/hour to gfsc-reader, gfsc-writer-pdb, codex-cert and llmhook NetworkPolicies


alfredolopez80:
## Summary

On `clerum-dev`, HCC sends about 230 accepted `update` calls per hour to four resource
families. Two of them are no-ops, because the object's `resourceVersion` does not change. The
other two change the spec on every call, which means the reconciler never converges. These
calls are about half of all HCC writes after #618 (≈442/h). No open issue covers them: #604,
#606 and #607 are about NetworkPolicies driven by McpServer/Context and about the Secret
informer.

Found while measuring #604 PR A (#618). The rates for gfsc-reader, gfsc-writer-pdb and the
llmhook NetworkPolicies are the same with the #610 image and the #618 image, so they are not
caused by either change.

## Measurement

Source: GKE Admin Activity audit log, principal = HCC service account, `update` calls with
status OK, counted per one-hour segment.

- Base: 13 segments between 2026-09-13T20:30Z and 2026-09-14T11:30Z (image `sha-9370f05`,
  #610).
- Post: P1–P3, 2026-09-14T12:34Z–15:34Z (image `sha-98840fb`, #618).
- There were no HCC restarts inside any segment.

| resource | base, per hour (13 segments) | P1 / P2 / P3 | does the object change? |
|---|---|---|---|
| `gfs/deployments/gfsc-reader` | 71–72 | — / — / 72 | yes: `metadata.generation` 119536 → 119543 across 7 updates (15:46:48Z–15:51:58Z) |
| `gfs/poddisruptionbudgets/gfsc-writer-pdb` | 71–72 | — / — / 72 | no: `resourceVersion` unchanged across 7 updates |
| `mcp-host/deployments/codex-cert` | 55–60 | 39 / 46 / 41 | no: `resourceVersion` unchanged across 3 updates |
| `llm-hooks/networkpolicies/llmhook-8204baab199521b5` and `llmhook-ecdea6ed62db0897` | 46–48 (both together) | — / — / 46 (23 each) | yes: generation 6296 → 6298 across 2 updates |

A "—" means the value was only broken down per resource for P3.

Other data points:

- `gfsc-reader` shows Deployment revision 479 with generation 119543. The spec changes on
  every call without creating a ReplicaSet: the newest ReplicaSet was created at
  2026-09-14T11:34:56Z (the #618 rollout), and the active one dates from 2026-09-03. The
[...]

PR #659, opened 2026-09-17
feat: add grok-subscription oauth-broker sibling of Codex

jozer-rami:
## Summary

Turn Codex subscription from a one-off ChatGPT integration into the first adapter of a reusable **oauth-broker** coding-plan product, then add **`grok-subscription`** as a sibling adapter for SuperGrok / X Premium+ / Grok Build.

This does **not** overload `xai` (that stays the metered API-key provider) and does **not** send SuperGrok traffic through `codex-llm-proxy`. Control API remains the only OAuth custodian. Access tokens reach only `grok-llm-proxy`. Refresh tokens never leave Control API.

### D2 deviation (recorded)

This PR clones the Codex oauth-broker **shape** into a sibling Grok adapter rather than extracting `SubscriptionGrantStore` / `projectOauthBrokerExecution` first. Jose chose that order, and the review accepted it. Wave 1 in this branch is parameterization (admin routes, dual-key annotations, live-target attest, D13). The remaining Codex-named modules are a follow-up extract, not a merge blocker for this adapter.

Wave 1 parameterization:

- `buildProviderMaps` / `OAUTH_BROKER_IDS` in `@clerum/llm-providers`
- live-target spend attestation (D6)
- at-most-one oauth-broker id (D13: Host CEL + API; recipes including cluster CEL)
- dual-read recipe annotations (`clerum.io/codex-connection-ref` + `clerum.io/subscription-connection-ref`)
- parameterized admin/PromptBridge paths

Wave 2 Grok adapter:

- sibling grant tables (migrations 0109–0113) and Grok RFC 8628 device-code OAuth (`auth.x.ai`, client_id in allowlisted `control-api/src/config.ts`)
- `grok-llm-proxy` with its own attempt contract, LIMITS (`maxToolCalls: 64`), and exact origins (`cli-chat-proxy.grok.com/v1/responses`; `/v1/models` is the **starting candidate**, probe-gated; never `api.x.ai`)
- split redeem, `llm:grok:execute`, Host `spec.model.connectionRef` persist for oauth-broker providers
- SDK `subscriptionBinding` minted independently of Codex `codexBinding`
- Control UI Grok pickers/hub/Desktop selector, gated on the capability probe
- empty Grok stays `unassigned` (never Codex empty → `deployment-default`)
- Codex #644/#648 ports: Grok Hosts default to `direct` tool presentation; grok-llm-proxy sends `strict: false`

## Review remediation (@alfredolopez80 comments A/B/C on `627f492`)

[...]

PR #673, opened 2026-09-17
fix(hcc): skip no-op PDB replaces and treat empty env or NP slices as absent

alfredolopez80:
## Summary
- Treat empty `env.value` (`''`) and empty NetworkPolicy `ingress`/`egress` slices as equivalent to omitted fields in HCC comparison normalizers, matching apiserver omitempty persistence (#629 PR 1).
- Add `podDisruptionBudgetMatchesDesired` and wire it as the `isUpToDate` gate on `K8sGfsApi.applyPodDisruptionBudget` so a matching writer PDB skips replace. `applyDeployment` is intentionally unchanged (PR 2).
- Unblock exact-head T2 `npm test` in rpc-proxy: bind `:hostRef([^/]+)` so a decoded `*` is 400 not Express splat 404, serialize Vitest, and reset in-process stream counters between tests.

## Test plan
- [x] T1/T2/T4/T5/T6 red before the HCC change, green after, vacuity FAIL when the rule/`isUpToDate` is removed
- [x] HCC suite: 2413 passed
- [x] `MINIKUBE_T2_PASS` `T0=PASS` `T1=PASS` `T2=PASS` on `clerum-fix-629-hcc-noop-normalizers-ffa72976` at `f69cd1fe5`
- [x] `gfs/gfsc-writer-pdb` generation stayed `1` / RV `15139` across two reads ≥5 min; 0 `Kubernetes resource updated` lines for that PDB
- [ ] Reviewer: confirm `applyDeployment` / T3/T7/T8 / reader gate stay out of this PR
- [ ] Reviewer: T2 distinct-case uses `_from` rather than k8s `from` — decide if the witness should use a real ingress rule

Closes nothing yet — this is PR 1 of #629. Reader Deployment no-op PUTs remain PR 2.


alfredolopez80:
@claude please do a full, adversarial review of this PR (issue #629, PR 1 only). Do not rubber-stamp. Walk every production hunk and every new test, challenge the omitempty lemma, and say explicitly what is in scope vs leftover for PR 2.

This comment is the implementer review brief plus findings already measured on the branch-owned Minikube profile. Please confirm, refute, or deepen each item with code citations.

## Intent

HCC was issuing no-op `replace` calls because comparison treated apiserver omitempty as drift:

1. `env[].value: ''` vs the live object that omitted `value`
2. NetworkPolicy `ingress: []` / `egress: []` vs omitted slices (after `policyTypes` default-fill)
3. Writer PDB (`gfsc-writer-pdb`) with no equality gate, so every reconcile PUT the same spec

PR 1 must fix comparison + the PDB gate only. `K8sGfsApi.applyDeployment` / reader / T3 / T7 / T8 / `Resync pass completed` are PR 2.

## What landed

### HCC production

- `normalizeContainerDefaults`: `if (env.value === '') delete env.value` on both sides of `deploymentMatchesDesired`.
- `normalizeNetworkPolicyForComparison`: **after** filling empty/omitted `policyTypes`, delete `spec.ingress` / `spec.egress` when `.length === 0`.
- New `podDisruptionBudgetMatchesDesired` — same shape as `serviceMatchesDesired` (`stripServerOwnedMetadata` + `canonicalizeValue`, fail-open-to-write, **no** spec defaults).
- `applyPodDisruptionBudget` now passes `mergeExisting: preserveObjectAnnotations` and `isUpToDate: podDisruptionBudgetMatchesDesired`.
- `applyDeployment` is unchanged (still no `isUpToDate`). That is required.

### Tests (existing files only)

| ID | File | Equal (must skip / match) | Distinct (must write / not match) |
| --- | --- | --- | --- |
| T1 | `deploymentMatchesDesired.test.ts` | `value: ''` vs omitted → `true` | `value: 'x'` vs omitted → `false` |
| T2 | `networkPolicyMatchesDesired.test.ts` | empty ingress vs omitted; empty egress + derived policyTypes vs omitted | ingress with a rule vs omitted → `false` |
| T4 | `gfsK8sApi.readFirst.test.ts` | live PDB == desired → 0 `replace`, `writeSkipsTotal` +1, GET 1× | `minAvailable` drift → 1 replace; 409 retry kept on a **drifted** live spec |
| T5 | `hostReconciler.readFirst.test.ts` | Host **without** `secretRef`, desired `value: ''`, live omits `value` → 0 Deployment replace, GET 1× | same Host, different image → 1 replace |
| T6 | `llmHookNetworkPolicyNoopGate.test.ts` | hook, no Hosts, desired `ingress: []`, live omits `ingress` → 0 NP replace, GET 1× | add referencing Host → 1 replace |

CMP-NP-13 was rewritten to agree with T2 (`egress: []` ≡ omitted). PDB `describe.each` is now `updates: false` like Service; 409 coverage moved to T4 drift specs.

Fixtures under `host-context-controller/src/__tests__/fixtures/629/` came from clerum-dev RO (PDB generation 1 / writer selector; NP spec is `podSelector` + `policyTypes: [Ingress]` with **no** `ingress`; `codex-cert` `CLERUM_LLM_SECRET_REF` has no `value`).

### rpc-proxy (T2 pre-gate, not #629 design)

Pre-gate `npm test` was flake/red under file parallelism:

- Express decoded `*` as a splat → 404 instead of 400. Route is now `/rpc/hosts/:hostRef([^/]+)/status/stream`.
- Module-level stream maps leaked across files. `resetHostStatusStreamRuntimeForTests()` + `beforeEach` reset; `"test": "vitest run --no-file-parallelism"`.
- `model-selection.test.ts` `afterEach` uses `mockReset()` so hoisted `vi.mock` survives (was `restoreAllMocks()`).
- Version `0.1.77` → `0.1.78` and `external-rest-api` `releaseManifest.rpcProxyVersion`.

Please treat this as a **necessary T2 unblock**, and say whether it should have been a follow-up PR. Do not let it hide HCC review.

## Local evidence (already run)

- Red before change: `.local-notes/implementacion-629-pr1-2026-09-17/01-tests-red-before.log` EXIT=1 (6 failed including T1/T2/T4/T5/T6).
- Green after: `02-tests-green-after.log` EXIT=0.
- Vacuity (rule / `isUpToDate` removed, then restored): `03` T1/T5, `04` T2/T6, `05` T4 — each EXIT=1 on the equal-case.
- HCC suite: `06-hcc-full-suite.log` — 110 files, **2413** passed, EXIT=0. Build EXIT=0.
- Canonical `make minikube-t2` on `clerum-fix-629-hcc-noop-normalizers-ffa72976`: `MINIKUBE_T2_PASS` `T0=PASS` `T1=PASS` `T2=PASS` `NP08_HCC_AUTHORIZATION=PASS` EXIT=0 at HEAD `f69cd1fe5`. Log `23-minikube-t2.log`.
- Extra PR1 cluster check: `gfsc-writer-pdb` generation **1** / RV **15139** at 20:47:30Z and 20:54:00Z. **0** `Kubernetes resource updated` lines for that PDB or any `llmhook-*`.
- This Minikube profile has **no** `llmhook-*` NetworkPolicies and one Host (`chatllm`) **with** secretRef. Live NP / no-secretRef Host generation was **not** observable here.
- `gfsc-reader` Deployment still gets `Kubernetes resource updated` about every 60s. That is the PR 2 bug. Do not “fix” it in this PR.

## Please verify in code (must-pass checklist)

1. **Safe-strip lemma.** `delete env.value` iff `value === ''`, applied identically to desired and live. Confirm it cannot hide `value: 'x'` vs omitted (T1 distinct). Confirm it does not delete `valueFrom`-only env entries.
2. **NP delete order.** Empty-slice delete must stay **after** `policyTypes` fill. If those two blocks are swapped, `egress: []` with omitted `policyTypes` would fill as Ingress-only for a different reason or fail T2 derived-policyTypes. Confirm TCP protocol strip still runs on remaining rules.
3. **Deny-all semantics.** With `policyTypes: ['Ingress']`, omitted `ingress` and `ingress: []` are the same persistable isolation (deny-all). Confirm we are not collapsing a real rule list to empty.
4. **PDB matcher.** No invented spec defaults (`unhealthyPodEvictionPolicy`, etc.). Fail-open on missing spec / thrown canonicalize. Labels / selector / `minAvailable` remain exact. `preserveObjectAnnotations` runs **before** `isUpToDate(next, existing)` so a foreign annotation is absorbed, not treated as permanent drift. A foreign **label** is still drift (same as Service/ConfigMap) — call that out if you disagree.
5. **Gate wiring.** `isUpToDate` is only on `applyPodDisruptionBudget`. `applyDeployment` must still PUT when the live object equals desired. T4 409 tests must use a **drifted** live spec (`minAvailable: 2`); if they used an equal live object the retry would skip and the test would lie.
6. **T5 Host.** `secretRef` omitted (not `'host-secret'`). Desired builder still emits `CLERUM_LLM_SECRET_REF` with `value: ''`. Live deletes `value` only. Image drift still writes. Witness is GET 1× / replace 0.
7. **T6.** Must **not** build live via `asApiserverNetworkPolicy(desired)` (that keeps `ingress: []` on both sides — the old NOOP-LLMNP-1 vacuity). Live must `delete live.spec.ingress`. Adding a Host must produce `ingress.length > 0` and one replace.
8. **T2 distinct witness.** The “real rule” object uses `_from`, which is **not** the Kubernetes `from` field. The matcher still returns false (extra key vs omitted ingress), so the test is green, but it does **not** prove a real peer rule. I want you to say whether this is a must-fix in PR 1 or an acceptable witness.
9. **No production logs added** in HCC. No `console.*` in HCC production. rpc-proxy `console.warn` on the stream path is pre-existing.
10. **Scope leak.** Confirm no edits to `hostReconciler` constructor around the old line 2887, `reconciler.ts` ~1305, GFS reader applyDeployment, or A1.3 harness strings.

## Findings I already have (agree / disagree with evidence)

### Not blocking (but say so)

- First HCC commit message says the tests land “before the production normalizers land”, but `utils.ts` / `gfsK8sApi.ts` are in the **same** commit. Message is slightly wrong; content is complete.
- HCC `npm run lint` has no ESLint v9 config (`LINT_EXIT=2`). Pre-existing package state, not introduced here.
- Mixing rpc-proxy + HCC in one PR couples two stories. Local T2 could not close without the rpc-proxy commit. Prefer keeping it unless you see a revert risk.
- Live `llmhook-*` generation and Hosts without `secretRef` were **absent** on the certifying Minikube profile. Equality is proven in unit/integration tests and by writer PDB generation stability, not by a live llmhook object.

### Worth a comment, probably not a rewrite

- T2 `_from` vs `from` (item 8).
- T5/T6 skip path asserts GET + no replace, not `writeSkipsTotal`. The plan allowed either. T4 asserts both.

### Must stay out

- Do not add an `isUpToDate` on `applyDeployment` in this PR.
- Do not “stabilize” `gfsc-reader` here.

## How I want the review written

Reply as a GitHub review (not just “LGTM”):

- Verdict: approve / comment / request-changes, with a one-line reason.
- File-by-file notes on `utils.ts`, `gfsK8sApi.ts`, the five test files, fixtures, and the rpc-proxy route/test/package.json/releaseManifest hunks.
- Any hidden write still possible for writer PDB or empty-ingress llmhook NPs.
- Any test that would stay green if the production rule were deleted (vacuity hole).
- Explicit list of PR 2 leftovers.

HEAD under review: `f69cd1fe58cf103e3207f7d086566ddc24e46f09` on `fix/629-hcc-noop-normalizers` → `dev`. Base `origin/dev` = `d6d98b09394d4afcef8e22382438d18bd95241cc` (0 behind, 2 ahead).

alfredolopez80 (commented):
Inline notes withdrawn. The only Claude request on this PR is the single consolidated issue comment.

zach88:
Took a deep pass through this one. The core #629 work in host-context-controller is solid — the PDB no-op gate, the empty-slice/empty-value normalizers are symmetric across desired/existing, match the apiserver's omitempty semantics, and are backed by tests that actually go red against the merge-base (reverted utils.ts + gfsK8sApi.ts and 11 tests fell over, so the guards are real). No correctness or security issues. Everything open is cleanup, and it's all concentrated in the rpc-proxy test-infra half.

🟡 **Approve with nits**

_This is my read from the review, not a formal GH approval._

Nothing blocking. Merge whenever.

## Optional (non-blocking)

The PR is mergeable without any of these — follow-up or your call.

1. **[R1-M1] `--no-file-parallelism` patches the symptom, not the cause** (`rpc-proxy/package.json:11`). This flag is the only load-bearing piece unblocking the suite, but it hides the root cause. Repro: with file parallelism on, the suite fails ~1 in 3 runs and the failure lands in `src/routes/sandboxUi.test.ts` — a file this PR doesn't even touch. Root cause is cross-file global state: several tests assign `globalThis.fetch = fetchMock` and never restore it (`src/__tests__/model-selection.test.ts:70,94,111,157,188`), so the poisoned fetch leaks into another file when they run concurrently. The real fix is save/restore `globalThis.fetch` in `afterEach` (or `vi.stubGlobal` + `vi.unstubAllGlobals`); then the flag is a safety net rather than the fix. Fine to keep as a stopgap if the debt gets tracked.

2. **[R1-L1] The `:hostRef([^/]+)` route regex is a functional no-op and its comment is wrong** (`rpc-proxy/src/routes/rpcHostStatusStream.ts:69-71`). Repro: checked against real Express — flat route vs regex return identical results for `%2A`→`*`, `%20`, `%2F`→`a/b` and normal names; and the `rejects host status stream for wildcard hostRef` test passes with the flat route. The 400 for `*` comes from the handler guard `isWildcardOrInvalidHostRef` (`/[*%]/`), which already existed at merge-base. The comment ("so a decoded `*` is a hostRef (400), not an Express splat (404)") claims routing behavior that doesn't happen here, and it's 1 of ~20 `:hostRef` routes carrying a regex while the sibling streams (`rpcHostActivityStream`, `rpcHostProgressStream`) rely on the handler guard alone. Suggest dropping the regex + comment; if there's a real app-level 404, add a test that reproduces it first.

3. **[R1-L2] Sibling stream route has the same module state but no reset helper** (`rpc-proxy/src/routes/rpcHostActivityStream.ts:11-14`). `rpcHostActivityStream` carries identical module-level state (`activeStreamCountsByUser`, `activeStreamCountsByUserHost`, `activeStreams`, `streamCounter`) and doesn't export a reset helper — the pattern only got applied to status-stream. Green today, but it's the sibling instance of the same isolation gap. Consider a shared counters module with one reset.

4. **[R1-L3] Misleading `_from`/ObjectSerializer comment in the T2 test** (`host-context-controller/src/__tests__/networkPolicyMatchesDesired.test.ts`, "T2: a real ingress rule…"). The comment says `ObjectSerializer maps _from to wire from`, but `networkPolicyMatchesDesired` doesn't serialize (just `structuredClone` + `canonicalizeValue` + `JSON.stringify`), so `_from` survives as-is. Repro: the `toBe(false)` assertion passes identically with garbage (`ingress:[{zzz:true}]`) or with the real wire key `from` — the witness is really "non-empty ingress array ≠ omitted", not "a real ingress rule". Still a valid guard for the empty-slice deletion, but the name/comment oversell it (this is the open point your body flags). Suggest fixing the comment; if you want to test a real rule, compare two NPs with distinct `from` rules so content matters.

5. **[R1-L4] Hand-trimmed PDB fixture doesn't exercise the status stripping** (`host-context-controller/src/__tests__/fixtures/629/gfsc-writer-pdb.json` + `k8s/gfsK8sApi.readFirst.test.ts` T4 skip). A real apiserver PDB read carries `status`, `resourceVersion`, `creationTimestamp`, `uid`, `managedFields`; all are missing here. The T4 skip test uses that trimmed fixture as both desired and live (only adds `resourceVersion`), so the `status`/`creationTimestamp`/`managedFields` stripping paths never run — the "real fixture" doesn't prove the gate survives an actual apiserver response. Stripping is shared and covered by other normalizers and the gate logic is correct, so it's low. Suggest deriving the fixture from an unredacted real read (or re-injecting `status` in the test's live).

## What checked out clean

- **Security: SAFE, zero findings.** The route change doesn't open an auth/scope bypass (`requireRpcAuth` + `requireScope('host:status:read')` run after match, before handler; ordering actually tightens since `*` used to 404 before auth and now hits auth first). The no-op gate doesn't mask NP/PDB drift (`[]`↔omitted are identical deny-all, `policyTypes` compared directly from pre-deletion egress length; PDB fails open on doubt). `resetHostStatusStreamRuntimeForTests` is test-only and unreachable in prod. Fixtures carry no real secrets. No new runtime deps.
- **HCC: 111 tests + `tsc --noEmit` clean.** T3 confirmed (11 tests go red at merge-base). PDB gate produces exactly the fixture shape, policy/v1 doesn't default-fill the spec so no loop/masking. `env.value===''` lives in the shared `normalizeContainerDefaults` so it covers every `deploymentMatchesDesired` consumer, and `value:''` is omitempty in `V1EnvVar` so the apiserver never emits it. Only PDB writer is `applyPodDisruptionBudget`, already gated — no ungated sibling.
- **rpc-proxy:** the reset helper covers all mutable module state; the `restoreAllMocks()`→`mockReset()` swap is neutral (no `vi.spyOn` in that file). Serialized suite: 37 files / 516 tests passing, deterministic across 3 runs.

One process note, your call: the PR bundles two unrelated services — the #629 core in host-context-controller and an independent rpc-proxy test-infra unblock. Splitting them would leave the #629 PR at a single service. Not blocking, just flagging.

alfredolopez80:
Synced onto current `origin/dev` (`32f7d0dca`, #659) and addressed the optional nits that survived an opposing pass:

- **R1-M1:** restored `globalThis.fetch` after the leaky files (`model-selection` plus the other raw `fetch` assignments). Dropped `--no-file-parallelism`. Suite is 37/516 green with file parallelism on.
- **R1-L1:** dropped the no-op `:hostRef([^/]+)` regex. Wildcard 400 stays on `isWildcardOrInvalidHostRef`.
- **R1-L2 (cheap half):** exported `resetHostActivityStreamRuntimeForTests` and call it from the activity-stream suite. Did **not** extract a shared counters module — no same-file flake, and Vitest isolate already contains cross-file module state.
- **R1-L3:** comment now names occupancy as the witness; added a distinct-`from`-rules case.
- **R1-L4:** T4 skip live object now carries apiserver `status` / `uid` / `creationTimestamp` / `managedFields` so stripping runs.

No follow-up issue for the shared-module idea — it does not earn a ticket on current evidence.

alfredolopez80:
@claude please do a **complete re-audit** of this PR at the current HEAD. Do not reuse the previous review as a rubber stamp. Walk the full diff against `dev` again: every production hunk, every test, and the post-zach follow-up. Challenge the same omitempty / PDB-gate lemmas as before, then independently confirm or refute the R1 nits that landed after your last pass.

This is still **issue #629, PR 1 only**. `applyDeployment` / gfsc-reader / T3 / T7 / T8 / `Resync pass completed` stay out.

## Why a new pass

HEAD moved since your last review:

- Merged current `origin/dev` (`32f7d0dca`, #659 grok-subscription). HCC is now `0.1.335`.
- Addressed zach88's optional nits that survived an opposing pass (`3781e2ca9`).
- The earlier rpc-proxy diagnosis (splat 404 / `--no-file-parallelism` as the fix) was wrong and has been corrected. Do not defend the old story.

HEAD under review: `3781e2ca945a8e9e086dc1a163586c0a4638371f` on `fix/629-hcc-noop-normalizers` → `dev`.

## What changed since the last Claude pass (must re-check)

### R1-M1 (was medium)

- `rpc-proxy` `"test"` is `vitest run` again — **no** `--no-file-parallelism`.
- `globalThis.fetch` is saved/restored in `model-selection.test.ts`, `sessions-passthrough.test.ts`, `session-rename-passthrough.test.ts`, and `sender-assignment.test.ts`.
- Local evidence: 37 files / 516 tests green with file parallelism on.

Please confirm the leak is actually closed (including `vi.restoreAllMocks()` files that previously left a poisoned `fetch` for `sandboxUi.test.ts` worker reuse). If another file still assigns `globalThis.fetch` without restore, name it.

### R1-L1

- Status stream route is flat `/rpc/hosts/:hostRef/status/stream`.
- `:hostRef([^/]+)` is gone. Wildcard 400 remains `isWildcardOrInvalidHostRef`.
- Confirm Express 4 match set is unchanged and auth/scope still run after match.

### R1-L2 (cheap half only)

- `resetHostActivityStreamRuntimeForTests` exists and is called from the activity-stream suite.
- **Intentionally not done:** a shared counters module for status+activity. Call that out if you think it is now a must-fix; otherwise agree it is out of PR 1.

### R1-L3

- T2 occupancy test no longer claims ObjectSerializer / `_from`.
- Wire key is `from`. There is a second case: distinct `from` rules must not match.

### R1-L4

- T4 skip live PDB now injects apiserver `status`, `uid`, `creationTimestamp`, `managedFields` so `stripServerOwnedMetadata` actually runs.
- Fixture file itself is still trimmed; say whether live injection is enough.

## Still must-pass (same as first brief)

1. Safe-strip lemma on `env.value === ''` — both sides, cannot hide `value: 'x'` vs omitted, does not delete `valueFrom`-only entries.
2. NP empty-slice delete stays **after** `policyTypes` fill. Deny-all `[]` ≡ omitted. Real rules must not collapse.
3. PDB matcher: no invented spec defaults; fail-open; `preserveObjectAnnotations` before `isUpToDate`; foreign **label** is still drift.
4. `isUpToDate` only on `applyPodDisruptionBudget`. `applyDeployment` still has none.
5. T4 409 path still uses a **drifted** live spec.
6. T5 Host omits `secretRef`; desired `value: ''`; live omits `value`; image drift still writes.
7. T6 live must not be `asApiserverNetworkPolicy(desired)`.
8. No new HCC production `console.*`.
9. No scope leak into hostReconciler constructor / GFS reader `applyDeployment` / A1.3 harness strings.

## How to write this review

Reply as a GitHub review (not just LGTM):

- Verdict: approve / comment / request-changes, one-line reason.
- File-by-file on `utils.ts`, `gfsK8sApi.ts`, the five test files, fixtures, and every rpc-proxy hunk still in the PR (route, resets, fetch restore, `package.json` test script, versions, `releaseManifest`).
- Explicitly list anything still green if the production rule were deleted (vacuity).
- Explicit PR 2 leftovers.
- Separate “still open after zach follow-up” from “fixed”.

Do not approve the old splat-404 / `--no-file-parallelism` story. That review comment is stale.

zach88:
Took another pass through this now that round 1's findings have fixes on top. Good news: everything from round 1 landed cleanly and the collateral check came back empty — no fix broke anything adjacent. The only thing left is a stale line in the PR description; the code itself is in good shape.

**🟡 Approve with nits**
- Optional / non-blocking: **1**

All five round-1 items are closed and there were zero regressions from the fixes. This is my read from the review, not a formal GH approval.

## Optional (non-blocking)
This can merge as-is; the item below is cleanup / author's call.

1. **[R2-L1] PR body's 3rd bullet no longer matches the code.** The Summary still says *"bind `:hostRef([^/]+)` so a decoded `*` is 400 not Express splat 404, serialize Vitest…"*, but round 2 reverted both of those:
   - The regex got dropped in `3781e2ca` — the route is plain `/rpc/hosts/:hostRef/status/stream` again (identical to dev), and the wildcard 400 comes from the `isWildcardOrInvalidHostRef` guard (`/[*%]/`) in `rpcHostStatusStream.ts`, not from routing.
   - Vitest isn't serialized anymore — `rpc-proxy/package.json:11` is back to `vitest run` (the `--no-file-parallelism` flag is gone). The real unblock is the per-file `globalThis.fetch` isolation.
   Suggest rewording that bullet to describe what actually shipped (plain route + existing guard, fetch-mock isolation per file). Not blocking, but the description is part of what gets reviewed so worth a 30-second edit.

A couple of trivial residuals from the round-1 closures, if you're in there anyway (neither reopens its finding):
- `networkPolicyMatchesDesired.test.ts` — the `_from` key is now correct, but the comment justifies it as a "reserved word"; it's actually just the OpenAPI generator convention.
- `fixtures/629/gfsc-writer-pdb.json` — the T4 PDB test compares fixture-vs-fixture and never calls `buildPodDisruptionBudget`, so a future builder field wouldn't be caught the way T5/T6 catch it (they use the real builders). Could build `desired` from the builder.

## What checked out clean
- **Round-1 findings all closed:** flag removed + `globalThis.fetch` isolated across 6 files (`3781e2ca`), sibling reset helper added covering all module state (`f5e60702`), `_from` fixture corrected (`892a3b36`), PDB fixture now exercises the real stripping path.
- **No regressions:** rpc-proxy suite green across 3 parallel runs + 4 `--sequence.shuffle` runs; touched files deterministic in isolation; wildcard/empty hostRef still return 400 with auth running before the handler.
- **HCC correctness:** 2426 tests pass. T3 verified — reverting only `utils.ts` + `gfsK8sApi.ts` to base makes exactly 8 regression tests fail (CMP-NP-13, empty ingress/egress, env value, Host/hook skip) and they pass again on head. Empty NP slices don't mask a differing `policyTypes` (defaulted before the slice deletion), and the PDB gate is fail-open on any spec/label diff so it can't mask drift.
- **Security:** no findings introduced here. The no-op NP/PDB gates still replace on any weakening of a managed resource, `env.value===''` normalization doesn't equate real values or `valueFrom` secret refs, and the reset helpers are unreachable in prod (env guard + no HTTP path).

zach88 (approved):

PR #676, opened 2026-09-18
feat(desktop): universal workspace tab strip with drawer and tab reorder

zach88:
## What

Introduces the **universal workspace tab strip** for the desktop app: a single global tab strip that owns chat, app, files, and settings tabs, replacing the old chat-only tabs and the `backToConversation` navigation mechanism. On top of that foundation it adds several UX refinements: app tabs that persist their in-app route and name themselves after the embed's live `document.title`, multi-instance files tabs deduped by live path, a shared right-rail shell hosting a universal chat drawer, a wider/resizable chat drawer, and drag/keyboard reordering of tabs.

## Why

The app previously had a chat-only tab model plus a separate "back to conversation" affordance, so non-chat surfaces (apps, files, settings) were not first-class tabs and could not coexist or be navigated uniformly. The universal store makes every surface a peer tab reachable from any other, which is the prerequisite for the drawer, route persistence, and reordering to behave consistently across kinds.

## How

- **Pure tab store.** `workspaceTabs.ts` is a pure algebra `(state, input) => state` owning tab identity and the active tab. It lifts out the old "never empty / re-seed a blank chat" invariant: the workspace may be empty and `activeTabId` is nullable. Rendering, route selection, and the drawer all derive from this store; controllers keep chat loading.
- **Route derivation seam.** The active tab drives `navItem`; operations that must not steal focus (drawer chat selection, tab reorder) leave `activeTabId` untouched so the route does not churn.
- **App tabs.** Persist the live in-app route across deactivation so reactivation re-mounts at it, and adopt the embed's `document.title` as the tab label (empty titles ignored).
- **Files tabs.** Multi-instance, deduped by live `gfsUri` at open time; the browser writes its leaf path back onto the tab as it navigates.
- **Chat drawer.** Lives on a shared right-rail shell, always docked (a native `WebContentsView` cannot be overlaid). Session-only width, resizable by drag and keyboard, default 525px and drag ceiling 1000px, with a dynamic cap that reserves a legible floor for the embed.
- **Tab reorder.** Native HTML5 drag-and-drop plus Alt+ArrowLeft/Right on the focused tab. Order is the array order of the store; `reorderWorkspaceTab` is a pure permutation that clamps the target index, preserves `activeTabId`, the id-set, and length, and returns the same reference on no-ops.

## Testing

- Reducer and strip suites for the recent additions run green locally: `vitest run workspaceTabs WorkspaceTabStrip` and `vitest run useChatDrawerResize ChatDrawer` (drawer width + reorder), including `fast-check` property tests for the reorder permutation (id-set/length/`activeTabId` preserved, target lands at `clamp(toIndex)`, no-op returns same ref, double-apply idempotent).
- The branch ships focused suites per area: navigation controller, app-controller workspace tabs, right-rail shell, files-page seed location, sandbox app-tab route persistence, and chat selection across route changes.
- Code review and security review were run on the width and reorder diffs (no blockers; security clean).

## Risks / Follow-ups

- **PR size (D1).** 58 files in a single service (`desktop-app`). It exceeds the soft budget but is one cohesive spec: the universal tab store is the shared foundation and the later commits (route persistence, drawer, files, reorder) build directly on it, so the 13 commits form an interdependent stack that would not compile if split by sub-feature after the fact. Kept as one PR for that reason; reviewers can follow it commit by commit.
- **Pre-existing renderer type debt.** `tsc --noEmit -p ui/tsconfig.json` reports ~30 type errors in unrelated renderer test files (e.g. `App.tsx` deep-link region, FilesPage/SettingsPage tests). These reproduce on the base and are not introduced here; the files this branch touches are clean. Candidate for a separate issue to add a real renderer typecheck gate.
- Tab order and drawer width are session-only by design (reset on relaunch), matching prior behavior.


josue-tech-94:
Round 1 review complete for `8d9d30aa2ea5dd731477d3af5ba1ee9e4dd001bb`.

**Decision: CHANGES REQUIRED**

There is one must-fix finding:

### R1-H1 — Stale sandbox close after app-tab reactivation

`desktop-app/ui/src/App.tsx:1158-1170` starts an async `sandboxUi.getLocation()` during app deactivation and later calls the singleton `sandboxUi.close()` without confirming that the original transition still owns the active embed.

Reproduced sequence:

1. Open App A
2. Switch App A → chat while delaying `getLocation()`
3. Reactivate App A
4. Resolve the old `getLocation()` request

The stale continuation can then overwrite the route and close the newly reopened native sandbox view.

**Required fix:** add lifecycle generation/view-identity ownership across the async route read and close. The stale continuation must not persist a route or close a view belonging to a newer activation. Preserve/check the returned `appRef` as part of that ownership validation.

**Required regression test:** cover both:

* App A → chat → App A before the old location request resolves
* App A → chat → App B before the old request resolves

Assert that the stale transition neither closes the current view nor overwrites its route.

Everything else reviewed in this round looked good, including workspace-tab/reorder invariants, files-tab dedup/path handling, drawer resize/accessibility behavior, title propagation, IPC/preload trust boundaries, and removal of `backToConversation`. Security review found no change-caused issue.

The focused suites, production builds, and full desktop suite passed. The existing renderer TypeScript error family reproduces at the merge base and remains out of scope for this PR.

Once `R1-H1` is fixed, we can review the corrective range against this reviewed head.

josue-tech-94:
Round 2 review complete for `2f75f6751dd73c2af1331c512e1031f3a1963d4b`.

**Decision: CHANGES REQUIRED**

Good news first: `R1-H1` is closed by `a0bc688b8`. The lifecycle-generation guard correctly prevents stale app-deactivation continuations from persisting an old route or closing a newly activated sandbox view. The regression tests also discriminate the previous reviewed head as expected.

There is one new must-fix item:

### R2-M1 — Distinct-app regression test does not actually use a distinct app

The test intended to cover **App A → chat → App B** currently launches another tab for the same `appRef` (`ns/app`). It therefore does not prove cross-app lifecycle isolation and is false-green for that requirement.

**Required fix:**

* Add a real second app fixture, e.g. `READY_APP_B`, with a distinct `appRef`.
* Exercise App A → chat → App B while the old `getLocation()` request is still pending.
* Assert App B remains the live tab/view and the stale App A continuation neither closes it nor overwrites its route.
* Run the test against both the current head and the previous reviewed head to preserve the T3 proof.

A scratch version using a genuinely distinct App B already passes on the current head and fails on the previous reviewed head, so the production fix itself appears correct; the committed regression coverage is what still needs correction.

found **no production regression from the R1-H1 corrective work**. Correctness and security otherwise look clean, and the focused suites, production builds, style rules, and full desktop suite all passed.

One process caution: `App.tsx` and `useNavigationController.ts` have each reached the F3 threshold of three `fix:` commits. No fourth corrective commit was observed, but further corrective patching in either file should not proceed without the required mini-spec/design step.

alfredolopez80:
@zach88 — independent multi-agent review of this PR at head `65fc25ec5`

**Repo:** evenfire-ai/evenfire · **PR:** #676 · **Title:** feat(desktop): universal workspace tab strip with drawer and tab reorder · **Base ← head:** `dev` ← `feat/universal-tabs` · **Author:** @zach88

**Findings by severity:** Blocker 0 · Critical 0 · High 2 · Medium 6 · Low 7

We ran five reviewers in parallel against a checkout of the PR head: the tab store and navigation; `App.tsx` and the UI components; the pages and the IPC contract; security; and tests. Every finding below was checked against the code. "Reproduced" means we confirmed it with a throwaway test or a mutation; "read" means we confirmed it by reading the code only.

## Test and typecheck status (Node v24.18.0, `npm ci` without `--ignore-scripts`)

| Command | Exit | Result |
|---|---|---|
| `npm run verify:electron` | 0 | Electron executable resolved |
| `npx vitest run` (main + ui) | 0 | 244 files, **2394 passed**, 0 failed, 0 pending |
| `tsc -p tsconfig.json --noEmit` (main) | 0 | 0 errors |
| `tsc -p ui/tsconfig.json --noEmit` (renderer) | 2 | Same pre-existing error set as the merge base `32f7d0dca`. The PR adds no errors and fixes one (`App.chatDrawer.test.tsx` TS2339) |
| lint | — | `desktop-app` has no lint script |

## Status of the earlier review rounds (@josue-tech-94)

- **R1-H1 (stale sandbox close): partly closed.** The lifecycle-generation guard works: removing it makes both regression tests fail. The round-1 requirement also said "Preserve/check the returned `appRef` as part of that ownership validation", and that part is not implemented. See M1.
- **R2-M1 (distinct-app regression test): closed.** `App.deepLinks.test.tsx:2003` now uses `READY_APP_B` (`ns/app-b`) and keeps B live. Moving the guard below the persist makes the route assertions fail.
- **F3 caution (three `fix:` commits in `App.tsx` / `useNavigationController.ts`):** noted. Several items below touch those files, so please follow the mini-spec step before another corrective commit.

## High

**H1. The deep-link handoff guard has no test.** `desktop-app/ui/src/App.tsx` ~1322/1326 (`closeActiveSandboxUiEmbedForHandoff`). *Reproduced.* We removed only the two `gen` checks in the handoff path, and `App.deepLinks` + `App.chatDrawer` still pass (71/71). The commit message says both sites are gated, but only the deactivation site is tested. If a deep-link handoff is superseded while `getLocation()` is in flight, the stale handoff can close the newer embed and no test fails. Fix: add a handoff variant of the stale-continuation test.

**H2. Closing the active tab loses keyboard focus.** `components/WorkspaceTabStrip/index.tsx:137-146`, wired from `App.tsx:472-485` (`handleCloseWorkspaceTab`). *Reproduced.* When a keyboard user closes the active tab with Enter or Space on `×`, `document.activeElement` becomes `document.body`, and the user has to Tab again from the top of the page. Keyboard reorder keeps focus because the keyed node survives; close has no equivalent and no test. Fix: focus the newly active tab (or the strip) after close.

## Medium

**M1. The saved route is not checked against its `appRef`.** `App.tsx` ~1148 (deactivation effect) and ~1317 (handoff) store `location?.routePath` on the outgoing tab and ignore `location.appRef`. *Read.* In an app → app switch, whether the route lands on the right tab depends only on the IPC ordering described in the comment (the `getLocation` read reaches main before the incoming `open()` passes its rAF gate). `getLocation` is not serialized through `enqueueSandboxUiLifecycle`. If that ordering changes, app B's route is saved on app A's tab. The route still stays inside A's `/view` prefix, so this is a correctness problem, not a sandbox escape. Fix: persist only when `location.appRef === outgoingTab.app.appRef`. This is also the missing half of R1-H1.

**M2. Embed-controlled tab titles are not sanitized.** `lib/workspaceTabs.ts:201-216` (`setAppTabTitle`); the path is `sandboxUiDriver.ts:486` → `ipc.ts:1747` → `App.tsx:1358`. *Reproduced.* The only processing is `trim()`. A 5020-character title containing U+202E (RLO), zero-width characters and BEL was stored unchanged. A plugin can name its tab "Settings" and, because app, files and settings tabs render the same way, pass it off as host UI to phish for credentials. Bidi overrides also affect the `Close ${title}` aria-label. React renders titles as text, so this is not XSS. Fix: strip Cc/Cf/bidi control characters, cap the length (for example 80), and show an app-kind marker on app tabs.

**M3. The "Chats" nav resumes the last chat by array position, not the last one active.** `hooks/domain/useNavigationController.ts:83-102` (`focusChatSection`). *Reproduced with `renderHook`.* It picks `[...tabs].reverse().find(kind === 'chat')`, but the same file already tracks `lastActiveChatTabId` by recency for the drawer. Scenario: open chat A, open chat B, activate A (or reorder B before A), go to Files, click Chats: you land on B (or A after the reorder), not the chat you left. Fix: prefer `lastActiveChatTabId` and fall back to the array scan only when it is stale.

**M4. The stale-continuation tests have no liveness witness.** `App.deepLinks.test.tsx:1999`, `:2031`. *Reproduced.* We disabled the deactivation read-then-close entirely, and both tests still pass: their assertions are all negative (`closeSandboxUi` not called, route not stale). Fix: add `expect(getSandboxUiLocation).toHaveBeenCalledTimes(1)` and check that the deferred read was consumed. Test 1 should also assert that A is live again.

**M5. The new main-process IPC surface has no tests.** `ipc.ts` (`sandboxUi:getLocation`, `sandboxUi:titleChanged`) and the `page-title-updated` listener in `sandboxUiDriver.ts:483-497`, including its `active?.view === view` guard. *Read.* The only driver test added covers `resolveSandboxUiSharePath`. Nothing would catch a regression that drops the guard and lets a replaced view relabel the new active tab.

**M6. `WorkspaceTab` is not a discriminated union.** `lib/workspaceTabs.types.ts:35-43`. *Read.* `kind` plus four optional payloads lets the compiler accept `{ kind: 'chat' }` with no `chat` payload, or a `chat` tab that carries an `app` payload. The invariant is enforced only in the constructors, so every accessor uses optional chaining, and a malformed tab becomes a silent `null`. For example, `focusChatSection` returns `null` for a chat tab that lacks `chat`. Fix: `{kind:'chat'; chat: …} | {kind:'app'; app: …} | …`.

## Low

- **L1.** `WorkspaceTabStrip` uses `role="toolbar"` with `aria-pressed` buttons instead of the ARIA tabs pattern (`tablist`/`tab`/`aria-selected`, arrow-key focus movement). `ChatSwitcher` in this same PR uses `listbox`/`option` correctly. Follow-up.
- **L2.** The `getLocation()`/`close()` failures in `App.tsx` ~1149-1162 and ~1318-1320 are swallowed without logging. Falling back to the default route is intended, but an IPC/driver failure looks the same as a normal out-of-prefix read. Add a structured log that omits the URL.
- **L3.** `page-title-updated` is forwarded on every event (`sandboxUiDriver.ts:486-491`). An embed that changes its title in a loop causes an IPC message and a renderer re-render per change. Debounce in main (for example 250 ms).
- **L4.** The comment at `App.tsx:383` still refers to "SandboxUiPage unmount cleanup", which this PR removes. The comment at `App.tsx:1100` describes the current behavior correctly.
- **L5.** `FilesPage` still declares and calls `onPendingGfsUriHandled` (`FilesPage.tsx:214,620`, `FilesPage.types.ts:22`), but `App.tsx` no longer passes it. This predates the PR; clean up in a follow-up.
- **L6.** More negative assertions without witnesses: `App.deepLinks.test.tsx:1863` (never checks that `selectTab(0)` switched tabs) and `WorkspaceTabStrip.test.tsx:383,418` (`onReorder` not called, with no positive check).
- **L7.** The property-test generator `workspaceTabs.test.ts:568` (`openOpArb`) only produces opens. It never produces `closeWorkspaceTab` or reconcile operations, always opens files at root, and never reuses ids, so those cases are covered only by the example tests.

## Verified clean

- Store algebra: `reorderWorkspaceTab` (pure permutation, clamped, same reference on no-op), `closeWorkspaceTab` (right-neighbor / clamp-left / empty workspace), files dedup by live `gfsUri`, blank-title rejection, and an exhaustive `mapKindToRoute`.
- Deactivation-effect generation guard: every `await` is re-checked. The title-change effect checks both `liveSandboxUiTabIdRef` and `appRef`.
- `useChatDrawerResize`: finite clamps (no `NaN` at panel width 0), idempotent drag teardown, cleanup on `active` flip and on unmount.
- IPC contract driver → appService → ipc → preload → `renderer.d.ts`: types and the optional `routePath` match at every hop. `assertTrustedSender` is present on the new channel, and listener unsubscription is wired.
- A persisted route cannot leave the app's sandbox: `normalizeSandboxUiRoute` rejects `//`, `\`, `?`, `#`, `javascript:`, `file:` and `..`, and the route is re-encoded under the fixed `/view` prefix.
- The FilesPage seed/write-back pair has no loop (the seed is captured once per tab id).
- `backToConversation` is fully retired, including the preload `DESKTOP_COMMAND_IDS` allowlist.
- Deep links open no privileged tabs, and the confirmation dialog is intact.
- The PR adds no `console.*` calls, no secret or URL logging, and no dependency changes. There are no `.only`/`.skip` tests.

## Recommendation

Nothing blocks the merge on severity. We suggest fixing these before merge because they are small and directly tied to the earlier review rounds: **H1** (add a handoff stale test), **M1** (the `appRef` check that R1-H1 asked for) and **M4** (a liveness witness on the R1-H1/R2-M1 tests). **H2**, **M2**, **M3**, **M5** and **M6** can be fixed in the PR or tracked as follow-ups; that is the author's and reviewers' decision. The Low items are optional.

---
_Read-only review: no code changes were pushed. 🤖 Generated with [Claude Code](https://claude.com/claude-code)_

zach88:
Thanks for the thorough pass. Round 3 is on the branch (head is now `19d722ea1`). We took the three you flagged for pre-merge, plus H2.

**H1 — handoff guard test** (`31601ecd9`, `19d722ea1`)
Added a handoff variant of the stale-continuation test: it defers the handoff's `getLocation()` read, supersedes it with a plain deactivation (which bumps the generation and closes the embed itself), then resolves the stale read and asserts the handoff added neither a stale persist nor a second close. Verified it goes red when the guard is removed.
Writing it turned up a self-inflicted overclaim: our first draft said it covered "both" generation checks, but only the post-read check is actually reachable — there's no `await` between the persist and the close, so that one check aborts both, and the pre-close check is a defensive re-check that can't fire on its own with today's control flow. Corrected the test's claim in `19d722ea1` (comment-only, assertions unchanged).

**M1 — appRef check before persisting the route** (`8a9c45f5d`)
The missing half of R1-H1. Both the deactivation effect and the handoff now persist `location.routePath` only when `location.appRef === outgoingAppRef` (captured synchronously from the outgoing tab, immutable). On an app→app switch where the incoming `open()` beats the read within a single activation generation, the mismatched read is now dropped instead of clobbering the outgoing tab. Repro test fails with the guard reverted.

**M4 — liveness witnesses** (`1bb76a72b`)
Added positive witnesses to the stale-continuation tests: `getSandboxUiLocation` called exactly once (the deferred read is consumed) and the surviving app is live again. Confirmed both go red if the read-then-close is disabled — previously disabling it left them green.

**H2 — keyboard focus on close** (`4a1c52c7f`)
Closing the active tab by keyboard now moves focus to the newly-active tab, or to the strip when the workspace is empty, gated on `event.detail === 0` so a mouse close doesn't steal focus. Asserted on `document.activeElement`.

Ran code + security + collateral reviewers over the round-3 diff (`95e71b49a..19d722ea1`): security clean, no regressions, full suite green (2398), renderer typecheck unchanged (the 31 errors are pre-existing).

On the same-`appRef` edge you and the security note both raised: confirmed pre-existing — the pre-fix code clobbered the same way, and the `appRef` guard can't tell two tabs of the same app apart. The real fix is keying by tab id / `webContentsId` rather than `appRef`; tracking that separately instead of widening this PR.

The rest — **M2** (title sanitization), **M3** (recency-based Chats nav), **M5** (IPC surface tests), **M6** (discriminated union), and the Low items — we're leaving as follow-ups for now. Happy to fold any of them in if you'd rather they land in this PR.

josue-tech-94:
Round 4 review complete for `1acc007d280019d0c6168cd4877f83bd54ffa116`.

**Decision: CHANGES REQUIRED / BLOCKED**

The earlier sandbox lifecycle-generation fix remains correct, but the PR is not merge-ready. Two carried findings have now escalated to process Blockers, and the latest corrective range introduced one additional route-persistence defect.

### R3-M1 — Blocker: Chats resumes the wrong conversation

`useNavigationController.ts` tracks `lastActiveChatTabId`, but `focusChatSection()` ignores it and selects the last chat by array order.

Reproduced:

1. Activate chat A.
2. Switch to Files.
3. Return to Chats.
4. Chat B is selected instead of chat A.

**Required:** prefer `lastActiveChatTabId`, falling back only when that ID is missing or stale. Add an observable producer-backed regression test asserting both the active tab ID and `{ agentRef, chatId }`.

### R3-M2 — Blocker: Untrusted app titles reach host chrome

The title propagation pipeline is wired correctly, but the renderer currently only applies `trim()`.

Plugin-controlled titles containing control characters, bidi overrides/invisible characters, or excessive length can reach workspace tab labels, accessible close-button names, and other host chrome.

**Required:** define one title-sanitization contract covering:

* control/bidi removal;
* bounded length;
* stable fallback behavior.

Add producer-backed tests for normal, empty, control/bidi, and overlong titles.

### R4-M1 — Medium: appRef mismatch clears an existing saved route

Introduced by `8a9c45f5d`.

The new app identity check correctly prevents App B's route from being stored for App A, but currently forwards `undefined` to `setAppTabSavedRoutePath()`. That helper interprets `undefined` as an authoritative clear.

So a foreign/stale location result can erase App A's previously saved route.

**Required:** distinguish:

* foreign/unknown location → preserve the existing route;
* authoritative current-app default route → clear the saved route.

Apply that distinction consistently in both deactivation and handoff paths.

### Follow-ups

* `R2-L1`: PR description remains stale.
* `R4-L1`: add direct title event identity/cleanup regression coverage across driver → IPC → preload.
* `R4-L2`: `WorkspaceTab` still uses optional payload fields rather than a discriminated union.

### Closed findings

The following are verified closed:

* `R1-H1` — lifecycle-generation / stale sandbox close
* `R2-M1` — genuine distinct-App-B regression fixture
* `R3-M3` — active-tab close restores keyboard focus
* `R3-M4` — stale-continuation tests now include liveness witnesses
* `R3-L1` — inaccurate test claim corrected

The bounded collateral review of:

`95e71b49a9c8999444fe313c87576aafd6d04eba..1acc007d280019d0c6168cd4877f83bd54ffa116`

confirmed those closures and found `R4-M1` as the new corrective regression. No additional lifecycle-close, focus, IPC, or listener-cleanup regression was found.

### Validation

The full desktop suite, production builds, style rules, Electron verification, focused lifecycle/navigation/workspace/IPC/drawer/files suites, and `git diff --check` all passed.

Renderer `tsc` remains non-zero at both the exact merge base and current head with the same pre-existing error family, so it remains a baseline issue rather than a new PR regression.

Native Electron/Playwright live-app validation was not run.

### Round-4 design stop

This PR has reached the design-stop threshold.

`App.tsx` now has four `fix:` commits in this PR and `useNavigationController.ts` has three. Please do **not** continue patching the shared navigation/lifecycle seam directly.

Before the next corrective cycle, write a mini-spec and decision table covering at minimum:

* active-chat selection / Chats resume behavior;
* app-view ownership;
* foreign `appRef` route reads;
* authoritative route clearing vs route preservation;
* stale async continuations and lifecycle generations.

Then implement the remaining findings from that design rather than adding another isolated patch.

The PR body also needs to be refreshed

josue-tech-94:
**Decision: CHANGES REQUIRED / BLOCKED**

Two technical items remain before the PR is ready for another review.

### R3-M2 — Title sanitization is still incomplete

The title-sanitization work improves the boundary, but it does not yet fully handle the intended directional/invisible-control cases, and the committed coverage does not verify the final host-visible outputs.

**Required:**

* correctly remove the unsafe control/bidi characters;
* preserve legitimate Unicode;
* retain the intended length bound and stable fallback;
* test the final workspace-tab label;
* test the final close-button accessible label.

This remains a security-review issue because sandbox-controlled title content is surfaced in host chrome and ARIA output.

### R5-M1 — Legitimate emoji ZWJ sequences are broken

The current sanitizer strips U+200D zero-width joiners.

For example:

`Build 👩‍💻` → `Build 👩💻`

This alters legitimate user-facing Unicode and is a regression introduced while addressing `R3-M2`.

**Required:** preserve legitimate Unicode shaping/joining characters such as emoji ZWJ sequences while still removing the unsafe directional/invisible controls covered by the sanitization policy.

Add a regression test proving valid emoji sequences remain intact.

### Closed / accepted

* `R3-M1` — chat navigation correctly resumes the tracked last-active chat.
* `R4-M1` — foreign app location reads preserve the existing saved route in both deactivation and handoff paths.

The earlier lifecycle, distinct-app, keyboard-focus, and stale-continuation findings remain closed.

### Non-blocking follow-ups

* `R4-L1` — direct title-event identity/cleanup coverage across driver → IPC → preload.
* `R4-L2` — stronger `WorkspaceTab` payload typing / discriminated-union invariant.

### Validation

The full desktop suite passes: **245 files / 2,412 tests**.

Main and renderer builds, Electron verification, focused navigation/IPC/drawer/files suites, style rules, and `git diff --check` also pass.

Renderer TypeScript remains non-zero at both the current head and exact merge base with the same pre-existing baseline family.

Native Electron/Playwright validation was not available.

zach88:
Addressed the round-6 title-sanitization blocker in `108c0f7d1`, but with a change of approach rather than another two ranges — I think that's the only way this stops recurring.

**Why not just add U+00AD and U+180E**

The sanitizer was an enumerated numeric-range denylist. That model never converges: there's always one more invisible code point, and each round has been adding the next one (round 5: bidi marks, word joiner, tag chars; round 6: soft hyphen, Mongolian vowel separator). Patching the reported instance instead of the class is exactly the loop we keep paying for. Adding two ranges would've closed this comment and reopened the same one next round.

**What it does now**

Replaced the denylist with Unicode's own definition of non-rendering text:

```
/(?!\s)\p{Cc}|(?![\p{Join_Control}\p{Variation_Selector}])\p{Default_Ignorable_Code_Point}|[￹-￻]/gu
```

- Strip the control category (`\p{Cc}`, minus whitespace — the `(?!\s)` guard keeps TAB/LF/CR so they collapse to a space instead of gluing words; NEL U+0085 is `\p{Cc}` and not `\s`, so it's stripped).
- Strip `\p{Default_Ignorable_Code_Point}` — the standard's set of "code points that should be ignored in rendering", i.e. the invisible set. **This covers U+00AD and U+180E by definition**, along with every bidi mark/override/isolate, zero-width formatter, BOM, tag char, Hangul filler, etc. It also *already excludes* the legitimate visible script formatters that real titles use (U+0600–U+0605, U+06DD ARABIC END OF AYAH, U+070F, Kaithi, Egyptian controls), so those survive with no per-code-point carve-out.
- Keep U+FFF9–U+FFFB (interlinear anchors) explicit — they hide text but Unicode doesn't classify them as Default_Ignorable. Written as `\u` escapes, never literal invisible bytes, so the denylist can't be mutated with no visible diff.

The payoff: any invisible code point is covered by definition. There's no "one more" to ask for — if a code point isn't caught, it's DI=false, meaning Unicode considers it visible.

**Two deliberate carve-outs (they're invisible but they *shape* the adjacent glyph):**

- `\p{Join_Control}` — ZWJ/ZWNJ. ZWJ fuses an emoji sequence into one glyph (`👩‍💻`); stripping it splits it into `👩💻`.
- `\p{Variation_Selector}` — VS16 (U+FE0F) is what makes `❤` render as the emoji `❤️`, keycaps (`1️⃣`) depend on it, and ideographic selectors pick CJK glyph variants. A blanket Default_Ignorable strip silently breaks all of these — worth calling out since it's a trap the property-based approach would otherwise walk into.

**On the "still no e2e assertion" part**

Added a producer-backed test that drives the real store border (`setAppTabTitle`) and asserts the *rendered* chrome for an overlong + bidi-override title: the workspace tab label text and the close button's accessible name both come out truncated and stripped (`WorkspaceTabStrip.test.tsx`). The invisible/shaping behavior is asserted end-to-end too (bidi/invisibles stripped, ZWJ emoji and variation selectors preserved).

**One thing that's intentionally out of scope**

A title made only of *visible-but-blank* glyphs (e.g. U+2800 BRAILLE PATTERN BLANK) renders as an empty-looking tab. No invisible-character filter can catch that — the glyph is visible, just empty — and it isn't a spoofing vector here: the label is free text shown to its own owner, never compared against a trusted string. So it's a UX edge, not a sanitizer gap; flagging it so it doesn't come back as "another invisible reaches the UI."

**Verification**

- One observable change vs. the old policy: U+1D173–17A (musical formatters) are now stripped. They're invisible and have no effect in a `nowrap` label; the test that asserted their preservation was protecting an idea ("don't touch Cf"), not a visible behavior, and was repointed at U+13430 (a visible Egyptian control that stays).
- Ran the old vs. new regex across the whole code space: the new policy is a strict superset for every real vector (0 gaps), stripping ~3765 additional code points, all invisible/reserved.
- Full renderer suite green; typecheck at baseline; the invisible-coverage test fails against the previous head (genuine regression coverage).

josue-tech-94 (approved):

Issue #680, opened 2026-09-18
Codex subscription: tool-call limit (maxToolCalls=32) is reported as retryable "Model Overloaded" and is failover-eligible


alfredolopez80:
## Problem

`codex-llm-proxy` aborts an upstream stream when the model emits more than `LIMITS.maxToolCalls` (32) tool calls in one response (`codex-llm-proxy/src/codexTransport.ts:344-356`, `consumeSse.acceptFrame`). The limit is Evenfire's own (`packages/llm-provider-attempt-contract/index.cjs:19`), not an upstream one: the proxy sends Codex `parallel_tool_calls = true` with no cap, so the model does not know the limit exists.

The abort is raised as `CodexTransportError('provider_unavailable', 'tool calls exceed 32')`. Tool-call frames are buffered until the response completes, so no bytes have been written to the client and `server.ts` answers HTTP 503 `{error:"provider_unavailable"}`. It logs only `codex_proxy_denied {code}`; the message is dropped.

`mcp-host` maps `provider_unavailable` to `LlmErrorCode.ModelOverloaded, retryable: true` (`mcp-host/src/llm/codexSubscription.ts`, `classifyError`, ~line 130). `mcp-host/src/llm/failover/classify.ts:48-49` classifies that as the `provider_unavailable` failover class, and the desktop shows "Model Overloaded" (`desktop-app/ui/src/constants/agents.ts:13`).

Consequences:
- The error is retried, although a retry of the same prompt has no reason to produce fewer calls.
- It is eligible to trigger provider failover.
- The operator is told the provider is overloaded when the cause is an Evenfire policy limit.
- No log records how many calls were emitted.

The same limit is also enforced on request history by `parseMessages` (`packages/llm-provider-attempt-contract/index.cjs:255`) in three places: control-api (`services/llmProviderAttemptAuthorizer.ts:626`), codex-llm-proxy (`codexTransport.ts:89`) and mcp-host (`codexSubscription.ts:263`). Raising only the stream check is therefore insufficient: the next request would be rejected.

## Evidence

Context: cluster `clerum`, project `eventfire-491421`, 2026-09-18 UTC. Host `andy-agent`, provider `codex-subscription`, model `gpt-5.6-sol` (selected per chat; the Host default is `gpt-5.6-luna`). 119 tools were presented in `mode=direct`.

[...]

PR #685, opened 2026-09-18
fix(codex): report tool_call_limit_exceeded end to end and raise limits to 64/256

alfredolopez80:
## Summary

PR-1 for #680 (Codex only). A Codex response with more than `maxToolCalls` function calls now fails as `tool_call_limit_exceeded` end to end, instead of surfacing as `provider_unavailable` / `LLM_MODEL_OVERLOADED` ("Model Overloaded") and being failed over:

- **Contract** (`packages/llm-provider-attempt-contract`): Codex `maxToolCalls` 32 → 64, `maxMessages` 128 → 256.
- **codex-llm-proxy**: the over-bound SSE response raises `CodexTransportError('tool_call_limit_exceeded', …, {limit, observed})`; the proxy maps it to HTTP 422, or to an SSE error frame when headers were already sent. Every attempt emits a `codex_proxy_attempt_finished` log line and failures increment `codex_proxy_attempt_failures_total{code}`.
- **mcp-host**: `classifyError` maps `tool_call_limit_exceeded` to `LlmErrorCode.ToolCallLimitExceeded` (not retryable, failover class `null`); a Host-side guard rejects a history longer than `maxMessages` with `request_limit_exceeded` → `LLM_CONTEXT_LENGTH_EXCEEDED` before a ticket is minted.
- **Desktop**: labels `LLM_TOOL_CALL_LIMIT_EXCEEDED` ("Too Many Tool Calls") and `LLM_CONTEXT_LENGTH_EXCEEDED` ("Conversation Too Long").
- **Docs/fixtures**: Codex and Grok contract docs, sanitized upstream contract fixture and its freeze test.
- **E2E**: the deterministic approved-tools upstream gains a `tool call limit probe` trigger (65 calls, any retry throws), and the Desktop Playwright journey asserts the label, one upstream completion, no retry and no connector call.

Grok parity (plan Step 7) is PR-2 and is not part of this PR.

## Commits (one per plan step, each `Refs #680`)

| Step | Commit | Subject |
|---|---|---|
| 1 | 50ca0cab4 | fix(contract): raise Codex maxToolCalls to 64 and maxMessages to 256 |
| 2 | 1a457fb54 | fix(codex-llm-proxy): report tool_call_limit_exceeded as 422 with attempt telemetry |
| 3 | 7d2bc2cfb | fix(mcp-host): classify Codex tool-call and history limits as non-retryable |
| 4 | ca34baa48 | fix(desktop): label tool-call limit and context length errors |
| 5 | 4ec5a4562 | docs(codex): document the 64/256 limits, tool_call_limit_exceeded and attempt telemetry |
| 6 | ed442902b | test(e2e): cover the Codex tool-call limit in the Desktop approved-tools journey |
| 6.4 (A3) | 594cafb98 | fix(e2e): accept kubectl's object-per-item dry-run output in approved-tools preparation |

## Step 0 facts

- **F1** No history trimmer keyed on `maxMessages` in mcp-host. Compaction is token-based at the agent level (`config.ts` compaction settings, `manageMessagesForIteration`) and runs before the provider builds the request, so the Step 3.4 guard sits directly before `parseCodexCompletionRequestV1`.
- **F2** No exhaustive `Record<LlmErrorCode, …>` / `[LlmErrorCode]` map in `mcp-host/src`, `control-api/src` or `desktop-app`.
- **F3** `classifyError` is called from `core/adapters/llmPortAdapter.ts:294` (chat path; entry point of the LlmError-level test), `workflow/failoverProvider.ts:208` and `pluginWorkloadSdk/promptBridge/llmBridge.ts:892`.
- **F4** `executeToolCalls` (`core/orchestration/toolUseLoopToolBatch.ts:193`) runs a batch sequentially (`for` loop at :216). No unbounded parallelism.
- **F5** `observeAttempt(outcome, operation)` increments the prom-client counter `codex_llm_proxy_attempts_total{outcome,operation}`, declared in `createProxyMetrics(register)`; tests read counters by scraping the admin `/metrics` text.
- **F6** There is no `test/server.test.ts`. The server telemetry cases went into `test/server.security.test.ts` (supertest over `createProxyApps`, fetchFn injection fakes the upstream); the logger is observed with `vi.spyOn(logger, 'info'|'warn')`.
- **F7** `errorTaxonomy` exists only in `tests/e2e/fixtures/codex-subscription/sanitized-upstream-contract.json`; `codex-llm-proxy/src` publishes no taxonomy. The `CODEX_TEST_UPSTREAM_*` variables are documented only in the inline comments of `tests/e2e/fixtures/codex-subscription/test-upstream/server.mjs`.
- **F8** No 32/128 literals for `maxToolCalls`/`maxMessages` in `workflow-recipes` or `control-api/src`; the limits flow from `LIMITS` in `packages/llm-provider-attempt-contract/index.cjs`.
- **F9** `LLM_MODEL_OVERLOADED` appears only in `desktop-app/ui/src/constants/agents.ts` (`AGENT_ERROR_CODE_LABELS`) and its render test. No zod enum or type lists LLM error codes between the Host and the Desktop; the code travels as a string. (This grep was recorded after the Step 4 commit; it confirmed Step 4 needed no further change.)
- **F11** Test commands: contract `node --test`; codex-llm-proxy `npm run typecheck:test && vitest run`; mcp-host and desktop-app per `scripts.test`.
- **F13** In `consumeSse.acceptFrame` the tool-call count check (`toolFrames.length >= LIMITS.maxToolCalls`) runs before the unknown-tool-name check, and every 65-call fixture uses a presented tool name. The pending-call key is `item_id || id || call_id`, so distinct `call_id`s are distinct entries.
- **F14** The loop returns `{type:'error', error}` (`toolUseLoop.ts:440`); `TaskExecutor.handleLoopResult` rethrows it (`agent/taskExecutor.ts:1201`); `toTaskError` (`:879`) passes an `LlmError` code through; `StateMachine.handleTaskFailure` (`agent/stateMachine.ts:1423`) stores it; Desktop reads `result.code` (`useAgentChatController.ts:2202`).
- **F15** `llmBridge.ts` failover decisions use `classifyFailoverClass(classified.code, classified.retryable)` (:401, :564, :642, :669, :759); failover tests live in the sibling file `promptBridge/llmBridge.failover.test.ts`.
- **F16** Lane entry point `scripts/e2e/run-codex-approved-tools.mjs`, driven by `make minikube-run-codex-approved-tools` (`prepare-codex-approved-tools.mjs run`). `tests/e2e/playwright/helpers/launch-desktop.ts:82-117` launches Electron with a `mkdtemp` `--user-data-dir`, asserts `app.getPath('userData')` equals it and removes it on close. The session token is stored via keytar in the macOS Keychain (service `Evenfire`), outside the user data dir; the test Electron binary is a different application from the installed signed Desktop, whose Keychain items are ACL-bound. No Keychain prompt was accepted during the run.

Facts 10 and 12 belong to the out-of-scope Steps 8.3/9.2.

## Tests (executed counts)

| Suite | Result |
|---|---|
| Contract `node --test` (Step 1) | 43/43 |
| codex-llm-proxy full suite (Step 2) | 13 files, 133/133 |
| codex-llm-proxy conformance (2.7) | 39/39 |
| codex-llm-proxy `server.security.test.ts` (2.7) | 26/26 |
| codex-llm-proxy `runtimePath.hermetic.e2e.test.ts` (2.7) | 5/5 |
| mcp-host full suite (Step 3) | 367 files, 4159 passed + 1 todo |
| mcp-host targeted (3.6) | 5 files 128/128; 2 files 43/43 |
| desktop-app `npm test` (Step 4, Node 24, `npm ci`, `verify:electron` EXIT 0) | 240 files, 2314 tests |
| `ChatThread.renderSemantics.test.tsx` (Step 4) | 6/6 |
| contract freeze `codex-subscription-contract-freeze.test.ts` (Step 5) | 2/2 |
| `approvedToolsUpstream.test.ts` (6.1) | 21/21 |
| `run-codex-approved-tools.test.mjs` (Addendum A2) | 17/17 |
| `prepare-codex-approved-tools.test.mjs` (Addendum A3) | 14/14 |
| CI approved-tools node:test set (`ci-public.yml`, A3) | 123/123 |
| Playwright typecheck, E2E Guardian static audit (Step 6) | EXIT 0, PASS (4 files) |

Every negative assertion has a liveness witness in the same test.

## Vacuity mutations (feature disabled, then restored and re-run green)

| Mutation (plan step) | Result | Failing tests |
|---|---|---|
| Step 1 contract | 1 of 43 fails | tool definition count never widens the independent assistant call limit |
| 2.7 conformance + server | 3 of 44 fail | validates the complete 65-call response before emitting executable calls; rejects 65 calls before any text with 422; rejects 65 calls after streamed text with an SSE frame |
| 2.7 hermetic runtime path | 2 of 5 fail | the hermetic 422 and SSE-frame cases |
| 3.6 failover engine | 3 of 53 fail | propagates a tool-call limit without switching or setting a cooldown; surfaces a 65-call stream through the port adapter as a non-retryable LlmError; classifies request limits as non-retryable, non-failover errors |
| Step 4: the two labels removed | 2 of 6 fail | the "Too Many Tool Calls" and "Conversation Too Long" cases |
| Step 5: taxonomy entry removed from the fixture | 1 of 2 fails | the freeze declaration test |
| 6.1: probe emits 64 calls instead of 65 | 2 of 21 fail | the 65-call probe case and the real `streamCodexCompletion` rejection |
| 6.1: retry detection disabled | 1 of 21 fails | the retry/continuation rejection case |
| A2: limit case expected in both modes | 3 of 17 fail | the mode matrix and both cross-mode rejection cases |
| A3: `parseDryRunItems` reverted to `JSON.parse(output).items` | 2 of 14 fail | the List/object-stream acceptance case and the rejection case |

## Step 6 Playwright lane (deterministic)

Running. `make minikube-run-codex-approved-tools` (deterministic mode) runs after the T2 certification on the same HEAD; this section will be updated with the JSON reporter summary (titles, passed, `failed=0`, `skipped=0`).

## Step 8 certification (T0/T1/T2)

Re-certification on the current HEAD `594cafb98` (after addendum A3, with the approved-tools fixtures re-acquired at that commit) is running; this section will be replaced with its verdict. Previous certification, before A3:

`make minikube-t2-preflight` ran first (planner only). `make minikube-t2` on HEAD `ed442902be69b3a696daceef4fd784835e45e3bc`, owned profile `clerum-fix-codex-tool-call-limit-taxonomy-53ec51f3`, planner `T2_PREFLIGHT_PASS` with `transition=full-bootstrap` (the profile had no pre-gate-sync state), EXIT 0:

```
MINIKUBE_T2_PASS
T0=PASS
T1=PASS
T2=PASS
NP08_HCC_AUTHORIZATION=PASS
Health=NOT_RUN
Playwright=NOT_RUN
evidence=.local-notes/infra/runs/20260918T163428Z-ed442902be69-12c6486c37d9499ba7ec834fbf0d6074/evidence.json
```

`Health` and `Playwright` are separate lanes; the Desktop Playwright journey is the Step 6 lane above, run on the same HEAD and profile after this certification.

## Plan addenda

- **A1** (Steps 2.6/2.7): the test upstream always wrote a `hello` text delta before tool calls, so a 65-call response reached the client after headers were sent (HTTP 200 + SSE error frame), not 422. The fixture gains `CODEX_TEST_UPSTREAM_OMIT_TEXT=1` (requires `CODEX_TEST_UPSTREAM_TOOL_CALL_COUNT`; unset keeps output byte-identical), and the hermetic runtime path covers 65 calls without text (422), 65 calls after text (SSE `tool_call_limit_exceeded` frame) and 64 calls (success, liveness witness).
- **A3** (Step 6.4): `make minikube-run-codex-approved-tools` failed before any fixture mutation. Preparation parsed `kubectl create --dry-run=server -f - -o json` on a `List` as one document with `.items`; kubectl client v1.36.2 prints one JSON object per item, concatenated, so `JSON.parse` threw (reproduced with a two-Namespace server dry-run on the owned profile). `parseDryRunItems` accepts exactly a `List` document or a stream of concatenated objects and rejects everything else. The code dates from a447bc9fe; no earlier run on this machine reached fixture preparation.
- **A2** `validateReport` expected titles depend on `APPROVED_TOOLS_UPSTREAM_MODE`; the tool-call limit case is required in deterministic mode only (only the deterministic upstream can produce it), and the runner test covers both modes.

## Findings (reported, not changed here)

- `llmBridge.ts:893-894` labels `PluginWorkloadError.reason` with `classifyFailoverClass(...) ?? 'provider_unavailable'`, so a non-failover error such as the tool-call limit is reported with reason `provider_unavailable`. The failover decision is correct. Introduced in 574a45761.
- `codexLlmProxyClient.ts:97` defaults a non-string upstream error code to `provider_unavailable`.
- `ingestSseBlock` swallows JSON parse errors of SSE blocks.
- `llmPortAdapter.handleProviderError` uses `console.log` instead of the mcp-host logger.
- `scripts/tests/test-codex-subscription-t0.sh` does not list `runtimePath.hermetic.e2e.test.ts` (it runs in CI through the `ci-public.yml` `npm test` matrix and `e2e-codex-subscription-runtime.sh`).
- `desktop-app/ui/tsconfig.json` has 38 pre-existing type errors; no gate runs it (the CI gate is the root `tsc --noEmit`, EXIT 0).
- The Desktop session token lives in the macOS Keychain outside the isolated E2E user data dir, and `migrateLegacyGlobalToken` deletes the legacy `session-token` account.
- A fresh worktree lacks `node_modules` in 28 packages; `make minikube-t2` stopped twice (`LOCAL_DEPENDENCY_MISSING` for `gfs-controller`, then a `tsc` failure in `packages/workflow-sdk` during `pre-gate-sync`) until `npm ci` ran in each. The local preflight only checks `control-api` and `gfs-controller`.
- `prepare-codex-approved-tools.mjs` `main().catch` replaces every preparation error with "Approved tools preparation failed; review bounded sanitized fixture evidence" and prints no message, so the A3 failure was only diagnosable with a preload that printed the rejection message. Not changed here.
- CI flake, unrelated to this diff: `scripts/tests/wrc-egress-dns-proxy.test.cjs` ("held-query overflow is bounded…") binds a TCP listener on the port number of a UDP ephemeral socket (`fixture()`, b4a78cb73) and failed once with `listen EADDRINUSE` in the "Shell script syntax (bash -n)" job (run 35371861859). It passes locally 11/11; the failed job was re-run.
- npm 11 prints `allow-scripts` warnings during `npm ci`; the Electron binary is still installed and `verify:electron` passes.


zach88:
Took a deep pass through this one across all the layers it touches (proxy, mcp-host, the shared contract package, desktop labels, and the e2e/fixtures). Good news up front: the end-to-end wiring is solid. I traced every seam by hand and the strings match byte for byte, so `tool_call_limit_exceeded` really does surface as "Too Many Tool Calls" and `request_limit_exceeded` as "Conversation Too Long" instead of the old "Model Overloaded" failover path. Nothing here blocks the merge; what's left is three small non-blockers.

**🟡 Approve with nits**
- Optional / non-blocking: 3

## Optional (non-blocking)

1. **[R1-L3] Loosening the shared limits needs control-api co-deployed.** This bumps `maxMessages` 128→256 and `maxToolCalls` 32→64 in `packages/llm-provider-attempt-contract` (`index.cjs:17,20`), but control-api isn't rebuilt in this diff and it re-validates the same package at authorize (`control-api/src/services/llmProviderAttemptAuthorizer.ts:626`, `parseCodexCompletionRequestV1`). During a rolling deploy, a request with 129–256 messages or 33–64 tool-calls in history clears the mcp-host guard (256) but an old control-api rejects it at authorize, which collapses to `invalid_request` and the user sees "Connection Error" instead of "Conversation Too Long". Fails closed and transient, but worth either co-deploying control-api + proxy + mcp-host, or calling out the deploy order in the PR body. Not a code bug: every layer references the same constant.

2. **[R1-L1] The 422 error body ships with an SSE content-type.** In the `http_status` branch (`codex-llm-proxy/src/server.ts:252`), the response body is `{ error: <code> }` but `Content-Type` is already `text/event-stream` from earlier in the handler, and `res.json()` won't overwrite it (reproduced with a minimal Express app). It's cosmetic for our actual consumer: mcp-host reads the code with `fetch().json()` (`codexLlmProxyClient.ts:97,132`) without branching on content-type, so the end-to-end goal holds. But it's incorrect HTTP hygiene and would bite any generic client that keys off content-type. One-liner: `res.setHeader('content-type', 'application/json')` before the `.json()`. (The sibling in-stream errors have the same shape but that's older code the diff doesn't make consequential, so leave those unless you want to fix the class in one go.)

3. **[R1-L2] Unbounded label on the new failure counter.** `codex_proxy_attempt_failures_total{code}` (`server.ts:226`, `metrics.ts:16-24`) is labeled with `mapped.code`, which for the `ControlApiClientError` branch comes from control-api's free-form `parsed.error` (`controlApiClient.ts:116`). Not attacker-controllable (upstream errors all map to constant literals; control-api is a trusted first-party with a bounded taxonomy), so cardinality is bounded today. Cheap hardening: run `mapped.code` through a closed allowlist with an `'other'` fallback before using it as a label, so a future control-api change can't widen it.

## What checked out clean

- **64/256 numbers agree everywhere:** `index.cjs` = `index.d.ts` = the Codex doc table = the freeze fixture. No stray `32`/`128` hardcoded in live code. Grok's code is untouched; only its doc line was clarified.
- **Both error codes trace end to end** with no typo/case mismatch, and mcp-host handles both report shapes (HTTP 422 and the SSE error frame).
- **Failover classification is right:** both new codes fall to `default: null`, so no retry, no cooldown, no cross-tenant poisoning. Liveness witness in the engine test (next call returns to primary).
- **The maxMessages guard runs before authorize and before the ticket is minted,** so a rejection doesn't burn quota; the contract parser re-rejects as defense in depth.
- **Telemetry is payload-free:** the `codex_proxy_attempt_finished` line logs only ids/hashes/counts, and the error frame/body carry just `{ code }`, no upstream tokens/headers. The security test asserts the forbidden keys are absent.
- **Desktop degrades gracefully:** `AGENT_ERROR_CODE_LABELS[errorCode] ?? 'Error'`, so an old client falls to "Error", not "Model Overloaded". Both new labels are covered by the render test.
- **`parseDryRunItems` (the A3 kubectl fix) is robust:** brace-depth scanner with string/escape state, stress-tested against concatenated objects, CRLF, nested braces, `"}{"`/backslash in strings, empty List, and it rejects arrays/junk/truncated input.
- **Suites green where runnable:** proxy 134/134, mcp-host 180/180 (area files), contract 43/43, e2e node tests 14/14 + 28/28, the minikube t2 gate scripts pass.

zach88 (approved):

alfredolopez80:
Thanks for the trace-by-hand pass. All three non-blockers are addressed, and one of them got more load-bearing since your review.

**[R1-L1] 422 served as `text/event-stream`** — fixed in c77b867e1. The completions handler stages `Content-Type: text/event-stream` and `Cache-Control: no-cache` before the stream starts, and `res.json()` keeps an existing content-type, so every pre-stream error (400/401/403/409/422/503), not just the new 422, went out with the SSE type. Nothing has been written on that branch yet, so the fix drops the staged `cache-control` and sets `application/json; charset=utf-8` before `res.status(...).json(...)`. The 422 and 400 cases in `server.security.test.ts` and the 403 `no_grant` and 422 cases in `runtimePath.hermetic.e2e.test.ts` now assert the JSON content-type and read `res.body`; with the header change removed all four fail with `text/event-stream; charset=utf-8`.

**[R1-L2] Unbounded `code` label** — fixed in da3328b6a, as the allowlist you suggested. The status map inside `mapError` became a module-level `ATTEMPT_ERROR_STATUS` (it also gained `sse_buffer_exceeded`, `invalid_receipt` and `conflict` at 503, so no HTTP status changed), and `failureLabel` records the code when it is in that map and `other` otherwise. The attempt log line still carries the raw code, so nothing is lost for debugging. Two new cases in `server.security.test.ts`: a `no_grant` denial counts under `no_grant`, and an unknown code counts under `other`, never appears in `/metrics`, and is still logged raw. Vacuity: returning the code unchanged fails the unknown case; always returning `other` fails the two known-code cases.

**[R1-L3] Deploy order** — documented in the PR body, and the window is wider than when you wrote it. A later review finding showed 64 tool calls is well below what a single Desktop turn produces, so fba84cb34 raised the contract to `maxToolCalls: 256`, `maxMessages: 1024`, `maxOutputTokens: 16384` (Grok stays 64/128). The reasoning is in a comment next to the constants: a turn of N calls adds N+1 messages, so N stays at or under `maxMessages/4`, and at that bound `maxOutputTokens` still leaves 64 output tokens per call in one response. So the rolling-deploy hazard you described now spans 257–1024 messages and 65–256 tool calls: an old `control-api` rejects those at authorize with `invalid_request` and the user sees a connection error instead of the right label. The PR body now says to deploy `control-api` and `codex-llm-proxy` before, or together with, `mcp-host`. No code change — as you noted, every layer reads the same constant.

**Where the boundary is actually proven.** `codexSubscription.test.ts` covers a successful batch of exactly 256, a 257-call batch rejected with `tool_call_limit_exceeded` before any executable tool is returned, and the same 257-call stream surfacing through the port adapter as a non-retryable `LlmError`; `engine.test.ts:225` asserts the failover engine does not retry it.

The matching Playwright boundary case (`codex-subscription-approved-tools.spec.ts:655`, a turn of exactly 256 calls) is written but **was not run**, and I want to be explicit about that rather than let `NOT_RUN` pass as a detail. That lane cannot pass on any branch today: its fixture requires the three `clerum__tool_*` discovery bridges in every payload, and `direct` presentation — the default since `b8c9cd6ab` — never registers them. Both commits are on `main` and neither is touched by this PR. That is #718, filed with the rest of the harness findings the attempt surfaced (#712, #714–#722).

Merged as `e77b9e08dfc2837194309de0151489a932b3539f`: 41/41 CI checks green, and `make minikube-t2` on the exact head `649f4d9ab` returned `MINIKUBE_T2_PASS` with `T0=PASS T1=PASS T2=PASS`, `T2_EXIT=0` (T1: 202/202 tests, 42/42 files, 0 pending) and `Playwright=NOT_RUN` for the reason above.

PR #686, opened 2026-09-18
fix(gfsc): name 0097_gfs_upload_sessions in the upload v2 readiness error

alfredolopez80:
## Summary

`assertUploadV2Ready` (`gfs-controller/src/index.ts`) failed startup with:

```
[gfsc] GFS_UPLOAD_V2_ENABLED requires migration 0091 upload tables
```

The number is stale. In `control-api/src/db.ts`:

- `0091` is `0091_gfs_desktop_operator_links` (unrelated to uploads).
- `gfs_upload_sessions` and `gfs_upload_parts` are created by `0097_gfs_upload_sessions` (`applyGfsUploadSessionSchema`).
- `0098_gfs_upload_cleanup_receipt` and `0099_gfs_upload_finalizing_recovery` only add columns to `gfs_upload_sessions`.

Changes:

- The error now reads `GFS_UPLOAD_V2_ENABLED requires the upload tables from migration 0097_gfs_upload_sessions`, naming the full version string so it can be grepped in `db.ts`.
- The doc comment on `applyGfsUploadCleanupSchema` (`control-api/src/services/gfsUploadSchema.ts`) said "already applied migration 0091"; it now names `0097_gfs_upload_sessions`.
- `control-api` package version bump from the pre-commit hook.

Both stale references were introduced in 97d1d13f24.

Other `0091` hits from `git grep -n 0091` were left unchanged because they refer to `0091_gfs_desktop_operator_links`, the historical `0091_oauth_grants_owner_generalization` rename chain, or a test UUID.

## Impact

Text-only. The readiness check (`to_regclass` on both tables), its conditions and its outcome are unchanged. The message only appears when `GFS_UPLOAD_V2_ENABLED=true` (default `false`; set by HCC on the writer only) and the upload tables are missing. No test asserts this message.

## Test plan

- [x] `gfs-controller`: `npx vitest run`, exit 0, 609 passed / 9 skipped / 618 total (two consecutive full runs).
- [x] `gfs-controller`: `npx tsc --noEmit`, exit 0.
- Note: a first full run had one timeout in `src/server.test.ts` ("delegates a route the serving plane OWNS", 5000 ms). Nothing imports `index.ts`; the file passed 3/3 in isolation and the next two full runs were green. Reporting it as a pre-existing timing flake, not caused by this change.


zach88:
Took a careful pass through this one. It's a text-only fix, so the whole review is really about whether the new migration reference is actually correct. It is — I verified every claim in the description against `control-api/src/db.ts` and the schema helpers, and it all holds up.

**🟢 Ready to approve**
- **Optional / non-blocking: 0**

## What checked out clean

- **The stale `0091` was genuinely wrong.** `0091` is `0091_gfs_desktop_operator_links` (db.ts:5906), nothing to do with uploads.
- **`0097_gfs_upload_sessions` is the right name.** `applyGfsUploadSessionSchema` (db.ts:5930) creates both `gfs_upload_sessions` and `gfs_upload_parts` (the `GRANT`/`REVOKE` block in `gfsUploadSchema.ts` covers both), which are exactly the two tables the `to_regclass` probe in `gfs-controller/src/index.ts:41` checks. New message is accurate.
- **The doc-comment fix is right too.** `applyGfsUploadCleanupSchema` is migration `0098`, which only adds the `cleanup_at` column on top of `0097`'s table, so "additive follow-up for installations that already applied 0097" is correct.
- **Class check on the whole `0091` pattern:** the remaining `0091` hits (`db.ts:2654` generic `0091+` comment, `db.ts:5906` the real registration, `db.ts:6001`/`6013` the oauth rename chain) are all legit. Both stale upload references are fixed and no sibling was left behind.
- **No test asserts the message** (grep on `*.test.ts`/`*.spec.ts` comes back empty), so nothing loses coverage and no repro test is needed for a text-only change.
- Version bump is consistent (`package.json` + `package-lock.json` both at `0.1.581`).

Nothing blocking, nothing to nit. Good to merge.

zach88 (approved):

PR #689, opened 2026-09-18
fix: correct stale control-api migration references and guard them

alfredolopez80:
## Summary

Several comments, error messages and docs cite control-api migrations by bare 4-digit number. After the renumbering in 5e6c990f8 (and in a few older cases, from the start), some of those numbers point at migrations that do something else, so an operator or reader following the hint looks at the wrong schema change.

Every reference was checked against the `apply` body in `control-api/src/db.ts` (`CONTROL_API_MIGRATIONS`), not the migration name alone.

### Wrong references fixed

| Location | Old | Migration that actually does it | Introduced in |
|---|---|---|---|
| `gfs-controller/src/authz/storeProbe.ts` blob schema / reader privileges | 0068 (`member_registration_credentials`) | `0071_gfs_immutable_blob_generations`: adds `blob_key`/`content_sha256` and their column grants | 5e6c990f8 |
| `storeProbe.ts` writer manifest/resource privileges | 0068 | `0071_gfs_immutable_blob_generations`: creates `gfs_blob_manifests` and grants SELECT/INSERT/UPDATE/DELETE to `gfs_controller` | 5e6c990f8 |
| `storeProbe.ts` reader has forbidden mutation privileges | 0069 (`member_registration_runtime_access`) | `0074_gfs_runtime_role_exact_contract`: rebuilds the exact reader envelope, including the `PUBLIC` revokes this check relies on | 5e6c990f8 |
| `storeProbe.ts` reader lacks SELECT / audit INSERT | `0048_gfs_permission_store` (the writer's migration) | `0072_gfs_reader_database_role` creates `gfs_controller_reader` and its grants | 5e6c990f8 |
| `control-api/src/db.ts:4413` (SQL comment inside 0061) | 0055 (`plugin_workload_sdk_grant_provider`) | `0062_governed_trace_runtime_roles`: redefines `governed_trace_prune_expired_events` and creates the runtime roles | 106efe7c4 |
| `control-api/src/services/pluginWorkloadSdkDb.ts:355` | 0027 (`control_admin_email_and_invitations`) | `0034_plugin_workload_sdk`: registers `applyPluginWorkloadSdkSchema` | 5d663a496 |
| `control-api/src/services/pluginWorkloadSdkSchema.ts` (SQL comment inside `CREATE TABLE IF NOT EXISTS plugin_workload_sdk_grants`) | 0050 (`host_wake_generations`) | `0055_plugin_workload_sdk_grant_provider`: adds the `provider` column whose NULL rows the comment describes | 106efe7c4 |
| `deploy/scripts/provision-control-api-runtime-roles.sh:4` | 0053 (`workflow_approval_medium_reply_in_threads`) | `0062_governed_trace_runtime_roles` creates `control_api_runtime`, `trace_maintenance_runtime`, `workflow_recipes_runtime`; later migrations adjust their grants | 106efe7c4 |
| `docs/deploy/gfs-permission-store.md:225` | reader 0071 | `0072_gfs_reader_database_role` (`CREATE ROLE gfs_controller_reader`) | 8f86579a0 |
| `control-api/test/gfsReadiness.realPostgres.integration.test.ts` (2 regexes) | `0068` alternative | `0071_gfs_immutable_blob_generations` / `0048_gfs_permission_store` | — |
| `gfsBlobMigration.realPostgres.integration.test.ts` / `db.gfsPermissionStoreMigration.test.ts` titles | "0068" / "0043" | the version each test actually asserts (0071 / 0048) | — |

Privilege-check errors (0071 blob schema, writer manifest privileges, reader/writer base grants) also name `0074_gfs_runtime_role_exact_contract`, which revokes every table and column privilege and rebuilds both envelopes, so it owns the grants a running cluster has today.

The references that were already correct (0073, 0092, 0095, 0096 in the probe; `index.ts` comment) now use the full version string as well. The message format follows #686.

The remaining bare references in `control-api/src`, `deploy`, `docs` and `scripts` (20 across 12 files) now use full version strings too, so the guard can forbid bare numbers in every scanned root. Each was checked against its migration; the `0050` row above is the only wrong one among them.

### Blast radius

Only error text, comments, docs, one e2e log line and tests change. No readiness condition, executable SQL, fail-closed order, success caching, grant or migration body changes.

- **Runtime:** `storeProbe.ts` now picks the grants migration named in the `can_read`/`can_audit` errors from `opts.storageRole` (typed `"writer" | "reader"`, the same ternary the probe already uses to choose its coherence SQL). The thrown message is the only output; `/readyz` returns it as text.
- **Consumers of the message text** (`git grep` over the repo): the gfs-controller unit tests, the control-api `realPostgres` readiness suites (`gfsReadiness`, `gfsReaderRole`), and the runbook table in `docs/deploy/gfs-permission-store.md`. All three are updated in this PR. The `realPostgres` regexes still match on `coherence check failed` / `permission denied`, which did not change. No alert rule, dashboard, script or log parser in the repo matches these strings.
- **SQL comments:** `db.ts` (inside 0061's query text) and `pluginWorkloadSdkSchema.ts` (a `--` comment inside `CREATE TABLE IF NOT EXISTS`, a no-op on existing databases). `schema_migrations` records only the version and no migration SQL is hashed or checksummed, so already-migrated databases are unaffected.
- **Deploy scripts:** comment-only changes in `provision-control-api-runtime-roles.sh` and `run-control-api-db-migration.sh` (0 non-comment lines in the diff).
- **`scripts/`:** comment changes in `minikube/full-setup.sh` and `minikube/pre-gate-sync.sh`; in `e2e/e2e-figure-d-multibot.sh`, 2 comments and the text of one `log` line, which nothing parses.
- **evenfire-infra** (`keyper-labs/evenfire-infra`, GitHub code search): no references to the probe messages, the guard test, the provision script, the e2e script or the runbook. It invokes `deploy/scripts/run-control-api-db-migration.sh`, whose change here is comment-only.

### Guard test

`control-api/test/migrationReferences.test.ts` scans `gfs-controller/src`, `control-api/src`, `deploy`, `docs` and `scripts`:
- Every version-shaped token (`NNNN_name`, including hex versions such as `00a4_…`) must be registered in `CONTROL_API_MIGRATIONS`, with or without the word "migration" in front of it.
- Only tokens inside a `legacyVersions: [...]` array are exempt. The same legacy name anywhere else fails, because that is exactly what a stale reference looks like (the old "0055" was 0062's legacy name).
- No scanned root may cite a migration by bare number. The check follows "migration(s)" across line breaks, `:`, backticks, parentheses, `–` ranges and `,` / `/` / `and` / `or` lists.
- Every root must yield at least one token (liveness witness); this replaces the fixed `>= 20` threshold. A missing root fails with `scanned root <root> does not exist; update SCANNED_ROOTS`. Two unit cases pin the bare-number and legacy-exemption matchers against known inputs.
- Not covered, as the test header states: whether a registered version is the right one for its context (`0071_…` where `0072_…` belongs passes), and bare numbers not introduced by "migration(s)".
- `control-api/test` is not scanned: it holds seed fixtures and legacy-name lists that are version-shaped on purpose.

It runs in the control-api matrix leg, which `ci-public.yml` runs on every PR with no paths filter.

Mutation checks (each applied, run, and restored byte-for-byte); all fail the guard:
- bare `migration 0068` in a `storeProbe.ts` error
- `migration:` + line break + `0048` in an `index.ts` comment
- unregistered `0072_gfs_reader_role` with no "migration" keyword
- stale named refs in `deploy`, `scripts` and `docs`
- legacy `0055_governed_trace_runtime_roles` cited in a `db.ts` comment
- the stale bare phrasings from review: `Migration 0055` in `db.ts`, `migration 0027` in `pluginWorkloadSdkDb.ts`, `migration 0053` in the provision script, `reader migration 0071` in the runbook, and a bare citation in `scripts/`
- vacuity: token pattern matches nothing; legacy exemption covers the whole file; bare-number scan disabled

A separate mutation hard-codes the writer's migration for both roles in `storeProbe.ts`: the two reader cases in `storeProbe.test.ts` fail.

## Verification (Node 24.18.0)

- gfs-controller: `npx tsc --noEmit` EXIT=0; `npx vitest run` EXIT=0, 613 passed / 9 skipped (skips existed before this change).
- control-api: `npx tsc --noEmit` EXIT=0; `npm test` EXIT=0, 423 files / 5496 tests passed, 196 skipped (the `realPostgres` suites, which run in the separate `control-api-migration` job).
- control-api `realPostgres` against a local PostgreSQL: EXIT=0, 40 files / 193 tests passed.
- `migrationReferences.test.ts`: 12 passed.
- `validate-release-version-bumps.mjs --previous origin/dev` and `update-desktop-release-manifest.mjs --validate-only`: EXIT=0.

The pre-commit hook bumped `control-api/package.json` to 0.1.586 (with its lockfile entry).


zach88:
Took a deep pass through this one. The reference corrections are all solid, security is clean, and the only runtime change (the role-specific grant hint in `storeProbe.ts`) checks out — I confirmed `0072_gfs_reader_database_role` really is the migration that grants the reader its `SELECT`/`INSERT`, so the reader hint points at the right place. The one thing worth a look is the guard test itself: it's narrower than its header suggests.

**🟡 Approve with nits**

None. Nothing here blocks the merge.

## Optional (non-blocking)

1. **[R1-M1] The guard doesn't cover the roots where this PR actually fixed the references** — `control-api/test/migrationReferences.test.ts:35,139-147`. The bare-number check (`MIGRATION_LIST` + `BARE_NUMBER`) only runs over `gfs-controller/src`. The other four roots (`control-api/src`, `deploy`, `docs`, `scripts`) are only pinned for full-name `VERSION_TOKEN`s. But the stale refs this PR corrects in `db.ts` ("Migration 0055"), `pluginWorkloadSdkDb.ts` ("0027"), `provision-control-api-runtime-roles.sh` ("0053") and `docs/deploy/gfs-permission-store.md` ("migration-0048") were all *bare numbers*. Repro: re-injecting those exact stale phrasings into `db.ts`, the provision script and the doc keeps the suite green (8 passed) — so a regression identical to what this PR just fixed would sail through in 4 of the files it touches. The guard meets its "won't reappear" goal only for gfs-controller. It's amplified by the header ("These checks pin every version-shaped token to a registered version") reading as broader coverage than there is. Cheapest fix with no downside: correct the header to say bare numbers outside `gfs-controller` aren't guarded, and add an assertion that the touched files carry no bare-number citation. The broader option (extend the bare-number scan to those four roots) is more robust but risks false positives on any stray 4-digit number in docs.

2. **[R1-L1] Per-root liveness witness is brittle** — `control-api/test/migrationReferences.test.ts:132-133`. `expect(references.length).toBeGreaterThan(0)` per root; `scripts` satisfies it today via just 3 files. If those stop citing a full-name migration, or `scripts/`/`docs/` gets reorganized, the guard fails spuriously on an unrelated change. Also `readdirSync` on a deleted root throws instead of reporting cleanly. Low probability — a comment or a softer guard covers it.

3. **[R1-L2] `VERSION_TOKEN` doesn't validate contextual correctness** — `control-api/test/migrationReferences.test.ts:24`. It only checks a token is *registered*, not that it's the *right* migration for the spot (`0071_…` cited where `0072_…` belongs passes). Inherent limitation and fine, but it's not stated in the header, so a reader may over-trust the guard. Worth folding into the header fix from item 1.

## What checked out clean

- **Role-specific hint** — `storageRole` is typed `"writer" | "reader"` and `config.ts:143-146` throws on anything else, so no third value falls silently into the reader branch. `can_read`/`can_audit` apply to both roles; `0048` (writer) / `0072` (reader) is correct.
- **Hardcoded hints match their checks** — `0074_gfs_runtime_role_exact_contract` revokes everything and rebuilds both envelopes, so it owns the reader's absence of mutation privileges (check at `:346`); `0071` defines the manifest schema (`:299`/`:361`); `0073`/`0092`/`0095`/`0096` all exist under their full names. Old `0068`/`0069` now point at `member_registration_*` — staleness confirmed.
- **Tests** — the new regexes embed the full migration name plus a trailing anchor, so they wouldn't match the old bare-number text; they genuinely pin the new strings. Assertions target the observable error message, not call counts. The new `0073` cases aren't tautological.
- **Guard bites where it covers** — bare `0068` and an unregistered `9999_nonexistent_thing` injected into `gfs-controller/src` fail 2 tests as expected.
- **Security** — `/readyz` messages carry only migration names and privilege descriptions, no DSN/password/token interpolation. The provision script diff is comment-only (0 executable lines). Fail-closed intact — only SUCCESS is cached, no readiness condition changed.
- **Suites** — `storeProbe.test.ts` 45 passed; `migrationReferences.test.ts` 8 passed; `db.gfsPermissionStoreMigration.test.ts` 9 passed. The `*.realPostgres.integration` suites need a live Postgres and weren't run; their regex changes match the new `storeProbe.ts` strings by inspection.

zach88 (approved):

alfredolopez80:
@zach88 thanks for the review. I addressed all three items in 719ae84f1.

**R1-M1: fixed.** The bare-number check now runs over all five scanned roots, not only `gfs-controller/src`. To make that pass, the 20 remaining bare citations in `control-api/src`, `deploy`, `docs` and `scripts` (12 files) now use full version strings. I reran your repro: `Migration 0055` in `db.ts`, `migration 0027` in `pluginWorkloadSdkDb.ts`, `migration 0053` in the provision script and `reader migration 0071` in the runbook each fail the guard now, and so does a bare citation injected into `scripts/`. I also added `–` ranges (`migrations 0109–0114`) to the matcher, and the matcher self-test now covers them.

Converting those references turned up one more stale one that my earlier audit had marked correct: `pluginWorkloadSdkSchema.ts` said rows written before "migration 0050" have a NULL provider. 0050 is `0050_host_wake_generations`; the `provider` column comes from `0055_plugin_workload_sdk_grant_provider`. It is a SQL comment inside `CREATE TABLE IF NOT EXISTS`, introduced in 106efe7c4, and it is fixed now.

On the header: the previous one did scope the bare check ("forbid bare numbers in gfs-controller"), but "pin every version-shaped token" read wider than it was. The new header says both checks apply in every scanned root and lists what they do not cover.

**R1-L1: kept by design, with a clearer failure.** The per-root witness stays: if a root stops yielding any token, the scan has most likely stopped reading its files, and I would rather see that fail. A missing root no longer throws a raw `ENOENT` from `readdirSync`; it fails with `scanned root <root> does not exist; update SCANNED_ROOTS`.

**R1-L2: documented.** The header now says the guard does not check whether a registered version is the right one for its context (`0071_…` where `0072_…` belongs passes), and does not catch bare numbers that are not introduced by "migration(s)".

Blast radius of the new commit: comments, docs, one SQL `--` comment inside `CREATE TABLE IF NOT EXISTS` and the text of one e2e `log` line. `schema_migrations` records only the version and no migration SQL is checksummed. `keyper-labs/evenfire-infra` does not reference any of the changed text; it only invokes `run-control-api-db-migration.sh`, and the change there is a comment.

Verification at 719ae84f1:
- control-api: `tsc` EXIT=0; `npm test` EXIT=0, 5496 passed / 196 skipped.
- control-api `realPostgres` against a local PostgreSQL: EXIT=0, 40 files / 193 passed.
- gfs-controller: `tsc` EXIT=0; `vitest` EXIT=0, 613 passed / 9 skipped.
- `migrationReferences.test.ts`: 12 passed. The earlier 10 mutations and the 5 new repros all fail the guard.
- Release validators: EXIT=0.

Issue #691, opened 2026-09-18
bug(hcc): reconcile_outcome 409 for recreated Hosts - v2 source id omits the Host uid that control-api hashes


alfredolopez80:
## Summary

On prod `clerum`, control-api rejects HCC `reconcile_outcome` telemetry with a 409 idempotency conflict
for one Host, `mcp-host/chatllm`, about every 150 s. The key is
`infrastructure_telemetry:hcc_internal_control:hcc-reconcile-outcome-v2:<sha>`.

Measured from Cloud Logging (read-only):
- 4,211 409 lines between 2026-09-16T07:31:15Z and 2026-09-18T19:23Z, all with the same key. That is 1,402 bursts of 3 plus one burst of 2; the 3 per burst come from the retries in #326.
- An earlier wave followed the same pattern: 3,168 lines of another key, for Host `alfredo-dev` at generation 1, from 2026-08-19 (the start of log retention) to 2026-08-21T09:33Z.

The key, recomputed locally over the tuple in `hccReconcileOutcomeSourceId`, matches exactly
`["mcp-host","chatllm",1,"ready",null,"deployed","ready","succeeded",null,"Host",null,...null]`.

Code read at `origin/dev@05aebc259`.

## Cause

**The Host uid is hashed by control-api but is not part of HCC's key.**

- HCC builds the source id from namespace, name, generation and payload fields
  (`host-context-controller/src/infrastructureTelemetryReporter.ts:63-87`). It does not include
  `metadata.uid`. `hostLookupReference` (`hostReconciler.ts:2157-2167`) sends only name, namespace and
  generation, although `HostCRD.uid` exists (`types.ts:567`).
- control-api resolves a binding for each event and puts it in the canonical hash. The binding includes
  `kubernetesUid: metadata.uid` from the live Host (`hccHealthTransitionBindingResolver.ts:112`).
  `occurredAt` is already excluded for `reconcile_outcome`.
- chatllm was deleted and recreated with the same name. The audit log shows a create on 09-11, a
  delete on 09-13 and a create on 09-16 06:26:45. A recreated Host starts at generation 1, so the new
  object produces a key that already exists, with a different uid in the binding. `append.ts:430`
  then throws `TracingIdempotencyConflictError`.

That `kubernetesUid` is the differing field is inferred by elimination. Every other hashed field is
either in the key or constant. The stored row itself was not read.

## Why only some passes collide (~150 s)

[...]

Issue #328, opened 2026-08-12
bug(hcc): reconcile_outcome telemetry silently lost (400) - gfs_* evidence keys are not allowlisted


jozer-rami:
## Summary

`enqueueReconcileOutcome` spreads the raw GFS token lifecycle evidence object into the
telemetry payload. Four of its keys are not in control-api's payload allowlist, so
control-api rejects the whole event with `400`.

Because `reconcile_outcome` events are enqueued **without** a `sourceEventId`, these events
cannot be deduplicated or replayed. They are **permanently lost**, not merely retried. This
is silent data loss in the infrastructure telemetry stream.

**~25 rejections per 15 minutes** on prod.

## Evidence (prod, 2026-08-12)

```
{"level":40,"svc":"control-api","status":400,
 "err":"tracing input contains a server-owned or monetary field: input.payload.gfs_subject"}
```

The GFS "reused" path is a common one, visible in HCC logs at the same time:

```
{"svc":"host-context-controller","module":"host-reconciler",
 "msg":"reusing mcp-host-runtime-token Secret","host":"chatllm","namespace":"mcp-host"}
```

## Root cause

`host-context-controller/src/hostReconciler.ts:1593`, inside `enqueueReconcileOutcome`:

```ts
this.enqueueHostTelemetry(host, 'reconcile_outcome', succeeded ? 'ready' : 'not_ready', {
  status: succeeded ? 'succeeded' : 'failed',
  phase: status.deployed ? 'deployed' : 'not_deployed',
  state: status.ready ? 'ready' : 'not_ready',
  ...(this.gfsTokenLifecycleEvidence.get(HostReconciler.gfsLifecycleEvidenceKey(host)) ?? {}),
})
```

The spread injects the whole `GfsTokenLifecycleEvidence` shape
(`hostReconciler.ts:319-324`):

```ts
type GfsTokenLifecycleEvidence = {
  gfs_subject: string
  gfs_outcome: 'minted' | 'rotated' | 'reused' | 'failed'
  gfs_old_host_uid?: string
  gfs_new_host_uid?: string
}
```

control-api's `assertSafeEventPayload` (`control-api/src/services/tracing/append.ts:233`) is
a strict **allowlist**, not a denylist:

```ts
for (const [key, fieldValue] of Object.entries(value)) {
  if (!SAFE_METADATA_KEYS.has(key) || !['string','number'].includes(typeof fieldValue) ...) {
[...]

Issue #327, opened 2026-08-12
bug(hcc): administrative outcome re-reported every reconcile causes permanent 409 idempotency conflict


jozer-rami:
## Summary

HCC re-reports an administrative outcome on **every** reconcile of an already-reconciled
Host. The submission carries a **stable** idempotency key but a **fresh** `occurredAt`,
and control-api folds `occurredAt` into the payload hash. The result is a permanent,
self-perpetuating `409` that can never succeed.

Highest-volume contributor to the HCC tracing rejection storm: **~35 per 15 minutes** on
prod, indefinitely.

## Evidence (prod, 2026-08-12)

```
{"level":40,"svc":"control-api","status":409,
 "err":"conflicting payload for administrative:hcc_internal_control:hcc-admin-outcome:85f5135a-52d9-4778-8897-7ee882de2c1a:2:succeeded"}
```

Database proof that the write is a genuine duplicate, not a lost write:

```sql
SELECT count(*) AS rows_stored, count(DISTINCT source_event_id) AS distinct_ids
FROM administrative_events WHERE source_event_id LIKE 'hcc-admin-outcome:%';
--  rows_stored | distinct_ids
--           21 |           21
```

**Every event landed exactly once.** Nothing is being lost; HCC is simply re-sending
forever. The event above (`85f5135a...:2:succeeded`, host `alfredo-dev`) was stored at
`2026-08-11 20:37:29` and was still being rejected at `15:13` the following day — **19
hours of retrying a row that already exists.**

Affects the 6 of 8 Hosts that carry a `clerum.io/administrative-intent-id` annotation
(`chatllm` and `trader` have none and never enqueue).

## Root cause

`host-context-controller/src/hostReconciler.ts:1606` builds a deliberately stable identity
but a moving timestamp:

```ts
this.administrativeOutcomeReporter?.enqueueHostOutcome({
  sourceEventId: `hcc-admin-outcome:${operationId}:${host.generation}:${outcome}`,
  occurredAt: this.now().toISOString(),   // <-- fresh on every reconcile
  ...
})
```

`enqueueAdministrativeOutcome` is called unconditionally from `enqueueReconcileOutcome`
(`hostReconciler.ts:1596`) whenever a reconcile succeeds:

```ts
if (succeeded) this.enqueueAdministrativeOutcome(host, 'succeeded', 'reconciled')
```

Neither `operationId` (an annotation) nor `host.generation` changes between reconciles of a
[...]

Issue #607, opened 2026-09-10
SecretInformer watches without resourceVersion, replaying every Secret on each reconnect (~114 no-op reconciles/hour)


alfredolopez80:
## Summary

`SecretInformer` opens its watch with an empty query object, so no `resourceVersion` is
sent. Every reconnect makes the apiserver replay a synthetic `ADDED` for every Secret in the
namespace, and each one drives a full McpServer reconcile. On `clerum-dev` that is roughly
114 reconciles per hour that correspond to no actual change.

Code verified against `origin/dev` = `c7741481a`, read with `git show c7741481a:<path>`.
Cluster state and log counts read on 2026-09-10.

## The call

`host-context-controller/src/informers/secretInformer.ts:151`:

```ts
const request = await this.watch.watch(path, {}, watchCallback, doneCallback)
```

No `resourceVersion`, so the watch starts from the current state rather than resuming, and
the apiserver replays the whole collection as `ADDED` events. No `timeoutSeconds` and no
`allowWatchBookmarks` either, so the reconnect cadence is set by whatever idle timeout sits
in front of the apiserver — the same transport behaviour described in #604.

The informer reconnects on both error and clean close (`:140-148`).

## The cost

Each replayed event reaches `reconcileByEnvSecret` (`k8sClient.ts:2563-2601`), which selects
every cached McpServer whose `spec.envSecret.name` matches and enqueues
`reconciler.reconcile(current)` for each.

On `clerum-dev`:

- 52 Secrets in `mcp-server`, and 52 of the 111 McpServers carry an `envSecret`.
- Measured over 3 hours of HCC logs: 7 watch reconnects and **343** occurrences of
  `Re-reconciling McpServer "<name>" after Secret "<name>" change` — consistent with one
  replay per Secret per reconnect.

That is ~114 reconciles per hour triggered by no change at all. Each one re-reads the
Secret, re-validates it, and re-runs `performReconcile`.

## Why it matters

This is write amplification of the same kind the #582/#584/#604 work is tracking, in a
component none of those issues covers. It is also a confounder for that work: the reconcile
counts attributed to McpServer watch churn include this traffic.

Unlike #604 the amplification is mostly reads and no-op writes rather than accepted deletes,
so the etcd cost is lower. It has not been measured against the audit log.

## Suggested direction

Not prescriptive. Resume the watch from the last observed `resourceVersion`, and handle
[...]

PR #692, opened 2026-09-18
fix(hcc,control-api): stop tracing idempotency 409s for admin outcomes and recreated Hosts

alfredolopez80:
## Summary

HCC submits governed tracing events whose idempotency keys do not cover every field control-api hashes. When the same key arrives with a different hash, control-api answers `409 tracing_idempotency_conflict`, and HCC's shared reporter queue retries every non-2xx twice (#326). This PR removes the two sources of those 409s, stops the `gfs_*` 400s, and makes HCC stop retrying deterministic rejections.

Measured on 2026-09-18 (read-only Cloud Logging):
- Admin-outcome 409s: 1,269/h on prod (9 keys), 630/h on dev (7 keys).
- Reconcile-outcome 409s: 4,211 on prod between 09-16 and 09-18, all with one key, all for `mcp-host/chatllm`. The audit log shows that Host was created on 09-11, deleted on 09-13 and created again on 09-16.

## Changes and why

**B. control-api: `occurredAt` is no longer hashed for HCC linked outcomes (#327)**
`administrativeEvents.ts`. The key `hcc-admin-outcome:<operationId>:<generation>:<outcome>` names a state fact. HCC has no stable observation time: it re-observes the outcome on every reconcile pass and after every restart, and sends `occurredAt: now()`. Because the hash included `occurredAt`, every resend became a conflict. The first `occurredAt` is still stored. `reasonCode`, `sourceStatusRef`, the payload and the server-resolved binding still distinguish a genuinely different claim. Every other submitter keeps `occurredAt` in the hash.

**A. HCC: each administrative outcome is sent once per process (#327)**
`administrativeOutcomeReporter.ts`. A bounded set (1,024 keys) holds each `sourceEventId` while it is queued, in flight, accepted or rejected as terminal. A repeated enqueue is counted as `deduplicated`, skipped, and moves the key to the most recently seen position; the least recently seen key is evicted first. Every drop (buffer full, retries exhausted, shutdown) deletes the key, so the next pass enqueues it again and no outcome is lost. `dedupeCapacity` must exceed the queue capacity, because the buffer holds `capacity` keys plus one in flight. Without this, the 3-4 fleet passes every 5 minutes (driven by watch closures, #604) would keep resending the same fact.

**D. HCC: a failed pass reports the failure, not the previous status (#327)**
`hostReconciler.ts`. On the exception path, `enqueueReconcileOutcome` read the status left by the previous pass. If that pass was ready, it emitted an administrative `succeeded` alongside the `failed`, and `reconcile_outcome` telemetry as `succeeded`/`ready`. That path now emits `reconcile_outcome` as `{status: failed, reason_code: reconcile_exception}` (plus the GFS transition, if the pass recorded one) and no administrative `succeeded`. A pass withdrawn by a benign supersession drops its GFS evidence: that exit emits no outcome, and the evidence map is keyed by namespace/name, so the next pass, or a Host recreated with the same name, would otherwise report it. The stateless suspension path also calls `reconcileCore` without emitting an outcome and does not clear the entry; that is tracked in #696.

**E. Host uid in the reconcile-outcome identity (#691)**
- HCC: `hostLookupReference` now carries `uid` on every telemetry type, including `health_transition`, which used to build its own reference. The reconcile-outcome source id tuple includes it, and the prefix becomes `hcc-reconcile-outcome-v3`.
- control-api: `hostLookupReference.uid` is optional. A non-string or empty value is rejected with 400. When the uid is present and does not match the live Host, the HCC binding resolver returns no binding, so an event queued for a deleted object is not bound to its successor.
- Current HCC does not send `uid`, and binding behaviour for those events is unchanged.

**C1. control-api: stable `code` on allowlisted 4xx responses (#326, partial)**
`errorHandler.ts`. `tracing_idempotency_conflict`, `unsafe_tracing_input` and `invalid_tracing_input` are forwarded as `code` in the response body and in the `forwarded_client_error` log. No other 4xx response changes. These are the tracing rejections that return the same answer for the same input; `invalid_tracing_input` comes only from request shape validation.

**C2. HCC: deterministic rejections are not retried (#326, partial)**
- `boundedOffPathReporter.ts`: adds `ReporterTerminalError` and a required `onTerminal` callback.
- `reporterHttpFailure.ts`: only `409` with `tracing_idempotency_conflict`, and `400` with `unsafe_tracing_input` or `invalid_tracing_input`, are terminal. Everything else stays retryable: 403, 5xx, a 4xx without a code, or a body that is not JSON. Only the `code` field of the body is read, and the body is never logged. When the code cannot be read, a warning records the status and the failure kind (`not_json`, `aborted`, `read_failed`).
- The results are counted as `conflict` and `rejected`: on `clerum_hcc_administrative_outcome_reporter_total` for administrative outcomes and on `clerum_hcc_infrastructure_telemetry_flushes_total` for telemetry. A `rejected` telemetry event is never stored, so it also counts in `clerum_hcc_infrastructure_telemetry_gaps_total`; a `conflict` does not, because a row exists for the key. That row has a different hash, so whether a conflict should count as a gap is open in #696.

**F. GFS token evidence travels as `transition` (#328)**
- The `gfs_*` payload keys were never accepted: they are outside the metadata allowlist, and nothing reads them.
- HCC now sends the outcome as `transition: gfs_token:<minted|rotated|reused|failed>`. `transition` is already accepted by ingest, by the DB CHECK and by both infrastructure read profiles, and HCC already uses it for `lifecycle:<state>`.
- `gfs_subject` is derivable from the binding. `gfs_new_host_uid` is now in the key and the binding through E. `gfs_old_host_uid` came from an unvalidated Secret annotation.
- control-api used the server-owned/monetary message for every `unsafe_tracing_input`. It now names the actual problem; the code stays `unsafe_tracing_input`:
  - `tracing input field is not permitted: <field>` for a payload key outside the allowlist or an unknown event field;
  - `tracing input value is not accepted: <field>` for an allowlisted key with a bad type, format or size;
  - `tracing input contains a server-owned or monetary field: <field>` only for that case.

## Tests

Each negative assertion has a liveness witness in the same test. Vacuity mutations were run and reverted:

| Mutation | Result |
|---|---|
| B: drop `occurredAt` for every principal / for no principal | red (other-submitter 409 test / HCC replay test) |
| A: disable the dedupe check | red, 4 tests |
| D: pass `administrative: true` on the exception path | red. A first version of this test failed the pass through a GFS mint error, which resets status first. It stayed green under the mutation and was rewritten to fail at RBAC, before any status write. |
| C2 queue: ignore `ReporterTerminalError` | red, 3 tests |
| C2 classification: nothing terminal | red |
| C1: stop forwarding `invalid_tracing_input` | red, 2 tests (handler, route) |
| C2: `invalid_tracing_input` not terminal | red |
| E: leave `uid` out of the tuple | red |
| E: `health_transition` builds its reference without `uid` again | red |
| C1: stop forwarding `code` in the global handler (route-level test) | red, 2 tests |
| F (HCC): stop setting `transition` | red |
| F (control-api): allowlist rejection with the old message | red |
| D: the exception path reports the previous status again | red, 4 tests |
| D: keep GFS evidence after a superseded pass | red |
| F (HCC): report `minted` as `reused` / `reused` as `rotated` / never record `failed` | red, 1 test each |
| A: dedupe only settled keys | red, 2 tests |
| A: release the key only after `retry_exhausted` | red (`buffer_full` test) |
| A: no refresh on a hit | red |
| A: accept `dedupeCapacity <= capacity` | red |
| C2: ignore the status of a terminal pair | red, 4 tests |
| C2: log nothing / one reason for an unreadable code | red, 3 tests / 2 tests |
| C2: no gap for `rejected` / a gap for `conflict` too | red |
| E: accept an empty `uid` | red |
| F (control-api): unknown event field / bad value with the server-owned message | red |

Gates:
- host-context-controller: `npm test` 112 files / 2,465 tests, build, `tsc --noEmit`.
- control-api: `vitest run --no-file-parallelism` 423 files / 5,501 tests passed (41 files skipped: the real-Postgres suites, which need a DSN and run in their own CI lane), build, `tsc --noEmit`.

Neither package has a working lint gate. control-api has no `lint` script. The host-context-controller `lint` script fails before this change because the package has no ESLint config. CI runs lint for neither package.

## Rollout

1. Deploy control-api first. The hash change, the optional `uid` and the `code` field are backward compatible with the current HCC.
2. Then deploy HCC.

Rows written before this change keep the old hash. After the HCC rollout, each old admin key gets one 409 per process, is marked terminal and is not sent again. The dedupe map lives in memory, so that single 409 comes back after each HCC restart, and after the key is evicted from the 1,024-entry map, until the Host's generation changes. No new v2 reconcile keys are produced, and no migration is needed.

After the deploy, verify with read-only Cloud Logging:
- admin-outcome 409s: at most one per live key after each HCC restart, then 0;
- `hcc-reconcile-outcome-*` 409s on prod: 0;
- `gfs_*` 400s: 0.

## Out of scope

- #329: a 403 from a stale intent generation stays retryable.
- #693: after the HCC rollout, require the uid for HCC references and reject unknown reference keys.
- #694: the administrative outcome key and binding carry no Host uid.
- #696: GFS token evidence recorded by the stateless suspension path is reported by a later pass or a recreated Host; plus two Low findings from the blast-radius review.
- #604 / #607: bookmarks on the remaining watches.
- These fallbacks are reported but not changed. Both turn a Kubernetes lookup failure into "Host not found":
  - `adminOperationBindingResolver.ts:60-63`: `catch { return [namespace, []] }`
  - `hccHealthTransitionBindingResolver.ts:84`: `catch { return null }`

Fixes #327
Fixes #691
Fixes #328
Refs #326


zach88:
Took a deep pass through this one across both sides of the contract (control-api tracing + HCC reporters). Solid PR: the scope is tight, the two contract changes are backward-compatible both ways, and the test suite is genuinely good (every negative assertion carries a liveness witness, and the vacuity mutations in the body check out). I only found two Low nits, and both are optional.

🟡 **Approve with nits**
- **Optional / non-blocking: 2**

## Optional (non-blocking)

This PR is mergeable as-is; both of these are follow-up or author's call.

1. **A deterministic `invalid_tracing_input` (400) isn't classified as terminal, so it gets retried and counted as a gap.** The `uid` validation this PR adds rejects a non-string/empty uid with `InvalidTracingInputError` (`invalid_tracing_input`, 400) in `control-api/src/services/tracing/routeSubmissionService.ts:307`, but the terminal classifier in `host-context-controller/src/reporterHttpFailure.ts:33-40` only treats `unsafe_tracing_input` (400) and `tracing_idempotency_conflict` (409) as terminal. An `invalid_tracing_input` is just as deterministic, so HCC retries it to the limit and books it as a gap in `infrastructureTelemetryGapsTotal` instead of settling it. In practice `host.uid` comes from a persisted k8s `metadata.uid` (always a non-empty UUID), so HCC never actually emits a bad uid — this only bites under schema drift, which is why it's Low. It's the uncovered sibling instance of exactly the class C2 fixes. If you want to close it: add `invalid_tracing_input` to `FORWARDABLE_CLIENT_ERROR_CODES` and to `TERMINAL_RESPONSES` (400, rejected).

2. **A couple of reporter test fixtures represent responses the real producer never emits (T1).** In `host-context-controller/src/reporterHttpFailure.test.ts` and `administrativeOutcomeReporter.test.ts`, some non-terminal cases use bodies like `{ code: 'invalid_tracing_input' }` / `{ code: 'tracing_binding_unavailable' }` and `(400, 'tracing_idempotency_conflict')`. `control-api/src/http/errorHandler.ts` only sets `code` in the body for the two allowlisted codes, so those bodies diverge from what the server actually sends. No behavioral impact — the retryable assertion matches reality, and the classifier is a pure function that's fair to test with synthetic inputs — flagging it as a note, not a bug.

## What checked out clean

Verified, not assumed:

- **B (occurredAt dropped from the hash):** gated on `principal.kind === 'hcc_internal_control' && input.kind === 'linked_outcome'`, derived from the JWT issuer, not the body — not reachable by a session/user client. Other submitters (`wrc_internal_control`) still hash `occurredAt`. First `occurredAt` is preserved. The replay regression test fails against the merge-base (T3 confirmed).
- **E (uid):** validated (non-string/empty → 400); absent = prior behavior (old-HCC compat). A forged uid can only *fail* the binding, never redirect to another Host — no IDOR/cross-tenant. Security pass came back clean.
- **A (dedupe LRU):** no outcome is lost — every drop deletes the key, accepted/terminal keep it, and eviction only produces a benign replay. `dedupeCapacity > capacity` enforced in the constructor.
- **C2:** the `return` after `onTerminal` doesn't stall the drain (the `finally` reschedules). Only (409, idempotency_conflict) and (400, unsafe_tracing_input) are terminal; 403/5xx/no-code/non-JSON stay retryable and bounded by `retryLimit`.
- **D (exception path):** all three post-`reconcileCore` exits clear the GFS evidence, and the exception path no longer emits a false administrative `succeeded`.
- **F + contract:** `transition: gfs_token:<state>` is accepted by control-api's metadata allowlist; the old `gfs_*` keys were never accepted (so this also fixes a latent 400). v2→v3 prefix has zero consumers. Both contract directions (old-client↔new-server and vice versa) hold.
- **Rollout:** the control-api-before-HCC order in the body is correct and confirmed in code — HCC's terminal classification needs a control-api that already forwards `code`.
- CI is green across the board (control-api, host-context-controller, Real Postgres integration, leak guards).

zach88 (approved):

alfredolopez80:
Thanks for the review. Both nits are addressed in 07acbe920.

1. **`invalid_tracing_input` is now terminal.** Every `InvalidTracingInputError` comes from request shape validation in `routeSubmissionService.ts`, so the same input always gets the same 400. control-api forwards it as `code`, and HCC settles `(400, invalid_tracing_input)` as `rejected`. It still counts as a telemetry gap, because the event is not stored. Tests: the handler and the real router both return the code, and the classifier treats the pair as terminal; `(409, invalid_tracing_input)` stays retryable. Vacuity mutations: removing the code from the handler allowlist fails 2 tests, and removing the classifier entry fails 1.

2. **Fixtures.** The administrative reporter's retryable cases no longer use bodies control-api never sends. The 403 with `tracing_binding_unavailable` is removed (the plain 403 case covers it), and the 400 case has no code. I kept the status-pinning cases in `reporterHttpFailure.test.ts` (`400 + tracing_idempotency_conflict`, `409 + unsafe_tracing_input`, and so on) on purpose: they test that the classifier matches the (status, code) pair rather than the code alone.

Issue #667, opened 2026-09-17
feat(files): integrate shared previews in Desktop and Control UI GFS browsers


alfredolopez80:
Coordination: #665.

## Problem and outcome

Connect the existing file browser and rendering components into a consistent human preview/download experience in Desktop and Control UI. Reuse existing Markdown/text/image rendering and Desktop HTML components after their safety requirements are met. Human viewing is independent of whether the selected model understands the file.

## Scope

- Consume the #666 FileReference/classifier contract and implement one shared preview-kind/policy registry, with framework-specific rendering adapters only where necessary.
- Integrate Markdown, plain text/code, HTML, JPEG, PNG and SVG with both GFS browsers and their preview modals. Code is displayed as inert text.
- Reuse the same classification for applicable Desktop attachment previews.
- Offer original-file download from both interfaces using existing authorized routes. Preserve original bytes, name and MIME metadata.
- Provide loading, empty, unsupported, too-large, denied, missing and stale-version states. Include a PDF registry entry with an honest download-only state until the human PDF viewer #668 ships.
- Support accessible modal focus, keyboard dismissal, descriptive titles and safe cleanup/cancellation when switching files or closing previews.

## Non-goals

PDF rendering or extraction, agent vision, new storage, WYSIWYG/office editing, arbitrary web browsing, script execution, changes to GFS grants, upload-limit increases or a replacement for existing working renderers.

## Dependencies

- Hard dependency: #666 FileReference/classifier contract for shared identity and classification.
- [#187](https://github.com/evenfire-ai/evenfire/issues/187) owns Desktop HTML network-denial hardening. Reuse and verify that policy in new Desktop and Control UI HTML entry points; HTML release is gated on equivalent protection, without duplicating [#187](https://github.com/evenfire-ai/evenfire/issues/187).
- [#585](https://github.com/evenfire-ai/evenfire/issues/585) owns per-child readability metadata. Consume it when present; absence means unknown, and the server still authorizes every read/download.
- [#538](https://github.com/evenfire-ai/evenfire/issues/538) owns native Desktop embed visibility coordination. Integrate with its modal mechanism; verify overlapping consent/deep-link/preview dialogs before releasing affected Desktop paths.
- [#586](https://github.com/evenfire-ai/evenfire/issues/586) metadata is optional UI enrichment, not a preview blocker.
[...]

PR #697, opened 2026-09-19
feat(desktop): GFS file explorer tree and in-tab file previews

zach88:
> Stacked on top of #676 (`feat/universal-tabs`). Base is `feat/universal-tabs`, not `dev`, so the diff shows only this unit's changes; retarget to `dev` once #676 merges.

## What
- Adds a GFS (EvenDrive) file explorer tree to the desktop-app sidebar, letting the user browse the shared filesystem without leaving the current view.
- Opens GFS file previews (image, markdown, video) in a dedicated workspace tab instead of a modal layout, riding on the universal workspace-tab store.
- Splits each preview into a presentational `Body` plus its container, and centralizes preview-kind resolution and download URL building in `lib/gfsPreview.ts` / `lib/gfsDownload.ts`.
- Sanitizes user-/plugin-controlled names (GFS entry names, files tab paths, preview tab titles) at the store border before they reach tab chrome.

## Why
The previous flow surfaced GFS files only through the Files page and a modal-style preview layout, with no persistent way to browse the tree. Moving previews into workspace tabs makes files first-class alongside chats and apps, and lets several previews stay open at once. Sanitizing at the store border keeps attacker-influenced filenames from injecting control characters into the tab strip chrome.

## How
- The file explorer tree lives under `SidebarNav/FileExplorerTree`; the sidebar fills available height with a contained scroll region so the tree does not push the nav off-screen.
- Preview tabs are seeded through the workspace-tab store; `FilePreviewPage` renders the correct `Body` by resolved preview kind, and the old `GfsPreviewWorkspaceLayout` is removed.
- Name sanitization is applied at the store setters (`setFilesTabPath`, preview/app title borders) rather than at each render site, so the invariant holds for every consumer of the tab title.

## Testing
`cd desktop-app/ui && npm test` over the touched suites:
- `lib/__tests__/workspaceTabs.test.ts`, `workspaceTabsRoute.test.ts` — title sanitization and route derivation.
- `lib/__tests__/gfsPreview.test.ts` — preview-kind resolution.
- `components/SidebarNav/FileExplorerTree/__tests__/FileExplorerTree.test.tsx`, `components/__tests__/SidebarNav.test.tsx` — tree rendering and name-in-chrome sanitization.
- `pages/__tests__/FilePreviewPage.test.tsx`, `FilesPage.test.tsx`, plus per-preview `*Body.test.tsx` (image/markdown/video).

## Risks / Follow-ups
- Size: 40 files (~13 of them tests), single service (`desktop-app/ui`), one cohesive feature — above the soft 25-file budget but not splittable without separating the tree from the previews it feeds. Kept as one unit deliberately.
- Depends on #676; must not merge before it. GitHub will retarget the base to `dev` automatically when #676 lands.


jozer-rami:
## Deep review (`d8ba85887` vs `origin/dev`)

Reviewed the unit-18 delta after #676 landed. Base is `dev`; the branch is 0 behind. Desktop CI is green. CodeQL already flagged the unused `fireEvent` import in `GfsImagePreviewBody.test.tsx`.

Independent lenses: correctness, security-tenancy, test-integrity, architecture/spec 18. No blockers (server ACL still denies unauthorized reads). Several majors should land before merge.

### What's solid

- `workspaceTabs` stays a pure algebra. Preview tabs dedupe by `gfsUri` with property tests. `mapKindToRoute('preview')` is wired.
- Only the **active** preview tab mounts `FilePreviewPage`. Switching unmounts the body and revokes object URLs.
- `resolveGfsPreview` is the shared Files + tree preview-vs-download decision. All three bodies size-guard listing size **and** downloaded bytes.
- New persistent chrome (tab titles, tree labels, toggle `aria-label`, download toast) runs `sanitizeAppTabTitle` (bidi, zero-width, controls, length).
- Markdown is the vanilla parser: no HTML, images become `[Image: alt]`, `javascript:` hrefs are nulled. Electron `will-navigate` is a second belt.
- Plugin **image** overlay is still the intended modal. `GfsPreviewWorkspaceLayout` is gone.

---

## Majors

### 1. Explorer treats `readable: false` rows as openable

[`FileExplorerTree/index.tsx` L257–L279](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/components/SidebarNav/FileExplorerTree/index.tsx#L257-L279)

FilesPage already refuses these rows (folder grant without inheritance) with a toast and a “No access” badge ([`FilesPage.tsx` L1020–L1028](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/pages/FilesPage.tsx#L1020-L1028), test at `FilesPage.test.tsx:432`). The tree never reads `node.readable`. The Files ⋯ Preview item at [`FilesPage.tsx` L1438–L1441](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/pages/FilesPage.tsx#L1438-L1441) has the same hole.

**Trigger:** Expand a folder whose grant does not inherit. Double-click or Enter `locked.md` / `sealed/` in the sidebar.

**Failure:** A preview tab opens and 403s, or `gfs.download` runs, or a files tab is created at an unreadable folder URI. The click looks allowed.

**Fix:** Gate `handleActivateFile` / `handleActivateFolder` (and the Files ⋯ Preview item) on `readable !== false`. Same toast as FilesPage. Mark the row in the tree.

---

### 2. Session revoke in the explorer looks like an empty drive

[`FileExplorerTree/index.tsx` L295–L306](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/components/SidebarNav/FileExplorerTree/index.tsx#L295-L306) · [`useGfsBrowserController.ts` L201, L387–L389](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts#L201)

`accessState` is per hook mount. Explorer, FilesPage, and FilePreviewPage each have their own. `revokeAccess()` clears the **shared** `gfsRoot` cache and sets **this** instance to `'revoked'`. The tree never branches on `accessState`. After `removeQueries`, `accessibleError` is gone and `hasMoreAccessible` is false, so the UI settles on “No shared files yet.”

**Trigger:** Expand the sidebar explorer, then hit a session 401 on discovery or `listChildren`. Files page shows “File access is not authorized” + Retry. Retry flips only the FilesPage instance.

**Failure:** Sidebar claims there are no shares. Retry on Files does not re-enable the still-mounted explorer (`enabled: … && accessState === 'active'` stays false). Collapse/re-expand is the only recovery.

**Fix:** One shared access flag (or query-client-scoped revoked state). Explorer renders the same unauthorized empty-state + `retryAccess()`. FilePreviewPage should not stand up a full browser only to watch `accessState`.

---

### 3. Plugin markdown/video `gfs://` leaves a files-tab landmine

[`App.tsx` L840–L859](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/App.tsx#L840-L859) · [`FilesPage.tsx` L537–L547, L562–L568](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/pages/FilesPage.tsx#L537-L547)

De-modalizing previews taught FilesPage `handleOpenGfsLink` to call `openFilePreview`, which now **switches the active tab**. The plugin handler still uses image-mime only and sends everything else to `openFilesSection(resource.gfsUri)`. The comment still says “Folders and non-previewable files.”

**Trigger:** In a plugin, open a `gfs://` `README.md` (or mp4).

**Failure:**
1. A files tab is created with `path =` that file URI.
2. FilesPage seed runs `handleOpenGfsLink` → preview tab opens → FilesPage unmounts.
3. Location reporting can persist the file URI on the leftover files tab (`FilesPage.tsx` L578–L588).
4. Selecting that files tab remounts FilesPage with the same seed and jumps to preview again.

Before this PR the same handoff stayed on one files view with a modal.

**Fix:** Plugin file opens go through `resolveGfsPreview` → `openPreviewSection`. Only folders / non-previewable files call `openFilesSection`. Do not persist a previewable file URI as a files-tab path.

---

### 4. Sticky preview payload + uncapped `arrayBuffer()` skips the metadata size-guard

[`workspaceTabs.ts` L310–L317](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/lib/workspaceTabs.ts#L310-L317) · [`GfsImagePreview/Body.tsx` L43–L47](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/components/GfsImagePreview/Body.tsx#L43-L47) (same in markdown L107–L109, video L39–L41) · [`appService.ts` L856–L861](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/src/appService.ts#L856-L861)

Focus-if-exists updates **title only**. `byteLength` / `fileKind` / `mimeType` freeze at first open. Bodies assert listing size **before** download, then `fetchBytes` does an uncapped `res.arrayBuffer()`. Product writes go up to 200 MiB.

**Trigger:**
1. Open `lure.png` at 1 KB → tab stores `byteLength: 1024`.
2. Same URI is replaced with a ~200 MB payload.
3. Switch away and back (only-active mount remounts with the old payload), or double-click again (store ignores the new listing size).
4. `assertGfsImagePreviewSize(1024)` passes; the renderer pulls ~200 MB; the actual-size assert fails after the fact.

Rename that keeps `gfsUri` (`notes.png` → `notes.md`) keeps the image body mounted until the tab is closed.

**Fix:** On the existing-tab branch, patch `preview: { gfsUri, fileKind, byteLength, mimeType }` from input. Cap the download itself (Content-Length or stream-and-abort). Treat listed size as a skip hint.

---

### 5. Explorer folder listings never refetch under production cache policy

[`FileExplorerTree/index.tsx` L71–L81](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/components/SidebarNav/FileExplorerTree/index.tsx#L71-L81) · [`queryClient.ts` L8–L16](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/lib/queryClient.ts#L8-L16)

Desktop queries default to `staleTime: Infinity`, `refetchOnMount: false`, `refetchOnWindowFocus: false`. The controller’s own `childrenQuery` therefore sets `refetchOnMount` / `refetchOnWindowFocus: 'always'` ([`useGfsBrowserController.ts` L247–L251, L284–L285](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts#L247-L251)). The tree copied `GfsMoveTreeFolder` and omitted those flags, on a **persistent** surface.

**Trigger:** Expand `Reports` in the sidebar. Another session adds a file. Focus the window, or collapse and expand `Reports` again.

**Failure:** The tree keeps the first page forever unless FilesPage happens to open that same folder (shared key; Files observer has the refetch flags). Sidebar-only browsing is the point of this PR.

**Fix:** Same `refetchOnMount: 'always'` and `refetchOnWindowFocus: 'always'` on the tree’s per-node `useInfiniteQuery`.

---

### 6. Unshare while a preview tab stays focused leaves bytes on screen

[`FilePreviewPage.tsx` L30–L48](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/pages/FilePreviewPage.tsx#L30-L48) · [`useGfsBrowserController.ts` L169–L180](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/hooks/domain/useGfsBrowserController.ts#L169-L180)

R-4 / §3.B.5 is implemented as “remount when **this page’s** `accessState` becomes `'revoked'`.” That flag is session-level. A 403 on `gfs.download` is a policy denial and does not flip it. There is no re-resolve while the tab stays active. `FilePreviewPage.test.tsx:96-134` only mutates a mocked `accessState`; it never 403s the same mount.

**Trigger:** Victim previews `secret.png`. Attacker revokes the grant. Victim leaves the preview tab focused.

**Failure:** Image stays; Copy still writes the PNG. Switching away and back remounts and 403s. Session 401 while mounted does remount (`reloadToken`). Per-file unshare is the common revoke and is uncovered. A persistent tab is a longer-lived surface than the old modal.

**Fix:** Re-`download` that `gfsUri` on focus (or a short interval / the deferred watcher). On 403, drop `previewUrl` / `sourceBlob`, revoke the object URL, disable Copy, show the reason.

---

### 7. Explorer leftover-height flex can collapse to 0px

[`styles.css` L1729–L1736, L1803–L1822](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/styles.css#L1803-L1822)

`.nav-file-explorer` / `.da-file-explorer` are `flex: 1 1 auto; min-height: 0` with contained scroll. Chat sessions (cap 5) and apps have **no** max-height and keep `min-height: auto`. `.nav-links` `overflow-y: auto` does not enlarge a 0-height flex sibling.

**Trigger:** Expand Chat sessions, several sandbox apps, and the Files explorer on a short window.

**Failure:** Files chevron is expanded and the tree is 0px tall. Scrolling the nav column does not reveal a usable tree. The style commit’s contained-scroll contract fails in the layout it was written for.

**Fix:** Give `.nav-file-explorer` a real min-height (~160px) so the nav-links fallback can scroll to it, and/or cap+scroll the sessions/apps lists.

---

### 8. Fail-closed tests would stay green if the wiring were deleted

| Production | Mutant that ships | Why the suite stays green |
|---|---|---|
| Tree `childError` → `onAuthorityFailure` ([`FileExplorerTree/index.tsx` L88–L90](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/components/SidebarNav/FileExplorerTree/index.tsx#L88-L90)) and download 401 catch (L280–L288) | Delete either | No tree test drives 401 / `accessState === 'revoked'` / `gfsRoot` purge. Controller tests cover the hook, not this extra child query |
| FilePreviewPage `handleDownloadError` → `handleAuthorityFailure` ([`FilePreviewPage.tsx` L41–L48](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/pages/FilePreviewPage.tsx#L41-L48)) | Make it a no-op | Hook is mocked; `handleAuthorityFailure` is `vi.fn(() => false)` and never expected. Remount test throws **403**, which is not session-authority on the operation surface |
| `enabled: isExpanded && accessActive` | `enabled: isExpanded` | Every tree test leaves `accessState === 'active'` |

The remount test **does** bite a dropped `reloadToken` (image gone, heading stays). Folder click→click→dblclick + `advanceTimersByTime(300)` **does** bite a dropped `clearTimeout`. Sanitizer tests use exact strings.

**Fix:** Tree: expand, `listChildren` 401, assert this instance is revoked and `gfsRoot` is gone; then assert a still-expanded node does not refetch. Preview: first download succeeds, second throws `401 Unauthorized`, assert session revoke (real hook or expect `handleAuthorityFailure`). Keep the existing remount test.

---

## Minors

- **`DOUBLE_CLICK_DELAY_MS = 250`** ([`FileExplorerTree/index.tsx` L20, L112–L123](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/components/SidebarNav/FileExplorerTree/index.tsx#L20)). OS double-click windows are ~500 ms. Clicks 300 ms apart toggle **and** activate. Tests only fire the sequence with no delay.
- **401 on the preview download remounts the body that already failed** ([`FilePreviewPage.tsx` L37–L53](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/pages/FilePreviewPage.tsx#L37-L53)). User sees error → Loading → error, plus a second download. Bump `reloadToken` only when bytes were already on screen.
- **Unused `fireEvent`** in [`GfsImagePreviewBody.test.tsx` L3](https://github.com/evenfire-ai/evenfire/blob/d8ba85887c91cebc24c207b79ab755841fb02ac6/desktop-app/ui/src/components/GfsImagePreview/GfsImagePreviewBody.test.tsx#L3) (CodeQL). Copy is display-only; markdown’s sister test actually clicks it.
- **`URL.revokeObjectURL` is stubbed and never expected** in the image/video/FilePreviewPage suites. Deleting the cleanup leaves blob bytes alive; pixels still disappear.
- **Enter tests** (`FileExplorerTree.test.tsx:265-286`) `fireEvent.keyDown` only. Native button click is never synthesized, so deleting `preventDefault` stays green.
- **`App.tsx` FilePreviewPage mount** is untested. Controller tests prove `navItem === 'preview'`; they do not prove only the active preview mounts.

---

## Follow-ups (out of this unit)

- This is not #667. No HTML preview, no Control UI, no PDF download-only registry.
- `role="tree"` without arrow keys; nested buttons inside `treeitem`.
- Explorer expand state dies when the accordion unmounts.
- Pre-existing FilesPage / plugin-modal raw names in chrome were left as tracked follow-ups in the sanitizer commits.

**Merge bar:** 1–5 before merge. 6–8 are the same class of fail-closed/honesty bugs the PR’s own comments treat as load-bearing; I would not ship the explorer without 1, 2, and 5.

josue-tech-94:
@zach88  Round 2 review is complete. This PR is currently **blocked**. There are 10 merge-gating items: 1 Blocker, 8 Highs carried from the previous review, and 1 Medium PR-description issue.

### Summary

* **R2-B1 — Blocker:** New IPC contract tests use hand-written fixtures instead of producer-backed data.
* **R1-H1 — High:** Tree rows marked unreadable can still be activated.
* **R1-H2 — High:** Revoked GFS access is shown as an empty drive and cannot recover in-place.
* **R1-H3 — High:** Plugin markdown/video handoff can leave a looping Files tab.
* **R1-H4 — High:** Preview-tab dedupe retains stale metadata and preview downloads are not bounded before materialization.
* **R1-H5 — High:** Persistent tree child listings are never revalidated.
* **R1-H6 — High:** Per-resource access revocation can leave already-loaded preview bytes visible.
* **R1-H7 — High:** Sidebar layout can collapse the explorer to zero usable height.
* **R1-H8 — High:** The intended fail-closed authority wiring is not actually exercised by the tests.
* **R2-M1 — Medium:** The PR description still claims the PR is stacked on #676, but GitHub currently reports the base as `dev`.

### Details

#### R2-B1 — IPC contract tests use hand-written fixtures

**Location:** `desktop-app/ui/src/components/SidebarNav/FileExplorerTree/__tests__/FileExplorerTree.test.tsx:21`

The new tree/preview tests construct preload-shaped values locally. Typing a mock with the IPC interface does not prove that the preload/main producer actually emits that shape.

That means an IPC serialization or producer-contract change could break production while these tests remain green.

**Required change:** derive the contract fixtures from the real producer or a faithful shared producer-backed builder/Electron integration seam. Keep locally constructed fixtures only for genuinely pure transformations.

**Validation needed:** prove producer-emitted data drives tree rendering, preview-kind selection, and download/error classification.

---

#### R1-H1 — Unreadable tree entries are still actionable

**Location:** `desktop-app/ui/src/components/SidebarNav/FileExplorerTree/index.tsx:257`

The listing can mark a node with `readable === false`, but folder and file activation do not enforce that flag. The Files page already rejects the equivalent condition.

As a result, an item that the backend explicitly says cannot be read still looks actionable and can start a bad navigation/preview flow before eventually failing.

**Required change:** gate folder activation, file activation, and any preview action on the readable state. Show the existing no-access/denial UX rather than opening the item.

**Validation needed:** cover unreadable files and folders through keyboard activation and double-click/click activation.

---

#### R1-H2 — Revoked access looks like an empty drive and cannot recover in-place

**Location:** `desktop-app/ui/src/components/SidebarNav/FileExplorerTree/index.tsx:295`

`useGfsBrowserController` keeps its access state local to each hook mount. When the tree receives a session-level authorization failure, its cache is removed, but the tree then renders as if there are simply no shared files.

The same mounted explorer does not recover correctly after the user regains a valid session/retries; currently it effectively needs to be remounted.

**Required change:** make revocation state shared/query-scoped, or otherwise synchronize it across mounts. The explorer should explicitly show an unauthorized/retry state rather than "No shared files yet."

**Validation needed:** return a session 401 from discovery/children, verify fetching is stopped/cache is cleared, then retry with valid access and confirm the still-mounted explorer recovers.

---

#### R1-H3 — Plugin markdown/video handoff can create a looping Files tab

**Location:** `desktop-app/ui/src/App.tsx:841`

The plugin handoff currently recognizes image previews separately, but markdown and video files are sent through `openFilesSection`.

`FilesPage` now recognizes those same files as previewable and immediately redirects them into preview tabs. This leaves behind a Files tab seeded with the previewable URI; selecting it can send the user straight back into the preview again.

**Required change:** use the shared `resolveGfsPreview` path for plugin resources as well. Images, markdown, video, and any other supported preview kind should open directly as preview tabs. Only folders/non-previewable files should seed Files.

**Validation needed:** cover plugin-initiated markdown and video resources and verify no Files tab remains seeded with their URI.

---

#### R1-H4 — Preview-tab dedupe preserves stale metadata and downloads are unbounded

**Location:** `desktop-app/ui/src/lib/workspaceTabs.ts:310`

When a preview tab with the same URI already exists, the dedupe path updates only its title. It retains the old preview kind, MIME type, size, and other preview payload.

If the resource at that URI has changed, the existing tab can therefore render using stale metadata.

Separately, preview bodies currently call `gfs.download()` before the real byte size is known, so an advertised preview-size limit does not actually prevent an oversized resource from being materialized in memory.

**Required change:** when deduping an existing preview URI, replace/update the complete preview payload, not just the title. Also enforce a bounded download path before materializing the full `ArrayBuffer`.

**Validation needed:** reopen the same URI after its name/kind/size changes and verify the tab receives the new complete payload. Add an over-limit case proving oversized bytes are rejected before full materialization.

---

#### R1-H5 — Tree child listings never revalidate

**Location:** `desktop-app/ui/src/components/SidebarNav/FileExplorerTree/index.tsx:71`

The application uses effectively persistent query data (`staleTime: Infinity`). The existing GFS controller compensates with mount/focus revalidation, but the new tree query does not carry those policies.

A user who primarily browses through the sidebar can therefore continue seeing removed files or miss newly added files after returning to the app.

**Required change:** give the tree child query the same revalidation semantics as the controller, including mount and window-focus refresh where appropriate.

**Validation needed:** change the producer result between fetches and verify collapse/re-expand and/or window focus refreshes the visible tree.

---

#### R1-H6 — Revoking one resource can leave its preview bytes visible

**Location:** `desktop-app/ui/src/pages/FilePreviewPage.tsx:30`

The preview reacts to session-level authority revocation, but a file-specific `403` is intentionally classified as a resource-policy denial rather than a session failure.

That means if the currently previewed file is unshared, its already-downloaded blob/object URL can remain visible, including copy capability, until the tab is switched or otherwise remounted.

This matters more now that previews are persistent tabs instead of short-lived modal content.

**Required change:** revalidate access to the focused preview through the appropriate focus/refresh/watcher mechanism. If that resource returns a denial, revoke object URLs, discard the loaded bytes, disable copy/actions, and render the denial state.

**Validation needed:** render a preview successfully, make the next resource authorization check return `403`, and confirm the preview content and copy capability disappear without switching tabs.

---

#### R1-H7 — Explorer can collapse to zero height

**Location:** `desktop-app/ui/src/styles.css:1803`

The explorer is currently the flexible region allowed to shrink while other expanded sidebar sections retain their automatic minimum heights.

On a sufficiently short window, the Files section can therefore appear expanded while its actual explorer area has effectively zero usable height.

**Required change:** guarantee a practical minimum explorer height and/or constrain competing sidebar lists into their own bounded scroll regions so the explorer cannot be squeezed away.

**Validation needed:** add a browser/layout assertion using a short sidebar/window height and verify the expanded explorer remains usable.

---

#### R1-H8 — Fail-closed behavior is not actually covered by the tests

**Location:** `desktop-app/ui/src/pages/__tests__/FilePreviewPage.test.tsx:12`

The preview test directly mutates a mocked `accessState`, but it does not prove that `handleAuthorityFailure` actually causes that transition. The tree tests likewise do not exercise a real discovery/child `401`.

So removing or breaking the authority-failure callback/query gate could leave these tests green while the actual fail-closed path is broken.

**Required change:** test the actual authority transition/observable effect rather than manually placing the mocked controller into the expected end state.

**Validation needed:** cover `401` from tree discovery/children and preview download, including cache purge, revoked rendering, and prevention of further fetches while revoked.

---

#### R2-M1 — PR description has stale base/dependency information

GitHub currently reports this PR against `dev`, and #676 is already incorporated. The description still says this PR is stacked on `feat/universal-tabs` and must wait for #676.

That can make reviewers use the wrong comparison and makes the merge dependency look different from the actual repository state.

**Required change:** replace the stale stacked-base paragraph with:

> This PR is currently based on `dev`. #676 is already incorporated and is not a merge prerequisite.

No code test is needed for this item; just verify the updated text against the current GitHub metadata.

### Validation note

The review verified the current GitHub metadata and complete diff, `git diff --check`, and the repository style rules.

The review also confirmed several good parts of the implementation: title sanitization is applied at the new store boundaries, Markdown stays on the safe existing renderer, preview tabs dedupe by GFS URI, and inactive previews clean up their object URLs.

josue-tech-94:
## Round 3 review

Round 3 review is complete at head `ecb4be453705babd43eda8e0d0f5b36813ed6fce`.


### Remaining must-fix items

- **R1-H4 — High:** Preview download limits are still controlled by the renderer rather than enforced by main IPC.

### Carried findings

| Finding | Status |
| --- | --- |
| `R1-H1` | Closed |
| `R1-H2` | Closed |
| `R1-H3` | Closed |
| `R1-H4` | **Open** |
| `R1-H5` | Closed |
| `R1-H6` | Closed |
| `R1-H7` | Closed |
| `R1-H8` | Closed |
| `R2-B1` | Closed |

### R1-H4 — Preview download limit is renderer-controlled

The new bounded-download path is an improvement, but the trust boundary is still in the wrong place.

`gfs:downloadPreview` accepts any positive safe-integer `maxBytes` from the renderer. Main IPC then forwards that value to the streaming implementation, which correctly stops when that caller-provided limit is reached.

The problem is that the renderer is the untrusted side of this boundary. It can request a much larger value than the intended preview policy, so main does not actually enforce a maximum preview size.

This was introduced by corrective commit:

`3efa15662efc3b71758001f3d60ad397a5c9f4e3`

**Required change:** enforce an authoritative maximum in the main process/IPC layer. The renderer may request a smaller limit if useful, but it must not be able to raise the limit above the application's allowed preview ceiling.

If different preview types have different limits, that policy should likewise be enforced on the trusted side rather than accepted from renderer input.

**Validation needed:** add coverage proving that an excessive renderer-supplied `maxBytes` is rejected or clamped by main and that the stream never materializes bytes beyond the trusted maximum.

josue-tech-94 (approved):

